PR #3, opened 2025-05-10
Improve installation instructions

SumanthRH:
# What does this PR do? 

Tiny PR to improve our installation instructions. Adds a Dockerfile for a quick start experience.


caoshiyi (approved):
LGTM!

bhks:
Hey  @SumanthRH, @caoshiyi  thanks for adding this. 

I am wondering if you could also add instruction on how to use this docker file.

**Couple of questions :**

- My understanding is that we will be using ray cluster to execute the training and if that is the case how would this docker file helpful. Given ray itself does not manage the containers.
- Are you thinking to run ray and everything on a single node within the container to replicate the training where ray head and ray workers all running within the same container ?
- I am trying to understand how did you run your training if its a multi node setup in ray cluster. If you could add some more details about training side of things that would be awesome.
- I am also assuming you are running the openhands server on the k8s nodes, Are you also running ray on k8s via KubeRay ? And if yes can you help us with what setup you had to do ?
Thanks

SumanthRH:
hi @bhks . Happy to help. 

> My understanding is that we will be using ray cluster to execute the training and if that is the case how would this docker file helpful. Given ray itself does not manage the containers.

I think it would be helpful to go over our architecture in the blog post: 
https://novasky-ai.notion.site/skyrl-v0 

TLDR is that LLM generated code is run on a remote server (separate from the training cluster). For each trajectory of the LLM, we run the code in a separate docker container in the remote server. 

> I am trying to understand how did you run your training if its a multi node setup in ray cluster. If you could add some more details about training side of things that would be awesome.

The training cluster is completely separate from the remote server running our OpenHands server. So you can set this up[ like a regular Ray cluster](docs.ray.io/en/latest/cluster/getting-started.html), clone and install SkyRL,  and run training. For installation, we have setup instructions here: https://github.com/NovaSky-AI/SkyRL/blob/main/INSTALL.md . In terms of our setup, we ran single/multi-node training on Anyscale. 


> I am also assuming you are running the openhands server on the k8s nodes, Are you also running ray on k8s via KubeRay ? And if yes can you help us with what setup you had to do ?

I think this is the same question as before, but basically training cluster can be managed with infra of your choice (self managed ray cluster, k8s, proprietary platform, etc). 

Hope that helps! If you have more questions, I would recommend we move this discussion to a separate Github issue for clarity!

bhks:
I think I understand now thank you so much man.

bhks:
I think it would be nice to put out the training cluster pieces into the architecture as well. I did read the blog post you guys have written and thank you for that.

I may create a pull request and let you review.

SumanthRH:
> I think it would be nice to put out the training cluster pieces into the architecture as well. I did read the blog post you guys have written and thank you for that.

> I may create a pull request and let you review.


@bhks  yes agreed I think what is missing is a full system diagram or a just a description of what is running where.  Let me see if we can add that. And contributions welcome, thank you!

bhks:
Exactly I had hard time reverse engineering things like 

1. Where is the training running 
2. why are we building docker, what would we do within docker ?
3. If I have a k8s ENV, the container we are building does it need to run on CPU/GPU ?
4. The data needed to execute also needs to be coppied to the docker container if the training is running within the container so that it can be pulled into the k8s pod and executed.
5. But if you are pulling everything into a node to run with virtual ENV then there is no need of containers.


These things were confusing to me when trying to understand. 

So yeah a step by step and system level architecture would be helpful.

PR #10, opened 2025-05-15
Add more concrete instruction on how to download dataset

pcmoritz:
Small fix to the README to make it clear how exactly to download the dataset and what the relation to the DATA_DIR exactly is. Uses the existing uv environment to do the download.


tyler-griggs (approved):
LGTM, thanks!

PR #29, opened 2025-06-25
fixing minor issues with examples

erictang000:
- Fix Multiply env example (need to register env inside ray worker)
- Fix configs + add some more comments for the run remote example
- add command for downloading the dataset for each script for clarity (since it may not be in the same folder)
- remove some comments

### DIFF
--- a/skyrl-train/docs/examples/new_env.rst
+++ b/skyrl-train/docs/examples/new_env.rst
@@ -152,14 +152,15 @@ We will create a new entrypoint for training with the ``multiply`` environment b
    :linenos:
    :caption: Environment registration
 
-   # Register the multiply environment.
-   register(
-      id="multiply",  # <-- The name of the environment.
-      entry_point="examples.multiply.env:MultiplyEnv",  # <-- The path to the environment class.
-   )
-
    @ray.remote(num_cpus=1)
    def skyrl_entrypoint(cfg: DictConfig):
+      # Register the multiply environment
+      # this needs to be done inside the ray driver worker
+      register(
+         id="multiply",  # <-- The name of the environment.
+         entry_point="examples.multiply.env:MultiplyEnv",  # <-- The path to the environment class.
+      )
+
       # make sure that the training loop is not run on the head node.
       exp = BasePPOExp(cfg)
       exp.run()
--- a/skyrl-train/examples/multiply/main_multiply.py
+++ b/skyrl-train/examples/multiply/main_multiply.py
@@ -9,15 +9,15 @@ from skyrl_train.utils import initialize_ray
 from skyrl_train.entrypoints.main_base import BasePPOExp, config_dir, validate_cfg
 from skygym.envs import register
 
-# Register the multiply environment (no need to modify the skyrl package).
-register(
-    id="multiply",
-    entry_point="examples.multiply.env:MultiplyEnv",
-)
-
 
 @ray.remote(num_cpus=1)
 def skyrl_entrypoint(cfg: DictConfig):
+    # Register the multiply environment inside the ray driver worker (no need to modify the skyrl package).
+    register(
+        id="multiply",
+        entry_point="examples.multiply.env:MultiplyEnv",
+    )
+
     # make sure that the training loop is not run on the head node.
     exp = BasePPOExp(cfg)
     exp.run()
--- a/skyrl-train/examples/text_to_sql/run_sql_deepspeed.sh
+++ b/skyrl-train/examples/text_to_sql/run_sql_deepspeed.sh
@@ -8,7 +8,6 @@ set -x
 DATA_DIR="$HOME/data/sql"
 
 
-# NOTE (sumanthrh): micro_train_batch_size and micro_forward_batch_size can be tuned
 uv run --isolated --frozen --extra vllm -m skyrl_train.entrypoints.main_base \
   trainer.algorithm.advantage_estimator="grpo" \
   data.train_data="['${DATA_DIR}/train.parquet']" \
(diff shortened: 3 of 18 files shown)


skyrl-train/docs/examples/new_env.rst
```diff
@@ -152,14 +152,15 @@ We will create a new entrypoint for training with the ``multiply`` environment b
...
-   # Register the multiply environment.
-   register(
-      id="multiply",  # <-- The name of the environment.
-      entry_point="examples.multiply.env:MultiplyEnv",  # <-- The path to the environment class.
-   )
-
    @ray.remote(num_cpus=1)
    def skyrl_entrypoint(cfg: DictConfig):
+      # Register the multiply environment
+      # this needs to be done inside the ray driver worker
```
SumanthRH: why? wouldn't the prevoius behavoiur be the same as `register` statements in  `skygym/envs/__init__.py` ?

erictang000: no (I tried it and got an error). 

Seems like those get reimported in envs/__init__.py inside the ray driver process, but here it won't

SumanthRH: Got it. It's because we make a dynamic change to a dependency, and the ray task does not execute all code in the script. 

```python
import ray 
import time

time.a = 1

# works
print("a attr", time.a)

@ray.remote
def test_func():
    # fails
    print("a attr", time.a)

ray.init()
ray.get(test_func.remote())
```

erictang000: yeah maybe we need to revisit this later since this might be annoying if we want to do something like spin out each agent loop instance as a ray remote task

SumanthRH:
@erictang000 can you merge from `main` and remove the redundant `skygym` symlink in this PR as well? This was left over from migration in #31

skyrl-train/docs/examples/new_env.rst
SumanthRH: nit: this is not the driver, ti's the entrypoint ray task

```suggestion
      # this needs to be done inside the entrypoint task
```

skyrl-train/docs/recipes/skyrl-sql.rst
```diff
@@ -51,7 +51,7 @@ Unzip `data.zip` in this folder, and set the corresponding `DB_PATH` in the trai
 Running the scripts 
 -------------------
 
-We provide a script `run_skyrl_sql.sh <../../skyrl/examples/skyrl-sql/run_skyrl_sql.sh>`_ for reproducing the results for SkyRL-SQL-7B. Make sure to substitute the `DB_PATH`  and `DATA_PATH` variables with your own.
```
SumanthRH, line 54: Relative paths will work?

skyrl-train/examples/multiply/main_multiply.py
```diff
@@ -9,15 +9,15 @@
...
-# Register the multiply environment (no need to modify the skyrl package).
-register(
-    id="multiply",
-    entry_point="examples.multiply.env:MultiplyEnv",
-)
-
 
 @ray.remote(num_cpus=1)
 def skyrl_entrypoint(cfg: DictConfig):
+    # Register the multiply environment inside the ray driver worker (no need to modify the skyrl package).
```
SumanthRH:
```suggestion
    # Register the multiply environment inside the entrypoint task (no need to modify the skyrl-gym package).
```

skyrl-train/examples/text_to_sql/run_sql_deepspeed.sh
```diff
@@ -46,6 +47,7 @@ uv run --isolated --frozen --extra vllm -m skyrl_train.entrypoints.main_base \
   generator.max_turns=5 \
   generator.sampling_params.temperature=0.6 \
   generator.sampling_params.top_p=0.95 \
+  ++environment.skygym.text2sql.db_path=$DB_PATH \
```
SumanthRH: So even if we have a structured config in `skygym.yaml` it won't work rn?

SumanthRH: We should be renaming to `skyrl-gym` ??

skyrl-train/examples/text_to_sql/run_sql_fsdp_2node.sh
```diff
@@ -50,6 +51,7 @@ uv run --isolated --extra vllm -m skyrl_train.entrypoints.main_base \
   generator.max_turns=5 \
   generator.sampling_params.temperature=0.6 \
   generator.sampling_params.top_p=0.95 \
+  ++environment.skygym.text2sql.db_path=$DB_PATH \
```
SumanthRH: `skyrl-gym`

skyrl-train/examples/text_to_sql/run_sql_deepspeed.sh
erictang000: yeah let me see if i can make that change instead

skyrl-train/docs/recipes/skyrl-sql.rst
erictang000: no they would not have worked before - updated `conf.py` to allow for linking within the `skyrl_train` folder

skyrl-train/examples/text_to_sql/run_sql_deepspeed.sh
erictang000: fixed

skyrl-train/skyrl_train/config/skyrl_gym_config/default.yaml
```diff
@@ -1,2 +1,3 @@
+# @package environment.skyrl_gym
```
SumanthRH, line 1: Great!

skyrl-train/docs/recipes/skyrl-sql.rst
SumanthRH: Thanks!

SumanthRH (approved):
Thanks!

PR #32, opened 2025-06-25
Improve installation instructions in README and docs

SumanthRH:
# What does this PR do?

Improves installation instructions in readme and docs.

The quick start on the README is meant for single node with 4 GPUs. (ray need not even be installed, because `ray.init` will start a cluster if not running)

The installation guide is also updated. I have verified that the quickstart works on a fresh instance.


skyrl-train/docs/getting-started/installation.rst
```diff
@@ -29,10 +30,11 @@ We provide a docker image with the base dependencies ``sumanthrh/skyrl-train-ray
 
     git clone https://github.com/novasky-ai/skyrl_train.git 
     cd skyrl-train
-    uv sync
+    uv sync --extra vllm
```
SumanthRH, line 33: `vllm` is explicitly chosen here because of the vllm subprocess + uv issue that can happen.

We can improve docs when we have better support for sglang

skyrl-train/README.md
```diff
@@ -27,17 +33,25 @@ cd SkyRL/skyrl-train
...
 ```
 
-You should now be able to run our example script:
+Then, prepare the dataset; 
+
+```bash
+uv run -- python examples/gsm8k/gsm8k_dataset.py
+```
+
+You should now be able to run our example script (assumes atleast 4 GPUs):
```
tyler-griggs: nit: "at least"

tyler-griggs (approved):

skyrl-train/README.md
SumanthRH: Bruh

PR #33, opened 2025-06-25
Doc fixes

tyler-griggs:
Made a pass through existing docs to add and fix links.

Made a few minor docs changes along the way.

Also I added `NUM_GPUS` and `LOGGER` as envvars at the top of the `run_gsm8k.sh` script. I imagine these two (and the data path) to be the main trip-ups for people trying this for the first time (e.g., not enough gpus, or no wandb api setup) so I wanted to make it easier to edit.


skyrl-train/docs/datasets/dataset-preparation.rst
```diff
@@ -93,9 +93,18 @@ Generally, only a single method (`make_map_fn`) must be implemented to convert t
         
         return process_fn
 
-Then, the mapping function is called on each samplie the dataset, and the final converted dataset is saved to a parquet file:
+Then, the mapping function is called on each sample the dataset, and the final converted dataset is saved to a parquet file:
```
CharlieFRuan:
```suggestion
Then, the mapping function is called on each sample in the dataset, and the final converted dataset is saved to a parquet file:
```

PR #35, opened 2025-06-25
Use `skyrl-gym` from PyPI for easier dependency management

SumanthRH:
# What does this PR do?

Uses `skyrl-gym` from PyPI for better dependency management. 

Note that in most cases, users will not need to modify `skyrl-gym` source code to use new environments with skyrl-train. 

We will add a local development guide shortly for those who want to modify `skyrl-gym` and use it with `skyrl-train`.


skyrl-gym/pyproject.toml
```diff
@@ -5,14 +5,14 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "skyrl-gym"
-version = "0.1.0"
-description = "A gymnasium environment for RL agents to interact with the sky"
```
SumanthRH, line 9: the previous description was ....interesting

tyler-griggs (approved):

PR #39, opened 2025-06-26
[Docs] Add more detailed example for Text2SQL

erictang000:
Adds doc for Text2SQL with some details on enabling multi-turn training with async rollouts:
image
image
image
image


skyrl-train/docs/examples/multi_turn_text2sql.rst
```diff
@@ -0,0 +1,166 @@
+Multi-Turn GRPO Text2SQL with Async Rollouts
+=====================================================
+
+In this example, we walk through how to train an effective multi-turn Text2SQL model (beating GPT-4o) with SkyRL-train and SkyRL-gym using async rollouts.
```
SumanthRH:
```suggestion
In this example, we walk through how to train an effective multi-turn Text2SQL model (beating GPT-4o) with SkyRL using async rollouts.
```

skyrl-train/docs/examples/multi_turn_text2sql.rst
```diff
@@ -0,0 +1,166 @@
...
+        trainer.micro_forward_batch_size_per_gpu=8 \
+        trainer.micro_train_batch_size_per_gpu=1 \
+        ... # Other parameters (see `examples/text_to_sql/run_skyrl_sql.sh` for the full script)
+
+- All we have to do to enable multi-turn training with async rollouts is to simply set ``generator.max_turns`` to the maximum number of turns we want the agent to take,
+  and to make sure ``generator.async_engine=true`` and ``generator.batched=false``. 
+
+- Chat templating and loss masking for multi-turn conversations are handled by the ``SkyRLGymGenerator`` class.
+
+  - In the above example, we set ``use_conversation_multi_turn=false`` to enforce that the multi-turn conversation is formatted as a single assistant response.
```
SumanthRH, line 145: This is actually mainly for reproduction. let's just clarify that? 

I know that we saw some perf differences, but we should clarify it's not a required setting to get multi-turn to work

SumanthRH: Actually, this doc need not focus on exactly reproduction. So this flag can be removed?

SumanthRH: Or rather, have a section about chat templating here

erictang000: I think it's just good to surface it somewhere other than just in the config, since it's something that we did have to set, and you will have to think about if you want to do multi-turn training

SumanthRH: Yes agreed - how about a separate section here on this config - `use_conversation_multi_turn`

The initial configuration for multit-turn can leave it to true and then you can go into what this is

SumanthRH (approved):

skyrl-train/docs/examples/multi_turn_text2sql.rst
SumanthRH: Let's take this in a future PR for now

PR #48, opened 2025-06-26
[fix] In docs and scripts, test->validation.parquet, and ["x"] -> "['x']"

CharlieFRuan:
Change `test.parquet` to `validation.parquet` across scripts and docs to stay consistent.

Change `["x"]` to `"['x']"` which is needed for bash to interpret lists with multiple items (we already do it in most places, only a few places missing)


erictang000 (approved):
nice, thanks!

PR #50, opened 2025-06-26
[Docs] add system overview doc (pt. 1)

tyler-griggs:
Mostly copying from the blog post. Second pass will actually give detail on the control flow between components, and perhaps blurbs on (or pointers to) data formats for transfers between components. 

Screenshot 2025-06-26 at 11 35 52 AM

Screenshot 2025-06-26 at 11 35 56 AM


skyrl-train/docs/getting-started/overview.rst
```diff
@@ -0,0 +1,55 @@
...
+   :align: center
+   :width: 80%
+
+The components' responsibilities are as follows:
+
+Trainer
+~~~~~~~
+Performs the optimization steps based on configured RL algorithm. Updates model parameters based on generated trajectories and their assigned rewards.
+
+- `Trainer Worker interface <https://github.com/NovaSky-AI/SkyRL/blob/main/skyrl-train/skyrl_train/workers/worker.py#L162>`_
```
SumanthRH, line 19: What do you think about adding the Trainer class here?

SumanthRH: Ok that is the controller. Got it. 

Isn't the main abstraction the ActorGroup and not Worker? Can we add ActorGroup here

PR #52, opened 2025-07-02
[trainer] Fix fsdp warmup steps + move warmup to optimizer config

erictang000:
# What does this PR do

Previously we weren't setting warmup steps correctly for FSDP: https://github.com/NovaSky-AI/SkyRL/blob/bafdbb5db6986e3ef844127c7b9608d3d4396779/skyrl-train/skyrl_train/distributed/fsdp_strategy.py#L277

This PR fixes that, and additionally updates warmup steps to be part of the optimizer config for policy/critic. Also moves the timer for eval out of the dataloader loop.

### DIFF
--- a/skyrl-train/skyrl_train/trainer.py
+++ b/skyrl-train/skyrl_train/trainer.py
@@ -139,17 +139,17 @@ class RayPPOTrainer:
         concat_env_extras: List[Dict[str, Any]] = []
         concat_uids: List[str] = []
         sampling_params = self.cfg.generator.eval_sampling_params
-        for _, prompts in enumerate(self.eval_dataloader):
-            prompts = self._remove_tail_data(prompts)
-            generator_input, uids = self._prepare_generator_input(
-                self.cfg.generator.eval_n_samples_per_prompt, prompts, sampling_params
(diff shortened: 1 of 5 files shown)


skyrl-train/skyrl_train/trainer.py
```diff
@@ -139,17 +139,17 @@ async def eval(self) -> Dict[str, float]:
         concat_env_extras: List[Dict[str, Any]] = []
         concat_uids: List[str] = []
         sampling_params = self.cfg.generator.eval_sampling_params
-        for _, prompts in enumerate(self.eval_dataloader):
-            prompts = self._remove_tail_data(prompts)
-            generator_input, uids = self._prepare_generator_input(
-                self.cfg.generator.eval_n_samples_per_prompt, prompts, sampling_params
-            )
-            with Timer("generate"):
+        with Timer("generate"):
```
SumanthRH: when moved outside the loop, the timer is redundant. We have a "eval" timer in the training loop. We can just remove this?

erictang000: removed!

PR #53, opened 2025-07-02
[generator] Add `max_env_workers` argument for generator thread pool executor

erictang000:
# What does this PR do
Previously, we were blocking the asyncio event loop on each `env.step()` call

https://github.com/NovaSky-AI/SkyRL/blob/160291f048fa719ba90ca14773102165c9b34eaa/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py#L269

This PR updates the `SkyRLGymGenerator` to use a simple `ThreadPoolExecutor` to allow trajectories to call synchronous env.step without blocking the main event loop

### Simple Test
Using a basic multi-turn env with 3 max turns that just calls time.sleep(3)
```
class TestEnv(BaseTextEnv):
    ...
    def step(self, action: str):
        ...
        time.sleep(3)
        ...
```
we see a speedup from > 30 mins (3s * 3 turns * 256 samples) to < 4 min
Before: 
image
After:
image

### DIFF
--- a/skyrl-train/skyrl_train/config/ppo_base_config.yaml
+++ b/skyrl-train/skyrl_train/config/ppo_base_config.yaml
@@ -193,9 +193,10 @@ generator:
   # TODO (erictang000): Show clear ablations for benefits of this on GSM8K or SQL.
   zero_reward_on_non_stop: false 
 
-environment:
-  env_class: "gsm8k"
   # number of background workers for env step calls. Set to 0 to disable background workers.
   max_env_workers: 10
+
+environment:
+  env_class: "gsm8k"
   # NOTE: environment specific defaults for environment.skyrl_gym are set at the following path:
   # skyrl_gym: config/skyrl_gym_config/default.yaml
--- a/skyrl-train/skyrl_train/config/ppo_base_config.yaml
+++ b/skyrl-train/skyrl_train/config/ppo_base_config.yaml
@@ -193,9 +193,6 @@ generator:
   # TODO (erictang000): Show clear ablations for benefits of this on GSM8K or SQL.
   zero_reward_on_non_stop: false 
 
-  # number of background workers for env step calls. Set to 0 to disable background workers.
-  max_env_workers: 32
-
 environment:
   env_class: "gsm8k"
   # NOTE: environment specific defaults for environment.skyrl_gym are set at the following path:
--- a/skyrl-train/skyrl_train/config/ppo_base_config.yaml
+++ b/skyrl-train/skyrl_train/config/ppo_base_config.yaml
@@ -195,5 +195,6 @@ generator:
 
 environment:
   env_class: "gsm8k"
+  max_env_workers: 10
   # NOTE: environment specific defaults for environment.skyrl_gym are set at the following path:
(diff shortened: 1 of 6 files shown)


skyrl-train/tests/gpu/test_skyrl_gym_generator.py
```diff
@@ -226,7 +228,7 @@ async def test_generator_multi_turn_text2sql():
             max_prompt_length=6000,
             max_input_length=29048,
             max_generate_length=3000,
-            data_path="./data/sql/train.parquet",
+            data_path="/home/ray/data/sql/train.parquet",
```
erictang000: note: with fixing gpu tests to run, will fix these paths to be relative, but hard coding for now since `./data` is wrong given our new data preprocessing scripts

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -193,6 +193,9 @@ generator:
   # TODO (erictang000): Show clear ablations for benefits of this on GSM8K or SQL.
   zero_reward_on_non_stop: false 
 
+  # number of background workers for env step calls. Set to 0 to disable background workers.
+  max_env_workers: 10
```
SumanthRH: This is a bit low? 

QQ: What is the expected improvement rn with the search env with this setting? And how does using say 32 max workers fare?

skyrl-train/tests/gpu/test_skyrl_gym_generator.py
SumanthRH: can you use `os.path.expanduser(~/data/...)`

skyrl-train/skyrl_train/config/ppo_base_config.yaml
erictang000: actually, it turns out that the issue I was seeing on the search env is not solved by this - it seems to be an issue on the search server side where it times out for all given requests once in a while (i.e. i can set max_env_workers to 16, but if timeout happens all 16 requests will timeout together and still block progress)

But yeah open to increasing the default here to 32 that seems reasonable

tyler-griggs: I agree with 32 as a good default. Should we move this to the `environment` sub-config below? I even almost wonder if we should keep this `skyrl_gym` specific and move it into the skyrl_gym config. Assuming a future where many custom generators exist with customized logic for how to execute environments, `max_env_workers` feels a bit too generic to put in the `generator` config. What do you think?

erictang000: yes good point, i did spend some time thinking about this as well - I had it in the `environment` sub config originally, then moved it to the `skyrl_gym` config (`skyrl_gym.max_env_workers`) but I landed on putting it in the generator config since I thought that the generator is ultimately what is spawning the env workers, and we have some other generator config arguments that are specific to our current generator class as well (`use_conversation_multi_turn`, `zero_reward_on_non_stop`). But I think you're right and we should start moving all of these behind `skyrl_gym` if it's specific to our generator implementation.

TLDR; I'll update it to put in the `skyrl_gym` config

tyler-griggs (approved):

PR #55, opened 2025-07-03
LLM-as-a-Judge example 

lynnliu030:


skyrl-gym/skyrl_gym/envs/llm_as_judge/env.py
tyler-griggs, line 1: How do you feel about moving this under `skyrl-train/examples/llm_as_a_judge` instead (similar to the example `multiply` environment)? I don't think people will use this environment, but rather just reference it as a example, so it might make more sense to be in the `examples` directory instead of actually in the `skyrl-gym` package.

tyler-griggs: Then, the `register` call in the file above would be in a file like `main_llm_judge.py` or similar (you can take a look `examples/multiply` for reference)

Issue #58, opened 2025-07-03
Make it possible to run SkyRL trainer without deepspeed


pcmoritz:
Currently, even if the FSDP backend is used, SkyRL will import deepspeed (e.g. from https://github.com/NovaSky-AI/SkyRL/blob/main/skyrl-train/skyrl_train/models.py#L8).

This is highly unfortunate and should be fixed, because it prevents running without the nvcc compiler at runtime.

pcmoritz:
This is just a one line fix and I already fixed it in my local branch, happy to submit a PR :)

pcmoritz:
PR is here now: https://github.com/NovaSky-AI/SkyRL/pull/59

pcmoritz:
Actually there seem to be more places where deepspeed is used, after 5 iterations of generation + training, I'm getting

```
[...]

PR #59, opened 2025-07-03
Make deepspeed optional, so it is not initialized if FSDP backend is used

pcmoritz:
This fixes https://github.com/NovaSky-AI/SkyRL/issues/58


pcmoritz:
I tested this out now (including the deepspeed setting) and it is working. Even though it makes running deepspeed slightly more involved, I think it is totally worth merging since it makes the default path (FSDP2) not depend on deepspeed and more importantly therefore not depend on having the full CUDA toolchain with nvcc installed :)

Issue #60, opened 2025-07-04
text2sql task reports an error


ccp123456789:
![Image](https://github.com/user-attachments/assets/8d9c8e0b-9a85-40f9-b3b6-a6ac35638318)

Executing run_sql_fsdp.sh, text2sql task reports an error

tyler-griggs:
Thanks for reporting this, I will take a look.

lynnliu030:
@ccp123456789 hi, just to clarify, have you tried to run exactly `run_sql_fsdp.sh` (same model, same data following the data download instructions)?

ccp123456789:
> Thanks for reporting this, I will take a look.

I modified utils.py, but I'm still getting the same error.

ccp123456789:
> Thanks for reporting this, I will take a look.

<img width="949" height="147" alt="Image" src="https://github.com/user-attachments/assets/6621791f-ea17-4135-983f-15de57bba6cf" />

Running the 'run_skyrl_sql.sh' script also produces a similar error.

tyler-griggs:
Let me ask a couple clarification questions:

1) Are you modifying the scripts (`run_sql_fsdp.sh` and `run_skyrl_sql.sh`) to use a different model, dataset, or other changes? This will help us ensure we can reproduce the issue more easily.
2) Could you confirm your package versions by running `uv pip list` inside of `SkyRL/skyrl-train` and pasting the results here.

Thanks!

ccp123456789:
> Let me ask a couple clarification questions:
> 
> 1. Are you modifying the scripts (`run_sql_fsdp.sh` and `run_skyrl_sql.sh`) to use a different model, dataset, or other changes? This will help us ensure we can reproduce the issue more easily.
> 2. Could you confirm your package versions by running `uv pip list` inside of `SkyRL/skyrl-train` and pasting the results here.
> 
> Thanks!

I only modified the paths for DATA_DIR and DB_PATH; Whether I use Qwen2.5-7B-Instruct or Qwen3-8B models, I get the same error.

When I execute 'uv pip list', I get the following results

<img width="375" height="563" alt="Image" src="https://github.com/user-attachments/assets/81f81769-c99c-4ff8-ad35-9a2828a311ef" />

<img width="373" height="610" alt="Image" src="https://github.com/user-attachments/assets/3b60ac2e-6837-4d44-a294-a17a3eddb7de" />

[...]

PR #61, opened 2025-07-04
[Fix] Minor SQL fixes

tyler-griggs:


skyrl-train/examples/text_to_sql/run_sql_fsdp.sh
```diff
@@ -7,7 +7,7 @@ set -x
 # bash examples/text_to_sql/run_sql_fsdp.sh
 
 DATA_DIR="$HOME/data/sql"
-DB_PATH="$HOME/path/to/db_files/"
+DB_PATH="$HOME/data/sql/db_files/data"
```
lynnliu030, line 10: Shall we keep this just `db_files` consistent with the other scripts in the `text_to_sql` folder?

tyler-griggs: If you follow the instructions in `sql.md` the unzipped `data.zip` file actually expands and automatically creates a `data` sub-directory. So, either we need to give directions to move all the contents of the unzipped data out of `data/` and into `db_files`, or just do this. What do you think?

lynnliu030: Doing this is fine! Then we should change all other bash scripts to be consistent as well in `text_to_sql` folder?

tyler-griggs: Good call! Done.

lynnliu030 (approved):
LGTM!

ccp123456789:
> This PR makes a couple clean-up changes to the SQL example and a minor fix to the SQL environment's utils to be more robust to malformed model responses.
> 
> 1. In `utils.py`, makes the parser robust to the case where a `</solution>` tag comes before a `<solution>` tag, which I suspect is the issue in [not enough values to unpack #60](https://github.com/NovaSky-AI/SkyRL/issues/60).
> 2. Add more extensive testing for the `utils.py` logic.
> 3. Clean up the SQL example instructions.

I modified utils.py, but I'm still getting the same error.

PR #63, opened 2025-07-05
Add Apache 2.0 License

tyler-griggs:


LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
pcmoritz, line 189: You could fill something like "2025 SkyRL authors" here before merging

tyler-griggs: Oh I thought this was just directions for how to apply the license to our work, so it was intended to remain unspecified here?

pcmoritz: Ah yes I think you are right about that!

PR #64, opened 2025-07-07
[Cleanup] Many small fixes and improvements

tyler-griggs:
Iterating on several small pieces of feedback from users, especially around clarity of documentation for creating a new environment.

This PR also changes the default setting of `generator.batched` to `false`. Reasoning: a couple users have been surprised by low performance on tool-use and multi-turn tasks. The root-cause was simply that `batched` was at its default setting of `true`. Because we focus on multi-turn, tool-use agentic tasks (rather than single-turn tasks), we should default this to `false`. 

Wandb link for gsm8k: https://wandb.ai/sky-posttraining-uc-berkeley/gsm8k/runs/navm8ull?nw=nwuserskyposttraining

### DIFF
--- a/skyrl-gym/README.md
+++ b/skyrl-gym/README.md
@@ -27,6 +27,6 @@ pip install -e .
 
 ## Documentation
 
-To build your first environment, see our [example](https://skyrl.readthedocs.io/en/latest/tutorials/new_env.html).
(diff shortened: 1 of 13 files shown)


erictang000 (approved):
nice, everything makes sense to me, thanks!

Could you double check/link to a training run (multiply or gsm8k with the latest changes) just to make sure nothing is breaking?

PR #68, opened 2025-07-08
[Generator] Add initial support for non-remote SGLang engine

CharlieFRuan:
This PR supports running SGLang engines locally.

TODO:
- [ ] Add unit tests
- [ ] Support TP > 1
- [ ] Support NCCL backend (CUDA IPC) for weight syncing; currently we can only use gloo
- [ ] Benchmark performance


skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
```diff
@@ -106,42 +116,89 @@ def create_ray_wrapped_inference_engines(
...
+            )
+        elif backend == "sglang":
+            # NOTE: there is no async / sync engine distinction in SGLang
+
+            # NOTE(Charlie): We need `torch.cuda.is_available()` to be True to import SGLang. Otherwise, it requires
+            # importing vllm. See https://github.com/sgl-project/sglang/blob/v0.4.8.post1/python/sglang/srt/layers/quantization/utils.py#L11-L17
+            # Similar comment: https://github.com/volcengine/verl/blob/9cc307767b0c787e8f5ef581dac929f7bde044ef/verl/workers/fsdp_workers.py#L520-L527
+            @ray.remote
+            def get_sglang_engine():
+                # A workaround to avoid importing vllm is to give this task a GPU.
```
CharlieFRuan, line 159: This is a bit hacky. An alternative is to make the Actor a wrapper of `SGLangInferenceEngine`, so that we only lazily import SGLang in the actor, which is when we are allocated with a GPU. Like this: https://gist.github.com/CharlieFRuan/c04b4bd8b064986bc4ba16e9734720eb

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
```diff
@@ -0,0 +1,342 @@
...
+        # is already running.`
+        # Same underlying implementation: https://github.com/sgl-project/sglang/blob/v0.4.8.post1/python/sglang/srt/model_executor/model_runner.py#L689
+        success, message = await self.engine.tokenizer_manager.init_weights_update_group(obj, None)
+        return success, message
+
+    async def update_named_weight(self, request: NamedWeightUpdateRequest) -> Tuple[bool, str]:
+        """Update named weights in SGLang engine."""
+        extras = request.get("extras")
+        if extras is not None and "ipc_handles" in extras:
+            # CUDA IPC -- Here we reuse SGLang's update_weights_from_tensor, but actually load the
```
CharlieFRuan, line 261: For CUDA IPC weight syncing, an alternative is to use SGLang's `Engine.collective_rpc()`, and patch a `update_weights_cuda_ipc()` to sgl's `Scheduler` (entity that carries out the RPC), mimicing our vLLM's implementation.

Tried it, but I found it hard to patch a method to SGLang's Scheduler without native support like vLLM's `worker_extension_cls`, since sgl instantiates Scheduler in subprocess and it can easily lose what we patched in the main process. Modifying source code might work, but likely the current solution is better?

skyrl-train/pyproject.toml
```diff
@@ -56,12 +56,16 @@ skyrl-gym = { path = "./skyrl-gym" , editable = true }
 deepspeed = [
     "deepspeed==0.16.5"
 ]
+cpu_ci_test = [
```
CharlieFRuan: This is added for CPU tests like `test_models.py` which requires flash attention. We can remove this once vllm and sglang depend on the same flash attention (hence moved back to the main `dependencies` rather than in each extra).

skyrl-train/pyproject.toml
```diff
@@ -73,16 +77,15 @@ docs = [
 ]
 vllm = [
     "vllm==0.8.5",
+    "flash-attn@https://github.com/Dao-AILab/flash-attention/releases/download/v2.7.4.post1/flash_attn-2.7.4.post1+cu12torch2.6cxx11abiFALSE-cp312-cp312-linux_x86_64.whl",
```
tyler-griggs: I am going to momentarily ignore the pyproject file in this first pass of reviews because I assume it will be rebased on #73

skyrl-train/examples/gsm8k/run_gsm8k.sh
```diff
@@ -12,7 +12,10 @@ DATA_DIR="$HOME/data/gsm8k"
 NUM_GPUS=4
 LOGGER="wandb"  # change to "console" to print to stdout
 
-uv run --isolated --extra vllm -m skyrl_train.entrypoints.main_base \
+BACKEND="vllm"
```
tyler-griggs: nit: can you make this "INFER_ENGINE_BACKEND" or similar? I know it's obvious, but just to keep the distinction between the many possible backends (e.g., training or environment backends)

skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
```diff
@@ -25,7 +25,10 @@ def tp_size(self):
         return ray.get(self.inference_engine_actor.tp_size.remote())
 
     async def generate(self, input_batch: InferenceEngineInput) -> InferenceEngineOutput:
-        return await self.inference_engine_actor.generate.remote(input_batch)
+        # NOTE(Charlie): need to specify keyword `input_batch=input_batch` to avoid Ray's error on
```
tyler-griggs: Makes sense! But we can probably remove this comment since you've added the fix already.

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
```diff
@@ -0,0 +1,355 @@
...
+    # Original SGLang code had: signal.signal(signal.SIGCHLD, sigchld_handler)
+    # But this fails in Ray actors since signal handlers only work in main thread
+
+
+# Apply the patch
+sglang.srt.entrypoints.engine._set_envs_and_config = _patched_set_envs_and_config
+
+
+# TODO(charlie): duplicate of setup_envvars_for_vllm, is it needed?
+def setup_envvars_for_sglang(kwargs, bundle_indices):
```
tyler-griggs, line 89: As far as I know, `distributed_executor_backend` is specific to vLLM's configuration. Also, it should only matter for TP>1. 

For `noset_visible_devices`, I have a feeling this also would only matter for TP>1 so it's hard to test now whether something like this will be needed.

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
```diff
@@ -0,0 +1,355 @@
...
+    except Exception as e:
+        raise ValueError(f"Failed to deserialize request data: {e}")
+
+    # Extract the request data
+    ipc_handles = request["extras"]["ipc_handles"]
+    dtype = request["dtype"]
+    _ = request["shape"]
+    weight_name = request["name"]
+
+    # Below is mostly copied from vllm_engine.py's update_weight_cuda_ipc
```
tyler-griggs: nit: can probably delete this comment. I believe the code is understandable on its own and we don't need to point to vllm_engine.py's impl to understand it.

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
```diff
@@ -0,0 +1,355 @@
...
+    handle = ipc_handles[physical_gpu_id]
+    func, args = handle
+    list_args = list(args)
+    # the key is to change device id to the current device id
+    # in case two processes have different CUDA_VISIBLE_DEVICES
+    list_args[6] = device_id
+    weight = func(*list_args)
+    model.load_weights([(weight_name, weight)])
+
+    # TODO(charlie): vllm has a torch.cuda.synchronize() call here, do we need it here?
```
tyler-griggs: I don't think we need it in either. You can delete this comment and not add it. For vLLM, especially because it comes at the end of method, if a synchronization was needed I think it should be the responsibility of the caller. But in the current code I don't think it's necessary.

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
```diff
@@ -0,0 +1,355 @@
...
+            )["input_ids"]
+
+        return prompt_token_ids, sampling_params
+
+    def _postprocess_outputs(self, outputs):
+        """Process SGLang outputs to match expected format."""
+        responses: List[str] = []
+        stop_reasons: List[str] = []
+
+        if not isinstance(outputs, list):
```
tyler-griggs: qq: in what cases is the type of `outputs` different? Is it not a list when only a single prompt is sent?

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
```diff
@@ -0,0 +1,355 @@
...
+            responses=responses,
+            stop_reasons=stop_reasons,
+        )
+
+    async def generate(self, input_batch: InferenceEngineInput) -> InferenceEngineOutput:
+        """Generate responses using SGLang engine."""
+        token_ids_prompts, sampling_params = self._preprocess_prompts(input_batch)
+
+        # Generate using SGLang's async method
+        # Otherwise using `await asyncio.to_thread(self.engine.generate)` will cause issues
```
tyler-griggs: nit: Nice to know, but can probably delete this comment! If they provide an async entrypoint it definitely makes sense to use it directly instead of wrapping with `to_thread` (which we'd only want to do with a synchronous method anyways)

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
```diff
@@ -0,0 +1,355 @@
...
+        obj = InitWeightsUpdateGroupReqInput(
+            master_address=master_addr,
+            master_port=master_port,
+            rank_offset=rank_offset,
+            world_size=world_size,
+            group_name=group_name,
+            backend=backend,
+        )
+
+        # NOTE(charlie): Call the async method on tokenizer_manager directly to avoid event loop
```
tyler-griggs, line 252: Makes sense! Nit: you can probably shorten the comment to just:

```
NOTE(charlie): Call the async method on tokenizer_manager directly to avoid event loop
        # conflicts.  Same underlying implementation: https://github.com/sgl-project/sglang/blob/v0.4.8.post1/python/sglang/srt/model_executor/model_runner.py#L689
```

In these past few comments, I'm basically suggesting removing the errors (which might change as implementations change) and just give high-level direction

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
tyler-griggs: I do think this current solution is better than patching. Let's now chat with SGLang folks to get their input in more detail on enabling per-TP worker code

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
```diff
@@ -0,0 +1,355 @@
...
+            request_tensor = tensor_array.cuda()
+
+            # Use SGLang's API to update weights with custom loader
+            request_tensor = [("ipc_request", request_tensor)]
+            obj = UpdateWeightsFromTensorReqInput(
+                serialized_named_tensors=[
+                    MultiprocessingSerializer.serialize(request_tensor) for _ in range(self._tp_size)
+                ],
+                load_format=CUSTOM_WEIGHT_LOADER_PATH,
+                flush_cache=False,  # TODO(charlie): should we flush cache?
```
tyler-griggs: Ideally, we only flush once and do it on the last call to `update_named_weight` in the batch of weight updates. But, I don't have a good intuition for what goes wrong if we don't flush the cache. Maybe update the todo to "flush cache on last weight update"?

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
```diff
@@ -0,0 +1,355 @@
...
+                name=request["name"], dtype=request["dtype"], shape=request["shape"]
+            )
+
+            # Call the underlying async method for the same reason as in `init_weight_update_communicator`
+            success, message = await self.engine.tokenizer_manager.update_weights_from_distributed(obj, None)
+            return success, message
+
+    async def wake_up(self, tags: Optional[List[str]] = None):
+        """Wake up the engine. For multi-stage waking up, pass in `"weight"` or `"kv_cache"` to tags."""
+        if tags is None:
```
tyler-griggs: nit: since `tags` defaults to `None` in `ResumeMemoryOccupationReqInput` anyways, can we just do `obj = ResumeMemoryOccupationReqInput(tags=tags)` without the `if` condition?

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
```diff
@@ -0,0 +1,355 @@
...
+            obj = ResumeMemoryOccupationReqInput()
+        else:
+            obj = ResumeMemoryOccupationReqInput(tags=tags)
+        # Call the underlying async method for the same reason as in `init_weight_update_communicator`
+        await self.engine.tokenizer_manager.resume_memory_occupation(obj, None)
+        print(
+            f"Free GPU memory after wake up with tags {tags if tags is not None else 'None'}: "
```
tyler-griggs: nit: Can you update this to specify this is coming from SGLang? E.g., `Free SGLang GPU memory after...`

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
```diff
@@ -0,0 +1,355 @@
...
+            + f"{torch.cuda.mem_get_info()[0] / 1024**2:.1f} MB"
+        )
+
+    async def sleep(self, tags: Optional[List[str]] = None):
+        """Put engine to sleep."""
+        if tags is None:
```
tyler-griggs: same question about `tags` as above

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
```diff
@@ -0,0 +1,355 @@
...
+            obj = ReleaseMemoryOccupationReqInput()
+        else:
+            obj = ReleaseMemoryOccupationReqInput(tags=tags)
+        # Call the underlying async method for the same reason as in `init_weight_update_communicator`
+        await self.engine.tokenizer_manager.release_memory_occupation(obj, None)
+        print(
+            f"Free GPU memory after sleep with tags {tags if tags is not None else 'None'}: "
```
tyler-griggs: same comment about adding "SGLang" to print

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
tyler-griggs: Though I guess we will anyways reset the prefix cache, which makes the same generic call to `flush_cache` so I guess it isn't a problem that we don't do this here.

skyrl-train/skyrl_train/trainer.py
```diff
@@ -157,6 +157,8 @@ async def eval(self) -> Dict[str, float]:
 
         # Extract data_sources from env_extras
         concat_data_sources = [env_extra.get("data_source") for env_extra in concat_env_extras]
+        vis = self.tokenizer.decode(generator_output["response_ids"][0])
+        print("example: ", vis)
```
tyler-griggs: I know you're just doing the same thing we did in the training loop, but can we actually update this print a bit:

E.g., `Eval output example: `

skyrl-train/skyrl_train/trainer.py
```diff
@@ -203,8 +205,14 @@ def train(self):
         self.weights_manager = InferenceWeightsManager(
             self.policy_model, self.inference_engine_client, self.cfg.trainer.placement.colocate_all
         )
+        # NOTE(Charlie): sglang's engine needs to sync weights after wake up. see https://github.com/sgl-project/sglang/issues/7939
+        # Change it to True after sglang fixes the issue.
+        eval_need_sync = self.cfg.trainer.placement.colocate_all and self.cfg.generator.backend == "sglang"
```
tyler-griggs: tiny nit: `eval_need_sync` -> `sync_before_eval`. The former just sounded a bit awkward to me :D

tyler-griggs (commented):
Adding a first round of comments

skyrl-train/tests/gpu/test_engine_generation.py
```diff
@@ -11,12 +20,13 @@
 from omegaconf import DictConfig
 from skyrl_train.inference_engines.base import InferenceEngineInput
 from skyrl_train.utils import initialize_ray
+from typing import Tuple
 
 model = "Qwen/Qwen2.5-1.5B-Instruct"
-tp_size = 2
+TP_SIZE = 2
```
tyler-griggs: nit: can we actually just move `tp_size` to the test parameterization, and default to 2 for vllm and 1 for sglang? It was bad design on my part to put this global `tp_size` here in the first place, and it's probably bad to have this default here and also an sglang-specific override later. 

Can you also change `model` to `MODEL`?

skyrl-train/tests/gpu/test_policy_local_engines_e2e.py
tyler-griggs, line 1: By the way, not sure if you saw it, but we did have a similar test for sglang under `tests/sglang`. It seems like your changes here replace it and we should delete the `tests/sglang` folder. We originally split `sglang` into its own folder so we could just run `uv run --isolated --extra dev --extra vllm pytest tests/gpu` and run all tests without having to separately run vllm and sglang. I don't know the right way to structure this long-term, but for now we rarely run all gpu tests like this and more often we manually choose some subset of tests, so I think it's fine.

skyrl-train/tests/gpu/test_policy_local_engines_e2e.py
```diff
@@ -18,8 +22,8 @@
 from skyrl_train.inference_engines.base import InferenceEngineInput
 from skyrl_train.entrypoints.main_base import config_dir
 
-model = "Qwen/Qwen2.5-1.5B-Instruct"
-tp_size = 2
+model = "Qwen/Qwen2.5-0.5B-Instruct"
+TP_SIZE = 2
```
tyler-griggs: Same nit as above -- can we move tp_size to test parameters with defaults of 2 and 1 for vllm and sglang, respectively.

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
CharlieFRuan: Output of this is not a list: `outputs = llm.generate("Hello")`

While output of this is a list: `outputs = llm.generate(["Hello"])`

I will remove this `if not isinstance(outputs, list):` since our `InferenceEngineInput`'s `prompts` and `prompt_token_ids` are always `List`

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
CharlieFRuan: It seems that we clear prefix cache before updating the weights. So I guess we do not need to flush cache here?

```python
    async def broadcast_to_inference_engines(self, inference_engine_client):
        use_prefix_cache = self.cfg.generator.enable_prefix_caching
        generator_dtype = str_to_torch_dtype(self.cfg.generator.model_dtype)
        cache_reset_task = None
        if use_prefix_cache and torch.distributed.get_rank() == 0:
            # clear prefix cache
            cache_reset_task = inference_engine_client.reset_prefix_cache()
```

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
CharlieFRuan: Ah yes, I can remove this if-else

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
CharlieFRuan: I see! I'll leave it here and address when we support TP > 1, would that be fine? It does not seem to affect current cases

skyrl-train/tests/gpu/test_policy_local_engines_e2e.py
CharlieFRuan: Ah yes, I'll delete `tests/sglang`. And what you said about being able to run an entire folder with `--extra vllm` makes sense. We could structure our future tests with `marks=pytest.mark.sglang` and `pytest.mark.vllm` and that should help keep both groups of tests in the same folder while able to run an entire folder.

CharlieFRuan:
@tyler-griggs addressed comments, rebased to main, and tested with GPU tests and `run_gsm8k.sh`. Ready for another round of review:)

skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
```diff
@@ -106,42 +114,92 @@ def create_ray_wrapped_inference_engines(
...
+            def get_sglang_engine():
+                # A workaround to avoid importing vllm is to give this task a GPU.
+                import os
+
+                before_cuda_visible_devices = os.environ.get("CUDA_VISIBLE_DEVICES", "")
+                os.environ["CUDA_VISIBLE_DEVICES"] = "0"
+                from skyrl_train.inference_engines.sglang.sglang_engine import SGLangRayActor
+
+                os.environ["CUDA_VISIBLE_DEVICES"] = before_cuda_visible_devices
+
```
SumanthRH, line 167: Seems like another issue with `CUDA_VISIBLE_DEVICES` patching in Ray for num_gpus=0 cc @pcmoritz

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
```diff
@@ -0,0 +1,334 @@
...
+            tensor_data = bytearray(data_with_marker)
+            tensor_data.extend(b"\x00" * (padded_size - data_size))
+            tensor_array = torch.frombuffer(tensor_data, dtype=torch.uint8)
+            request_tensor = tensor_array.cuda()
+
+            # Use SGLang's API to update weights with custom loader
+            request_tensor = [("ipc_request", request_tensor)]
+            obj = UpdateWeightsFromTensorReqInput(
+                serialized_named_tensors=[
+                    MultiprocessingSerializer.serialize(request_tensor) for _ in range(self._tp_size)
```
SumanthRH, line 284: QQ: I don't know much about weight sync in sglang. Does `MultiprocessingSerializer.serialize`  take tensor on GPU and serialize it via pickle in sglang? If so, I'm wondering why the conversion base64 encoding + .cuda op was needed?

SumanthRH:
cc @tyler-griggs Let's try to get this PR in today

skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
pcmoritz: We are going to fix it, see https://github.com/ray-project/ray/issues/54868

SumanthRH: Great!

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
CharlieFRuan: - The `serialize()` will have to take in a list of `(str, torch.Tensor)` (i.e. named tensors) so that it matches what the codepath in SGLang expects when they deserialize it
- We want to pass our `request` (which includes ipc handle info) down to the codepath so our customized `update_weight_cuda_ipc()` can handle it in a per-TP-worker fashion
- However our request is not a tensor, so we base64 encoded it and store in a `torch.unit8` Tensor

Not sure if I answered your question...

SumanthRH: Okay got it. 

I am wondering why the tensor needs to be on gpu though? I am not too familiar with sglang internals, but it looks like the Engine class runs the tokenizer manager and the TP workers are in a different process. Does `MultiprocessingSerializer.serialize` expect a gpu tensor?

CharlieFRuan: Oh yes you are right, it should be a CPU tensor. Changed and tested. Thanks for the catch!

skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
```diff
@@ -106,42 +114,92 @@ def create_ray_wrapped_inference_engines(
...
+                tokenizer=tokenizer,
+            )
+        elif backend == "sglang":
+            # NOTE: there is no async / sync engine distinction in SGLang
+
+            # NOTE(Charlie): We need `torch.cuda.is_available()` to be True to import SGLang. Otherwise, it requires
+            # importing vllm. See https://github.com/sgl-project/sglang/blob/v0.4.8.post1/python/sglang/srt/layers/quantization/utils.py#L11-L17
+            # Similar comment: https://github.com/volcengine/verl/blob/9cc307767b0c787e8f5ef581dac929f7bde044ef/verl/workers/fsdp_workers.py#L520-L527
+            @ray.remote
+            def get_sglang_engine():
```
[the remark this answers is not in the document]
CharlieFRuan: We won't be able to allocate 0.01 GPU to schedule this task

skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
SumanthRH: QQ: Why put this code in a ray task instead of running this on the same process again? you can do the same kind of cuda visible devices save , patch and restore ??

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
SumanthRH: sgtm

skyrl-train/skyrl_train/inference_engines/sglang/sglang_engine.py
```diff
@@ -0,0 +1,333 @@
...
+    name, tensor = named_tensors[0]
+    if name != "ipc_request":
+        raise ValueError(f"Expected IPC request tensor name to be 'ipc_request', got: {name}")
+
+    # Convert tensor to bytes, then decode and deserialize
+    tensor_bytes = tensor.cpu().numpy().tobytes()
+    end_marker = b"__END_OF_REQUEST__"
+    end_index = tensor_bytes.find(end_marker)
+    if end_index == -1:
+        raise ValueError("End marker not found in tensor data")
```
SumanthRH, line 127: Nice!

SumanthRH (approved):
Thanks! 

Left a nit - trying to understand why sglang import happens in  a separate ray task - but beyond that LGTM. 

Some CPU tests are failing. Let's make sure CPU tests and the new GPU tests pass.

skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
CharlieFRuan: If we make the following change (basically what you suggested):
image

The logging when running `run_gsm8k.sh` would be:
```
(skyrl_entrypoint pid=2559350) CHARLIE before_cuda_visible_devices: 
(skyrl_entrypoint pid=2559350) CHARLIE after_cuda_visible_devices: 0
(skyrl_entrypoint pid=2559350) torch.cuda.is_available(): False
```
and SGLang will report error due to no GPU available. However `test_policy_local_engine_e2e.py` works.

I believe this is because, when running E2E, we have already imported `torch` in this process before this `CUDA_VISIBLE_DEVICES` change, so pytorch already has its opinion on what resources are / are not available.

To prove this, for the current solution, if we simply do
```python
    @ray.remote
    def get_sglang_engine():
        import torch  # newly added
        print(f"CHARLIE torch.cuda.is_available(): {torch.cuda.is_available()}")  # prints False
        import os

        before_cuda_visible_devices = os.environ.get("CUDA_VISIBLE_DEVICES", "")
        os.environ["CUDA_VISIBLE_DEVICES"] = "0"
        print(f"CHARLIE torch.cuda.is_available(): {torch.cuda.is_available()}")  # prints False
```

It would also fail the same way, as SGLang's codepath will think there is no GPU (which uses torch to judge).

CharlieFRuan:
This is ready again. Able to run the entire `skyrl-train/tests/gpu/test_policy_local_engines_e2e.py`. The CPU test is fixed by itself for some reason (not sure if flaky or fixed unintentionally), was some sentencepiece/tiktoken missing dependency issue.

skyrl-train/skyrl_train/utils/utils.py
```diff
@@ -261,6 +266,13 @@ def get_physical_gpu_id():
 def initialize_ray(cfg: DictConfig):
     # TODO(sumanthrh): introduce a debug mode and add debugging flags like `CUDA_LAUNCH_BLOCKING` here
     env_vars = {}
+
+    # NOTE (charlie): See https://github.com/vllm-project/vllm/blob/c6b0a7d3ba03ca414be1174e9bd86a97191b7090/vllm/worker/worker_base.py#L445
+    # and https://docs.vllm.ai/en/v0.9.2/usage/troubleshooting.html?h=nccl_cumem_enable#known-issues
+    # Same for SGLang as we set `NCCL_CUMEM_ENABLE` to 0 in `sglang_engine.py`'s _patched_set_envs_and_config
+    if cfg.generator.weight_sync_backend == "nccl":
+        env_vars["NCCL_CUMEM_ENABLE"] = "0"
```
CharlieFRuan, line 274: comment for visibility

CharlieFRuan:
current 1.5B Qwen2.5 gsm8k run on 4xH1001
image

skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
tyler-griggs: (leaving this to @SumanthRH )

tyler-griggs (approved):
I'm leaving one open thread to @SumanthRH , otherwise approve!

skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
SumanthRH: Got it. Thanks @CharlieFRuan

PR #67, opened 2025-07-08
Minor cleanup pt 2.

tyler-griggs:
Another round of minor cleanups.

PR #69, opened 2025-07-08
Revert "Minor cleanup pt 2."

tyler-griggs:
Reverts NovaSky-AI/SkyRL#67

Users are getting errors with the update to `_init_tool_groups` (from `init_tool_groups`) related to fixed skyrl-gym. Will revert this and fix.

```
File "/tmp/ray/session_2025-07-08_01-27-39_405126_2918156/runtime_resources/working_dir_files/_ray_pkg_32090e37b3aa13c5/examples/exp/search/env.py", line 31, in
 __init__
    self._init_tool_groups([self.tool_group])
    ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'ExpSearchEnv' object has no attribute '_init_tool_groups'. Did you mean: 'init_tool_groups'?. Did you mean: '_return_value'?
```


pcmoritz:
Nice catch! For the time being, it might be worth configuring the pyproject.toml in sky-trainer to use the checked out version of sky-gym, so we can move faster on these kinds of things and not wait for a release. I believe uv should support that.

EDIT: Seems like we also need to symlink it

PR #71, opened 2025-07-09
[Installation] Use skyrl-gym as a symlink for easier development; Add a developer guide

SumanthRH:
# What does this PR do?

1. Shifts to our older way of using skyrl-gym as a symlink. This is so that developers can easily modify skyrl-gym. Previously I shifted to using skyrl-gym from PyPI for cleaner installs in #35 , but then modifications don't show up properly. 

*Is there a better way ?* : There could be, but this is the simplest. Note that we require two things:  using the latest skyrl-gym during training and modifications getting shipped appropriately with Ray. To meet both, symlink is an easy option. 

2. Adds a developer guide: Development instructions were mostly missing. Added a simple developer guide. There are two kinds of people who will be developing on skyrl-train: researchers/engineers customizing skyrl-train for their use-case and contributors looking improve skyrl-train in some way (and an overlap of both). 

Preview: 

Screenshot 2025-07-09 at 9 47 11 AM

image


skyrl-train/docs/getting-started/development.rst
```diff
@@ -0,0 +1,69 @@
...
+Pre-requisities
+---------------
+
+Follow the :doc:`installation guide <installation>`. Make sure that the installation works with our :doc:`quick start example <quickstart>`.
+
+
+Modifying the code
+-------------------
+
+- Are you adding a new environment or task? Follow the :doc:`new task tutorial <../tutorials/new_env>`. Add your custom code to a folder in ``examples/`` similar to `examples/multiply <https://github.com/NovaSky-AI/SkyRL/blob/main/skyrl-train/examples/multiply>`_.
```
tyler-griggs: Two nits:

``examples/`` -> ``skyrl-train/examples``?

Should we add a note that you can add this folder any where, but we usually add new code in ``examples/``? Just so it's clear that this isn't a requirement. E.g. something like:

"Write your custom code in any location, although we typically develop in the ``skyrl-train/examples/`` folder similar to..."

tyler-griggs: Could we add one more option:

"- Are you creating a custom Generator (ex: porting an agent harness, custom trajectory generation logic, etc.)?  Same as the above, you can write your custom code any where, and we typically use ``skyrl-train/examples/``. 

Philipp and I are working on adding a better example of doing this (which the SWE-bench integration will also have). At that time, I'll write a doc on it. I just want to ensure we are mentioning that as an option everywhere because it will be a common path.

SumanthRH: Done!

PR #65, opened 2025-07-07
[WIP] SearchR1 reproduction fixes

erictang000:
# What does this PR do?
Fixes for searchR1 repro

PR #73, opened 2025-07-09
[Dependencies] Upgrade to torch 2.7 

SumanthRH:
# What does this PR do?

Upgrades to torch 2.7. The upgrade is somewhat implicit because we upgrade the inference engines and then let them pin the required torch versions. (vllm uses torch 2.7.0 and sglang uses 2.7.1)

- This PR also upgrades CUDA to 12.8. 

TODO: 
- [x] Test sglang after upgrade 
- [x] Publish new docker image to dockerhub


skyrl-train/pyproject.toml
```diff
@@ -72,20 +73,26 @@ docs = [
...
+    "vllm==0.9.2",
     # NOTE (sumanthrh): We explictly use a flashinfer wheel from their index. 
     # The wheels on PyPI don't come with pre-compiled kernels and the package will JIT compile them at runtime (terribly slow).
-    "flashinfer-python@https://github.com/flashinfer-ai/flashinfer/releases/download/v0.2.5/flashinfer_python-0.2.5+cu124torch2.6-cp38-abi3-linux_x86_64.whl#sha256=43d767b912c0c43a04be99595e0123eab9385fc72530a2874b5fb08e3145c0be",
+    "flashinfer-python@https://download.pytorch.org/whl/cu128/flashinfer/flashinfer_python-0.2.6.post1%2Bcu128torch2.7-cp39-abi3-linux_x86_64.whl",
 ]
 sglang = [
-    "sglang[srt,openai]==0.4.6.post4",
+    "sglang[srt,openai]==0.4.8.post1",
     "torch-memory-saver>=0.0.5",
```
CharlieFRuan: I think you can do `sglang[srt,openai,torch_memory_saver]==0.4.8.post1`

SumanthRH: Thanks!

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -172,7 +172,7 @@ generator:
   # whether to use a conversation based format for multi-turn generations
   # if false, append multi-turn model responses and env observations to the original assistant response
   # if true, each multi-turn model response and env observations is stored in a separate assistant/user message respectively
-  use_conversation_multi_turn: false
+  use_conversation_multi_turn: true
```
SumanthRH, line 175: This should be `true` by default. This is because the standard way should be using multi-turn conversations for observation. We had noticed some perf degradation for qwen 7B models if be used `true` instead of `false`, but that seems model specific. 

 sglang has also some issues with the flag -I think their `/completions` endpoint behaves differently in some way, because of which I got empty responses.  
 
 I will make sure to add this caveat with sglang to the docs as well.

skyrl-train/skyrl_train/inference_engines/sglang/sglang.patch
SumanthRH, line 1: Patch is no longer needed after sglang upgrade

skyrl-train/skyrl_train/inference_engines/remote_inference_engine.py
```diff
@@ -120,7 +118,7 @@ async def update_named_weight(self, request: NamedWeightUpdateRequest):
                 f"{self.url}/{weight_update_method}",
                 json={
                     "name": request["name"],
-                    "dtype": torch_dtype_to_str(request["dtype"]),
```
SumanthRH, line 123: minor edit here: since the inference engines expect this to be a string anyways, I have changed the datatype to be string for consistency

skyrl-train/skyrl_train/workers/fsdp/fsdp_worker.py
```diff
@@ -127,7 +126,7 @@ async def broadcast_to_inference_engines(self, inference_engine_client):
                         inference_engine_client.update_named_weight(
                             {
                                 "name": name,
-                                "dtype": generator_dtype,
+                                "dtype": self.cfg.generator.model_dtype,
```
SumanthRH, line 129: switched to directly providing the string version

SumanthRH:
This Pr should be ready for review now. 

I've tested:
- vLLM with colocation and remote server (with FSDP2)
- vLLM with colocation with DeepSpeed
- Sglang remote server with FSDP2 and DeepSpeed

and things have worked. 

I haven't done careful perf benchmarking given no major changes

skyrl-train/examples/remote_inference_engine/run_vllm_server.sh
```diff
@@ -2,7 +2,9 @@
 # bash examples/remote_inference_engine/run_vllm_server.sh
 set -x
 
-uv run --isolated --extra vllm -m skyrl_train.inference_engines.vllm.vllm_server \
+# NOTE (sumanthrh): Currently, there's an issue with distributed executor backend ray for vllm 0.9.2.
+# For standalone server, we use mp for now. 
+CUDA_VISIBLE_DEVICES=4,5,6,7 uv run --isolated --extra vllm -m skyrl_train.inference_engines.vllm.vllm_server \
```
SumanthRH, line 7: This fails rn with vllm 0.9.2:

```bash
uv run --isolated --extra vllm -m  vllm.entrypoints.api_server     --model Qwen/Qwen2.5-1.5B-Instruct     --tensor-parallel-size 4     --host 127.0.0.1     --port 8001     --seed 42     --max-model-len 4096     --enable-prefix-caching     --enable-chunked-prefill     --dtype bfloat16     --gpu-memory-utilization 0.9     --enable-sleep-mode     --max-num_batched_tokens 8192     --max-num-seqs 1024     --trust-remote-code     --distributed-executor-backend ray
```

There's some issue here that we can dig into later. as such the remote server can use any backend so I'm just using mp backend

skyrl-train/pyproject.toml
```diff
@@ -72,20 +74,28 @@ docs = [
...
 ]
 vllm = [
-    "vllm==0.8.5",
+    "vllm==0.9.2",
     # NOTE (sumanthrh): We explictly use a flashinfer wheel from their index. 
     # The wheels on PyPI don't come with pre-compiled kernels and the package will JIT compile them at runtime (terribly slow).
-    "flashinfer-python@https://github.com/flashinfer-ai/flashinfer/releases/download/v0.2.5/flashinfer_python-0.2.5+cu124torch2.6-cp38-abi3-linux_x86_64.whl#sha256=43d767b912c0c43a04be99595e0123eab9385fc72530a2874b5fb08e3145c0be",
+    "flashinfer-python@https://download.pytorch.org/whl/cu128/flashinfer/flashinfer_python-0.2.6.post1%2Bcu128torch2.7-cp39-abi3-linux_x86_64.whl",
+    "torch==2.7.0",
+    "torchvision"
```
SumanthRH, line 82: Added explicit torch and torchvision dependencies

Torch versions need not be specified but it's kinda helpful to know what we're dealing with. Does not hurt since `uv ` dependency resolution will catch issues. 

the main reason is being able to use torch compiled for cu12.8. By default, the version used is 12.6

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
```diff
@@ -86,10 +87,9 @@ def init_weight_update_communicator(
             f"rank={rank}, world_size={world_size}, group_name={group_name}",
         )
 
-    def update_weight(self, name, dtype, shape):
-        import torch
-
+    def update_weight(self, name: str, dtype: str, shape: List[int]):
         """Broadcast weight to all vllm workers from source rank 0 (actor model)"""
+        dtype = str_to_torch_dtype(dtype)
```
SumanthRH, line 92: vllm 0.9.2 change AFAIK. Previously we passed string direclty here but now `model_config.dtype` is of type `torch.dtype`

skyrl-train/skyrl_train/utils/utils.py
```diff
@@ -200,6 +200,9 @@ def validate_cfg(cfg: DictConfig):
             "`offload_after_step=False` is not supported for DeepSpeed, please set `offload_after_step` to `true` for both policy and critic"
         )
 
+    if cfg.generator.backend == "sglang" and not cfg.generator.use_conversation_multi_turn:
+        raise NotImplementedError("`use_conversation_multi_turn=False` is not supported for SGLang backend")
```
SumanthRH, line 204: Encountered an issue with sglang and `use_conversation_multi_turn=False` so it's not implemented rn

CharlieFRuan:
Should we bump sglang to 0.4.9.post1? It seems to have some RL-related fixes: 
- https://github.com/sgl-project/sglang/pull/7733
- https://github.com/sgl-project/sglang/pull/7812

I could also do it in #68

SumanthRH:
@CharlieFRuan Interesting point!

Regarding the version bump:

For this PR, I want to simply use a torch 2.7 compatible sglang version. The fixes for RL training you mentioned are for colocated training, and should not affect non-colocated training AFAIK. We only support non-colocated rn so this should be fine.

> I could also do it in https://github.com/NovaSky-AI/SkyRL/pull/68

This is best, since you are introducing colocated training. Btw, another issue with going from sglang 0.4.8 -> 0.4.9 is that flash-infer version gets upgraded from 0.2.6 -> 0.2.7 . flashinfer is switching over to not releasing cuda specific wheels from 0.2.7: https://github.com/flashinfer-ai/flashinfer/issues/1139#issuecomment-2964776207 , so we'll need to carefully test out how slow JIT compilation is (it was pretty bad in previous versions). Ideally we avoid JIT compiled as long as possible.

skyrl-train/examples/text_to_sql/run_skyrl_sql.sh
```diff
@@ -52,6 +52,7 @@ uv run --isolated --extra vllm -m skyrl_train.entrypoints.main_base \
   generator.async_engine=true \
   generator.batched=false \
   environment.env_class=text2sql \
+  generator.use_conversation_multi_turn=false \
```
SumanthRH, line 55: Since #65  has landed, the same fix needs to be made for the search example

skyrl-train/skyrl_train/config/ppo_base_config.yaml
tyler-griggs: I am tracking this and other sglang feature support tasks here: https://github.com/NovaSky-AI/SkyRL/issues/82

If there's more to say about this issue and what you observed, could you dump it into the issue?

skyrl-train/examples/remote_inference_engine/run_sglang_server.sh
```diff
@@ -0,0 +1,12 @@
+# Launches sglang server for Meta-Llama-3-8B-Instruct on 2 GPUs.
```
tyler-griggs: nit: it's Qwen2.5-1.5B-Instruct (not llama), and defaults to 4 gpus (not 2)

tyler-griggs (approved):
I ran colocated vllm and remote sglang, no issues!

skyrl-train/skyrl_train/config/ppo_base_config.yaml
SumanthRH: Yes sg. Let me in fact dive into this separately. The main issue is probably due to some difference in the implementation of the `/v1/completions` API in sglang vs vllm

SumanthRH: The empty response comes up due to our defaults (which might have to change as well)

skyrl-train/examples/remote_inference_engine/run_sglang_server.sh
SumanthRH: fixed. thanks

PR #74, opened 2025-07-09
[skyrl-gym] GSM8k - LLM Judge example

lynnliu030:
Create a new GSM8k env with LLM judge example

WanDB run:
image

### DIFF
--- a/skyrl-train/examples/llm_as_a_judge/llm_judge_env.py
+++ b/skyrl-train/examples/llm_as_a_judge/llm_judge_env.py
@@ -9,7 +9,13 @@ import re
 PROMPT = """
 You are a strict math evaluation assistant.
 
-Compare the following **gold** and **predicted** math solutions. Your job is to determine if the predicted solution is mathematically correct and arrives at the same final answer as the gold solution.
+Compare the following **gold** and **predicted** math solutions. Your job is to determine if the predicted solution is mathematically correct and if the predicted solution ends with a line of the form:
+
+#### <number>
+
+You must only give a score of "1" if:
+- The final line of the predicted solution **ends with `#### <number>`**, and
+- The number **matches the final answer in the gold solution** exactly.
 
 Instructions:
 - You may provide internal reasoning or explanation before giving your final judgment.
@@ -21,10 +27,7 @@ or
 
 ### Final Score: 0
 
-Where:
-- Use "1" only if the predicted solution is mathematically correct and leads to the same final answer as the gold solution.
-- Use "0" otherwise.
-- Do not include any explanation after the final score.
+Do not include any explanation after the final score.
 """
 
 
@@ -57,8 +60,6 @@ class GSM8kLLMJudgeEnv(BaseTextEnv):
                 model=self.model, messages=[{"role": "user", "content": message}]
             )
             reply = response.choices[0].message.content.strip()
-
-            print(f"LLM Judge response: {reply}")
 
             # Try to parse score from "### Final Score: x"
             match = re.search(r"### Final Score:\s*([01](?:\.0)?)", reply)


skyrl-train/skyrl_train/config/skyrl_gym_config/default.yaml
```diff
@@ -2,4 +2,8 @@
 # number of background workers for env step calls. Set to 0 to disable background workers.
 max_env_workers: 32
 text2sql:
-  db_path: "/home/ray/default/sql_data"
\ No newline at end of file
+  db_path: "/home/ray/default/sql_data"
+
+llm_as_a_judge:
+  model: "gpt-4o-mini"
+  OPENAI_API_KEY: "openai_api_key"
```
SumanthRH: api keys should be read from env vars and not expected in config files

SumanthRH: model being read from config is good. In fact you should also add `base_url` here for people serving local models with vLLM / sglang

skyrl-train/examples/llm_as_a_judge/env.py
```diff
@@ -0,0 +1,64 @@
...
+        assert "ground_truth" in extras["reward_spec"], "ground_truth is required in reward_spec field"
+        self.ground_truth = extras["reward_spec"]["ground_truth"]
+
+        # Set up OpenAI client
+        openai_api_key = env_config.OPENAI_API_KEY
+        self.llm_judge_client = OpenAI(api_key=openai_api_key)
+        self.model = env_config.model
+
+    def _get_reward(self, action: str) -> float:
+        prompt = f"""
```
SumanthRH: nit: `prompt` should live outside the function as a global var for since it's easier to see.

skyrl-train/examples/llm_as_a_judge/env.py
```diff
@@ -0,0 +1,64 @@
...
+            Answer:
+        """
+
+        try:
+            response = self.llm_judge_client.chat.completions.create(
+                model=self.model,
+                messages=[{"role": "user", "content": prompt}],
+                max_tokens=1,
+            )
+            reply = response.choices[0].message.content.strip()
```
SumanthRH: Nit: ideally, this will allow for CoT + final answer. This is to (1) improve judge performance (2) support weaker open source models that might not follow the prompt correctly and thus mess up the first token. 

So it can be CoT + final answer in a expected format ("### Final Score: 1.0")  and then we parse the answer. 

This seems okay as a first step tho

SumanthRH (commented):
Looks good overall as a first step! Could you also add some screenshots of a successful run with gpt 4o mini as a judge for gsm8k?

skyrl-train/examples/llm_as_a_judge/env.py
lynnliu030: Sounds good!

skyrl-train/examples/llm_as_a_judge/env.py
```diff
@@ -0,0 +1,70 @@
...
+    def __init__(self, env_config: DictConfig, extras: Dict[str, Any] = {}):
+        super().__init__()
+
+        assert "reward_spec" in extras, "reward_spec field is required"
+        assert "ground_truth" in extras["reward_spec"], "ground_truth is required in reward_spec field"
+        self.ground_truth = extras["reward_spec"]["ground_truth"]
+
+        # Set up OpenAI client
+        openai_api_key = os.getenv("OPENAI_API_KEY")
+        self.llm_judge_client = OpenAI(api_key=openai_api_key)
```
SumanthRH: shouldn't we read the base_url from the `env_config` and pass it along here?

```suggestion
        self.llm_judge_client = OpenAI(base_url=env_config.base_url, api_key=openai_api_key)
```

skyrl-train/skyrl_train/config/skyrl_gym_config/default.yaml
```diff
@@ -1,10 +1,16 @@
 # @package environment.skyrl_gym
 # number of background workers for env step calls. Set to 0 to disable background workers.
 max_env_workers: 32
+
 text2sql:
   db_path: "/home/ray/default/sql_data"
+
+llm_as_a_judge:
+  model: "gpt-4o-mini"
+  base_url: "http://localhost:8000/v1"  # local endpoint
```
SumanthRH: um the base url should be default `None`? that should hit OpenAi's api server

```suggestion
  base_url: null  # or a local endpoint: http://localhost:8000/v1
```

SumanthRH (changes requested):
thanks! minor fixes

skyrl-train/examples/llm_as_a_judge/run_llm_judge.sh
```diff
@@ -0,0 +1,56 @@
...
+  generator.batched=true \
+  generator.n_samples_per_prompt=5 \
+  generator.gpu_memory_utilization=0.8 \
+  trainer.logger="$LOGGER" \
+  trainer.project_name="gsm8k" \
+  trainer.run_name="gsm8k_llm_as_a_judge" \
+  trainer.resume_mode=null \
+  trainer.ckpt_path="$HOME/ckpts/gsm8k_1.5B_ckpt" \
+  environment.env_class=llm_as_a_judge \
+  environment.skyrl_gym.llm_as_a_judge.model="gpt-4o-mini" \
```
SumanthRH, line 56: I just ran the command and it looks like the env isn't being picked up because the gsm8k dataset prep includes an "env class" column with value "gsm8k"

skyrl-train/examples/llm_as_a_judge/llm_judge_env.py
```diff
@@ -0,0 +1,72 @@
...
+        message = PROMPT + f"\n\nGOLD SOLUTION:\n{self.ground_truth}\n\nPREDICTED SOLUTION:\n{action}\n\nAnswer:"
+
+        try:
+            response = self.llm_judge_client.chat.completions.create(
+                model=self.model, messages=[{"role": "user", "content": message}]
+            )
+            reply = response.choices[0].message.content.strip()
+
+            # Try to parse score from "### Final Score: x"
+            match = re.search(r"### Final Score:\s*([01](?:\.0)?)", reply)
```
tyler-griggs, line 65: Why try to parse from "### Final Score: x" since we explicitly prompted it to only return 0 or 1?

skyrl-train/examples/llm_as_a_judge/main_llm_judge.py
```diff
@@ -0,0 +1,36 @@
...
+import hydra
+from omegaconf import DictConfig
+from skyrl_train.utils import initialize_ray
+from skyrl_train.entrypoints.main_base import BasePPOExp, config_dir, validate_cfg
+from skyrl_gym.envs import register
+
+
+@ray.remote(num_cpus=1)
+def skyrl_entrypoint(cfg: DictConfig):
+    # Register the multiply environment inside the entrypoint task (no need to modify the skyrl-gym package).
```
tyler-griggs: nit: `llm_as_a_judge`, not `multiply`

skyrl-train/examples/llm_as_a_judge/run_llm_judge.sh
```diff
@@ -0,0 +1,57 @@
+set -x
+
+# Colocated GRPO training+generation for Qwen2.5-Coder-7B-Instruct on SkyRL-SQL-653 data.
```
tyler-griggs: this is 1.5b (not 7b) and is gsm8k with a judge (not sql data)

skyrl-train/examples/llm_as_a_judge/run_llm_judge.sh
```diff
@@ -0,0 +1,57 @@
...
+# Uses 1 node with 8 GPUs.
```
tyler-griggs: this uses 4 gpus by default, not 8

skyrl-train/examples/llm_as_a_judge/gsm8k_dataset_judge_env.py
tyler-griggs, line 1: nit: could you update this file name to `gsm8k_with_judge_dataset.py`? Our unspoken standard for dataset preprocessing scripts is they end in `dataset`. If it ends in `env`, it sounds like this is an environment implementation, not the dataset preprocessing.

tyler-griggs (commented):
Left a few small comments and nits!

skyrl-train/examples/llm_as_a_judge/llm_judge_env.py
lynnliu030: Update the prompt!

SumanthRH (approved):
Thanks! 

Before merging it would be good to confirm that the provided example script runs after filling in the API keys

PR #75, opened 2025-07-10
[skyrl-gym] search env cleanup

erictang000:
# What does this PR do
Cleans up `_parse_action` to just parse the action:
```python
def _parse_action(self, action: str) -> List[Optional[str]]:
        match = None
        if "<search>" in action and "</search>" in action:
            match = re.search(r"<search>(.*?)</search>", action, re.DOTALL)
        return [match.group(1)] if match else [None]
```

since we are always using 

```python
tool_group_name = self.tool_group.get_name()
tool_name = self.tool_group.get_tool_names()[0]
```


tyler-griggs (approved):
Thanks! If you haven't already, could you just launch a training run for 1 step to sanity check that this all works?

erictang000:
> Thanks! If you haven't already, could you just launch a training run for 1 step to sanity check that this all works?

image
training is running!

PR #78, opened 2025-07-10
Minor cleanup

tyler-griggs:
No modifications to logic, just editing docs, adding type hints, etc.


skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -49,169 +49,6 @@ def __init__(
         else:
             self.env_executor = None
 
-    def _update_engine_input_chat_history(
```
tyler-griggs, line 52: note: not deleted, just moved lower in the file to make agent_loop & generate more accessible. Users trying to ready through the generator kept getting stuck here unnecessarily

pcmoritz (approved):

PR #80, opened 2025-07-12
Very simple local coding sandbox example

pcmoritz:
**This PR adds a simple sandbox example using the Guix package manager**

This change introduces a basic sandbox environment powered by Guix (https://guix.gnu.org/), a package manager that comes with thousands of pre-built software packages that you can easily customize and compile.

**How it works**

Guix makes it easy to set up development environments with all the right dependencies. For example, if you want to work on a Linux package like Inkscape, you can run `guix shell --development inkscape` and it automatically gives you a shell with everything needed to build that software.

You can also create custom environments by running `guix shell -m manifest.scm`, where the `manifest.scm` file lists exactly which packages and dependencies you want. Think of this like Python's `uv run` command, but instead of just handling Python projects, it can set up complete Linux development environments for any type of software:

`manifest.scm` (in Guix) = `pyproject.toml` (in Python)
`guix shell` (for any Linux software) = `uv run` (for Python only)

This gives developers a consistent way to create isolated, reproducible development environments for any kind of project.

**How to use it**

The docker image / environment you are working in needs to have guix installed, you can
e.g. install it by running

```shell
wget https://guix.gnu.org/install.sh
chmod +x install.sh
sudo ./install.sh
```

For the	following commands, we assume that you are in the `sky-train/examples/simplecoder` directory. It is worth running the following command once to initialize the packages (this ensures the following commands won't time out):

```shell
guix shell -m manifest.scm -- sh
```

To run the example, first clone the test repository
```shell
git clone https://github.com/SWE-agent/test-repo
```

and then you can run the "agent" using
```shell
python simplecoder.py
```


skyrl-train/examples/simplecoder/simplecoder.py
```diff
@@ -0,0 +1,305 @@
...
+
+if __name__ == "__main__":
+    import os
+    import simplecoder
+
+    manifest = os.path.abspath("manifest.scm")
+    working_dir = os.path.abspath("test-repo")
+    executor = simplecoder.GuixExecutor(working_dir, manifest)
+
+    coder = simplecoder.SimpleCoder(os.environ["OPENAI_KEY"], "o4-mini", executor)
```
SumanthRH: nit: might be better to rename to `OPENAI_API_KEY`, we already use it in other envs like #55

SumanthRH (commented):
Looks great! 

I'm also able to follow the instructions to get this running easily. 

Discussed offline and we'll have the same PR description as a README for running with guix in `examples/simplecoder`. 

In a future PR, we will integrate with `skyrl-train` so that we have an actual training script to show. This can start off simple with livecodebench and then we can also move to complex tasks.

SumanthRH (approved):

Issue #70, opened 2025-07-08
Cuda Memory Error After Saving Checkpoint


howard-yen:
Hello, I'm following the GSM8K training script, and I'm getting Cuda memory error in the training step after a checkpoint is saved. 
When running the script, it's able to complete all the training steps until the step where it saves the checkpoint. The checkpoint seems to be saved correctly and I'm able to restart the run and resume from training at that step.
I'm using Qwen3-8B as the model, and decreased the mini batch size accordingly. This is on a 8 x H100 node so I also changed the number of GPUs to 8. 

The error log is here and I would be happy to provide more information as needed, thanks!
```
(skyrl_entrypoint pid=716206) 2025-07-08 19:50:10.227 | INFO     | skyrl_train.trainer:train:310 - Finished: 'save_checkpoints', time cost: 29.84s
Training Step Progress: 100%|██████████| 20/20 [1:58:31<00:00, 722.48s/it]
(skyrl_entrypoint pid=716206) 2025-07-08 19:50:34.677 | INFO     | skyrl_train.trainer:train:232 - Started: 'step'
(skyrl_entrypoint pid=716206) 2025-07-08 19:50:35.931 | INFO     | skyrl_train.weights_manager:__enter__:56 - Started: 'sync_weights_to_inference_engines'
(AsyncVLLMInferenceEngine pid=719401) INFO 07-08 19:50:35 [executor_base.py:226] It took 1.185601 seconds to wake up tags ['weights'].
(skyrl_entrypoint pid=716206) 2025-07-08 19:50:39.509 | INFO     | skyrl_train.weights_manager:__enter__:56 - Finished: 'sync_weights_to_inference_engines', time cost: 3.58s
(skyrl_entrypoint pid=716206) 2025-07-08 19:50:39.509 | INFO     | skyrl_train.weights_manager:__enter__:60 - Started: 'offload_policy_model_to_cpu'
(AsyncVLLMInferenceEngine pid=719401) INFO 07-08 19:50:39 [block_pool.py:264] Successfully reset prefix cache


(skyrl_entrypoint pid=716206) 2025-07-08 19:50:40.385 | INFO     | skyrl_train.weights_manager:__enter__:60 - Finished: 'offload_policy_model_to_cpu', time cost: 0.88s
(skyrl_entrypoint pid=716206) 2025-07-08 19:50:40.452 | INFO     | skyrl_train.trainer:train:232 - Finished: 'step', time cost: 5.78s
[...]

PR #94, opened 2025-07-16
[FIX] Garbage collect temp buffers after checkpoint

tyler-griggs:
## What does this PR do?
Resolve issue where offloading optimizers failed after checkpointing. 

As reported in #70, OOMs can occur during inference engine wake-up after checkpointing. The root-cause was the `state_dict` materialization in `save_ckpt` created temporary buffers that were not garbage collected before we try to wakeup the inference engine kv cache, causing an OOM. 

This PR executes the garbage collection and resolves the OOM issue.

## Tests
Added a GPU test to check for successful offloading after checkpointing. It fails before this PR.

## What's next?
We should switch to using pytorch's distributed checkpointing APIs for checkpointing, which is much simpler.

### DIFF
--- a/skyrl-train/tests/gpu/test_worker_offload.py
+++ b/skyrl-train/tests/gpu/test_worker_offload.py
@@ -15,7 +15,6 @@ from skyrl_train.entrypoints.main_base import config_dir
 from skyrl_train.training_batch import TrainingOutputBatch
 
 MODEL_NAME = "Qwen/Qwen2.5-1.5B-Instruct"
-CKPT_PATH = "$HOME/ckpts/test/"
 
 
 def get_test_actor_config() -> DictConfig:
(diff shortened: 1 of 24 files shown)


skyrl-train/tests/gpu/test_worker_offload.py
```diff
@@ -288,3 +290,73 @@ async def test_cpu_offload_correctness(cfg, worker_type, strategy):
...
+
+@pytest.mark.parametrize(
+    "strategy",
+    [
+        "deepspeed",
+        "fsdp",
+        "fsdp2",
+    ],
+)
+def test_offload_after_ckpt(strategy):
```
SumanthRH, line 303: QQ: does the test fail before this PR?

SumanthRH (approved):
Left a minor comment. Thanks!

skyrl-train/tests/gpu/test_worker_offload.py
tyler-griggs: Yes indeed! Should have mentioned that

SumanthRH: Great

Issue #96, opened 2025-07-17
Provide HTTP Endpoint for inference engine client


tyler-griggs:
Many agent scaffolds assume access to an OpenAI API compatible HTTP endpoint to get inference engine rollouts. Currently, we provide the generic [InferenceEngineClient](https://github.com/NovaSky-AI/SkyRL/blob/194d4f1cb9ea444eef35a5f20e7e843343b23733/skyrl-train/skyrl_train/inference_engines/inference_engine_client.py#L11) (which exposes the `InferenceEngineInterface`) to agent stacks to get model rollouts from `generate()`. This client should also expose an HTTP endpoint.

PR #101, opened 2025-07-18
[Generator] Add FastAPI server for inference engine client

CharlieFRuan:
This is a first stab for https://github.com/NovaSky-AI/SkyRL/issues/96

We merge this PR to a separate branch `http-inf_engine-openai` for now for users to build on top of.

The intended use case is when users have their existing agent harness and want to do rollout with the existing code while using SkyRL.

After setting the following configs:
```
generator:
  use_http_server_inference_engine_client: true
  http_server_inference_engine_client_host: "127.0.0.1"
  http_server_inference_engine_client_port: 8000
```
user can post OpenAI requests to `127.0.0.1:8000` in tools like `openai.OpenAI` or `litellm`. You can refer to `skyrl-train/examples/rollout_with_http_server_litellm.py` on how to use this feature.

Underlying-ly, when these configs are set, we launch a FastAPI server when calling `InferenceEngineClient.__init__()`, and only return when the server is ready.

We support OpenAI API, including features like `response_format`, but not features like `stream` or `tools` yet. You can see the supported / unsupported fields in `openai_api_protocol.py`.

### This PR is verified with:
- Running `run_gsm8k.sh` with `skyrl_gym_generator.py` but with `generate_with_http_server()` rather than `inference_engine_client.generate()`. Note this `generate_with_http_server()` is only intended for testing
- Unit tests `tests/gpu/test_http_inference_engine_client.py` and `skyrl-train/tests/cpu/http/test_openai_request_utils.py`
- `rollout_with_http_server_litellm.py` to check compatibility with tools like LiteLLM

However, how to balance such ease of use with token-in-token-out (https://github.com/NovaSky-AI/SkyRL/issues/123) is still under discussion. 

### TODOs:
- [x] Clean up `openai_api_protocol.py` -- be clear about the fields we do not support right now
- [x] CPU unit tests for `convert_inference_output_to_openai()` and `convert_openai_to_inference_input()`
- [x] Add a config flag that will spin up this server; test with `skyrl_gym_generator.py` E2E on `run_gsm8k.sh`
- [x] Address `trajectory_id`
- [x] Be more pragmatic about translating ChatCompletionRequest' sampling-related params to our `sampling_params` in `launch_inference_engine_http_server.py`
- [x] Check error handling
- [x] Test E2E with SkyRLGymGenerator with HTTP server, make sure performance not affected
  - Before 0.5B run_gsm8k takes 100sec for each step's generate, now it takes 130sec. It should be the performance difference between adding each request to the vLLM engine (with HTTP server) vs. batched generate with an offline engine.

### Quick note:
- If we use `OpenAI.chat.completions.create()` in `skyrl_gym_generator.py` without setting a maximum limit on the number of HTTP requests, it will hang. Using `aiohttp.ClientSession().post` does not have this issue. Revisit once this issue is encountered again.

### DIFF
--- a/skyrl-train/examples/rollout_with_http_server_litellm.py
+++ b/skyrl-train/examples/rollout_with_http_server_litellm.py
@@ -38,6 +38,7 @@ TP_SIZE = 1
 SERVER_PORT = 8123
 SERVER_HOST = "127.0.0.1"
 
+
 def get_test_actor_config() -> DictConfig:
     """Get base config with test-specific overrides."""
     with hydra.initialize_config_dir(config_dir=config_dir):
@@ -79,7 +80,6 @@ def init_to_simulate_trainer():
 
     # Wait for server to be ready using the helper method
     wait_for_server_ready(host=SERVER_HOST, port=SERVER_PORT, max_wait_seconds=30)
-
 
 
 def agent_loop(prompt: str, base_url: str):
@@ -128,12 +128,12 @@ def main():
         print(f"outputs[0]: {outputs[0]}")
         for output in outputs:
             assert len(output) == 3 * 2
-        
 
     finally:
         # You do not need to call this in your custom generator either.
         shutdown_server(host=SERVER_HOST, port=SERVER_PORT, max_wait_seconds=5)
         ray.shutdown()
+
 
 if __name__ == "__main__":
     main()
--- a/skyrl-train/examples/rollout_with_http_server_litellm.py
+++ b/skyrl-train/examples/rollout_with_http_server_litellm.py
@@ -4,6 +4,14 @@ Example of using LiteLLM with SkyRL's HTTP inference server.
 Note that `init_to_simulate_trainer()` is not needed in your custom generator, we
 do it here to simulate what the trainer will instantiate.
 
+Set the following configs in your bash script:
+```
+generator:
+  use_http_server_inference_engine_client: true
+  http_server_inference_engine_client_host: "127.0.0.1"
+  http_server_inference_engine_client_port: 8000
+```
+
 Also note that `trajectory_id` is important for better trajectory routing for prefix cache reuse.
 
 Run with:
--- a/skyrl-train/examples/rollout_with_http_server_litellm.py
+++ b/skyrl-train/examples/rollout_with_http_server_litellm.py
@@ -1,0 +1,131 @@
+"""
+Example of using LiteLLM with SkyRL's HTTP inference server.
+
+Note that `init_to_simulate_trainer()` is not needed in your custom generator, we
+do it here to simulate what the trainer will instantiate.
+
+Also note that `trajectory_id` is important for better trajectory routing for prefix cache reuse.
+
+Run with:
+`uv run --isolated --extra dev --extra vllm --extra litellm python examples/rollout_with_http_server_litellm.py`
+"""
+
+import threading
+import ray
+import hydra
+from omegaconf import DictConfig
+from litellm import completion
+from concurrent.futures import ThreadPoolExecutor
+from skyrl_train.inference_engines.launch_inference_engine_http_server import (
+    serve,
+    wait_for_server_ready,
+    shutdown_server,
+)
+from skyrl_train.entrypoints.main_base import config_dir
+from tests.gpu.test_policy_vllm_e2e import init_inference_engines
+from uuid import uuid4
+
+MODEL = "Qwen/Qwen2.5-0.5B-Instruct"
+TP_SIZE = 1
+SERVER_PORT = 8123
+SERVER_HOST = "127.0.0.1"
+
+def get_test_actor_config() -> DictConfig:
+    """Get base config with test-specific overrides."""
+    with hydra.initialize_config_dir(config_dir=config_dir):
+        cfg = hydra.compose(config_name="ppo_base_config")
+
+        # Override specific parameters
+        cfg.trainer.policy.model.path = MODEL
+        cfg.trainer.critic.model.path = ""
+        cfg.trainer.placement.policy_num_gpus_per_node = TP_SIZE
+        cfg.generator.async_engine = True
+        cfg.generator.num_inference_engines = 1
+        cfg.generator.inference_engine_tensor_parallel_size = TP_SIZE
+        cfg.generator.run_engines_locally = True
+
+        return cfg
+
+
+def init_to_simulate_trainer():
+    cfg = get_test_actor_config()
+    cfg.trainer.placement.colocate_all = True  # Use colocate for simplicity
+    cfg.generator.weight_sync_backend = "nccl"
+    cfg.trainer.strategy = "fsdp2"
+
+    client, _ = init_inference_engines(
+        cfg=cfg,
+        v1=True,
+        use_local=True,
+        async_engine=cfg.generator.async_engine,
+        tp_size=cfg.generator.inference_engine_tensor_parallel_size,
+        colocate_all=cfg.trainer.placement.colocate_all,
+    )
+
+    # Start server in background thread using serve function directly
+    def run_server():
+        serve(client, host=SERVER_HOST, port=SERVER_PORT, log_level="warning")
+
+    server_thread = threading.Thread(target=run_server, daemon=True)
+    server_thread.start()
+
+    # Wait for server to be ready using the helper method
+    wait_for_server_ready(host=SERVER_HOST, port=SERVER_PORT, max_wait_seconds=30)
+
+
+
+def agent_loop(prompt: str, base_url: str):
+    chat_history = [
+        {"role": "user", "content": prompt},
+    ]
+    max_turns = 3
+    trajectory_id = uuid4().hex
+
+    while True:
+        response = completion(
+            model=f"openai/{MODEL}",  # Add openai/ prefix for custom endpoints
+            messages=chat_history,
+            api_base=base_url,
+            temperature=0.7,
+            max_tokens=100,
+            # NOTE(Charlie): this is needed for better trajectory routing for prefix cache reuse
+            trajectory_id=trajectory_id,
+        )
+        chat_history.append(response.choices[0].message)
+        # dummy multi-turn chat
+        max_turns -= 1
+        if max_turns <= 0:
+            break
+        else:
+            chat_history.append({"role": "user", "content": "Repeat what you just said."})
+    return chat_history
+
+
+def main():
+    try:
+        # 1. This method simulates what the trainer will do. When you set
+        # `generator.use_http_server_inference_engine_client=True` in bash script,
+        # you do not need to write these code at all. In your custom generator,
+        # you can simply post request to `base_url`. The server will be ready automatically
+        # before `CustomGenerator.generate()` is called.
+        init_to_simulate_trainer()
+        base_url = f"http://{SERVER_HOST}:{SERVER_PORT}/v1"
+
+        # 2. Pretend we have 500 tasks, and run one agent loop for each task parallely.
+        prompts = ["Hello, how are you?"] * 500
+        with ThreadPoolExecutor() as executor:
+            output_tasks = [executor.submit(agent_loop, prompt, base_url) for prompt in prompts]
+            outputs = [task.result() for task in output_tasks]
+        print(f"len(outputs): {len(outputs)}")
+        print(f"outputs[0]: {outputs[0]}")
+        for output in outputs:
+            assert len(output) == 3 * 2
+        
+
+    finally:
+        # You do not need to call this in your custom generator either.
+        shutdown_server(host=SERVER_HOST, port=SERVER_PORT, max_wait_seconds=5)
+        ray.shutdown()
+
+if __name__ == "__main__":
+    main()
--- a/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
+++ b/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
@@ -2,7 +2,7 @@ import asyncio
 import copy
 from uuid import uuid4
 import skyrl_gym
-from typing import List, Dict, Any, Optional, Tuple
+from typing import List, Dict, Any, Optional, Tuple, Hashable
 import numpy as np
 from concurrent.futures import ThreadPoolExecutor
 from tqdm.asyncio import tqdm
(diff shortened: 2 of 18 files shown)


skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -34,10 +41,46 @@ def __init__(
...
+                daemon=True,
+            )
+            self._server_thread.start()
+            wait_for_server_ready(
+                host=self.http_server_inference_engine_client_host,
+                port=self.http_server_inference_engine_client_port,
+                max_wait_seconds=30,
+            )
+            # Store the base URL for direct HTTP requests
+            self.base_url = f"http://{self.http_server_inference_engine_client_host}:{self.http_server_inference_engine_client_port}"
```
tyler-griggs, line 58: Why is the Generator class responsible for spinning up the server, rather than the client spinning up the server itself?

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -49,6 +92,55 @@ def __init__(
...
+                    host=self.http_server_inference_engine_client_host,
+                    port=self.http_server_inference_engine_client_port,
+                    max_wait_seconds=5,
+                )
+                if hasattr(self, "_server_thread") and self._server_thread.is_alive():
+                    self._server_thread.join(timeout=5)
+            except Exception:
+                pass
+
+    async def _openai_generate(
```
tyler-griggs: I'm also wondering whether this logic should sit inside the the inference engine client -- any reason why not? This code seems very generic and not specific to the skyrl gym generator, so we'd prefer that users creating a new custom generator don't have to re-implement this, right?

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
```diff
@@ -15,8 +15,10 @@ class InferenceEngineClient(InferenceEngineInterface):
     Note that InferenceEngineClient sub-classes InferenceEngineInterface so it can be used as if talking to a single engine.
     """
 
-    def __init__(self, engines: List[InferenceEngineInterface]):
+    def __init__(self, engines: List[InferenceEngineInterface], generator_config: Dict[str, Any]):
         self.engines = engines
+        self.generator_config = generator_config
```
tyler-griggs, line 21: qq: is this used?

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
```diff
@@ -15,8 +15,10 @@ class InferenceEngineClient(InferenceEngineInterface):
...
+        self.model_name = generator_config.get("model_name")
```
tyler-griggs, line 22: qq: is this only used in tests? If yes, is it actually needed here?

skyrl-train/skyrl_train/inference_engines/launch_inference_engine_http_server.py
```diff
@@ -0,0 +1,272 @@
...
+    """Set the global inference engine client."""
+    global _global_inference_engine_client
+    global _global_uvicorn_server
+    global _global_backend
+    _global_inference_engine_client = inference_engine_client
+    _global_uvicorn_server = uvicorn_server
+    _global_backend = backend
+
+
+def convert_openai_to_inference_input(request: ChatCompletionRequest, backend: str) -> InferenceEngineInput:
```
tyler-griggs, line 62: Based on your work here, do you think we should update the InferenceEngineInput/Output types to better match ChatCompletionRequest/Response?

skyrl-train/skyrl_train/inference_engines/openai_api_protocol.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+class ChatMessage(BaseModel):
+    """OpenAI chat message format."""
+
+    role: str
+    content: str
+
+
+class ChatCompletionRequest(BaseModel):
```
tyler-griggs, line 33: Is there a reason not to directly use `openai.types.chat` to get these types? Is it too heavy-weight for some reason?

tyler-griggs (commented):
I left a handful of questions in this first pass. My primary question is about the intended user experience of creating a custom generator. Ideally, a user just flips a config flag or in the custom generator calls into a single method (probably exposed by the inference engine client), and receives the URL to an OpenAI API compatible endpoint. The current example code in the `skyrl_gym_generator.py` is more than we'd like to make a user implement.

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
CharlieFRuan: Good point, moved the spinning up / shutting down into `InferenceEngineClient`, triggered based on generator config.

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
CharlieFRuan: > don't have to re-implement this

Agreed, moved to a static helper method `generate_with_http_server()`.

> sit inside the the inference engine client

If users are calling `inference_engine_client.xxx()`, they might as well just not use the HTTP server right...

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
CharlieFRuan: If we are spinning up the HTTP server inside the InferenceEngineClient, then we are using those related fields (e.g. host, port, etc.)

Besides, OpenAI requests have `model` field. I used the `model_name` field, got from `generator_config` to check in `_handle_completion()`:

```
if _global_inference_engine_client.model_name != request.model:
```

Or we could just ignore the `model` field in OpenAI request, but that seems implicit and perhaps we should help user check whether the model field aligns with the model loaded?

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
CharlieFRuan: See response above

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
```diff
@@ -15,10 +16,79 @@ class InferenceEngineClient(InferenceEngineInterface):
...
+
+    def __getstate__(self):
+        """
+        Override to avoid pickling the server thread.
+        Needed when passing InferenceEngineClient as an argument to async_run_ray_method().
+        """
+        state = self.__dict__.copy()
+        # Thread objects are not picklable – just drop the reference.
+        state["_server_thread"] = None
+        return state
```
CharlieFRuan, line 98: Not sure if this is good practice, but needed if we want to spin up server inside inference engine client. Alternative is to spin up server in `main_base.py`

skyrl-train/skyrl_train/inference_engines/launch_inference_engine_http_server.py
```diff
@@ -0,0 +1,323 @@
...
+) -> InferenceEngineOutput:
+    """Generate responses using direct ClientSession.post calls with the InferenceEngineClient HTTP server."""
+    prompts = input_batch.get("prompts")
+    trajectory_ids = input_batch.get("trajectory_ids")
+    sampling_params = input_batch.get("sampling_params")
+    if trajectory_ids is not None:
+        assert len(prompts) == len(trajectory_ids), "prompts and trajectory_ids must have the same length"
+
+    # Use aiohttp session for direct HTTP requests (no concurrency limiting)
+    async with aiohttp.ClientSession(timeout=aiohttp.ClientTimeout(total=None)) as session:
```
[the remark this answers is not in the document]
SumanthRH: 300 is certainly not enough, and we should be careful in introducing such a timeout value for generation. 

`generate_with_http_server` can be launched in an asynchronous fashion where the policy continues training and never awaits the coroutine. In this case, a total timeout of 300 will lead to timeout errors.

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
```diff
@@ -15,10 +16,79 @@ class InferenceEngineClient(InferenceEngineInterface):
...
+
+                shutdown_server(
+                    host=self.http_server_inference_engine_client_host,
+                    port=self.http_server_inference_engine_client_port,
+                    max_wait_seconds=5,
+                )
+                if hasattr(self, "_server_thread") and self._server_thread.is_alive():
+                    self._server_thread.join(timeout=5)
+            except Exception as e:
+                print(f"Error shutting down HTTP server: {e}")
```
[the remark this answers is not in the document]
CharlieFRuan: Agreed. We should simply move the code to `teardown()`. Since `_handle_termination` is not called anywhere yet, will leave the `__del__` for now. Graceful shutdown in general is left as a TODO. See notes on `atexit.register(self._handle_termination)` in `worker.py` for more. Besides, `fsdp_worker.py` does not have a `_handle_termination()` yet.

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -108,7 +126,14 @@ async def agent_loop(
...
                 )
-            engine_output = await self.inference_engine_client.generate(engine_input)
+            if self.use_http_server_inference_engine_client:
+                engine_output = await generate_with_http_server(
+                    base_url=self.base_url,
+                    model_name=self.model_name,
+                    input_batch=engine_input,
+                )
+            else:
+                engine_output = await self.inference_engine_client.generate(engine_input)
```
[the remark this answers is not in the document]
CharlieFRuan: Addressed.

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -49,6 +92,55 @@ def __init__(
         else:
             self.env_executor = None
 
+    def __del__(self):
```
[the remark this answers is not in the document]
CharlieFRuan: See discussion at gemini review for more on this.

skyrl-train/skyrl_train/inference_engines/openai_api_protocol.py
SumanthRH: Using the data models from OpenAI directly would also mean that our server is fully compatible with all the output fields, which can be very misleading. I think as long as we support a subset (field names are same etc), this is fine

skyrl-train/examples/rollout_with_http_server_litellm.py
SumanthRH, line 1: This is great. Can we actually create a folder called `http_server` in `examples/` ? 

```bash
http_server/
   |- rollout_with_http_server_litellm.py # standalone for showing the feature
   |- run_with_http_server.sh # bash script for E2E training with GSM8K and http server
```

skyrl-train/examples/rollout_with_http_server_litellm.py
```diff
@@ -0,0 +1,139 @@
...
+Note that `init_to_simulate_trainer()` is not needed in your custom generator, we
+do it here to simulate what the trainer will instantiate.
+
+Set the following configs in your bash script:
+```
+generator:
+  use_http_server_inference_engine_client: true
+  http_server_inference_engine_client_host: "127.0.0.1"
+  http_server_inference_engine_client_port: 8000
+```
```
SumanthRH: This part is best just put in a standalone bash script...it's confusing to mention this here since this script is meant to be standalone

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
SumanthRH: Yes this is fine. Server thread will be alive as long as the original inference engine client object is alive in the training loop, ray actors can freely interact with it.

skyrl-train/skyrl_train/inference_engines/launch_inference_engine_http_server.py
```diff
@@ -0,0 +1,323 @@
...
+    # Chat completion endpoint
+    @app.post("/v1/chat/completions", response_model=ChatCompletionResponse)
+    async def chat_completion(request: ChatCompletionRequest, raw_request: Request):
+        return await handle_chat_completion(request, raw_request)
+
+    # Health check endpoint
+    @app.get("/health")
+    async def health_check():
+        return {"status": "healthy"}
+
```
SumanthRH:
```suggestion
    # Health check endpoint
    # All inference engine replicas are initialized before creating `InferenceEngineClient`, and thus
    # we can start receiving requests as soon as the FastAPI server starts
    @app.get("/health")
    async def health_check():
        return {"status": "healthy"}

```

SumanthRH: Thought we'd add this since this is a difference from say vllm

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -34,10 +35,27 @@ def __init__(
...
+        self.http_server_inference_engine_client_port = generator_cfg.get(
+            "http_server_inference_engine_client_port", 8000
+        )
+
+        if self.use_http_server_inference_engine_client:
+            # Store the base URL for direct HTTP requests
+            self.base_url = f"http://{self.http_server_inference_engine_client_host}:{self.http_server_inference_engine_client_port}"
+        else:
+            self.base_url = None
+
```
SumanthRH, line 61: Is `use_conversation_multi_turn` supported with the http server? I think not, because we'd need a `/completions` endpoint that supports token in, token out. 

Can you raise an error here if `use_conversation_multi_turn=False` with http server?

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
SumanthRH: Yeah I think having it be fully OpenAI compatible is a good first step, before introducing such overrides

CharlieFRuan:
Thanks for the review; addressed all comments.

GSM8K timing on 1 A100 for 0.5B:
image
This performance discrepancy I think is expected because we are breaking a previously batched N request into N HTTP requests.

Reward is identical:
image

Still running Search R1, which should have much less performance difference.

CharlieFRuan:
For SearchR1 (max_turn=2, 1xH100, 0.5B model) we have roughly the same time cost for the generate step between HTTP server and `.generate()`.

CharlieFRuan:
for future reference:
- the litellm example needs this fix: https://github.com/NovaSky-AI/SkyRL/pull/153
- A generator config check: https://github.com/NovaSky-AI/SkyRL/pull/155

skyrl-train/skyrl_train/inference_engines/launch_inference_engine_http_server.py
```diff
@@ -0,0 +1,325 @@
...
+    )
+
+
+async def handle_chat_completion(request: ChatCompletionRequest, raw_request: Request) -> ChatCompletionResponse:
+    """Handle chat completion request."""
+    if _global_inference_engine_client is None:
+        raise HTTPException(
+            status_code=HTTPStatus.INTERNAL_SERVER_ERROR, detail="Inference engine client not initialized"
+        )
+    if _global_inference_engine_client.model_name != request.model:
```
CharlieFRuan, line 104: quick note that, users currently need to set `request.model` to whatever they passed in as `cfg.trainer.policy.model.path`. This could be a local address, etc. vLLM has a config for it. See this for more https://github.com/vllm-project/vllm/issues/13257

PR #102, opened 2025-07-21
[GPU CI 1/N] Init GPU CI on Anyscale

SumanthRH:
# What does this PR do?

FIrst PR in enabling GPU CI. 

This PR adds a `SkyRL-GPU` workflow that will run GPU CI on Anyscale. We will basic unit and integration tests running on a 4xL4 instance in this workflow. Currently, the CI will run post-merge on `main`  . For pull requests, it requires manual trigger by a maintainer. (so it should not run on every commit even for maintainers, since that can be wasteful). 

In this PR, I've only ported over one of our existing gpu tests - `test_models.py` to run on CI . The plan is to gradually add more and more tests after we've made them more efficient and compatible to be run on a 4xL4 node.


.github/workflows/gpu_ci.yaml
```diff
@@ -0,0 +1,45 @@
+name: SkyRL-GPU
+
+on: 
+  push: 
+    branches: 
+      - main 
+      - gpu_ci # TESTING
```
SumanthRH: Will be removed before merging the PR.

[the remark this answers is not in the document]
SumanthRH: Yep, see: https://github.com/NovaSky-AI/SkyRL/pull/102#discussion_r2218039051

skyrl-train/ci/anyscale_gpu_ci.yaml
```diff
@@ -0,0 +1,10 @@
+name: my-job
```
tyler-griggs: nit: a more meaningful name?

skyrl-train/tests/gpu/gpu_ci/test_models.py
```diff
@@ -126,7 +126,7 @@ def cleanup(self):
         dist.destroy_process_group()
 
 
-def test_actor_model_fwd_with_sequence_parallelism():
+def test_actor_model_fwd_with_sequence_parallelism(ray_init_fixture):
```
tyler-griggs, line 129: In the future, will we move all GPU tests to take this `ray_init_fixture`? If so, do we want to run all GPU tests with `NCCL_P2P_DISABLE` and `NCCL_SHM_DISABLE`? I guess we need this for the L4s running CI, but setting these flags could miss some issues that we would otherwise catch while running tests in a non-CI environment. 

Can we set these envvars only for CI, but otherwise leave them unset when we run GPU tests manually / locally?

tyler-griggs (approved):

skyrl-train/ci/anyscale_gpu_ci.yaml
SumanthRH: done!

skyrl-train/tests/gpu/gpu_ci/test_models.py
SumanthRH: This is a good point!

One of the main purposes of the GPU CI is to catch all issues early for developers, and trying to make sure that we don't have to run E2E tests for every single PR. 

With this goal, it would indeed be bad if there's some issue we couldn't find because of the flags, and the only way to see it is if you run an E2E test. 

I've now guarded the flags with the standard `CI` env var. If there's a corner case where things break in CI but work locally due to this, we should be able to replicate CI setup with this `CI=true` flag.

SumanthRH: GPU tests passing in CI and the P2P env vars are disabled: 

image

PR #103, opened 2025-07-21
Make ray dependency lower-bounded

bthecohen:
We are trying to run SkyRL on a managed ray cluster running version 2.46.0. Making this dependency a lower bound should allow for this.

### DIFF
--- a/skyrl-train/pyproject.toml
+++ b/skyrl-train/pyproject.toml
@@ -29,7 +29,7 @@ dependencies = [
     "accelerate",
     "torchdata",
     "omegaconf",
-    "ray==2.44.0",
+    "ray>=2.44.0",
     "peft",
     "debugpy==1.8.0",
     "hf_transfer",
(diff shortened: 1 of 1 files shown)


SumanthRH:
Hi @bthecohen , thank you for the PR. 

I believe this PR doesn't fix the real issue of people wanting to run SkyRL on existing ray clusters with a different ray version. 
By default, adding the >= and then re-running `uv lock` doesn't update the lockfile. So in practice, users would  delete the lockfile and then re-run. However, this will grab the latest ray version - which is 2.48 at the moment. Since your ray cluster uses 2.46, this would again cause a conflict. 

I think the real solution for users with a different ray version is to simply do:

```bash
uv run ..... --with ray==2.46 skyrl_train.entrypoints.main_base ....
```

This is a manual override for the version in `pyproject.toml`, and they will be able to grab the exact version they need.
For making this easier, we will also document how to override ray version.

Does this sound reasonable?

SumanthRH:
Hi @bthecohen , I've made #105 PR to document how to override ray version. In the short term, we will have only one ray version for which SkyRL will be tested. When we have more bandwidth, we will try to have tests against different ray versions to ensure a better experience.

I hope this works

bthecohen:
Makes sense, thanks!

PR #106, opened 2025-07-24
[dependencies] Upgrade ray to 2.48.0

erictang000:
# What does this PR do
Upgrades ray to 2.48.0, which allows us to remove the pip install vllm in the Dockerfile as a fallback for when uv + vllm does not resolve dependencies with the vllm + ray backend correctly.

We leave the previous Dockerfile in `docker/Dockerfile.ray244` for backwards compatibility

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM anyscale/ray:2.44.0-slim-py312-cu128
+FROM anyscale/ray:2.48.0-slim-py312-cu128
 
 RUN sudo apt-get update -y && sudo apt-get install -y wget kmod libxml2 build-essential libnuma-dev
 
@@ -17,8 +17,3 @@ RUN sudo apt-get update \
 RUN sudo apt update && sudo apt install --fix-broken && sudo apt install -y default-jre-headless openjdk-8-jdk \
     && sudo apt-get clean \
     && sudo rm -rf /var/lib/apt/lists/*
-
-# NOTE: vllm installation in base environment is needed for uv + vLLM to work
-RUN pip install vllm==0.9.2 --extra-index-url https://download.pytorch.org/whl/cu128 \
-    && pip install ray==2.44.0 \
-    && rm -rf ~/.cache/pip
--- a/skyrl-train/docs/getting-started/installation.rst
+++ b/skyrl-train/docs/getting-started/installation.rst
@@ -128,13 +128,27 @@ Running on an existing Ray cluster
 
 For running on an existing Ray cluster, you need to first make sure that the python version used is 3.12. 
 
-We recommend using Ray version 2.48.0 and above for the best experience. However, SkyRL-Train is compatible with any Ray version 2.44.0 and above (except 2.47.0 and 2.47.1 -- which we do not recommend due to an issue in the uv + Ray integration). Since we use a uv lockfile to pin dependencies, the best way to run SkyRL-Train on a custom Ray version (say 2.46.0) would be to override the version at runtime with the ``--with`` flag. For example, to run with Ray 2.46.0, you can do:
+Ray >= 2.48.0
+~~~~~~~~~~~~~
+
+We recommend using Ray version 2.48.0 and above for the best experience. In this case, you can simply use the ``uv run`` command to get training started.
+
+.. code-block:: bash
+
+    uv run -m skyrl_train.entrypoints.main_base ...
+
+Ray < 2.48.0
+~~~~~~~~~~~~
+SkyRL-Train is compatible with any Ray version 2.44.0 and above (except 2.47.0 and 2.47.1 -- which we do not recommend due to an issue in the uv + Ray integration). 
+Since we use a uv lockfile to pin dependencies, the best way to run SkyRL-Train on a custom Ray version (say 2.46.0) would be to override the version at runtime with the ``--with`` flag. 
+For example, to run with Ray 2.46.0, you can do:
 
 .. code-block:: bash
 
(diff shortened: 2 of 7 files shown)


skyrl-train/docs/getting-started/installation.rst
```diff
@@ -8,21 +8,21 @@ Requirements
 
 We use `uv <https://docs.astral.sh/uv/>`_ to manage dependencies. We also make use of the `uv` and `ray` integration to manage dependencies for ray workers. 
 
-If you're running on an existing Ray cluster, make sure to use Ray 2.44.0 and Python 3.12.
+If you're running on an existing Ray cluster, make sure to use Ray 2.48.0 and Python 3.12.
```
SumanthRH: This should actually link to the appropriate section on "Running on an existing Ray cluster" . Can you fix this here?

SumanthRH (approved):
Left a nit

skyrl-train/docs/getting-started/installation.rst
erictang000: updated!

Also updated the running on an existing ray cluster section since we no longer include the `pip install vllm` in the dockerfile, and if we're saying to use ray>=2.44.0, then that should be added somewhere

docker/Dockerfile
```diff
@@ -17,8 +17,3 @@ RUN sudo apt-get update \
 RUN sudo apt update && sudo apt install --fix-broken && sudo apt install -y default-jre-headless openjdk-8-jdk \
     && sudo apt-get clean \
     && sudo rm -rf /var/lib/apt/lists/*
-
-# NOTE: vllm installation in base environment is needed for uv + vLLM to work
-RUN pip install vllm==0.9.2 --extra-index-url https://download.pytorch.org/whl/cu128 \
-    && pip install ray==2.44.0 \
-    && rm -rf ~/.cache/pip
```
SumanthRH, line 24: One of the friction points with a removal right now is that people with existing ray clusters with ray < 2.48 will be affected - it would make skyrl harder to run and we are pretty much forcing them to upgrade. Could we provide another Dockerfile  "Dockerfile.ray244" as an example Dockerfile for older ray version, and include the vllm installation there?

skyrl-train/docs/getting-started/installation.rst
```diff
@@ -122,15 +121,25 @@ Finally, you can initialize a Ray cluster using the following command (for singl
...
 .. code-block:: bash
 
     uv run .... --with ray==2.46.0 -m skyrl_train.entrypoints.main_base ...
 
+For ray versions >= 2.44.0 but < 2.48.0, you need to install vllm in the base pip environment, and then re-install ray to your desired version to ensure that the uv + Ray integration works as expected.
+
+.. code-block:: bash
+
+    pip install vllm==0.9.2 --extra-index-url https://download.pytorch.org/whl/cu128
+    pip install ray==2.46.0
```
SumanthRH: I have missed some updates to the Dockerfile in `main`. The current docker image I have at `sumanthrh/skyrl-train-ray-2.44.0-py3.12-cu12.8:latest` has the following dependencies: 

```suggestion
.. code-block:: bash

    pip install vllm==0.9.2 --extra-index-url https://download.pytorch.org/whl/cu128
    pip install ray==2.46.0 omegaconf==2.3.0 loguru==0.7.3 jaxtyping==0.3.2 pyarrow==20.0.0
```

It would be good to confirm things work with this as well.

docker/Dockerfile
SumanthRH: The "Dockerfile.ray244" file should have the following :

```bash
# NOTE: vllm installation in base environment is needed for uv + vLLM to work
RUN pip install vllm==0.9.2 --extra-index-url https://download.pytorch.org/whl/cu128 \
    && pip install ray==2.44.0 omegaconf==2.3.0 loguru==0.7.3 \
    && rm -rf ~/.cache/pip
RUN pip install pyarrow==20.0.0 jaxtyping==0.3.2 && rm -rf ~/.cache/pip
```

erictang000: added

skyrl-train/docs/getting-started/installation.rst
SumanthRH, line 1: Can the doc reference the new Dockerfile.ray244 file in "Running on an existing Ray cluster"? Also we should have two sections for ray 2.48 and above and the ray less than 2.48 so that it's a bit more clear. As in for ray greater than 2.48 it should be like no change and just the uv command

erictang000: added

skyrl-train/docs/getting-started/installation.rst
```diff
@@ -122,15 +121,39 @@ Finally, you can initialize a Ray cluster using the following command (for singl
...
+For running on an existing Ray cluster, you need to first make sure that the python version used is 3.12. 
+
+Ray >= 2.48.0
+~~~~~~~~~~~~~
+
+We recommend using Ray version 2.48.0 and above for the best experience. In this case, you can simply use the ``uv run`` command to get training started.
+
+.. code-block:: bash
+
+    uv run -m skyrl_train.entrypoints.main_base ...
```
SumanthRH: Maybe we should future proof this? My original intention was to say that for ray>=2.48, all they need is to use `uv run --with`:

```
uv run .... --with ray==2.xx.yy
```

But for ray less than that, you need the vLLM stuff

Wdyt?

erictang000: makes sense, updated

SumanthRH (approved):
Thanks! WE can revisit installation docs once there's a release after 2.48

PR #108, opened 2025-07-24
Add check for whether p2p access is supported - allows code to run on L4/L40S after #73 upgrade to cuda 12.8

erictang000:
# Overview
After #73, the main code path no longer runs on GPUs without P2P support (potentially due to cuda 12.8 upgrade?) - an error would be thrown like

```bash
torch.distributed.DistBackendError: NCCL error in: /pytorch/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp:3353, unhandled cuda error (run with NCCL_DEBUG=INFO for details), NCCL version 2.26.2
ncclUnhandledCudaError: Call to CUDA function failed.
Last error:
Cuda failure 217 'peer access is not supported between these two devices'
```

This PR adds a check for whether peer access is supported (using torch/cuda) between all GPUs on a node to the ray initialization, and sets relevant NCCL env vars to allow the code to run on these machine types.

```python
if not peer_access_supported():
        logger.info("Peer access is not supported, disabling P2P and SHM")
        env_vars["NCCL_P2P_DISABLE"] = "1"
        env_vars["NCCL_SHM_DISABLE"] = "1"
```

Example running on L40S:
image

### DIFF
--- a/skyrl-train/skyrl_train/utils/utils.py
+++ b/skyrl-train/skyrl_train/utils/utils.py
@@ -267,7 +267,7 @@ def initialize_ray(cfg: DictConfig):
             )
             env_vars["VLLM_USE_V1"] = "1"
             env_vars["VLLM_ENABLE_V1_MULTIPROCESSING"] = "0"
-        
+
     if not peer_access_supported():
         logger.info("Peer access is not supported, disabling P2P and SHM")
         env_vars["NCCL_P2P_DISABLE"] = "1"
@@ -341,14 +341,15 @@ def print_mem(tag: str, mem: dict):
         f"Total: {format_gib(mem['total'])}"
     )
 
+
 def peer_access_supported():
     if not torch.cuda.is_available():
         return False
-    
+
     device_count = torch.cuda.device_count()
     if device_count < 2:
         return False
-    
+
     # Check P2P access between all GPU pairs
     for i in range(device_count):
         for j in range(device_count):
@@ -357,5 +358,5 @@ def peer_access_supported():
                 can_access = torch.cuda.can_device_access_peer(i, j)
                 if not can_access:
                     return False
-    
+
     return True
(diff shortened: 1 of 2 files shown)


skyrl-train/tests/gpu/gpu_ci/conftest.py
```diff
@@ -3,23 +3,39 @@
...
 from functools import lru_cache
+import torch
 
 
 @lru_cache(5)
 def log_once(msg):
     logger.info(msg)
     return None
 
+def peer_access_supported():
```
erictang000: duplicated code here since I didn't want to introduce an import to `skyrl-train` in the conftest.py file

pcmoritz: I'm curious why not, is that a problem?

skyrl-train/skyrl_train/utils/utils.py
```diff
@@ -268,6 +268,11 @@ def initialize_ray(cfg: DictConfig):
             env_vars["VLLM_USE_V1"] = "1"
             env_vars["VLLM_ENABLE_V1_MULTIPROCESSING"] = "0"
 
+    if not peer_access_supported():
```
SumanthRH: This will disable P2P all the time for multi node. We should discuss.

SumanthRH (changes requested):

skyrl-train/skyrl_train/utils/utils.py
SumanthRH: I think it would be best if `peer_acccess_supported` has it's own self contained ray session - schedules on a gpu worker node (use num gpus per node I guess),  runs the check, and shuts down the ray session. We use this info to initialize our new ray session  for the training run

skyrl-train/tests/gpu/gpu_ci/conftest.py
erictang000: hmm, previously had some dependency issues when this file was more bloated, but i think you're right it should be ok now - updated to just import the same util function

SumanthRH (approved):
Thanks! Could you add a screenshot for logs on L4/L40s for reference in this PR?

pcmoritz (approved):

erictang000:
> Thanks! Could you add a screenshot for logs on L4/L40s for reference in this PR?

Added to PR description, posting in thread here too: 

image

PR #109, opened 2025-07-25
[Trainer] Support per-token rewards in trainer

SumanthRH:
# What does this PR do?

Supports per-token rewards from the generator. Currently, the postprocess_generator_output function is hardcoded to only support per-sequence rewards, but we should be able to easily support per-token rewards. 

I've also added a test to make sure that the postprocessing function is tested. This is placed in a new file for ease of use.


tyler-griggs (approved):
Thanks!

PR #112, opened 2025-07-26
fix issue with #108 that broke gpu ci

erictang000:
missed an argument in `gpu_ci/conftest.py` for `peer_access_supported()` - fix for gpu ci to run 

Passing now with update:
image

And main runs (and disables p2p access) correctly:
image

### DIFF
--- a/skyrl-train/ci/gpu_ci_run.sh
+++ b/skyrl-train/ci/gpu_ci_run.sh
@@ -1,2 +1,3 @@
 #!/usr/bin/env bash
+export CI=true
 uv run --directory . --isolated --extra dev --extra vllm pytest -s tests/gpu/gpu_ci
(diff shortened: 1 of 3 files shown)


skyrl-train/ci/gpu_ci_run.sh
```diff
@@ -1,3 +1,2 @@
 #!/usr/bin/env bash
-export CI=true
```
SumanthRH: This is meant to be set. Revert?

 https://docs.pytest.org/en/stable/explanation/ci.html

skyrl-train/tests/gpu/gpu_ci/conftest.py
```diff
@@ -16,7 +16,8 @@ def ray_init_fixture():
     if ray.is_initialized():
         ray.shutdown()
     env_vars = {}
-    if not peer_access_supported():
+    # hard code to 4 gpus for CI environment for now
+    if not peer_access_supported(max_num_gpus_per_node=4):
```
SumanthRH: Can just use 2? I don't think we should be too precise in checking every pair in any case

erictang000: sure

SumanthRH (approved):
Thanks

PR #115, opened 2025-07-26
[Trainer] Support registering custom advantage estimators

tyler-griggs:
## What does this PR do? 

Adds an `AdvantageEstimatorRegistry` to support custom advantage estimation methods without modifying the skyrl-train package. 

Added `examples/algorithm/custom_advantage_estimator` folder to give quick example of how to register a custom adv est function.

## Tests
Adding cpu test to ensure registration works.

### DIFF
--- a/skyrl-train/examples/algorithm/custom_advantage_estimator/main_custom_adv_est.py
+++ b/skyrl-train/examples/algorithm/custom_advantage_estimator/main_custom_adv_est.py
@@ -1,0 +1,58 @@
+"""
+uv run --isolated --extra vllm -m examples.algorithm.custom_advantage_estimator.main_custom_adv_est
+"""
+
+import ray
+import hydra
+import torch
+import numpy as np
+from omegaconf import DictConfig
+from skyrl_train.utils import initialize_ray
+from skyrl_train.entrypoints.main_base import BasePPOExp, config_dir, validate_cfg
+
+
+# Example of custom advantage estimator: "simple_baseline"
+def compute_simple_baseline_advantage(
+    token_level_rewards: torch.Tensor, response_mask: torch.Tensor, index: np.ndarray, **kwargs
+):
+    """
+    A simple custom advantage estimator that uses response-level rewards
+    and computes advantages against a simple baseline.
+
+    This is just an example - replace with your own logic.
+    """
+    with torch.no_grad():
+        response_rewards = (token_level_rewards * response_mask).sum(dim=-1, keepdim=True)
+
+        # Simple baseline: use the mean reward across the batch
+        baseline = response_rewards.mean()
+        advantages = (response_rewards - baseline) * response_mask
+        returns = advantages.clone()
+
+        return advantages, returns
+
+
+@ray.remote(num_cpus=1)
+def skyrl_entrypoint(cfg: DictConfig):
+    # Import inside the remote function so it's available in the Ray worker
+    from skyrl_train.utils.ppo_utils import AdvantageEstimatorRegistry
+
+    # Register the custom advantage estimator
+    AdvantageEstimatorRegistry.register("simple_baseline", compute_simple_baseline_advantage)
+
+    # make sure that the training loop is not run on the head node.
+    exp = BasePPOExp(cfg)
+    exp.run()
+
+
+@hydra.main(config_path=config_dir, config_name="ppo_base_config", version_base=None)
+def main(cfg: DictConfig) -> None:
+    # validate the arguments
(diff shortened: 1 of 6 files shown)


skyrl-train/examples/algorithm/custom_advantage_estimator/main_custom_adv_est.py
```diff
@@ -0,0 +1,58 @@
...
+        advantages = (response_rewards - baseline) * response_mask
+        returns = advantages.clone()
+
+        return advantages, returns
+
+
+@ray.remote(num_cpus=1)
+def skyrl_entrypoint(cfg: DictConfig):
+    # Import inside the remote function so it's available in the Ray worker
+    from skyrl_train.utils.ppo_utils import AdvantageEstimatorRegistry
```
tyler-griggs: Note: this was the cleanest approach I could find to get the registration to work on the Ray worker. Previously I just had decorated the method at the top of this `main` file, but then the registry is not picked up by the Ray workers.

skyrl-train/skyrl_train/utils/utils.py
```diff
@@ -140,10 +140,6 @@ def validate_cfg(cfg: DictConfig):
         cfg.trainer.sequence_parallel_backend == "ulysses"
     ), f"only ulysses is supported as of now, got {cfg.trainer.sequence_parallel_backend}"
 
-    assert cfg.trainer.algorithm.advantage_estimator in (
```
tyler-griggs, line 143: Note: this is an unfortunate deletion. The tricky part is that the adv est registration has to happen on a Ray worker (or passed into the ray workers), but our config validation currently happens _before_ calling into the head ray process, so we can't actually check the configured `advantage_estimator` against the contents of the registry.

This will be a problem with any registries we later add, so I opened an issue on this: #116 . One option is to move config validation to inside the ray head worker.

erictang000: hmm yeah, I guess we don't run any validation for env names for the same reason... this makes sense for now.

skyrl-train/examples/algorithm/custom_advantage_estimator/main_custom_adv_est.py
erictang000: does the import need to happen inside the `skyrl_entrypoint` as well? or just the register call

skyrl-train/examples/algorithm/custom_advantage_estimator/main_custom_adv_est.py
```diff
@@ -0,0 +1,58 @@
+"""
```
erictang000, line 1: do you think we should add this example to a docs page? Maybe it's time for us to initialize a docs section for "Algorithms", and this small example could be the "Custom algorithms" page under that section. 

We can add explainers of the relevant configs to tune for each algorithm in each of the doc pages (i.e. for PPO, GRPO, DAPO, GSPO, etc..) - wdyt?

At the very least should update the https://github.com/NovaSky-AI/SkyRL/blob/main/skyrl-train/docs/configuration/config.rst config page to mention the AdvEstimator registry

erictang000 (commented):
makes sense, left a question about docs to include with this

skyrl-train/skyrl_train/utils/utils.py
tyler-griggs: Yeahh. Updated the issue (#116) with the env name validation point, too.

skyrl-train/examples/algorithm/custom_advantage_estimator/main_custom_adv_est.py
tyler-griggs: Oh no it doesn't. Moving it out.

skyrl-train/examples/algorithm/custom_advantage_estimator/main_custom_adv_est.py
tyler-griggs: Yes good call. Let me take this as a todo? I am working on one more DAPO feature and on GSPO. I will wrap these up in the next couple days and it will be a good time to introduce an "Algorithms" section.

tyler-griggs: Created issue at #119

tyler-griggs:
Thanks! Resolved comments. I want to get one more DAPO feature in, then will write an Algorithms doc.

erictang000 (approved):
makes sense, thanks!

Issue #97, opened 2025-07-17
Loading saved HF weights errors `Cannot find model module.`


tyler-griggs:
## What's the issue?

After saving HF model weights on intermediate steps, loading them back in a new training run (or with `vllm serve`) gives the error:
```
ValueError: Cannot find model module. 'FSDPQwen2ForCausalLM' is not a registered model in the Transformers library (only relevant if the model is meant to be in Transformers) and 'AutoModel' is not present in the model config's 'auto_map' (relevant if the model is custom).
```

The hf model_config, the architecture is:
```
"architectures": [
  "FSDPQwen2ForCausalLM"
],
```

So it seems like the HF model is still FSDP-wrapped (or labeled as FSDP-wrapped).

Simply overriding the architecture `FSDPQwen2ForCausalLM` to `Qwen2ForCausalLM` seems to work, so it might just be incorrect architecture labeling. 

## Script used
```
uv run --isolated --frozen --extra vllm -m skyrl_train.entrypoints.main_base \
  trainer.algorithm.advantage_estimator="grpo" \
  data.train_data="['${DATA_DIR}/train.parquet']" \
  data.val_data="['${DATA_DIR}/validation.parquet']" \
  trainer.policy.model.path="Qwen/Qwen2.5-3B-Instruct" \
  trainer.placement.colocate_all=true \
  trainer.strategy=fsdp2 \
  trainer.policy.optimizer_config.max_grad_norm=0.5 \
  trainer.placement.policy_num_gpus_per_node=4 \
  trainer.placement.ref_num_gpus_per_node=4 \
  generator.model_dtype=bfloat16 \
  generator.num_inference_engines=4 \
  generator.inference_engine_tensor_parallel_size=1 \
  trainer.epochs=2 \
  trainer.update_epochs_per_batch=1 \
  trainer.train_batch_size=256 \
  trainer.policy_mini_batch_size=256 \
  trainer.micro_forward_batch_size_per_gpu=4 \
  trainer.micro_train_batch_size_per_gpu=4 \
  trainer.max_prompt_length=1024 \
  generator.max_input_length=4096 \
  generator.sampling_params.max_generate_length=500 \
  trainer.policy.optimizer_config.lr=1e-6 \
  trainer.policy.optimizer_config.num_warmup_steps=10 \
  trainer.algorithm.use_kl_loss=true \
  trainer.ckpt_interval=100000 \
  generator.backend=vllm \
  generator.run_engines_locally=true \
[...]

PR #118, opened 2025-07-28
[bug] Loading saved HF weights errors

erictang000:
Addresses #97

### DIFF
--- a/skyrl-train/tests/gpu/test_save_load_model.py
+++ b/skyrl-train/tests/gpu/test_save_load_model.py
@@ -13,8 +13,14 @@ import os
 import shutil
 import tempfile
 from omegaconf import DictConfig
+import json
 
-from tests.gpu.utils import init_worker_with_type, make_dummy_experience, get_model_logits_from_actor
+from tests.gpu.utils import (
+    init_worker_with_type,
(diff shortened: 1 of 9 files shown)


skyrl-train/tests/gpu/test_save_load_model.py
```diff
@@ -43,7 +50,7 @@ def get_test_actor_config(strategy: str) -> DictConfig:
         "fsdp2",
     ],
 )
-def test_save_load_hf_model(strategy):
+def test_save_load_hf_model(ray_init_fixture, strategy):
```
SumanthRH, line 53: Nit:

```python
try:
    ...
finally: 
    ray.shutdown()
```

in this test is no longer needed with the fixture?

SumanthRH (approved):
Thanks! Could you add a screenshot for the save model test passing for all strategies?

erictang000:
image

skyrl-train/tests/gpu/test_save_load_model.py
erictang000: removed the ray.shutdown() - kept the try-finally for the cleaning up of tmp directories

PR #120, opened 2025-07-29
[Trainer] GSPO support

bthecohen:
This PR adds support for [Group Sequence Policy Optimization (GSPO)](https://arxiv.org/abs/2507.18071), the hotness du jour from Alibaba Qwen. The implementation in this PR is loosely based on [this one](https://github.com/huggingface/trl/pull/3775) from TRL. It adds an `importance_sampling_level` config option which can be `token` (PPO/GRPO) or `sequence` (GSPO).

I ran a short/small GSM8k run with Qwen2.5-0.5B and the loss curves look okay:
image

However, I had to hack a few things to get this to run on Datadog's cloud infra (including changing some dependency versions) so I'd encourage one of the maintainers to reproduce these results locally before merging.

### DIFF
--- a/skyrl-train/docs/configuration/config.rst
+++ b/skyrl-train/docs/configuration/config.rst
@@ -291,7 +291,7 @@ Algorithm Configuration
       advantage_batch_normalize: false
       value_head_prefix: "value_head"
       policy_loss_type: "regular" # "regular", "dual_clip", "gspo", or customizable with PolicyLossRegistry
-      loss_reduction: "token_mean" # "token_mean", "sequence_mean" (ignored for GSPO)
+      loss_reduction: "token_mean" # "token_mean", "sequence_mean"
 
       # GAE parameters
       lambd: 1.0
@@ -319,7 +319,7 @@ Algorithm Configuration
 
   - ``regular``: Vanilla PPO loss with token-level importance sampling
   - ``dual_clip``: Dual clip PPO loss proposed in `this paper <https://arxiv.org/pdf/1912.09729>`_
-  - ``gspo``: Group Sequence Policy Optimization with sequence-level importance sampling for improved training stability (`arXiv:2507.18071 <https://arxiv.org/abs/2507.18071>`_)
+  - ``gspo``: `Group Sequence Policy Optimization <https://arxiv.org/abs/2507.18071>`_ with sequence-level importance sampling for improved training stability. Implements "GSPO-token" variant from the paper.
   - Custom policy losses can be registered with the ``PolicyLossRegistry``
 
 - ``algorithm.loss_reduction``: Type of loss reduction to use. Options are ``token_mean`` and ``sequence_mean``. ``token_mean`` matches token-level loss introduced by `DAPO <https://dapo-sia.github.io/>`_. ``sequence_mean`` computes per-sequence avg token loss, then averages over the batch.
--- a/skyrl-train/docs/configuration/config.rst
+++ b/skyrl-train/docs/configuration/config.rst
@@ -322,7 +322,7 @@ Algorithm Configuration
   - ``gspo``: Group Sequence Policy Optimization with sequence-level importance sampling for improved training stability (`arXiv:2507.18071 <https://arxiv.org/abs/2507.18071>`_)
   - Custom policy losses can be registered with the ``PolicyLossRegistry``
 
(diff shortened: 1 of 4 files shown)


bthecohen:
One thing a bit unexpected about this run was that the clip ratio was actually _lower_ for GSPO, which is the opposite of what the paper says:

image

A few hypotheses about why this could be, in roughly descending order of likelihood IMO:

- I didn't run the training long enough (in the paper, they report the clip ratios averaged over a full post-training run). It does seem like towards the end the ratio was going up for GSPO
- There's something different about the model / data / hyperparams in my setup that led to the opposite behavior of what the Qwen folks saw
- There's a bug in how we're calculating the clip ratio
- There's a bug in my GSPO implementation

@tyler-griggs I'd be curious to see if you see the same thing with a bigger model / longer run

erictang000:
hi @bthecohen, thanks for the PR! I'll try testing this out and let you know what I see!

erictang000:
btw @bthecohen what `eps_clip_low`/`eps_clip_high` values did you use for the runs? It seems like the comparison in the paper is between 0.2/0.27 for GRPO and 4e-4/3e-4 for GSPO (https://x.com/ChujieZheng/status/1948933507696525392), so maybe that explains the diff in clip ratio?

bthecohen:
Ah yeah that's it. I left this as the default (so 0.2/0.2 for both runs).

bthecohen:
One additional thing to note that I realized this morning is that this implementation isn't quite correct when you have token-level advantages - for that case I think we should be using "GSPO-token" (equations 13 and 14 in the paper). As long as the per-token advantages are all the same within a sequence, they are equivalent but given that SkyRL now supports returning advantages as a list per-token, I think we'll either want to:

- add explicit support for gspo-token with the stop-gradient correction
- give a warning that the current implementation shouldn't be used with per-token advantages

I didn't notice this at first because the TRL implementation doesn't support it. WDYT, is it worth blocking this PR on getting that working? Given the focus on multi-turn rollouts it might be important.

erictang000:
yeah I think adding explicit support for gspo-token would be good given that the regular gspo case is just a subset of it where advantages are all identical - maybe we could do something similar to the structure of [this verl PR](https://github.com/volcengine/verl/pull/2775), and add the GSPO loss function as a separate policy loss function to avoid bloating the loss logic as we add more algorithm support?

I'll draft a quick PR to update the logic to make adding a new policy loss cleaner

bthecohen:
I agree having modular policy loss functions is a good idea in general, but for the sake of argument we still might want the various PPO  flavors to share the same class...I'm just thinking people want to mix and match the GSPO stuff with the DAPO loss reduction, which could lead to some code duplication. An alternative approach would to be keep one class for PPO/GRPO/GSPO, but factor out some of the subcomputations into private methods to make the `forward` less bloated.

I don't feel super strongly about this – if you'd prefer it with separate classes I'm happy to rebase off of your PR. Let me know, I'm also happy to help with the refactor if I can. And revise this to use GSPO-token.

On a separate note, from that PR we have the lead author of GSPO saying that it might only matter for MoE models, or only in the SFT -> RL setting:

image

bthecohen:
Actually, I guess verl solves this by just factoring out the loss reduction into a helper function. Although, for GSPO maybe only sequence-mean reduction makes sense?

erictang000:
hey @bthecohen finished updating the logic for adding custom/new policy losses in #126, could you take a look and let me know if it makes sense to you?

once that's merged then gspo/gspo-token would just be a new policy loss implementation here (and we could reuse the loss reduction function with sequence mean hard coded)

erictang000:
hey @bthecohen, now that #126 is merged, can we rebase this PR and add GSPO token? 

happy to help/run any longer test runs for it as well!

bthecohen:
Yep, working on it now! Given the refactor, and the fact that I'm changing the algorithm, I may decide to close this PR and reopen rather than rebasing. But I'll see if I can keep it in here to keep the discussion thread linear.

bthecohen:
@erictang000 I've updated this to use the new refactoring and to use GSPO token. New training results on my end:

image

The GSPO run here is using the clipping values from the paper. The GRPO run is still using the default hyperparams from the SkyRL example, so not exactly as in the paper.

bthecohen:
The clip ratio is now much higher for GSPO, as expected 
image

bthecohen:
Also, the eval curves look extremely close between GSPO and GRPO
image

bthecohen:
Update: ran a longer GSPO run, looks quite stable 
image

PR #121, opened 2025-07-29
Fix MLFlow logging

bthecohen:
This is a small change to make the MLFlow integration work. Currently this fails with a Pandas error when trying to flatten an Omega dict; we need to convert to a regular Python dictionary.

Can confirm this works on our MLFlow setup:
image

### DIFF
--- a/skyrl-train/docs/configuration/config.rst
+++ b/skyrl-train/docs/configuration/config.rst
@@ -121,7 +121,7 @@ Logging and Debugging Configuration
     dump_data_batch: false
     dump_eval_results: true
 
-- ``logger``: Logger to use. Currently, we support ``wandb`` and ``console``. ``console`` will simply log metrics to the console. 
+- ``logger``: Logger to use. Currently, we support ``wandb``, ``mlflow``, and ``console``. ``console`` will simply log metrics to the console. 
 - ``project_name``: Name of the project in WandB.
 - ``run_name``: Name of the run in WandB.
 - ``dump_data_batch``: Whether to dump the data batch to a file. This is useful for debugging. When ``true``, the data batch will be dumped to a file in the ``export_path`` directory. The training batch at global step ``N`` is saved to ``self.cfg.trainer.export_path / "dumped_data" / global_step_N_training_input``
--- a/skyrl-train/skyrl_train/utils/tracking.py
+++ b/skyrl-train/skyrl_train/utils/tracking.py
@@ -21,6 +21,7 @@ from functools import partial
 from pathlib import Path
 from typing import Any, Dict, List, Union
 from loguru import logger
+from omegaconf import DictConfig, OmegaConf
 import pprint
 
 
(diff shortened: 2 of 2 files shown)


erictang000 (approved):
lgtm thanks for the change!

skyrl-train/docs/configuration/config.rst
```diff
@@ -121,7 +121,7 @@ Logging and Debugging Configuration
     dump_data_batch: false
     dump_eval_results: true
 
-- ``logger``: Logger to use. Currently, we support ``wandb`` and ``console``. ``console`` will simply log metrics to the console. 
+- ``logger``: Logger to use. Currently, we support ``wandb``, ``mlflow``, and ``console``. ``console`` will simply log metrics to the console.
```
SumanthRH, line 124: For now, could you add the env vars needed for MLFlow right here ? We can have a better guide on MLFlow later. I'm guessing it's `MLFLOW_TRACKING_TOKEN`  and `MLFLOW_TRACKING_URI`?

SumanthRH (commented):
Thanks! 

How about we add `MLFLOW_TRACKING_TOKEN`  and `MLFLOW_TRACKING_URI` to the set of exported env vars at ray init? In the future, we probably want to switch to always using a .env file passed to uv , but this would be nice to have so that the quickstart is easy. 

here's an example for what we did with wandb: 

https://github.com/NovaSky-AI/SkyRL/blob/321b8fa90cc23bc3e194f7c6f91095023d65f039/skyrl-train/skyrl_train/utils/utils.py#L284-L288

skyrl-train/docs/configuration/config.rst
SumanthRH: oops just saw that the PR is merged. we can take this up!

etnbrd:
Hey @SumanthRH 👋 
I could submit a PR either to add the `MLFLOW_TRACKING_TOKEN` and `MLFLOW_TRACKING_URI`, or to read them from an env file.
But for the latter, I'd like to know what you mean by standardize on env files?
https://github.com/NovaSky-AI/SkyRL/blob/321b8fa90cc23bc3e194f7c6f91095023d65f039/skyrl-train/skyrl_train/utils/utils.py#L284

Is it as you said here?
> In the future, we probably want to switch to always using a .env file passed to uv

Would it be an `.env` file passed with `--env-file` as documented here: https://docs.astral.sh/uv/concepts/configuration-files/#env
I'm not sure to follow how these env variables would be passed to ray, though.

SumanthRH:
Hi @etnbrd . Thanks for your interest in SkyRL! 


> Would it be an .env file passed with --env-file as documented here: https://docs.astral.sh/uv/concepts/configuration-files/#env
I'm not sure to follow how these env variables would be passed to ray, though.

Yes that's what I meant! Just using `--env-file .env` with uv is enough, there's nothing special needed for ray. all ray tasks and actors will inherit the env variables in the env file because of our uv + ray integration. 

However, we'll probably have to update all our examples to use this env file. For now, maybe you can submit a PR to detect  if `MLFLOW_TRACKING_TOKEN ` and `MLFLOW_TRACKING_URI` are set on the driver, and if so propagate to all workers in `ray.init` , similar to `WANDB_API_KEY`.

etnbrd:
Here's the PR: https://github.com/NovaSky-AI/SkyRL/pull/135

I removed the TODO comment because, in my understanding, relying on env files wouldn't make this unnecessary anymore, but I'd be happy to put it back in the PR if I'm misunderstanding something.

SumanthRH:
Thanks @etnbrd , will leave comments in the PR!

Issue #116, opened 2025-07-26
Improve passing registries into the Ray workers


tyler-griggs:
It is currently challenging to pass registries into Ray workers _and_ validate their configuration. This is because registration must happen on a Ray worker for it to be usable during training, but config validation happens before calling into the Ray worker.

One option is to move config validation into the entrypoint of the head worker (e.g., `skyrl_entrypoint`).

erictang000:
partially closed by https://github.com/NovaSky-AI/SkyRL/pull/126 for algorithm support, but we should improve this for the SkyGym envs in a similar way (back the gym registry with named ray actor?)

PR #126, opened 2025-07-30
[Trainer/Algorithm] Support registering custom policy loss functions + refactor adv estimator registry to allow registration outside ray workers

erictang000:
# Overview
- Adds support for registering custom policy loss functions, similar to #115, 
- Refactors the policy loss to be a function in `ppo_utils.py` instead of a (`nn.Module` in `worker.py`)
- Introduces a breaking change in renaming `trainer.algorithm.ppo_loss_type` to `trainer.algorithm.policy_loss_type`
- Addresses Issue #116 by creating a new `BaseFunctionRegistry` class that uses a [named actor](https://docs.ray.io/en/latest/ray-core/actors/named-actors.html) to support the following pattern:

```python
# Example of custom policy loss: "simple_baseline"
def compute_simple_baseline_policy_loss(
    log_probs: torch.Tensor,
    ...
):
    return torch.randn(1, device=log_probs.device), 0.0

# Register the custom policy loss - outside of the ray worker
PolicyLossRegistry.register("simple_baseline", compute_simple_baseline_policy_loss)


@ray.remote(num_cpus=1)
def skyrl_entrypoint(cfg: DictConfig):
    exp = BasePPOExp(cfg)
    exp.run()


@hydra.main(config_path=config_dir, config_name="ppo_base_config", version_base=None)
def main(cfg: DictConfig) -> None:
    # validate the arguments
    validate_cfg(cfg)

    initialize_ray(cfg)

    ray.get(skyrl_entrypoint.remote(cfg))
```
  this change was necessary for `PolicyLossRegistry` to be accessible, since the worker `actor_loss_fn` attribute is set in `init_model` within the `worker` actor, which is a ray actor created from within the skyrl_entrypoint ray task (and registering within the entrypoint wouldn't propagate down another layer).
   - updates AdvantageEstimatorRegistry to extend the same `BaseFunctionRegistry` class

Example runs:
Custom advantage (mean of reward)
image

Custom policy loss (reinforce - just (-logprobs * advantages).mean())
image

### DIFF
--- a/skyrl-train/skyrl_train/utils/ppo_utils.py
+++ b/skyrl-train/skyrl_train/utils/ppo_utils.py
@@ -446,6 +446,8 @@ def register_policy_loss(name: Union[str, PolicyLossType]):
 
 def sync_registries():
     """Sync the registries with the ray actor once ray is initialized"""
+    if not ray.is_initialized():
+        raise ValueError("Ray is not initialized, cannot sync registries")
     PolicyLossRegistry.sync_with_actor()
     AdvantageEstimatorRegistry.sync_with_actor()
     logger.info("Synced registries to ray actor")
--- a/skyrl-train/docs/algorithms/custom_algorithms.rst
+++ b/skyrl-train/docs/algorithms/custom_algorithms.rst
@@ -1,8 +1,8 @@
 Implementing Custom Algorithms
-=============================
+==============================
 
 SkyRL-Train provides a registry system for easily implementing custom algorithms (advantage estimators, policy loss) without modifying the core codebase. 
-Full documentation for the registry system can be found in the :doc:`registry API <../api/registry>`.
+The API for the registry system can be found in the :doc:`registry API <../api/registry>`.
 Example scripts of using the registry can be found in at :code_link:`examples/algorithm/`.
 
 Registering a Custom Advantage Estimator


skyrl-train/skyrl_train/workers/deepspeed/deepspeed_worker.py
```diff
@@ -93,13 +95,8 @@ def init_model(self, model_id_or_path):
...
 
         # set ppo loss function
-        self.actor_loss_fn = PolicyLoss(
-            self.cfg.trainer.algorithm.eps_clip_low,
-            self.cfg.trainer.algorithm.eps_clip_high,
-            self.cfg.trainer.algorithm.clip_ratio_c,
-            loss_type=self.cfg.trainer.algorithm.ppo_loss_type,
-            loss_reduction=self.cfg.trainer.algorithm.loss_reduction,
-        )
+        policy_loss_func = PolicyLossRegistry.get(self.cfg.trainer.algorithm.policy_loss_type)
```
erictang000, line 98: this is broken right now because we have 2 levels of ray worker nesting here :( so registering in the entrypoint doesn't propagate to here...

bthecohen: The simplest way to fix this is probably to make the registry a [named actor](https://docs.ray.io/en/latest/ray-core/actors/named-actors.html) and fetch it inside the training worker. Another option would be to write stuff directly to the ray [object store](https://docs.ray.io/en/latest/ray-core/objects.html).

skyrl-train/skyrl_train/utils/utils.py
```diff
@@ -290,6 +286,10 @@ def initialize_ray(cfg: DictConfig):
         env_vars["LD_LIBRARY_PATH"] = os.environ["LD_LIBRARY_PATH"]
     ray.init(runtime_env={"env_vars": env_vars})
 
+    # sync the registries with the ray actor now that ray is initialized
```
erictang000: in the BaseFunctionRegistry, there's a check for 

`if not ray.is_initialized():` to guard creation of the named actor for the registry - this is because otherwise on import of `skyrl_train.utils`, ray.init will be called without the env vars that we specify here (which require the cfg object to resolve)

Once `ray.init` is called, we then want to create a named actor for the registry and sync the local changes to that actor so it is accessible from all ray processes - so we need to call sync here to update the local changes. 

This was the cleanest solution I could think of here for getting ray initialized with all the env vars we want and also the registries available everywhere.

skyrl-train/skyrl_train/utils/utils.py
```diff
@@ -183,10 +184,13 @@ def validate_cfg(cfg: DictConfig):
             # for local engines or sglang, we disable
             cfg.generator.override_existing_update_group = "disable"
 
-    assert cfg.trainer.algorithm.ppo_loss_type in (
-        "regular",
-        "dual_clip",
-    ), f"invalid ppo_loss_type: {cfg.trainer.algorithm.ppo_loss_type}. Must be one of `['regular', 'dual_clip']`"
+    assert (
```
erictang000, line 187: this validation means that users have to register their functions **outside** of the ray entrypoint in their scripts like:

```python

# Register the custom policy loss
PolicyLossRegistry.register("reinforce", compute_reinforce_policy_loss)


@ray.remote(num_cpus=1)
def skyrl_entrypoint(cfg: DictConfig):
    exp = BasePPOExp(cfg)
    exp.run()
```

since config validation happens outside of the ray entrypoint as well. I think this is the natural way people would be inclined to do just noting it here.

skyrl-train/skyrl_train/utils/ppo_utils.py
```diff
@@ -216,6 +158,292 @@ def masked_whiten(values, mask, shift_mean=True):
...
+        if not found_locally and not found_in_actor:
+            raise ValueError(f"{cls._function_type} '{name}' not registered")
+
+
+class AdvantageEstimator(Enum):
+    GAE = "gae"
+    GRPO = "grpo"
+
+    def __str__(self):
+        return self.value
```
bthecohen: I think you can avoid this by subclassing [StrEnum](https://docs.python.org/3/library/enum.html#enum.StrEnum)

skyrl-train/skyrl_train/utils/ppo_utils.py
```diff
@@ -216,6 +158,292 @@ def masked_whiten(values, mask, shift_mean=True):
...
+        AdvantageEstimatorRegistry.register(name, func)
+        return func
+
+    return decorator
+
+
+def register_policy_loss(name: Union[str, PolicyLossType]):
+    """Decorator to register a policy loss function."""
+
+    def decorator(func: Callable):
```
bthecohen, line 436: Should use https://docs.python.org/3/library/functools.html#functools.wraps here

skyrl-train/skyrl_train/utils/ppo_utils.py
```diff
@@ -216,6 +158,292 @@ def masked_whiten(values, mask, shift_mean=True):
...
+    clip_pg_losses1 = loss
+    if config.policy_loss_type == "dual_clip":
+        pg_losses3 = -advantages * config.clip_ratio_c
+        clip_pg_losses2 = torch.min(pg_losses3, clip_pg_losses1)
+        loss = torch.where(advantages < 0, clip_pg_losses2, clip_pg_losses1)
+    loss = reduce_loss(loss, loss_mask, loss_reduction)
+    return loss, clip_ratio
+
+
+def reduce_loss(loss: torch.Tensor, loss_mask: Optional[torch.Tensor], loss_reduction: str) -> torch.Tensor:
```
bthecohen: can we make loss_reduction a literal type for static checking here?

skyrl-train/skyrl_train/utils/ppo_utils.py
```diff
@@ -216,6 +158,292 @@ def masked_whiten(values, mask, shift_mean=True):
...
+    _function_type = "Function"
+
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__(**kwargs)
+        cls._functions = {}
+        cls._ray_actor = None
+        cls._synced_to_actor = False
+
+    @classmethod
+    def _get_or_create_actor(cls):
```
bthecohen, line 202: This should probably raise if Ray isn't initialized

skyrl-train/skyrl_train/utils/ppo_utils.py
```diff
@@ -216,6 +158,292 @@ def masked_whiten(values, mask, shift_mean=True):
...
+            # Only try to create actors if Ray is initialized
+            if ray.is_initialized():
+                # Use get_if_exists to create actor only if it doesn't exist
+                cls._ray_actor = RegistryActor.options(name=cls._actor_name, get_if_exists=True).remote()
+        return cls._ray_actor
+
+    @classmethod
+    def _sync_local_to_actor(cls):
+        """Sync all local functions to Ray actor (one-time when Ray becomes available)."""
+        if cls._synced_to_actor or not ray.is_initialized():
```
bthecohen: again, I think you might want to raise here if ray isn't initialized, otherwise the training could fail later when we try to get a function from the registry

bthecohen (commented):
Left a few small comments, but overall this approach looks good to me and gives a clean API for researchers.

skyrl-train/skyrl_train/utils/ppo_utils.py
```diff
@@ -216,6 +158,292 @@ def masked_whiten(values, mask, shift_mean=True):
...
+
+    @classmethod
+    def register(cls, name: Union[str, Enum], func: Callable):
+        """Register a function."""
+        # Convert enum to string if needed
+        if isinstance(name, Enum):
+            name = name.value
+
+        if name in cls._functions:
+            raise ValueError(f"{cls._function_type} '{name}' already registered")
```
SumanthRH, line 286: could there be a case where the same function is being  registered multiple times because of `.register` line being run in the driver + a worker process? 

Currently, I think not, because the worker processes will have a fresh import of `ppo_utils.py` (like say a FSDPWorker, or even `skyrl_entrypoint`) and thus the registry state is empty, and the `.register` code will only run on the driver.

skyrl-train/skyrl_train/utils/ppo_utils.py
SumanthRH: +1

skyrl-train/skyrl_train/utils/ppo_utils.py
```diff
@@ -216,6 +158,292 @@ def masked_whiten(values, mask, shift_mean=True):
...
+                        func = cloudpickle.loads(func_serialized)
+                        cls._functions[name] = func
+                    except Exception as e:
+                        # If deserialization fails, skip this function
+                        logger.error(f"Error deserializing {name} from actor: {e}")
+                        raise e
+
+    @classmethod
+    def register(cls, name: Union[str, Enum], func: Callable):
+        """Register a function."""
```
SumanthRH: Please add a detailed docstring for register, we should document behavior in both cases with and without the named actor being present.

skyrl-train/skyrl_train/utils/ppo_utils.py
```diff
@@ -216,6 +158,292 @@ def masked_whiten(values, mask, shift_mean=True):
...
+        if ray.is_initialized():
+            actor = cls._get_or_create_actor()
+            if actor is not None:
+                # Serialize the function using cloudpickle
+                func_serialized = cloudpickle.dumps(func)
+                ray.get(actor.register.remote(name, func_serialized))
+
+    @classmethod
+    def get(cls, name: str) -> Callable:
+        """Get a function by name."""
```
SumanthRH: Similar comment on docstring as above. Pretty sure registries will be used heavily so we should go the extra mile

skyrl-train/tests/cpu/utils/test_ppo_utils.py
```diff
@@ -149,110 +151,247 @@ def test_fixed_kl_controller():
...
+        if "cross_process_test" in PolicyLossRegistry.list_available():
+            PolicyLossRegistry.unregister("cross_process_test")
+        if "cross_process_test_2" in PolicyLossRegistry.list_available():
+            PolicyLossRegistry.unregister("cross_process_test_2")
+        if "cross_process_adv_test" in AdvantageEstimatorRegistry.list_available():
+            AdvantageEstimatorRegistry.unregister("cross_process_adv_test")
+        PolicyLossRegistry._ray_actor = None
+        PolicyLossRegistry._synced_to_actor = False
+        AdvantageEstimatorRegistry._ray_actor = None
+        AdvantageEstimatorRegistry._synced_to_actor = False
```
SumanthRH: Shouldn't the named actor be cleaned up here (or i'm missing something) ? good time to use the ray init fixture probably.

skyrl-train/tests/cpu/utils/test_ppo_utils.py
```diff
@@ -149,110 +151,247 @@ def test_fixed_kl_controller():
...
+
+        assert torch.allclose(result[0], test_rewards * 2)
+        assert torch.allclose(result[1], test_rewards * 3)
+
+    finally:
+        # Clean up
+        if "named_actor_test" in AdvantageEstimatorRegistry.list_available():
+            AdvantageEstimatorRegistry.unregister("named_actor_test")
+        AdvantageEstimatorRegistry._ray_actor = None
+        AdvantageEstimatorRegistry._synced_to_actor = False
```
SumanthRH: Another point on cleanup, since ther'es no ray.shutdown call here

SumanthRH (commented):
Looks good overall. There's some implicit behaviour around how registry access works with and without the presence of the named actor, and it would be good to atleast document.

skyrl-train/docs/configuration/config.rst
SumanthRH, line 1: can you add the registry classes to the API reference please?

skyrl-train/tests/cpu/utils/test_ppo_utils.py
erictang000: added a ray.kill on the actor - the current cpu test ray init fixture is session level so didn't want to call `ray.shutdown` here

PR #132, opened 2025-08-04
[trainer] add more robust generation output validation

erictang000:
# Overview
Adds a `validate_generation_output` function in `trainer_utils.py` with more robust validation of generation output format. Specifically, given 
```
class GeneratorOutput(TypedDict):
    prompt_token_ids: List[List[int]]
    response_ids: List[List[int]]
    rewards: Union[List[float], List[List[float]]]
    loss_masks: List[List[int]]
    stop_reasons: Optional[List[str]]
    rollout_metrics: Optional[Dict[str, Any]]
```

We expect
- all list attributes should have the same length  and be the same length as the input batch of prompts at dim=0
- non zero length lists
- response_ids, loss masks, and rewards (if token level rewards) should be the same length
- the sum of loss masks should be non-zero (logging a warning if it is not)

verified gsm8k run still works:
image

### DIFF
--- a/skyrl-train/tests/cpu/test_trainer_utils.py
+++ b/skyrl-train/tests/cpu/test_trainer_utils.py
@@ -1,3 +1,7 @@
+"""
+uv run --isolated --extra dev pytest tests/cpu/test_trainer_utils.py
+"""
+
 from skyrl_train.utils.trainer_utils import (
     run_on_each_node,
     cleanup_old_checkpoints,
@@ -5,7 +9,9 @@ from skyrl_train.utils.trainer_utils import (
     sanitize_data_source,
     calculate_per_dataset_metrics,
     dump_per_dataset_eval_results,
+    validate_generator_output,
 )
+from skyrl_train.generators.base import GeneratorInput, GeneratorOutput
 from typing import Union
 import ray
 import os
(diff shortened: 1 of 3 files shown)


skyrl-train/tests/cpu/test_trainer_utils.py
```diff
@@ -312,3 +319,133 @@ def test_dump_per_dataset_eval_results_comprehensive(mock_file):
...
+        prompt_token_ids=[[1, 2, 3, 4], [5, 6], [7, 8, 9]],
+        response_ids=[[10, 11, 12], [13, 14], [15]],
+        rewards=[[0.5, 0.6, 0.7], [0.8, 0.9], [1.0]],
+        loss_masks=[[1, 1], [1], [1, 1]],  # loss masks are not the same length as response ids
+        stop_reasons=["stop", "length", "stop"],
+        rollout_metrics={"metric1": 0.5, "metric2": 0.6},
+    )
+
+    with pytest.raises(
+        AssertionError, match="Response ids and loss masks must have the same length, for sample 0 got 3 and 2"
```
SumanthRH:
```suggestion
        AssertionError, match="Response ids and loss masks must have the same length"
```

make the test less brittle

skyrl-train/tests/cpu/test_trainer_utils.py
```diff
@@ -312,3 +319,133 @@ def test_dump_per_dataset_eval_results_comprehensive(mock_file):
...
+    ):
+        validate_generator_output(input_batch, generator_output)
+
+    generator_output["loss_masks"] = [[1, 1, 1], [1, 1], [1]]  # add correct loss masks
+    generator_output["rewards"] = [[0.5, 0.6], [0.8], [1.0, 2.0]]  # add incorrect rewards
+    with pytest.raises(
+        AssertionError, match="Token rewards and response ids must have the same length, for sample 0 got 2 and 3"
```
SumanthRH:
```suggestion
        AssertionError, match="Token rewards and response ids must have the same length"
```

SumanthRH (approved):
Looks good, thanks!

PR #133, opened 2025-08-06
[algorithm] Support Dr. GRPO + refactor where policy/critic loss functions are set

erictang000:
# Overview
## Dr GRPO
Adds `loss_reduction`: `seq_mean_token_sum_norm ` option, and `grpo_norm_by_std` option to support Dr. GRPO

So to run Dr. GRPO, set: 

```yaml
trainer:
 algorithm:
   grpo_norm_by_std: false
   loss_reduction: "seq_mean_token_sum_norm"
...
```

Example run:
image

Reward looks similar

image

Magnitude of policy loss is lower as expected (since we are normalizing by a larger constant rather than taking the mean)

## Refactor where Critic/Policy Loss are set
Changes ppo critic `ValueLoss` to just a function instead of a `nn.Module` for consistency with `policy_loss`, and adds new algorithm field to cfg that require evaluating field values  in `utils::validate_cfg` (this runs before entrypoint code, allowing users to modify the cfg further by subclassing `BasePPOExp`)

PPO example still running after this refactor:
image


skyrl-train/docs/configuration/config.rst
```diff
@@ -322,7 +323,13 @@ Algorithm Configuration
   - ``gspo``: `Group Sequence Policy Optimization <https://arxiv.org/abs/2507.18071>`_ with sequence-level importance sampling for improved training stability. Implements "GSPO-token" variant from the paper.
   - Custom policy losses can be registered with the ``PolicyLossRegistry``
 
-- ``algorithm.loss_reduction``: Type of loss reduction to use. Options are ``token_mean`` and ``sequence_mean``. ``token_mean`` matches token-level loss introduced by `DAPO <https://dapo-sia.github.io/>`_. ``sequence_mean`` computes per-sequence avg token loss, then averages over the batch.
+- ``algorithm.loss_reduction``: Type of loss reduction to use. Options include:
+
+  - ``token_mean``: computes average loss over all valid tokens in the batch. Used in `DAPO <https://dapo-sia.github.io/>`_.
+  - ``sequence_mean``: computes per-sequence avg token loss, then averages over the batch.
+  - ``max_seq_len_normalized_mean``: computes the sum of token losses for each sequence, normalizes by the max sequence length, and then averages over the batch. This is used in `Dr. GRPO <https://arxiv.org/abs/2503.20783>`_.
```
SumanthRH: Just use `seq_mean_token_sum_norm` , and then explain what the normalization is.... the current name `max_seq_len_normalized_mean`  is pretty unclear

skyrl-train/skyrl_train/workers/worker.py
```diff
@@ -618,6 +620,18 @@ def _normalize_mini_batch_size(self):
...
+    def set_actor_loss_fn(self):
+        policy_loss_func = PolicyLossRegistry.get(self.cfg.trainer.algorithm.policy_loss_type)
+        config = OmegaConf.create(self.cfg.trainer.algorithm)
+        # NOTE (erictang000): this is the max sequence length including the prompt, since max response length
+        # per batch can be variable based on the prompt length. This is used to normalize the loss for
+        # max_seq_len_normalized_mean loss reduction. Potentially revisit this if we update to use a
+        # fixed max response budget.
+        config.max_seq_len = (
+            self.cfg.generator.max_input_length + self.cfg.generator.sampling_params.max_generate_length
+        )
```
SumanthRH: This value being set inside each rank rather than being passed from the training loop is bad....  it would be good to have this be set in the trainer so that it can also be easily modified (ex: I want to change this max seq len logic) . It's actually algorithm related, but it's hidden deep inside the base worker class

skyrl-train/examples/algorithm/drgrpo/run_drgrpo_gsm8k.sh
SumanthRH, line 1: Could you use as many settings as possible as in : https://github.com/sail-sg/understand-r1-zero

I think it will take more work in providing the exact "Dr. GRPO" recipe with reward function, etc... so we can add a disclaimer in the script and make sure all training settings we can configure (like kl loss, ppo loss function used, etc) are as close as possible

SumanthRH (changes requested):
Thanks! Just a minor comment on where `max_seq_len` is being set

skyrl-train/examples/algorithm/drgrpo/run_drgrpo_gsm8k.sh
erictang000: hmmm, i would say that the two changes that are the core of the Dr. GRPO recipe are the ones we update, and it's accurate to say we are using their method by just doing those. We aren't providing a script to reproduce their results.

image

erictang000: I think kl loss and ppo loss function are reasonable ones to make sure are the same, but i would say we aren't providing a "recipe" per se

SumanthRH: Yeah makes sense, I agree we don't have to aim for the exact dr. Grpo recipe!

erictang000: seems like kl = 0 
image

and i don't see dual clipping in oat: https://github.com/sail-sg/oat/blob/7174a3783fd07d981df8a1edd8bb09cfedfbb843/oat/algorithms/ppo.py#L541

 so regular ppo loss

skyrl-train/docs/configuration/config.rst
erictang000: changed

skyrl-train/skyrl_train/workers/worker.py
erictang000: good point, refactored the overall implementation of getting loss functions to be selected in the trainer by adding this:

```python
class RayPPOTrainer:
     ...
     def get_loss_functions(self) -> Tuple[Callable, Callable]:
        ...
        return policy_loss_func, critic_loss_func 
```

and adding a setter function for loss function in the policy/critic workers that gets called during `build_models`. Should make things more extensible.

skyrl-train/docs/configuration/config.rst
```diff
@@ -329,7 +330,13 @@ Algorithm Configuration
...
   - Custom policy losses can be registered with the ``PolicyLossRegistry``
 
-- ``algorithm.loss_reduction``: Type of loss reduction to use. Options are ``token_mean`` and ``sequence_mean``. ``token_mean`` matches token-level loss introduced by `DAPO <https://dapo-sia.github.io/>`_. ``sequence_mean`` computes per-sequence avg token loss, then averages over the batch.
+- ``algorithm.loss_reduction``: Type of loss reduction to use. Options include:
+
+  - ``token_mean``: computes average loss over all valid tokens in the batch. Used in `DAPO <https://dapo-sia.github.io/>`_.
+  - ``sequence_mean``: computes per-sequence avg token loss, then averages over the batch.
+  - ``seq_mean_token_sum_norm``: computes the sum of token losses for each sequence, normalizes by the max sequence length, and then averages over the batch. This is used in `Dr. GRPO <https://arxiv.org/abs/2503.20783>`_.
+
+- ``algorithm.grpo_norm_by_std``: Whether to normalize advantages by the standard deviation in GRPO. This is set to ``false`` in `Dr. GRPO <https://arxiv.org/abs/2503.20783>`_.
```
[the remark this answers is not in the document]
erictang000: lol

erictang000:
addressed comments ready for re-review @SumanthRH

skyrl-train/skyrl_train/trainer.py
```diff
@@ -413,6 +415,28 @@ def _prepare_generator_input(
...
+            self.cfg.generator.max_input_length + self.cfg.generator.sampling_params.max_generate_length
+        )
+
+        # policy loss function
+        policy_loss_func = PolicyLossRegistry.get(self.cfg.trainer.algorithm.policy_loss_type)
+        policy_loss_func = partial(policy_loss_func, config=algorithm_config)
+
+        # critic loss function
+        critic_loss_func = partial(ppo_critic_loss, config=algorithm_config)
+        return policy_loss_func, critic_loss_func
```
SumanthRH: QQ: The trainer can just `set` the loss function in the policy loss registry, and then all the actors can just `get` it as before? Maybe I missed something, but why do this approach? 

I do like the wildcard `config` entry being an argument to each loss function, making it more flexible, but then the pattern should be that PolicyLossRegistry is still the source of truth, with configs being passed to be in the trainer like here?

erictang000: updated per offline discussion!

erictang000:
@SumanthRH made changes to move algorithm config modification to `validate_cfg` and pass in wild card config without partial in policy/critic workers, ready for re-review

skyrl-train/skyrl_train/utils/ppo_utils.py
```diff
@@ -538,20 +560,33 @@ def gspo_policy_loss(
...
     if loss_reduction == "token_mean":
         # sum over *all* valid tokens, divide by total valid-token count
         loss = masked_mean(loss, loss_mask)
     elif loss_reduction == "sequence_mean":
         # per-sequence token-mean (dim=-1), then batch-mean
         loss = masked_mean(loss, loss_mask, dim=-1).mean()
+    elif loss_reduction == "seq_mean_token_sum_norm":
+        # per-sequence token-sum, normalized by the max sequence length, then batch mean
+        # this is the Dr. GRPO loss reduction to avoid length bias by normalizing by a constant
+        assert max_seq_len is not None, "max_seq_len must be provided for seq_mean_token_sum_norm loss reduction"
```
SumanthRH, line 583:
```suggestion
        # NOTE: max_seq_len is computed as generator_cfg.max_input_length + generator_cfg.sampling_params.max_generate_length by default
        assert max_seq_len is not None, "max_seq_len must be provided for seq_mean_token_sum_norm loss reduction"
```

While this function doesn't know about `max_input_length` or `generator_cfg`,  it would be nice to just say what this is, instead of me having to follow along and find where it's set

skyrl-train/docs/configuration/config.rst
```diff
@@ -329,7 +330,13 @@ Algorithm Configuration
   - ``gspo``: `Group Sequence Policy Optimization <https://arxiv.org/abs/2507.18071>`_ with sequence-level importance sampling for improved training stability. Implements "GSPO-token" variant from the paper.
   - Custom policy losses can be registered with the ``PolicyLossRegistry``
 
-- ``algorithm.loss_reduction``: Type of loss reduction to use. Options are ``token_mean`` and ``sequence_mean``. ``token_mean`` matches token-level loss introduced by `DAPO <https://dapo-sia.github.io/>`_. ``sequence_mean`` computes per-sequence avg token loss, then averages over the batch.
+- ``algorithm.loss_reduction``: Type of loss reduction to use. Options include:
+
+  - ``token_mean``: computes average loss over all valid tokens in the batch. Used in `DAPO <https://dapo-sia.github.io/>`_.
+  - ``sequence_mean``: computes per-sequence avg token loss, then averages over the batch.
+  - ``seq_mean_token_sum_norm``: computes the sum of token losses for each sequence, normalizes by the max sequence length, and then averages over the batch. This is used in `Dr. GRPO <https://arxiv.org/abs/2503.20783>`_.
```
SumanthRH: Add what the max sequence length is here as well?

PR #134, opened 2025-08-06
[SkyAgent] Add initial docs

SumanthRH:
# What does this PR do?

Adds initial documentation for SkyAgent. 

We are still actively cleaning this package up, but I thought initial documentation will be helpful for anyone who stumbles across this. 

The documentation folder is still in `skyrl-train`, and much of the docs also refer to "SkyRL" when they are really referring to "SkyRL-train", so to avoid any confusion, I have just added this as a simple page on the sidebar. We need to make the docs be mono-repo wide and structure it better but I'm leaving it for a future PR.


caoshiyi (commented):
LGTM! Thanks @SumanthRH

PR #135, opened 2025-08-08
Forward mlflow env vars to ray runtime env

etnbrd:
This PR forward the `MLFLOW_TRACKING_URI` and `MLFLOW_TRACKING_TOKEN` environment variable to the ray runtime env during its initialisation.

### DIFF
--- a/skyrl-train/skyrl_train/utils/utils.py
+++ b/skyrl-train/skyrl_train/utils/utils.py
@@ -281,11 +281,17 @@ def initialize_ray(cfg: DictConfig):
         env_vars["NCCL_P2P_DISABLE"] = "1"
         env_vars["NCCL_SHM_DISABLE"] = "1"
 
-    # TODO: this can be removed if we standardize on env files.
-    # But it's helpful for a quickstart
     if os.environ.get("WANDB_API_KEY"):
         logger.info("Exporting wandb api key to ray runtime env")
         env_vars["WANDB_API_KEY"] = os.environ["WANDB_API_KEY"]
+
+    if os.environ.get("MLFLOW_TRACKING_URI"):
+        logger.info("Exporting mlflow tracking uri to ray runtime env")
+        env_vars["MLFLOW_TRACKING_URI"] = os.environ["MLFLOW_TRACKING_URI"]
+
+    if os.environ.get("MLFLOW_TRACKING_TOKEN"):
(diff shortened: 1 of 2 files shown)


skyrl-train/skyrl_train/utils/utils.py
```diff
@@ -281,12 +281,18 @@ def initialize_ray(cfg: DictConfig):
...
         logger.info("Exporting wandb api key to ray runtime env")
         env_vars["WANDB_API_KEY"] = os.environ["WANDB_API_KEY"]
 
+    if os.environ.get("MLFLOW_TRACKING_URI"):
+        logger.info("Exporting mlflow tracking uri to ray runtime env")
+        env_vars["MLFLOW_TRACKING_URI"] = os.environ["MLFLOW_TRACKING_URI"]
+
+    if os.environ.get("MLFLOW_TRACKING_TOKEN"):
+        logger.info("Exporting mlflow tracking token to ray runtime env")
+        env_vars["MLFLOW_TRACKING_TOKEN"] = os.environ["MLFLOW_TRACKING_TOKEN"]
```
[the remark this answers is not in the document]
etnbrd: I considered looping all of the env variables to forward to the ray runtime env, even including `WANDB_API_KEY` and `SKYRL_LD_LIBRARY_PATH_EXPORT`, but thought that keeping the implementation as a list of statements would be more readable. In this case, I don't believe code duplication is an issue. This iteration would add unnecessary complexity.

skyrl-train/skyrl_train/utils/utils.py
```diff
@@ -281,12 +281,18 @@ def initialize_ray(cfg: DictConfig):
         env_vars["NCCL_P2P_DISABLE"] = "1"
         env_vars["NCCL_SHM_DISABLE"] = "1"
 
-    # TODO: this can be removed if we standardize on env files.
-    # But it's helpful for a quickstart
```
SumanthRH: let's keep this TODO? 

Using `.env` file and then using `uv run ... --env-file .env` will propagate env vars to all ray worker processes (all tasks and actors) so adding the vars in ray.init runtime env will not be needed

etnbrd: Oh indeed! I thought `--env-file` would just provide environment variables. I missed the uv+ray integration you were talking about. Thanks!

SumanthRH (approved):
Thanks!

PR #137, opened 2025-08-08
[algorithm] add rloo and reinforce++ advantage estimators + improve KL penalty handling

erictang000:
# Overview
## Advantage Estimators
- just ports implementations from verl, which is what we have for GRPO and GAE as well + adds example scripts for each
## KL Type Refactor
Config Updates - adds `kl_estimator_type` and marks `use_kl_estimator_k3` and `use_abs_kl` to be deprecated. These will still be respected to override `kl_estimator_type` prior to deprecation if they are set. Adds support for `k1` and `k2` types, since `REINFORCE++` uses the k2 estimator and suggests it is less biased than k3 (see below)
```yaml
kl_estimator_type: "k3" # "k1", "k2", "k3", "abs" - see http://joschu.net/blog/kl-approx.html for details
use_kl_estimator_k3: false # to be deprecated, use kl_estimator_type="k3" instead
use_abs_kl: false # to be deprecated, use kl_estimator_type="abs" instead
```

Reinforce++:
image
image

## KL Controller Support
Adds the kl_ctrl config, which is implemented as follows:
```
kl_ctrl: # only used if use_kl_in_reward is true (not applied in the case of use_kl_loss=true) - uses kl_loss_coef as the initial KL coefficient
      type: "fixed" # "fixed" or "adaptive"
      kl_target: 0.1 # target KL divergence for adaptive KL controller
      horizon: 10000 # controls the update rate of the adaptive KL controller
```
Example with `kl_ctrl.type`=`adaptive` + rloo - kl coeff is being modified

image
image


skyrl-train/skyrl_train/trainer.py
```diff
@@ -753,6 +753,7 @@ def compute_advantages_and_returns(self, data: TrainingInputBatch) -> TrainingIn
             response_mask=data["response_mask"],
             index=data.metadata["uids"],
             adv_estimator=self.cfg.trainer.algorithm.advantage_estimator,
+            config=self.cfg.trainer.algorithm,
```
erictang000, line 765: add this so that users can add additional fields to cfg and have it passed to their custom adv estimator similar to policy loss taking in config wildcard

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -77,10 +77,13 @@ trainer:
     sequence_parallel_size: 1
   algorithm:
     advantage_estimator: "grpo"  # "grpo", "gae", or customizable with AdvantageEstimatorRegistry
-    kl_target: null
-    init_kl_coef: 0.0
```
erictang000, line 81: these two above were previously unused in the code so just directly removing

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -77,10 +77,13 @@ trainer:
...
-    use_kl_estimator_k3: true
-    use_abs_kl: false
+    kl_ctrl: # only used if use_kl_in_reward is true (not applied in the case of use_kl_loss=true) - uses kl_loss_coef as the initial KL coefficient
+      type: "fixed" # "fixed" or "adaptive"
+      kl_target: 0.1 # target KL divergence for adaptive KL controller
+      horizon: 10000 # controls the update rate of the adaptive KL controller
```
erictang000, line 83: i thought this was needed for RLOO but I misread the verl config - but not a bad addition to have, we had the AdaptiveKLController code already previously

skyrl-train/tests/cpu/test_trainer.py
```diff
@@ -74,6 +74,7 @@ def dummy_config():
                     "advantage_estimator": "grpo",
                     "use_kl_estimator_k3": False,
                     "use_abs_kl": False,
+                    "kl_estimator_type": "k1",
                     "init_kl_coef": 0.2,
```
tyler-griggs: nit: `init_kl_coef` is removed right? Should this be updated to `kl_loss_coef`?

skyrl-train/skyrl_train/utils/ppo_utils.py
```diff
@@ -91,23 +105,29 @@ def compute_approx_kl(
...
-        log_ratio = log_ratio.abs()
+    if kl_estimator_type == "k1":
+        kld = log_ratio
+    elif kl_estimator_type == "abs":
+        kld = log_ratio.abs()
+    elif kl_estimator_type == "k2":
+        kld = 0.5 * log_ratio.square()
+    # J. Schulman. Approximating kl divergence, 2020.
+    # URL http://joschu.net/blog/kl-approx.html.
+    elif kl_estimator_type == "k3":
```
tyler-griggs, line 116: Just to sanity check here, is the reason that log_ratio is negated here (`kl = -log_ratio`) because we computed "new_log_probs - old_log_probs" and John's blog computed "old_log_probs - new_log_probs"? The reason I ask is the code above takes the exp() of -log_ratio, but John's blog takes exp() of log_ratio, so just want to confirm I'm understanding correctly.

skyrl-train/tests/cpu/test_trainer.py
erictang000: yes good call

skyrl-train/skyrl_train/utils/ppo_utils.py
erictang000: yes! that's right

so for k1 we were using `log_ratio` where he used `-log_ratio`(and for abs and k2 the sign doesn't matter).

Now that you bring it up though... maybe doing `-(log_probs - log_probs_base)` might have some minor numerical differences from `log_probs_base - log_probs`, so updated the code to just directly do those instead

please sanity check me to make sure that looks right

PR #138, opened 2025-08-09
Optimize tuple unpacking in skyrl_gym_generator.generate()

davenpi:
Hey! 👋🏾

Love what you're building with SkyRL - the modular architecture makes the system level challenges super clear!

I was exploring the codebase and noticed an opportunity to optimize the tuple unpacking in `skyrl_gym_generator.py`. The current implementation iterates through `all_outputs` 5 separate times and uses `sum()` for list concatenation, which has O(n²) complexity.

## What I Changed
Replaced the multiple iterations with a single `zip(*all_outputs)` call to transpose the data in one pass.

**Before:**
```python
responses = sum([[output[0]] for output in all_outputs], [])
rewards = sum([[output[1]] for output in all_outputs], [])
# ... 3 more similar lines
```

**After:**
```python
responses, rewards, stop_reasons, loss_masks, prompt_token_ids = map(list, zip(*all_outputs))
```

## Performance Impact
I wrote a quick benchmark script and the speedup scales nicely with batch size:
- Small batches (10 items): ~5x faster
- Large batches (1000 items): ~50x faster (2.51ms → 0.05ms)

The optimization maintains identical functionality - just more efficient execution. Happy to discuss any questions!

### DIFF
--- a/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
+++ b/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
@@ -289,7 +289,10 @@ class SkyRLGymGenerator(GeneratorInterface):
             mininterval=5,
         )
 
-        responses, rewards, stop_reasons, loss_masks, prompt_token_ids = map(list, zip(*all_outputs))
+        if not all_outputs:
+            responses, rewards, stop_reasons, loss_masks, prompt_token_ids = [], [], [], [], []
+        else:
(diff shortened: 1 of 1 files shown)


tyler-griggs (approved):
Hi @davenpi , thanks for your interest in SkyRL! I'm glad you've enjoyed using it and have had a chance to start digging into the codebase. 

LGTM

PR #143, opened 2025-08-12
[Fix] Add NCCL_CUMEM_ENABLE=0 for vllm to address weight sync error

CharlieFRuan:
Before this PR, running `{colocate_all=false, local vllm engine, NCCL weight_sync_backend, any TP size}` would run into the following issue: https://gist.github.com/CharlieFRuan/18f78a803ff72fa67132d2b80b4f48c4

This is because vLLM sets `NCCL_CUMEM_ENABLE=0`, and we need to match that. See this doc for more https://docs.vllm.ai/en/v0.9.2/usage/troubleshooting.html?h=nccl_cumem_enable#known-issues


SumanthRH:
Thanks @CharlieFRuan

PR #144, opened 2025-08-12
[Fix] Reset registries in BaseFunctionRegistry.sync_with_actor() when needed and fix registry reset

CharlieFRuan:
This PR does two things.

### Change 1
Currently, when we run unit tests like `test_policy_vllm_e2e.py`, in the same Python process, we do `ray.init()` and `ray.shutdown()` for each unit test. However, the states of `BaseFunctionRegistry` persist across each init/shutdown. This makes attributes like `BaseFunctionRegistry._actor_handle` outdated -- as  the actor is killed by `ray.shutdown()`.

This leads to the Registry mistakenly thinking the actor is still alive, leading to error in `BaseFunctionRegistry` for subsequent unit tests.

To fix this, we check whether the actor is still alive with `ray.get_actor(cls._actor_name)` in `BaseFunctionRegistry.sync_with_actor()`. If dead, we reset the attributes accordingly.

An alternative fix is to call `BaseFunctionRegistry.reset()` before each `ray.shutdown()` in unit tests -- this seems less robust in my opinion.

### Change 2
While fixing this, realized that `BaseFunctionRegistry.reset()` has a small issue (this method is not used in non-test code anyway). For a stale actor, `ray.kill(cls._ray_actor)` would directly crash on a C++ level, which cannot be caught by `except Exception`. This can be observed with the newly added `test_registry_reset_after_ray_shutdown()`. Before this PR, this unit test runs into: 

```bash
2025-08-12 02:48:38,218 INFO worker.py:1927 -- Started a local Ray instance.
2025-08-12 02:48:38,264 INFO packaging.py:588 -- Creating a file package for local module '/home/ubuntu/SkyRL/skyrl-train'.
2025-08-12 02:48:38,327 INFO packaging.py:380 -- Pushing file package 'gcs://_ray_pkg_51aa048fab463db9.zip' (3.07MiB) to Ray cluster...
2025-08-12 02:48:38,339 INFO packaging.py:393 -- Successfully pushed file package 'gcs://_ray_pkg_51aa048fab463db9.zip'.
[2025-08-12 02:48:39,449 C 1324650 1324650] actor_manager.cc:55:  Check failed: it != actor_handles_.end() Cannot find an actor handle of id, fdbd48b711397d054bcf399101000000. This method should be called only when you ensure actor handles exists.
*** StackTrace Information ***
/home/ubuntu/.cache/uv/builds-v0/.tmpQDkgsz/lib/python3.12/site-packages/ray/_raylet.so(+0x151a6ea) [0x718cf071a6ea] ray::operator<<()
```

To fix this, we use `actor = ray.get_actor(cls._actor_name)` before killing it.

### DIFF
--- a/skyrl-train/skyrl_train/utils/utils.py
+++ b/skyrl-train/skyrl_train/utils/utils.py
@@ -9,7 +9,6 @@ from ray.util.placement_group import placement_group, PlacementGroupSchedulingSt
 from skyrl_train.utils.ppo_utils import (
     AdvantageEstimatorRegistry,
     PolicyLossRegistry,
-    reset_registries,
     sync_registries,
 )
 
@@ -331,7 +330,6 @@ def initialize_ray(cfg: DictConfig):
     ray.init(runtime_env={"env_vars": env_vars})
 
     # create the named ray actors for the registries to make available to all workers
-    reset_registries()
     sync_registries()
 
 
--- a/skyrl-train/skyrl_train/utils/utils.py
+++ b/skyrl-train/skyrl_train/utils/utils.py
@@ -6,7 +6,12 @@ import torch
 from loguru import logger
 from omegaconf import DictConfig, OmegaConf
 from ray.util.placement_group import placement_group, PlacementGroupSchedulingStrategy, PlacementGroup
-from skyrl_train.utils.ppo_utils import AdvantageEstimatorRegistry, PolicyLossRegistry, sync_registries
+from skyrl_train.utils.ppo_utils import (
+    AdvantageEstimatorRegistry,
+    PolicyLossRegistry,
+    reset_registries,
+    sync_registries,
+)
 
 
 class Timer:
@@ -326,6 +331,7 @@ def initialize_ray(cfg: DictConfig):
     ray.init(runtime_env={"env_vars": env_vars})
 
     # create the named ray actors for the registries to make available to all workers
+    reset_registries()
     sync_registries()
 
 
--- a/skyrl-train/tests/cpu/utils/test_ppo_utils.py
+++ b/skyrl-train/tests/cpu/utils/test_ppo_utils.py
@@ -438,3 +438,43 @@ def test_registry_named_actor_creation():
 
     finally:
         AdvantageEstimatorRegistry.reset()
+
+
+def test_registry_reset_after_ray_shutdown():
+    """
+    Test that the registry resets properly after ray is shutdown.
+
+    This mimics when we run multiple unit tests in a row with ray inits and shutdowns.
+    """
+
+    def _register_func_and_verify():
+        """Register a function and verify it works."""
+
+        def test_func(**kwargs):
+            rewards = kwargs["token_level_rewards"]
+            return rewards * 2, rewards * 3
+
+        AdvantageEstimatorRegistry.register("named_actor_test", test_func)
+        retrieved = AdvantageEstimatorRegistry.get("named_actor_test")
+        assert retrieved == test_func
+        actor = ray.get_actor("advantage_estimator_registry")
+        assert actor is not None
+
+    try:
+        import ray
+
+        # 1. Initialize ray and register function
+        if not ray.is_initialized():
+            ray.init()
+        _register_func_and_verify()
+
+        # 2. Shutdown ray
+        ray.shutdown()
+
+        # 3. Initialize ray, reset registry, and register function
+        ray.init()
+        AdvantageEstimatorRegistry.reset()
+        _register_func_and_verify()
+
+    finally:
+        ray.shutdown()
--- a/skyrl-train/skyrl_train/utils/ppo_utils.py
+++ b/skyrl-train/skyrl_train/utils/ppo_utils.py
@@ -387,7 +387,7 @@ class BaseFunctionRegistry:
             try:
                 actor = ray.get_actor(cls._actor_name)  # this raises exception if the actor is stale
                 ray.kill(actor)
-            except Exception:
+            except ValueError:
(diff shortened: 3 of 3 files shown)


skyrl-train/tests/gpu/test_policy_vllm_e2e.py
```diff
@@ -46,19 +47,7 @@ async def run_vllm_inference(client, prompts):
...
-                "NCCL_CUMEM_ENABLE": "0",
-                "NCCL_P2P_DISABLE": "0",
-                "CUDA_LAUNCH_BLOCKING": "1",
-                "VLLM_USE_V1": "1" if v1 else "0",
-                "VLLM_ENABLE_V1_MULTIPROCESSING": "0",
-                "PYTORCH_NVML_BASED_CUDA_CHECK": "1",
-            }
-        },
-    )
+    initialize_ray(cfg)
```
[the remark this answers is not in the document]
CharlieFRuan: This is expected.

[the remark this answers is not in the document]
SumanthRH: QQ: by changing the ray init  can you get reinit error now from ray if we run the tests in tests/gpu? 

As such this e2e shud be run in a standalone pytest session, so the change is fine

SumanthRH (approved):
Thanks

skyrl-train/skyrl_train/utils/utils.py
```diff
@@ -319,6 +324,7 @@ def initialize_ray(cfg: DictConfig):
     ray.init(runtime_env={"env_vars": env_vars})
 
     # create the named ray actors for the registries to make available to all workers
+    reset_registries()
```
SumanthRH: resetting the registry just before sync registeries will unregister any functions added in the driver? registering will not work anymore?

SumanthRH (changes requested):
QQ

skyrl-train/skyrl_train/utils/utils.py
SumanthRH: Can you maybe add the full stack trace of when you observed the crash with gpu tests? 

ideally we just call ray.shutdown after each gpu test properly, so we init cleanly ( because we'll have gpu actors). The named actors will be attached to the ray session, so they will be killed by ray automatically when we do so. Can you maybe provide the stack trace of when you observed the error? 

let's just fix the GPU tests with ray.shutdown() at exit

CharlieFRuan: As discussed offline, the original fix I had is bad and killed the purpose of registries in the first place. Will fix.

The stack trace fails silently, but after printing it is because of `available = ray.get(actor.list_available.remote())`, where the actor is a reference to an actor in the previous ray session (i.e. the previous unit test).

I think the issue is that, despite `ray.shutdown()`, the Registries Python objects persist -- since we are running multiple unit tests in the same Python process, each unit test has ray init and shutdown. This creates misalignment between ray's state and `BaseFunctionRegistry`'s attributes (e.g. the `_ray_actor` points to a stale actor in the previous ray session).

Essentially, we wish the `BaseFunctionRegistry`'s states to be updated when `ray.shutdown()` is called. I will create a helper `shutdown()` for unit tests.

CharlieFRuan: Alternatively, at the start of `sync_with_actor()`, I check whether `_ray_actor` is legit, and if not I set it to `None` and `_synced_to_actor` to `False`. I like this solution better, will try

skyrl-train/tests/gpu/test_policy_vllm_e2e.py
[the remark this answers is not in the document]
CharlieFRuan: > QQ: by changing the ray init can you get reinit error now from ray if we run the tests in tests/gpu?

We'd have to parametrize `initialize_ray()` with `ignore_reinit_error`.

In this test we do `finally: ray.shutdown()` so I think it is fine.

The discrepancy between `ray.init()`'s arguments and `initialize_ray()` have led to various GPU test failing & debuggings, so I think reusing would be the best

skyrl-train/skyrl_train/utils/utils.py
CharlieFRuan: > Alternatively, at the start of sync_with_actor(), I check whether _ray_actor is legit, and if not I set it to None and _synced_to_actor to False. 

Fixed.

skyrl-train/tests/gpu/test_policy_vllm_e2e.py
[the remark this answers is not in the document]
CharlieFRuan: expected, same as the previous gemini review

skyrl-train/tests/cpu/utils/test_ppo_utils.py
```diff
@@ -438,3 +438,43 @@ def test_func(**kwargs):
...
+        # 2. Shutdown ray
+        ray.shutdown()
+
+        # 3. Initialize ray, reset registry, and register function
+        ray.init()
+        AdvantageEstimatorRegistry.reset()
+        _register_func_and_verify()
+
+    finally:
+        ray.shutdown()
```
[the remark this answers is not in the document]
CharlieFRuan: I don't think it is needed since we basically reset the actor handle when syncing

skyrl-train/skyrl_train/utils/utils.py
SumanthRH: This is better. Thanks @CharlieFRuan !

SumanthRH (approved):

PR #146, opened 2025-08-13
[GPU CI] Add new workflow for GSM8k e2e convergence test

erictang000:
# Overview
Adds a workflow to schedule an anyscale job to run basic GSM8k training for a single epoch (7 steps) on 4xL4 GPUs. Training currently takes 1 hour and 10 minutes. 

Scheduled to run once a day a little bit after midnight PST.

Logs to a WandB project so we can track correctness and any regressions in speed
image

Anyscale job: 
image

### DIFF
--- a/.github/workflows/gpu_e2e_ci.yaml
+++ b/.github/workflows/gpu_e2e_ci.yaml
@@ -1,16 +1,9 @@
-name: SkyRL-GPU
+name: SkyRL-GPU-E2E-CI
 
 on:
-  push:
-    branches:
-      - main
-    paths:
-      - 'skyrl-train/**'
-      - '!skyrl-train/docs/**'
-      - '!skyrl-train/examples/**'
-      - '.github/workflows/**'
+  schedule:
+    - cron: '5 8 * * *'   # Every day at 08:05 UTC (~00:05 PST / ~01:05 PDT)
   workflow_dispatch:
-
 
 permissions:
   checks: write   # for status checks to appear
@@ -18,7 +11,7 @@ permissions:
 
 jobs:
   
-  skyrl_tests:
+  skyrl_gpu_e2e_test:
     runs-on: ubuntu-latest
     defaults:
       run:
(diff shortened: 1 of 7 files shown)


.github/workflows/gpu_e2e_ci.yaml
```diff
@@ -0,0 +1,46 @@
...
+          activate-environment: true
+      - name: Install basic dependencies
+        run: uv pip install anyscale==0.24.79 typer==0.9.0
+      - name: Install envsubst
+        run: sudo apt-get update && sudo apt-get install -y gettext-base
+      - name: Basic convergence test
+        env:
+          ANYSCALE_CLI_TOKEN: ${{ secrets.ANYSCALE_CLI_TOKEN }}
+          ANYSCALE_HOST: https://console.anyscale.com
+          WANDB_API_KEY: ${{ secrets.WANDB_API_KEY }}
```
erictang000, line 42: requires WandB API key here so we can track runs over time - will add to secrets after merging this PR

skyrl-train/ci/anyscale_gpu_e2e_test.yaml
```diff
@@ -0,0 +1,12 @@
+name: skyrl-train-gpu-e2e-test
+entrypoint: bash ci/gpu_e2e_test_run.sh
+image_uri: erictang000/skyrl-train-ray-2.48.0-py3.12-cu12.8 # (Optional) Exclusive with `containerfile`.
+cloud: sky-anyscale-aws-us-east-1
+ray_version: "2.48.0"
+compute_config: l4_ci 
+working_dir: . # (Optional) Use current working directory "." as the working_dir. Can be any local path or remote .zip file in cloud storage.
+env_vars:
+  RAY_OVERRIDE_JOB_RUNTIME_ENV: "1"
+  WANDB_API_KEY: $WANDB_API_KEY
```
SumanthRH, line 10: these two  env vars (job override and the wandb api key) are actually not needed given that we export WANDB_API_KEY at init time? just need to be set at driver?

skyrl-train/tests/gpu/gpu_e2e_test/gsm8k_colocate.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# The anyscale job's working_dir is the repo root, so we can use relative paths.
+bash examples/gsm8k/run_gsm8k.sh \
+  trainer.epochs=1 \
```
SumanthRH, line 6: Can we make the CI test script standalone please? 

I don't want changes in the example to affect our CI runs. Also, it is better to isolate our E2E test scripts from our example scripts

erictang000: hmm, gemini suggested the opposite and I think I agree - I'd say that the example should always work given it's our main entrypoint for quickstart in the docs, and I think it makes sense to test that it's working

skyrl-train/ci/anyscale_gpu_e2e_test.yaml
erictang000: I don't think that's true? how will the anyscale job have access to the wandb key if we don't include it here?

skyrl-train/tests/gpu/gpu_e2e_test/gsm8k_colocate.sh
SumanthRH: oh yeah for gsm8k i agree. sg

skyrl-train/ci/anyscale_gpu_e2e_test.yaml
SumanthRH: Oh that's right

SumanthRH: the bash script is run inside the job, but the github ci will have the secret. sg

SumanthRH (approved):

PR #148, opened 2025-08-14
[Gym] Add AIME 

lynnliu030:


skyrl-train/examples/algorithms/dapo/run_dapo_aime.sh
```diff
@@ -0,0 +1,86 @@
...
+  trainer.eval_batch_size=1024 \
+  trainer.eval_before_train=false \
+  trainer.eval_interval=5 \
+  trainer.update_epochs_per_batch=1 \
+  trainer.train_batch_size=1024 \
+  trainer.policy_mini_batch_size=256 \
+  trainer.micro_forward_batch_size_per_gpu=64 \
+  trainer.micro_train_batch_size_per_gpu=64 \
+  trainer.ckpt_interval=10 \
+  trainer.max_prompt_length=512 \
```
SumanthRH: Let's use the same settings as DAPO did? 

https://github.com/volcengine/verl/blob/bd756c15c8873e40cfbb5ef2a4d1d6cadb4ea272/recipe/dapo/run_dapo_qwen2.5_32b.sh#L17-L18

SumanthRH: I think the only difference here is like batch size and model arch, but beyond that we should match all other settings

skyrl-train/examples/algorithms/dapo/run_dapo_aime.sh
```diff
@@ -0,0 +1,86 @@
...
+OVERLONG_BUFFER_LEN=512
+OVERLONG_BUFFER_PENALTY_FACTOR=1.0
+
+# other DAPO parameters
+USE_KL_LOSS=false
+TEMPERATURE=1.0
+TOP_P=1.0
+EVAL_TOP_P=0.7
+CLIP_RATIO_C=10.0
+MAX_RESPONSE_LENGTH=1024
```
SumanthRH: This one as well: https://github.com/volcengine/verl/blob/bd756c15c8873e40cfbb5ef2a4d1d6cadb4ea272/recipe/dapo/run_dapo_qwen2.5_32b.sh#L17-L18

SumanthRH (commented):
Looks good overall!

For a sanity check on the reward: Could we run evaluation loop before training once for a base model on AIME and makes sure the numbers look fine? maybe Qwen3-8B or [DeepSeek-R1-0528-Qwen3-8B](https://huggingface.co/deepseek-ai/DeepSeek-R1-0528-Qwen3-8B) ?

skyrl-train/examples/algorithms/dapo/run_dapo_aime.sh
lynnliu030: Sounds good, will match their setup

skyrl-gym/skyrl_gym/envs/aime/utils.py
```diff
@@ -0,0 +1,273 @@
...
+    """
+    if strict_box_verify:
+        correct, pred = is_correct_strict_box(solution_str, answer, pause_tokens_index)
+        return correct == 1, pred
+
+    correct, pred = is_correct_minerva(solution_str, answer)
+    return correct, pred
+
+
+def compute_score(
```
SumanthRH, line 246: Could we also have a few tests with typical corner cases for this? can be examples straight from aime

lynnliu030: Added!

skyrl-gym/skyrl_gym/envs/aime/env.py
```diff
@@ -0,0 +1,31 @@
...
+        assert "reward_model" in extras, "reward_model field is required"
+        assert "ground_truth" in extras["reward_model"], "ground_truth is required in reward_model field"
+        self.ground_truth = extras["reward_model"]["ground_truth"]
+
+    def step(self, action: str) -> BaseTextEnvStepOutput:
+        done = True  # always done after one step
+
+        # NOTE(shu): manually parse the action with end_string |im_end|
+        # Might want the env to have a tokenizer and being able to take tokens for special reward calculation
+        action = action.split("<|im_end|>")[0]
```
SumanthRH: is this needed? this EOS token will be model specific ( this will work only for qwen then). Can probably make our parsing be more robust

skyrl-gym/tests/test_aime.py
```diff
@@ -0,0 +1,24 @@
+import skyrl_gym
+import pytest
+from omegaconf import DictConfig
+
+
+@pytest.mark.parametrize(
```
SumanthRH, line 6: Another point on eos token: we should also test for different eos tokens here

SumanthRH (commented):
Thanks! Would be good to remove hardcoded eos token rn, and also incorporate eos token in tests

skyrl-gym/skyrl_gym/envs/aime/env.py
```diff
@@ -0,0 +1,27 @@
+from skyrl_gym.envs.base_text_env import BaseTextEnv, BaseTextEnvStepOutput
+from skyrl_gym.envs.aime import utils
+from typing import Dict, Any
+from omegaconf import DictConfig
+
+
+class AIMEEnv(BaseTextEnv):
+    """
+    Environment for Math execution tasks.
```
SumanthRH, line 9: Add a disclaimer on the parsing here as well please, since we use `<` now to detect eos token.

SumanthRH (approved):
Thanks! Before merging, it would be great to add the eval accuracy we get with the AIME env for a 8B or smaller model rn and compare with reported

PR #145, opened 2025-08-12
[FlashRL 1/N] Add support for truncated importance sampling

SumanthRH:
# What does this PR do? 

Adds support for truncated importance sampling proposed in https://fengyao.notion.site/flash-rl 

To support this, we need to add a an optional `rollout_logprobs` fields at the inference engine. Note that because the default path has no use for rollout logprobs it is not obtained by default and is populated as `None`. There is some ugliness with just being able to handle both these cases (no logprobs and with logprobs) but it's fine for now. 

TODO: 
- [x] Run E2E GPU tests for vllm and sglang
- [x] Test agent loop path with tis (requires token in, token out fix - not taken up for now)

Issue #123, opened 2025-07-29
Follow token-in-token-out protocol for inference engines


CharlieFRuan:
Currently, whether in the offline batched codepath or the agent loop codepath, SkyRL gets output from inference engines in strings (i.e. the inference engine detokenizes internally), and re-tokenizes the string output to prepare for the input into the training pipeline:
- batched: https://github.com/NovaSky-AI/SkyRL/blob/19b8ca58a23be8cae39dfc1ad3d65c7b76d72d27/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py#L221
- agent loop: https://github.com/NovaSky-AI/SkyRL/blob/19b8ca58a23be8cae39dfc1ad3d65c7b76d72d27/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py#L500

This is susceptible to inconsistency between the token IDs LLMs actually generated and the detokenized-then-tokenized token IDs.

This short script can illustrate the inconsistency, where the assertion at the end can fail (also see section below):  https://gist.github.com/CharlieFRuan/9c3961847176c6e447504e34c35e543f

A fail-safe solution is to always ask the inference engines to return token IDs, use it as-is to prepare for training input. We only de-tokenize for goals such as tool call parsing.

For our remote LLM engines (i.e. `vllm_server.py` and `sglang_server.py`), this means that SGLang will use the `http://localhost:{port}/generate` endpoint, and vLLM is pending this PR to land https://github.com/vllm-project/vllm/pull/22587

### What is token-in-token-out and why do we need it
- Essentially it means that, during the process of converting LLM engines’ output to trainer’s input, we do not want to re-tokenize the engines’ text output to token IDs, as this can cause 2 types of misalignments:
  1. Misalignment between what the LLM actually generated, and what the trainer thinks the LLM generated
  2. In multi-turn setting, misalignment between output token IDs from turn N, and input token IDs to turn N+1
- Consider the following example to illustrate the first type of misalignment
- Say we have 4 vocabs, mapping from token ID to string:
  - `0: <`
  - `1: search`
  - `2: <search`
  - `3: >`
- During rollout, the underlying LLM generated `0, 1, 3`
- `vLLMEngine` returned string `<search>` in `InferenceEngineOutput` all the way back to `Generator` (where the detokenization of token IDs to string is done by the vLLM engine internally)
[...]

PR #152, opened 2025-08-16
[Generator] Support token-in-token-out rollout

CharlieFRuan:
This PR ensures token-in-token-out rollout in most codepaths in SkyRL. The motivation of having token-in/out is described in this issue https://github.com/NovaSky-AI/SkyRL/issues/123

### 1. Making sure the inference engines are doing token-in-token-out (returning `InferenceEngineOutput.response_ids`)
- While `response_ids` were already added for local vLLM engine in https://github.com/NovaSky-AI/SkyRL/pull/145, this PR supports this field for {vllm, sglang} x {remote engine, local engine}.
  - SGLang remote engine:
    - with `--skip-tokenizer-init` passed to SGLang engine creation, we can use `/generate` endpoint, which can take in batch input via HTTP, performing token-in-token-out (almost exactly the same as the python `.generate()`)
  - SGLang local engine: same thing
  - vLLM remote engine:
    - While `/completions` can take token-in, it does not do token-out. vLLM's `/generate` is demo-only according to its doc. So vLLM server endpoint does not support token-in-token-out yet
    - Instead, we do token-in, and prepare `response_ids` by re-encoding the text output from vLLM for a fake token-out
    - This will be fixed after this PR lands in vLLM: https://github.com/vllm-project/vllm/pull/22587
  - That is, SGLang uses `/generate` and vLLM uses `/completions` (neither use `/chat/completions`)
- We always return both `response_ids` and `responses` in `InferenceEngineOutput`, but users when writing their own generator should use `response_ids` (either to prepare next turn's input or `GeneratorOutput`) to respect token-in-token-out. `responses` should only be used to parse for tool calls (like a "read-only" thing)

### 2. Making sure SkyRLGymGenerator respects token-in-token-out
After making sure the InferenceEngineClient is token-in-token-out and returns token IDs to the generator, we need to make sure the generator itself respects token-in-token-out while conforming to (multi-turn) chat templates.
- There are 2 codepaths in SkyRLGymGenerator, `generate_batched()` for single-turn, and `agent_loop()` for both single-turn and multi-turn
  - `generate_batched()` now supports token-in-token-out by simply passing `InferenceEngineOutput.response_ids` to `GeneratorOutput`
  - `agent_loop()` is more tricky, elaborated below

`SkyRLGymGenerator` has 3 ways of managing context in `agent_loop()`, configured by `cfg.generator.use_conversation_multi_turn` and whether `get_custom_chat_template()` returns None (currently it means that the model is Qwen3).
1. Always re-tokenize chat history, when `use_conversation_multi_turn==True` and `get_custom_chat_template() is not None` 
2. `use_conversation_multi_turn == True`, but has no custom chat template
3. `use_conversation_multi_turn == False`

This PR ensures token-in-token-out for codepaths 2 and 3 (but not 1) and make the 3 codepaths more distinct and clearer in the code.

#### 2.1. Always re-tokenize chat history (does not support token-in-token-out)
- This codepath currently serves Qwen3 models, where we always re-tokenize the chat history for each turn
- There are two reasons for this:
  - Qwen3 removes non-last turn thinking tokens
  - Able to get `["assistant_masks"]` and `["input_ids"]` from the final tokenized chat history with the help of `{% generation %}` and `{% endgeneration %}` tags in the jinja template
- That is, this codepath rollouts Qwen3 by following the inference chat template, and returns only the last-turn thinking tokens to Generator for the training pipeline
- Comment:
  - It is debatable whether this is the best method to train Qwen3 -- will revisit this codepath in a future PR
  - TODO: Users should be able to train Qwen3 following 2.2 by default, and follow 2.1 by setting a config flag

#### 2.2 `use_conversation_multi_turn == True` without re-tokenizing chat history each turn
- In this codepath, the agent loop does the following
  1. Tokenize dataset's prompt to initialize `input_ids`
  2. Feed `input_ids` to LLM engine, get `output_ids` out (thanks to change 1)
  3. `input_ids += output_ids` (a.k.a. token-in-token-out) -- the next turn's input IDs are precisely what the LLM generated
  4. Tokenize observations got from SkyRL-Gym's environment output (i.e. `env.step()`), and append to `input_ids` -- this is a bit tricky, explained later below
  5. Repeat 2-4 until `env.step()` marks done

- To ensure that the observation is correctly tokenized, we follow the delta-based method proposed [here](https://jybsuper.github.io/posts/multiturn_tokenization/#the-breakthrough-fixed-base-approach)
  - We instantiate `self.base_conversation_token_ids` in `__init__()` and use that to get `observation_ids`
  - One tricky thing is how Qwen models will do: `<|assistant|>something<|im_end|>\n<|user|>something...`
  - That is, after the assistant generates the EOS token `<|im_end|>`, the chat template adds a `\n`
  - We need to manually add this `\n` by making sure `self.base_conversation_token_ids` ends with the EOS token ID
  - For more, see changes in `tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py`

#### 2.3 `use_conversation_multi_turn == False`
- This PR makes this codepath support token-in-token-out without much change (replace `self.tokenizer.encode(output)` with `output_ids` from `InferenceEngineClient`).
- See `_get_next_input_ids_with_single_turn_chat_template()` on what this codepath does

## Tests
- GPU tests
  - `tests/gpu/test_skyrl_gym_generator.py` (besides search and text2sql)
  - `test_engine_generation.py` works (both remote and local)
- CPU tests all passed
  - `tests/cpu.generator/test_skyrl_gym_generator.py`: this PR changes this test a lot because we no longer re-tokenize the LLM output (i.e. the mock tokenizer would be used much less). We should re-write this CPU test and only use a mock LLM, but use an actual tokenizer
  - `tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py`
  - We should consider testing more than Qwen and Llama3 tokenizers

## TODOs
TODOs till ready for a review
- [x] Support token-in-token-out for `SkyRLGymGenerator.generate_batched()`
- [x] Support token-in-token-out for `SkyRLGymGenerator.agent_loop()`
- [x] Make remote engines return tokens
- [x] Thoroughly test each codepath

Future PRs
Tracked by https://github.com/NovaSky-AI/SkyRL/issues/179
- CPU testing
  - [ ] Do not mock tokenizer in `skyrl-train/tests/cpu/generators/test_skyrl_gym_generator.py`, use an actual tokenizer. This should make the test cleaner. Can also parametrize tokenizer and test with different models.
  - [ ] Consider adding tokenizers beyond Qwen and Llama
  - [ ] Add single-turn code path to `tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py`
- [ ] Deprecate post-processed action for SkyRL Gym to ensure token-in-token-out
- [ ] Re-consider Qwen3 training, should it follow the "always re-tokenize chat history" codepath? Maybe by default make it follow 2.2 and only follow 2.1 with a config flag
- [ ] Add a documentation page detailing the behavior of SkyRLGymGenerator, where the content is mainly this PR description and the current chunks of comments. With the doc, we can remove those chunks of comments in the file and keep it more lightweight
- [ ] Consider maintaining string-based `chat_history` in `agent_loop()` even when token-in-token-out, with the goal of doing a sanity check of equivalence between the maintained `input_ids` and `chat_history`

### DIFF
--- a/skyrl-train/skyrl_train/entrypoints/main_base.py
+++ b/skyrl-train/skyrl_train/entrypoints/main_base.py
@@ -61,12 +61,13 @@ def create_ray_wrapped_inference_engines_from_config(cfg: DictConfig, colocate_p
     )
 
 
-def create_remote_inference_engines_from_config(cfg: DictConfig):
+def create_remote_inference_engines_from_config(cfg: DictConfig, tokenizer):
     # TODO(tgriggs): We may want a separate config for the model name in case it's different from the name used in the OpenAI API
     return create_remote_inference_engines(
         urls=cfg.generator.remote_inference_engine_urls,
         model_name=cfg.trainer.policy.model.path,
         engine_backend=cfg.generator.backend,
+        tokenizer=tokenizer,
         tensor_parallel_size=cfg.generator.inference_engine_tensor_parallel_size,
         sampling_params=get_sampling_params_for_backend(cfg.generator.backend, cfg.generator.sampling_params),
     )
@@ -247,7 +248,7 @@ class BasePPOExp:
         if self.cfg.generator.run_engines_locally:
             inference_engines = create_ray_wrapped_inference_engines_from_config(self.cfg, self.colocate_pg, tokenizer)
         else:
-            inference_engines = create_remote_inference_engines_from_config(self.cfg)
+            inference_engines = create_remote_inference_engines_from_config(self.cfg, tokenizer)
 
         inference_engine_client = InferenceEngineClient(inference_engines)
 
--- a/skyrl-train/skyrl_train/entrypoints/main_base.py
+++ b/skyrl-train/skyrl_train/entrypoints/main_base.py
@@ -6,7 +6,7 @@ uv run --isolated --extra vllm -m skyrl_train.entrypoints.main_base
 
 from ray.util.placement_group import placement_group, PlacementGroup
 
-from transformers import AutoTokenizer
+from transformers import AutoTokenizer, PreTrainedTokenizerBase
 from skyrl_train.dataset import PromptDataset
 from skyrl_train.utils import validate_cfg
 
@@ -36,7 +36,7 @@ config_dir = str(Path(__file__).parent.parent / "config")
 __all__ = ["BasePPOExp", "config_dir"]
 
 
-def create_ray_wrapped_inference_engines_from_config(cfg: DictConfig, colocate_pg, tokenizer):
+def create_ray_wrapped_inference_engines_from_config(cfg: DictConfig, colocate_pg, tokenizer: PreTrainedTokenizerBase):
     from skyrl_train.inference_engines.ray_wrapped_inference_engine import create_ray_wrapped_inference_engines
 
     return create_ray_wrapped_inference_engines(
@@ -61,7 +61,7 @@ def create_ray_wrapped_inference_engines_from_config(cfg: DictConfig, colocate_p
     )
 
 
-def create_remote_inference_engines_from_config(cfg: DictConfig, tokenizer):
+def create_remote_inference_engines_from_config(cfg: DictConfig, tokenizer: PreTrainedTokenizerBase):
     # TODO(tgriggs): We may want a separate config for the model name in case it's different from the name used in the OpenAI API
     return create_remote_inference_engines(
         urls=cfg.generator.remote_inference_engine_urls,
--- a/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
+++ b/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
@@ -39,8 +39,27 @@ class SkyRLGymGenerator(GeneratorInterface):
         self.use_conversation_multi_turn = generator_cfg.use_conversation_multi_turn
 
         # This is used to format observation in the chat template.
-        # See https://github.com/volcengine/verl/blob/6bbbff13a131dac9f29411f180ccabe2fe64e786/verl/experimental/agent_loop/tool_agent_loop.py#L196
-        self.system_prompt_ids = tokenizer.apply_chat_template([{}], add_generation_prompt=False, tokenize=True)
+        # Similar to https://github.com/volcengine/verl/blob/6bbbff13a131dac9f29411f180ccabe2fe64e786/verl/experimental/agent_loop/tool_agent_loop.py#L196
+        self.base_conversation = [
+            {"role": "system", "content": "You are a helpful assistant."},
+            {"role": "user", "content": "I am a user."},
+            {"role": "assistant", "content": "I am an assistant."},
+        ]
+        self.base_conversation_token_ids = tokenizer.apply_chat_template(
+            self.base_conversation,
+            add_generation_prompt=False,
+            tokenize=True,
+        )
+        # We cut before the last eos token in base_conversation_token_ids so that the tokens after
+        # the assistant eos token can be included as part of the observation_ids.
+        # In Qwen, there is an extra \n after the eos token that we want to capture in observation_ids.
+        if self.tokenizer.eos_token_id in self.base_conversation_token_ids:
+            last_eos_token_index = (
+                len(self.base_conversation_token_ids)
+                - 1
+                - self.base_conversation_token_ids[::-1].index(self.tokenizer.eos_token_id)
+            )
+            self.base_conversation_token_ids = self.base_conversation_token_ids[: last_eos_token_index + 1]
 
         # optionally use custom chat template to get loss masks (i.e. for Qwen3)
         self.custom_chat_template = get_custom_chat_template(model_name)
@@ -109,7 +128,9 @@ class SkyRLGymGenerator(GeneratorInterface):
         chat_history, _ = env.init(chat_history)
         input_ids = self.tokenizer.apply_chat_template(
             chat_history,
-            add_generation_prompt=True,
+            # otherwise the prompt_ids and response_ids will both include the generation prompt
+            # when retokenize_chat_history is True, due to how `response_encodings["input_ids"]` works.
+            add_generation_prompt=not retokenize_chat_history,
             tokenize=True,
         )
 
@@ -177,6 +198,7 @@ class SkyRLGymGenerator(GeneratorInterface):
 
         env.close()  # does nothing for now
 
+        # TODO(Charlie): this makes the prompt_ids include the generation prompt, is this what we want?
         prompt_ids = input_ids[:initial_prompt_length]
         if retokenize_chat_history:
             response_encodings = self.tokenizer.apply_chat_template(
@@ -504,8 +526,13 @@ class SkyRLGymGenerator(GeneratorInterface):
 
         # 2. apply chat template for observations, also generate generation prompt for next turn
         if len(new_obs) > 0:
-            observation_ids = self.tokenizer.apply_chat_template(new_obs, add_generation_prompt=True, tokenize=True)
-            observation_ids = observation_ids[len(self.system_prompt_ids) :]
+            # For Qwen, this will generate `\n<|user|>Some observation<|im_end|>\n`. Note that the first `\n`
+            # is generated due to `self.base_conversation_token_ids[: last_eos_token_index + 1]` in `__init__()`.
+            observation_ids = self.tokenizer.apply_chat_template(
+                [*self.base_conversation, *new_obs],
+                add_generation_prompt=True,
+                tokenize=True,
+            )[len(self.base_conversation_token_ids) :]
             input_ids += observation_ids
             loss_mask += [0] * len(observation_ids)
         else:
--- a/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
+++ b/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
@@ -231,8 +231,6 @@ class SkyRLGymGenerator(GeneratorInterface):
                 response_ids.append(self.tokenizer.eos_token_id)
                 loss_mask.append(1)
 
-        # TODO(Charlie): do we need to truncate generation_prompt_ids of the last round?
-
         # need to truncate loss mask correctly for responses that go to max length
         if self.max_turns > 1:
             # max total resp length = max tokens (max length of final turn generation) + max_input_length (max input for any generation turn) - len(original prompt)
@@ -293,8 +291,7 @@ class SkyRLGymGenerator(GeneratorInterface):
         truncated_logprobs: Optional[List[List[float]]] = [] if logprobs is not None else None
 
         for i, (response, response_ids, env) in enumerate(zip(responses, all_response_ids, envs)):
-            # step on function and compute reward
-            # assert self.tokenizer.decode(response_ids) == response  # TODO(Charlie): remove this
+            # step on environment and compute reward
             env_step_output: BaseTextEnvStepOutput = env.step(response)
             reward = env_step_output["reward"]
             rewards.append(reward)
--- a/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
+++ b/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
@@ -124,6 +124,7 @@ class SkyRLGymGenerator(GeneratorInterface):
                     prompts=[chat_history], trajectory_ids=[trajectory_id], sampling_params=sampling_params
                 )
             else:
+                # Token-in-token-out.
                 engine_input = InferenceEngineInput(
                     prompt_token_ids=[input_ids], trajectory_ids=[trajectory_id], sampling_params=sampling_params
                 )
@@ -155,18 +156,17 @@ class SkyRLGymGenerator(GeneratorInterface):
             # 3. Update states: input ids, loss_mask, chat_history, etc.
             # Three ways of managing input
             if retokenize_chat_history:
-                # 1. We always re-tokenize the entire chat history every turn and at the end.
+                # a. We always re-tokenize the entire chat history every turn and at the end.
                 chat_history, chat_end_index, input_ids = self._get_next_input_ids_by_retokenizing_chat_history(
                     chat_history, chat_end_index, output, new_obs
                 )
             elif self.use_conversation_multi_turn:
-                # 2. Token-in-token-out is ensured. Follow multi-turn chat history format.
+                # b. Token-in-token-out. Follow multi-turn chat history format.
                 input_ids, loss_mask = self._get_next_input_ids_with_multiturn_chat_template(
-                    input_ids, loss_mask, output_ids, new_obs
+                    input_ids, loss_mask, output_ids, new_obs, done
                 )
             else:
-                # 3. Token-in-token-out is ensured. Each step of the assistant is appended to the
-                # same single assistant message.
+                # c. Token-in-token-out. All steps/observations are appended to a single assistant message.
                 loss_mask, input_ids, rollout_logprobs = self._get_next_input_ids_with_single_turn_chat_template(
                     output_ids, new_obs, loss_mask, input_ids, rollout_logprobs
                 )
@@ -191,6 +191,7 @@ class SkyRLGymGenerator(GeneratorInterface):
             response_ids = response_encodings["input_ids"]
         else:
             response_ids = input_ids[initial_prompt_length:]
+            assert len(loss_mask) == len(response_ids), "loss_mask and response_ids should have the same length"
 
         if not self.use_conversation_multi_turn:
             # we might need to add the eos token to the response ids
@@ -261,7 +262,7 @@ class SkyRLGymGenerator(GeneratorInterface):
 
         for i, (response, response_ids, env) in enumerate(zip(responses, all_response_ids, envs)):
             # step on function and compute reward
-            assert self.tokenizer.decode(response_ids) == response  # TODO(Charlie): remove this
+            # assert self.tokenizer.decode(response_ids) == response  # TODO(Charlie): remove this
             env_step_output: BaseTextEnvStepOutput = env.step(response)
             reward = env_step_output["reward"]
             rewards.append(reward)
@@ -411,6 +412,9 @@ class SkyRLGymGenerator(GeneratorInterface):
                 rewards[i] = 0.0
         return rewards
 
+    # ----------------------------------------------------------------------------
+    # Three methods of managing chat history and input ids in `agent_loop()`
+    # ----------------------------------------------------------------------------
     def _get_next_input_ids_by_retokenizing_chat_history(
         self,
         chat_history: ConversationType,
@@ -450,6 +454,7 @@ class SkyRLGymGenerator(GeneratorInterface):
         loss_mask: List[int],
         output_ids: List[int],
         new_obs: ConversationType,
+        done: bool,
     ):
         """
         Update the chat history, loss mask, and input ids given a new model response and observation.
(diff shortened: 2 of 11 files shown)


skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -52,6 +52,40 @@ def __init__(
...
+        # `_get_next_input_ids_with_multiturn_chat_template()` will be `\n<|im_start|>user\nObservation here<|im_end|>\n`.
+        # Note the `\n` at the final assistant turn will still be missing, but this is fine.
+        # See tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py for more details.
+        if self.tokenizer.eos_token_id in self.base_conversation_token_ids:
+            last_eos_token_index = (
+                len(self.base_conversation_token_ids)
+                - 1
+                - self.base_conversation_token_ids[::-1].index(self.tokenizer.eos_token_id)
+            )
+            self.base_conversation_token_ids = self.base_conversation_token_ids[: last_eos_token_index + 1]
```
CharlieFRuan, line 86: Want to highlight this change, slightly unsure about this

skyrl-train/skyrl_train/inference_engines/remote_inference_engine.py
```diff
@@ -29,45 +31,87 @@ def __init__(
...
-                    output_tasks.append(session.post(f"{self.url}/v1/chat/completions", json=payload, headers=headers))
-            else:  # prompt_token_ids is not None
+
+            if self.engine_backend == "vllm":
+                # vLLM does not support /generate, use /completions instead
+                payload = sampling_params.copy()
+                payload["model"] = self.model_name
                 for p_ids in prompt_token_ids:
                     payload["prompt"] = p_ids
                     output_tasks.append(session.post(f"{self.url}/v1/completions", json=payload, headers=headers))
```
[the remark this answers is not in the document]
CharlieFRuan: Wow I did not know this, good catch gemini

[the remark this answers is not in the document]
CharlieFRuan: @gemini-code-assist two questions:
- If I pass multiple prompts in the `prompt` field to vLLM's `/v1/completions`, does the returned list of `CompletionResponse` follow the same order of the `prompt`, being request-response 1-to-1?
- Is it the same for SGLang's `/generate` endpoint?

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -523,16 +587,17 @@ def _update_engine_input_token_ids(
...
         Args:
-            output: str
+            output_ids: List[int]
             new_obs: ConversationType
             loss_mask: List[int]
             input_ids: List[int]
         Returns:
             loss_mask: List[int]
             input_ids: List[int]
+            logprobs: Optional[List[int]]
```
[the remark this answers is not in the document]
CharlieFRuan: Fixed

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -379,18 +445,53 @@ def _zero_reward_if_not_stop(self, rewards: List[float], stop_reasons: List[str]
...
-        input_ids: List[int],
         output: str,
         new_obs: ConversationType,
     ):
         """
         Update the chat history, loss mask, and input ids given a new model response and observation.
 
+        loss_mask is not maintained because we get it at the end of the trajectory with
+        `response_encodings["assistant_masks"]`.
+        """
```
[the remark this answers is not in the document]
CharlieFRuan: Fixed

skyrl-train/skyrl_train/inference_engines/remote_inference_engine.py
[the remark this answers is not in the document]
CharlieFRuan: Fixed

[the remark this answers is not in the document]
CharlieFRuan: @gemini-code-assist I just pushed a commit to address this, please take another look to see if it looks good

skyrl-train/skyrl_train/inference_engines/base.py
```diff
@@ -14,9 +14,15 @@ class InferenceEngineInput(TypedDict):
 
 
 class InferenceEngineOutput(TypedDict):
```
tyler-griggs, line 16: I have small nits on this paragraph just for brevity / clarity, but will draft something in next round of updates :)

skyrl-train/skyrl_train/inference_engines/remote_inference_engine.py
```diff
@@ -29,45 +30,88 @@ def __init__(
...
+                response = await resp.json()
+
+        # 3. Parse outputs
+        outputs = []
+        output_ids = []
+        finish_reasons = []
+
+        if self.engine_backend == "vllm":
+            for i, choice in enumerate(response.get("choices", [])):
+                # Since n=1, so choice of index i represents the output of `pompt[i]`
```
tyler-griggs: Nit: "Since n=1, index i represents the output for `prompt[i]`"

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -52,6 +52,40 @@ def __init__(
         if getattr(self.generator_cfg.sampling_params, "logprobs", None) is not None and not self.generator_cfg.batched:
             raise ValueError("`sampling_params.logprobs` should be `None` if `batched` is `False`")
```
tyler-griggs, line 54: nit: Updating the below paragraph for brevity:

`base_conversation` is used when `use_conversation_multi_turn==True and custom_chat_template==None` to correctly format and tokenize observations. 
Follows https://jybsuper.github.io/posts/multiturn_tokenization/#the-breakthrough-fixed-base-approach

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -52,6 +52,40 @@ def __init__(
...
 
+        # Code below is only for use_conversation_multi_turn==True and custom_chat_template==None
+        # This is used to format observation in `_get_next_input_ids_with_multiturn_chat_template()`.
+        # Similar to https://github.com/volcengine/verl/blob/6bbbff13a131dac9f29411f180ccabe2fe64e786/verl/experimental/agent_loop/tool_agent_loop.py#L196
+        # Follows https://jybsuper.github.io/posts/multiturn_tokenization/#the-breakthrough-fixed-base-approach
+        self.base_conversation = [
+            {"role": "system", "content": "You are a helpful assistant."},
+            {"role": "user", "content": "I am a user."},
+            {"role": "assistant", "content": "I am an assistant."},
```
tyler-griggs, line 61: qq: why include the assistant message here? Can you not just use system and user?

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -86,6 +130,7 @@ async def agent_loop(
         trajectory_id = uuid4().hex
         done = False
 
+        # Instantiate chat_history and chat_end_index, which are only used for retokenize_chat_history.
```
tyler-griggs: nit: "... only used if retokenize_chat_history==True"

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -94,28 +139,33 @@ async def agent_loop(
         chat_history, _ = env.init(chat_history)
         input_ids = self.tokenizer.apply_chat_template(
             chat_history,
-            # if we are keeping the chat history in token ids, we have to add the generation prompt to the original prompt
-            add_generation_prompt=not self.use_conversation_multi_turn,
+            # otherwise the prompt_ids and response_ids will both include the generation prompt
```
tyler-griggs: nit: "If retokenize_chat_history==True, we avoid including the generation prompt in both the prompt_ids and response_ids"

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -379,18 +445,59 @@ def _zero_reward_if_not_stop(self, rewards: List[float], stop_reasons: List[str]
...
         """
         Update the chat history, loss mask, and input ids given a new model response and observation.
 
+        loss_mask is not maintained because we get it at the end of the trajectory with
+        `response_encodings["assistant_masks"]`.
+
+        Returns:
+            chat_history: The updated chat history.
+            chat_end_index: The updated chat end index.
+            input_ids: The new input IDs after tokenizing the chat history, mainly used to check if
```
tyler-griggs: nit: I'd remove "mainly used to check if the input length is too long for next turn" because the use can always change. It makes sense that this would return `input_ids` so I don't think an explanation is necessary.

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -379,18 +445,59 @@ def _zero_reward_if_not_stop(self, rewards: List[float], stop_reasons: List[str]
...
+    def _get_next_input_ids_with_multiturn_chat_template(
+        self,
+        input_ids: List[int],
+        loss_mask: List[int],
+        output_ids: List[int],
+        new_obs: ConversationType,
+        done: bool,
+    ):
+        """
+        Update the chat history, loss mask, and input ids given a new model response and observation.
```
tyler-griggs: nit: this is the same description as the previous method -- should we change this to add at least some (very brief) explanation of the difference / use-case?

skyrl-train/skyrl_train/entrypoints/main_base.py
```diff
@@ -61,12 +61,13 @@ def create_ray_wrapped_inference_engines_from_config(cfg: DictConfig, colocate_p
     )
 
 
-def create_remote_inference_engines_from_config(cfg: DictConfig):
+def create_remote_inference_engines_from_config(cfg: DictConfig, tokenizer):
```
tyler-griggs: nit: can you type hint?

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -52,6 +52,40 @@ def __init__(
...
+        # <|im_start|>user\nI am a user.<|im_end|>\n
+        # <|im_start|>assistant\nI am an assistant.<|im_end|>\n
+
+        # Note that there is a `\n` in assistant's line before next user's messsage starts.
+        # If we do token-in-token-out, there is no way for LLM engine to generate `\n` since the
+        # EOS token is `<|im_end|>`. Therefore, we need to add it during `observation_ids`.
+        # By cutting `\n` out in `base_conversation_token_ids`, `observation_ids` in
+        # `_get_next_input_ids_with_multiturn_chat_template()` will be `\n<|im_start|>user\nObservation here<|im_end|>\n`.
+        # Note the `\n` at the final assistant turn will still be missing, but this is fine.
+        # See tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py for more details.
```
tyler-griggs, line 79: These explanations are important and useful, but I actually think they should be in a doc with a single comment linking to them. They do add quite a lot of blot to the fast-growing generator file :D I think the skyrl gym generator has reached sufficient complexity to add a doc detailing it's behavior. At the least, we could start with a briefer doc on "multi-turn chat tokenization" and discuss the three possible methods you have here. What do you think?

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
tyler-griggs: (You answered offline, but could be good to add a brief explanation in this comment just for visibility)

CharlieFRuan: Added in the comments

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
CharlieFRuan: Yes, a doc should definitely be needed. I will add that to a follow-up PR

tyler-griggs: Got it, thanks! Once added, can you please remove some of the comments here and just link to the doc?

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -52,6 +52,39 @@ def __init__(
         if getattr(self.generator_cfg.sampling_params, "logprobs", None) is not None and not self.generator_cfg.batched:
             raise ValueError("`sampling_params.logprobs` should be `None` if `batched` is `False`")
 
+        # base_conversation is used when `use_conversation_multi_turn==True and custom_chat_template==None` to
+        # correctly format and tokenize observations. Using both user and assistant messages as most observations are user messages.
```
tyler-griggs: nit: I feel like you may have added "Using both user and assistant messages as most observations are user messages." because of my question, but I actually think adding the comment here makes it a little confusing because it's not clear what is meant by "Using". I'd recommend just removing this line, and our doc (and the one linked) should be the source of truth for why base conversation is needed and how it works.

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -126,15 +175,31 @@ async def agent_loop(
             done = env_step_output["done"]
 
             if env_step_output.get("postprocessed_action", None) is not None:
+                # TODO(Charlie): come back to this, we should deprecate postprocessed action
+                print(
+                    "WARNING: postprocessed action may violate token-in-token-out. Ideally you "
+                    "post-process it in the token space rather than string space, and you can do "
+                    "it in SkyRLGymGenerator."
```
tyler-griggs: nit: what do you mean by "you can do it in SkyRLGymGenerator"? Do you mean you can do token-based post-processing in SkyRLGymGenerator? Because we don't actually have a hook for that, so someone would have to customize the generator (ie, copy and paste the file, then make edits), right?

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -143,8 +208,9 @@ async def agent_loop(
 
         env.close()  # does nothing for now
 
+        # TODO(Charlie): this makes the prompt_ids include the generation prompt, is this what we want?
```
tyler-griggs: Yes I believe it is. Primarily because maintaining the generation prompt is "on policy". If we remove those tokens, it will not actually reflect what the model observed when making its generation.

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -126,15 +175,31 @@ async def agent_loop(
             done = env_step_output["done"]
 
             if env_step_output.get("postprocessed_action", None) is not None:
+                # TODO(Charlie): come back to this, we should deprecate postprocessed action
+                print(
+                    "WARNING: postprocessed action may violate token-in-token-out. Ideally you "
+                    "post-process it in the token space rather than string space, and you can do "
+                    "it in SkyRLGymGenerator."
+                )
```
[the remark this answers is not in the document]
CharlieFRuan: will replace print with logger for the entire project in a future PR

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -428,78 +532,42 @@ def _update_engine_input_chat_history(
...
+            # is generated due to `self.base_conversation_token_ids[: last_eos_token_index + 1]` in `__init__()`.
+            observation_ids = self.tokenizer.apply_chat_template(
+                [*self.base_conversation, *new_obs],
+                add_generation_prompt=True,
+                tokenize=True,
+            )[len(self.base_conversation_token_ids) :]
+            input_ids += observation_ids
+            loss_mask += [0] * len(observation_ids)
+        else:
+            # TODO(Charlie): or assert done? i.e. can we still prompt another turn if no observation?
```
tyler-griggs: I think that's fine. Want to remove TODO, or are you wanting to look into it further?

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator.py
```diff
@@ -11,9 +11,19 @@
...
+
+
 # TODO (erictang000): clean up the mocking for tests in this file
 @pytest.fixture
 def mock_tokenizer():
+    """
+    A mock tokenizer that encodes any non-empty string to `[1,2,3,4]`.
+    For chat template, if `tokenize=False`, concatenate the content of each message.
+    If `tokenize=True`, return `[1,2,3,4]` for each message.
+    """
```
tyler-griggs, line 26: nit: update [1,2,3,4] to MOCK_TOKENIZER_ENCODED_IDS? Just in case `MOCK_TOKENIZER_ENCODED_IDS` gets modified later.

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator.py
```diff
@@ -438,38 +460,36 @@ def mock_step_never_done(output):
             metadata={},
         )
 
+    # We start with initial prompt len 4 due to mock_apply_chat_template
+    # Each turn, obesrvation is 4 tokens due to mock_encode
```
tyler-griggs: very nitty nit: obesrvation -> observation

skyrl-train/tests/gpu/test_skyrl_gym_generator.py
```diff
@@ -205,8 +209,8 @@ async def run_generator_end_to_end(
 @pytest.mark.parametrize(
     ("use_async_engine", "batched", "n_samples_per_prompt", "num_inference_engines", "tensor_parallel_size"),
     [
-        (False, True, 5, 2, 1),
-        (True, False, 5, 1, 2),
+        (False, True, 5, 1, 1),  # tests SkyRLGymGenerator.generate_batched for single-turn
+        (True, False, 5, 1, 1),  # tests SkyRLGymGenerator.agent_loop for single-turn
```
tyler-griggs: qq: Why reduce these to 1? Just to make the test lighter-weight? Running with multiple inference engines and TP>1 can catch sneaky errors, but I guess we should expect other tests (e.g., the generation and inference engine tests) to catch this.

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
CharlieFRuan: yep will do!

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
CharlieFRuan: Ah yes, they'd have to hack SkyRLGymGenerator to do token-space post-processings. I'll say solutions coming soon instead then

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
CharlieFRuan: Before this PR the generation prompt is in `response_ids` (with proper maskings). After this PR it is in `prompt_ids`, hence the comment here. I am guessing it is equivalent?

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
CharlieFRuan: Not really sure what to further look into. I'll just remove the TODO and keep it as is then

skyrl-train/tests/gpu/test_skyrl_gym_generator.py
CharlieFRuan: Oh right, I'll change it back. Forgot to revert this change when running it in my 1xGPU machine

CharlieFRuan:
(Will in-place edit this comment / PR description to add full results and details when runs finish.)

E2E run with search-r1 with 4 turns.

`use_conversation_multi_turn == False` (codepath 2.3):
image

`use_conversation_multi_turn == True` (codepath 2.2):
image

tyler-griggs (approved):
amazing

Issue #147, opened 2025-08-13
Support ReasoningGym


tyler-griggs:
The reasoning gym repo (https://github.com/open-thought/reasoning-gym) contains a large set of verifiable reasoning environments for training. This issue tracks support for training in ReasoningGym environments on top of SkyRL.

I expect this integration to largely require:

1. Handling the dataset format. ReasoningGym uses procedurally generated datasets (e.g., generated game maps), whereas SkyRL currently expects a complete dataset handed to the training stack at the start of training.
2. Handling scoring of the completed generations. ReasoningGym already provides methods to score the model's output on some task, and this method should be used to compute rewards.

Reasoning Gym has example integrations [with other training stacks](https://github.com/open-thought/reasoning-gym/tree/main/examples) that will serve as useful references. 

## TODOs
- [ ] Create a `ReasoningGymDataset` class that formats the procedurally generated ReasoningGym datasets into the format that SkyRL supports. See the ReasoningGym+Verl integration [example](https://github.com/open-thought/reasoning-gym/blob/main/examples/veRL/grpo_train.py#L27) for reference, and see SkyRL's [dataset format docs](https://skyrl.readthedocs.io/en/latest/datasets/dataset-preparation.html).
- [ ] Use the scoring methods provided by ReasoningGym to compute the rewards

tyler-griggs:
@SASHASINGH66 is picking this up!

PR #160, opened 2025-08-18
Reasoning gym integration

SASHASINGH66:
### What does this PR do

This PR creates a `ReasoningGymDataset` class that transforms Reasoning Gym's procedurally generated datasets into the format that SkyRL expects. Also created an `Env` to support training using Reasoning Gym datasets and their built-in `score_answer` method.

I also added a script to generate training and validation datasets as well as bash script to allow users to train a model using their Reasoning Gym datasets.

### Issue
Issue: https://github.com/NovaSky-AI/SkyRL/issues/147

### Testing
I was able to successfully run training using my script

8C1BE9B4-A622-4659-BDB6-420AE0A93019_1_201_a
Added logs to see that the `_get_reward` function was correctly returning a reward using the in-built `score_answer` function, in this example, the correct response was 92 but the model responded with 27 so the reward was 0.


tyler-griggs:
This is great! Could you add a folder `reasoning_gym` under `skyrl_train/examples` that holds all files needed to actually launch a training run with a `reasoning_gym` task? E.g., we'll want a `run_reasoning_gym.sh` (similar to `run_gsm8k.sh`) for actually launching training. Ideally a user can run 1-2 commands (e.g., 1 command to generate a dataset if needed, 1 more to launch training) and get started with training.

skyrl-gym/skyrl_gym/envs/reasoning_gym/dataset.py
```diff
@@ -0,0 +1,197 @@
...
+
+from typing import Optional, Dict, Any, Union
+from datasets import Dataset, Features, Value, LargeList
+
+import reasoning_gym
+import json
+from reasoning_gym.coaching.experiment import Experiment
+from reasoning_gym.dataset import ProceduralDataset
+from reasoning_gym.utils import extract_answer
+REASONING_GYM_AVAILABLE = True
```
tyler-griggs: nit: is this used, or can we remove it?

skyrl-gym/skyrl_gym/envs/reasoning_gym/dataset.py
```diff
@@ -0,0 +1,184 @@
...
+            prompt.append({"role": "user", "content": question})
+
+            # Extract ground truth answer
+            ground_truth = entry.get("answer", "")
+            if not ground_truth:
+                solution = entry.get("solution", "")
+                if solution:
+                    try:
+                        ground_truth = extract_answer(solution, tag_name="answer")
+                    except Exception:
```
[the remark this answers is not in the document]
tyler-griggs: ^^ This is a fair point. Did you add this `try`/`except` because you observed occasional errors in `extract_answer`? If so, it could be good to catch those errors more specifically. At the least, we should probably be logging / printing a warning. If answers are silently not being extracted properly, that may be surprising/confusing to a user.

skyrl-gym/skyrl_gym/envs/reasoning_gym/dataset.py
```diff
@@ -0,0 +1,184 @@
...
+            found_answer = extract_answer(model_output, tag_name="answer")
+        except Exception:
+            found_answer = model_output
+
+        if hasattr(self.data_source, "score_answer"):
+            try:
+                reward = self.data_source.score_answer(found_answer, entry=original_entry)
+                return float(reward)
+            except Exception:
+                pass
```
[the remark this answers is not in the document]
tyler-griggs: Similar to the comment above. Is there a reason you added a `try`/`except` here? Did you observe errors? If not, do we need this? If so, we should at least log the errors.

skyrl-gym/skyrl_gym/envs/reasoning_gym/dataset.py
```diff
@@ -0,0 +1,184 @@
...
+        Write the SkyRL-formatted dataset to a Parquet file.
+        
+        Args:
+            path (str): The file path to write the dataset to.
+        """
+        self.skyrl_dataset.to_parquet(path)
+
+
+def make_reasoning_gym_dataset(
+    data_source,
```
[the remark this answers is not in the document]
tyler-griggs: ^^ Yes please

skyrl-gym/skyrl_gym/envs/reasoning_gym/env.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+    def _get_reward(self, action: str) -> float:
+        """
+        Calculate reward using ReasoningGym's built-in scoring logic.
+        """
+        try:
+            found_answer = extract_answer(action, tag_name="answer")
+        except Exception:
+            found_answer = action
```
tyler-griggs, line 43: Same comment as above.

skyrl-gym/skyrl_gym/envs/reasoning_gym/env.py
```diff
@@ -0,0 +1,64 @@
...
+                return float(reward)
+            except Exception as e:
+                return 0.0
+        
+    def step(self, action: str) -> BaseTextEnvStepOutput:
+        """
+        Process one step in the reasoning environment.
+        For reasoning tasks, we typically complete in one step.
+        """
+        done = True  # Most reasoning tasks complete in one step
```
tyler-griggs, line 54: Are ReasoningGym environments only single-turn / single-step? If not, then we will want to support multi-turn (can see the other envs for examples of this). I'd also take a look at the other ReasoningGym integrations (Verifiers, Verl, etc.) to see whether multi-turn is needed here.

skyrl-train/examples/reasoning_gym/run_reasoning_gym.sh
```diff
@@ -0,0 +1,44 @@
+export INFERENCE_BACKEND="vllm"
+export DATA_DIR="$HOME/data/reasoning_gym"
+export LOGGER="console"
+
```
tyler-griggs, line 4: nit: Could you please add a brief comment on what's launched by this script (see `run_gsm8k.sh` for an example)?

tyler-griggs (commented):
Thanks for the updates! The two main points are:

a) Try to minimize the use of `try`/`except` because it can silence errors that should be raised.

b) Determine whether reasoning gym requires multi-turn

skyrl-gym/skyrl_gym/envs/reasoning_gym/env.py
SASHASINGH66: OOh good catch! Yeah I looked into the examples again and it does seem like ReasoningGym environments are only single-turn/single-step. In particular, looking at the [Verl](https://github.com/open-thought/reasoning-gym/blob/02b7fac86358f7ef6239608b0b738a5a03ecfe9e/examples/veRL/grpo_train.py#L188) example, it seems like the score is only computed once for each response and returned

skyrl-gym/skyrl_gym/envs/reasoning_gym/dataset.py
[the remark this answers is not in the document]
SASHASINGH66: Yep, I think we should do try/except here since it's possible for the solution to not have <answer></answer> tags in which case extract_answer throws an exception. I'll add logging to capture this when the exception is thrown

skyrl-train/examples/reasoning_gym/run_reasoning_gym.sh
SASHASINGH66: yep!

skyrl-train/examples/reasoning_gym/run_reasoning_gym.sh
```diff
@@ -0,0 +1,49 @@
...
+# PPO training+generation for Qwen2.5-0.5B-Instruct on ReasoningGym.
+
+# uv run examples/reasoning_gym/reasoning_gym_dataset.py --filepath $HOME/data/gsm8k
```
tyler-griggs: nit: I believe the path is `/dataset.py`, and the argument is `file_path`

skyrl-gym/skyrl_gym/envs/reasoning_gym/dataset.py
```diff
@@ -0,0 +1,190 @@
...
+            prompt.append({"role": "user", "content": question})
+
+            ground_truth = entry.get("answer", "")
+            if not ground_truth:
+                solution = entry.get("solution", "")
+                try:
+                    ground_truth = extract_answer(solution, tag_name="answer")
+                except Exception as e:
+                    print(f"Warning: Error extracting answer between <answer></answer> tags from solution, using entire solution as ground truth: {e}")
+                    ground_truth = solution.strip()
```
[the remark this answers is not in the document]
SASHASINGH66: seems like print is generally used throughout the codebase! so resolving this one

skyrl-gym/skyrl_gym/envs/reasoning_gym/test_reasoning_gym.py
```diff
@@ -0,0 +1,56 @@
...
+        procedural_dataset=dataset,
+        developer_prompt="You are a helpful assistant. Solve the problem step by step and provide the final answer inside <answer> tags."
+    )
+    example = reasoning_gym_dataset[0]
+    print("Question:", example["extra_info"]["question"])
+    print("Ground Truth:", example["reward_spec"]["ground_truth"])
+    partial_output = "The total number of legs is 99. Final answer: <answer>99</answer>"
+    print("score_generation: ", reasoning_gym_dataset.score_generation(0, partial_output))
+
+def test_spiral_matrix():
```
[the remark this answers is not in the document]
SASHASINGH66: fixed

skyrl-gym/skyrl_gym/envs/reasoning_gym/env.py
```diff
@@ -0,0 +1,62 @@
...
+        """
+        found_answer = extract_answer(action, tag_name="answer")
+
+        if self.reasoning_gym_data_source and hasattr(self.reasoning_gym_data_source, 'score_answer'):
+            try:
+                reward = self.reasoning_gym_data_source.score_answer(found_answer, entry=self.original_entry)
+                return float(reward)
+            except Exception as e:
+                print("Warning: Error scoring answer, returning 0.0")
+                return 0.0
```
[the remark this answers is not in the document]
SASHASINGH66: fixed to print the exception as well

skyrl-train/examples/reasoning_gym/run_reasoning_gym.sh
```diff
@@ -0,0 +1,49 @@
...
+# --dataset_name leg_counting --size 10000 --developer_prompt "You are a helpful assistant that can solve problems. Place your final answer between <answer></answer> tags."
```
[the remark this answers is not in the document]
SASHASINGH66: fixed!

skyrl-train/examples/reasoning_gym/run_reasoning_gym.sh
```diff
@@ -0,0 +1,49 @@
...
+  generator.async_engine=true \
+  generator.batched=true \
+  environment.env_class=gsm8k \
+  generator.n_samples_per_prompt=2 \
+  generator.gpu_memory_utilization=0.8 \
+  trainer.logger="$LOGGER" \
+  trainer.project_name="gsm8k" \
+  trainer.run_name="gsm8k_test_small" \
+  trainer.resume_mode=null \
+  trainer.ckpt_path="$HOME/ckpts/gsm8k_0.5B_ckpt" \
```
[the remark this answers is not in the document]
SASHASINGH66: fixed!

skyrl-gym/skyrl_gym/envs/reasoning_gym/dataset.py
```diff
@@ -0,0 +1,190 @@
...
+    def score_generation(self, idx: int, model_output: str) -> float:
+        """Score a single generation using ReasoningGym."""
+        skyrl_entry = self.skyrl_dataset[idx]
+        original_entry = skyrl_entry["extra_info"]["dataset_entry"]
+        original_entry = json.loads(original_entry)
+        try:
+            found_answer = extract_answer(model_output, tag_name="answer")
+        except Exception as e:
+            print(f"Warning: Error extracting answer between <answer></answer> tags from model output, scoring the entire model output: {e}")
+            found_answer = model_output
```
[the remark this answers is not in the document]
SASHASINGH66: seems like print is generally used throughout the codebase!

skyrl-train/examples/reasoning_gym/run_reasoning_gym.sh
```diff
@@ -0,0 +1,49 @@
...
+  trainer.max_prompt_length=512 \
+  generator.sampling_params.max_generate_length=256 \
+  trainer.policy.optimizer_config.lr=5.0e-6 \
+  trainer.algorithm.use_kl_loss=true \
+  generator.backend=$INFERENCE_BACKEND \
+  generator.run_engines_locally=true \
+  generator.weight_sync_backend=nccl \
+  generator.async_engine=true \
+  generator.batched=true \
+  environment.env_class=gsm8k \
```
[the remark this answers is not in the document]
SASHASINGH66: fixed!

skyrl-gym/skyrl_gym/envs/reasoning_gym/env.py
```diff
@@ -0,0 +1,62 @@
...
+
+        assert "extra_info" in extras, "extra_info field is required"
+        self.extra_info = extras["extra_info"]
+        self.data_source = extras["data_source"]
+        
+        dataset_name = self.extra_info.get("dataset_name")
+        size = self.extra_info.get("size")
+        seed = self.extra_info.get("seed")
+        
+        self.reasoning_gym_data_source = create_dataset(dataset_name, size=size, seed=seed)
```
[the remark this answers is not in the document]
SASHASINGH66: this is a good callout, unfortunately storing the actual dataset itself in the SkyRL dataset entries is not very easy to do, so im working on a fix for this right now

skyrl-gym/skyrl_gym/envs/reasoning_gym/env.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+
+    def __init__(self, env_config: DictConfig, extras: Dict[str, Any] = {}):
+        super().__init__()
+
+        assert "extra_info" in extras, "extra_info field is required"
+        self.extra_info = extras["extra_info"]        
+        
+        self.registry_key = self.extra_info.get("registry_key")
+        self.reasoning_gym_data_source = get_dataset_from_registry(self.registry_key)
```
SASHASINGH66: did this kinda hacky fix to not have to recreate the dataset every time the environment is instantiated

skyrl-gym/skyrl_gym/envs/reasoning_gym/dataset.py
```diff
@@ -0,0 +1,203 @@
+"""ReasoningGym dataset integration for SkyRL training."""
+
+from typing import Optional, Dict, Any, Union
+from datasets import Dataset, Features, Value, LargeList
+
+import reasoning_gym
```
lynnliu030: This is not used, can be removed?

lynnliu030 (commented):
Leave some minor comment; btw @tyler-griggs shall the `reasoning_gym` dependency be included into the pyproject? would be good to have some instructions and doc for this too.

skyrl-gym/skyrl_gym/envs/reasoning_gym/env.py
```diff
@@ -0,0 +1,70 @@
...
+
+        try:
+            self.original_entry = json.loads(self.extra_info["dataset_entry"])
+        except (json.JSONDecodeError, TypeError):
+            self.original_entry = self.extra_info["dataset_entry"]
+
+
+    def _get_reward(self, action: str) -> float:
```
tyler-griggs, line 34: Hi @SASHASINGH66 , could you briefly explain how scoring works here? What are the two different scoring methods here? What's the difference between `reasoning_gym_dataset` and `reasoning_gym_data_source`? I'm also trying to understand this first check for `hasattr(self, 'reasoning_gym_dataset')` -- I don't see `self.reasoning_gym_dataset` ever being set, where is it expected to be initialized? 

These may all just be reasoning gym concepts, but I wanted to understand better!

SASHASINGH66: to score reasoning gym tasks, i'm using the input (procedurally generated) reasoning gym dataset's built-in `score_answer` method (example in the first code block under the [quickstart](https://github.com/open-thought/reasoning-gym) section) 

the hasattr(self, 'reasoning_gym_dataset') shouldn't be there, that was from before when i was generating the whole original RG dataset in the env and using the dataset's `score_answer` method (removing this and pushing again now)

the new logic starts at `if self.reasoning_gym_data_source`, `self.reasoning_gym_data_source` here is a reference to the ReasoningGymDataset class instance, and im using the instance's `score_generation` method which just calls the originally passed in RG dataset's `score_answer` method (ill rename `reasoning_gym_data_source` to `reasoning_gym_skyrl_dataset` for clarity)

skyrl-train/examples/reasoning_gym/run_reasoning_gym.sh
SASHASINGH66: i think i fixed this in the latest commit!

skyrl-gym/skyrl_gym/envs/reasoning_gym/dataset.py
```diff
@@ -0,0 +1,193 @@
...
+            }),
+            "extra_info": Features({
+                "index": Value("int32"),
+                "dataset_entry": Value("string"),
+                "question": Value("string"),
+                "solution": Value("string"),
+                "dataset_name": Value("string"),
+                "size": Value("int32"),
+                "seed": Value("int32"),
+                "data_source_serialized": Value("string"),  # Store serialized data source
```
SASHASINGH66, line 117: storing the serialized procudurally generated dataset in the SkyRL entry so that it can be used in the env class for reward calculation

skyrl-gym/skyrl_gym/envs/reasoning_gym/env.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(self, env_config: DictConfig, extras: Dict[str, Any] = {}):
+        super().__init__()
+
+        assert "extra_info" in extras, "extra_info field is required"
+        self.extra_info = extras["extra_info"]
+        
+        
+        # Deserialize the data source from the stored serialized version
+        data_source_serialized = self.extra_info.get("data_source_serialized")
+        self.reasoning_gym_data_source = pickle.loads(base64.b64decode(data_source_serialized))
```
SASHASINGH66, line 26: deserializing the procedural dataset here to call it's built-in `score_answer` method for reward calculation

skyrl-gym/skyrl_gym/envs/reasoning_gym/dataset.py
SASHASINGH66: yep!

skyrl-gym/skyrl_gym/envs/reasoning_gym/env.py
```diff
@@ -0,0 +1,63 @@
...
```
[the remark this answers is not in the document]
SASHASINGH66: i don't think JSON can be used here, since the ReasoningGym object isn't a python serializable object, @tyler-griggs do you think we should worry about security here, or is the current state fine?

skyrl-gym/skyrl_gym/envs/reasoning_gym/dataset.py
```diff
@@ -0,0 +1,190 @@
...
+                "method": Value("string"),
+                "ground_truth": Value("string"),
+            }),
+            "extra_info": Features({
+                "index": Value("int32"),
+                "dataset_entry": Value("string"),
+                "question": Value("string"),
+                "solution": Value("string"),
+                "data_source_serialized": Value("string"),  # Store serialized data source
+            })
```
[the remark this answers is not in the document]
SASHASINGH66: fixed

skyrl-train/examples/reasoning_gym/generate_datasets.py
```diff
@@ -0,0 +1,58 @@
...
+    
+    os.makedirs(output_dir, exist_ok=True)
+    
+    training_dataset = generate_dataset(dataset_name, size=size, developer_prompt=developer_prompt)
+    training_dataset.write_to_parquet(os.path.join(output_dir, "train.parquet"))
+    print(f"Dataset with {len(training_dataset)} examples saved to {output_dir}/train.parquet")
+
+    validation_dataset = generate_dataset(dataset_name, size=size, developer_prompt=developer_prompt)
+    validation_dataset.write_to_parquet(os.path.join(output_dir, "validation.parquet"))
+    print(f"Dataset with {len(validation_dataset)} examples saved to {output_dir}/validation.parquet")
```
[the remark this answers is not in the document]
SASHASINGH66: fixed, allowing users to optionally specify seeds for both training and validation sets

skyrl-gym/skyrl_gym/envs/reasoning_gym/env.py
```diff
@@ -0,0 +1,63 @@
...
+        Calculate reward using ReasoningGym's built-in scoring logic.
+        """
+        
+        try:
+            found_answer = extract_answer(action, tag_name="answer")
+        except Exception as e:
+            print(f"Warning: Error extracting answer between <answer></answer> tags from model output, scoring the entire model output: {e}")
+            found_answer = action
+
+        reward = self.reasoning_gym_data_source.score_answer(found_answer, entry=self.original_entry)
```
SASHASINGH66, line 45: following up on your question about scoring here @tyler-griggs since i've changed it a bit

im now serializing the original procedurally generated dataset in the `ReasoningGymDataset` class and storing that serialized object in the skyRL dataset entry. Then I deserialize that object in this `Env` class and use Reasoning Gym's in-built `score_answer` method to get the model score.

skyrl-gym/skyrl_gym/envs/reasoning_gym/dataset.py
```diff
@@ -0,0 +1,184 @@
...
+                try:
+                    ground_truth = extract_answer(solution, tag_name="answer")
+                except Exception as e:
+                    print(f"Warning: Error extracting answer between <answer></answer> tags from solution, using entire solution as ground truth: {e}")
+                    ground_truth = solution.strip()
+
+
+            entry_str = json.dumps(entry)
+            skyrl_entry = {
+                "data_source": "reasoning_gym",
```
tyler-griggs, line 80: Please make this "reasoning_gym" + self.dataset_name so that the distinction is clear between environments within reasoning_gym. For context, this will be displayed in wandb or the console for eval metrics (ie, we break evals down per data source)

skyrl-train/examples/reasoning_gym/generate_datasets.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+    
+    return rg
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--dataset_name", type=str, default="leg_counting")
+    parser.add_argument("--size", type=int, default=200)
+    parser.add_argument("--developer_prompt", type=str, default="Think step by step. Return ONLY the final numeric answer wrapped in <answer> tags, like <answer>42</answer>.")
+    parser.add_argument("--file_path", type=str, default="~/data/reasoning_gym")
```
tyler-griggs, line 59: Please update this to `output_dir` to be consistent with other files (and update everywhere this is used, such as run_reasoning_gym.sh).

skyrl-train/examples/reasoning_gym/run_reasoning_gym.sh
```diff
@@ -0,0 +1,49 @@
+export INFERENCE_BACKEND="vllm"
```
tyler-griggs, line 1: Please reorganize the top of this file to:

```
set -x

# PPO training+generation for Qwen2.5-0.5B-Instruct on ReasoningGym.

# uv run --with reasoning_gym -- examples/reasoning_gym/generate_datasets.py --output_dir $HOME/data/reasoning_gym \
# --dataset_name leg_counting --size 10000 --developer_prompt "You are a helpful assistant that can solve problems. Place your final answer between <answer></answer> tags."
# bash examples/reasoning_gym/run_reasoning_gym.sh

INFERENCE_BACKEND="vllm"
DATA_DIR="$HOME/data/reasoning_gym"
LOGGER="console"
NUM_GPUS=1
```

The `set -x` prints out the bash commands, which is helpful for the logs. Also, the env vars are what users will most likely modify, so it's nice to have it right next to the `uv run` command

Also, please plumb `NUM_GPUS` to all the relevant fields (num_inference_engines, num_gpus_per_node)

skyrl-train/examples/reasoning_gym/run_reasoning_gym.sh
```diff
@@ -0,0 +1,49 @@
...
+export DATA_DIR="$HOME/data/reasoning_gym"
+export LOGGER="console"
+
+# PPO training+generation for Qwen2.5-0.5B-Instruct on ReasoningGym.
+
+# uv run examples/reasoning_gym/generate_datasets.py --file_path $HOME/data/reasoning_gym
+# --dataset_name leg_counting --size 10000 --developer_prompt "You are a helpful assistant that can solve problems. Place your final answer between <answer></answer> tags."
+# bash examples/reasoning_gym/run_reasoning_gym.sh
+
+uv run --isolated --extra $INFERENCE_BACKEND -m skyrl_train.entrypoints.main_base \
```
tyler-griggs, line 11: Users will get an error without `reasoning_gym` installed. We will need to add `--with reasoning_gym` to the uv command (note that I also added it to the comment above)

skyrl-train/examples/reasoning_gym/run_reasoning_gym.sh
```diff
@@ -0,0 +1,49 @@
...
+  trainer.placement.colocate_all=true \
+  trainer.strategy=fsdp2 \
+  trainer.placement.policy_num_gpus_per_node=1 \
+  trainer.placement.ref_num_gpus_per_node=1 \
+  generator.num_inference_engines=1 \
+  generator.inference_engine_tensor_parallel_size=1 \
+  trainer.epochs=1 \
+  trainer.eval_batch_size=32 \
+  trainer.eval_before_train=true \
+  trainer.eval_interval=1 \
```
tyler-griggs, line 25: Let's up this to 5

skyrl-train/examples/reasoning_gym/run_reasoning_gym.sh
```diff
@@ -0,0 +1,49 @@
...
+  trainer.update_epochs_per_batch=1 \
+  trainer.train_batch_size=32 \
+  trainer.policy_mini_batch_size=8 \
+  trainer.micro_forward_batch_size_per_gpu=4 \
+  trainer.micro_train_batch_size_per_gpu=4 \
+  trainer.ckpt_interval=1 \
```
tyler-griggs, line 31: And this down to -1 (ie, no checkpointing)

skyrl-train/examples/reasoning_gym/run_reasoning_gym.sh
```diff
@@ -0,0 +1,49 @@
...
+  trainer.placement.policy_num_gpus_per_node=1 \
+  trainer.placement.ref_num_gpus_per_node=1 \
+  generator.num_inference_engines=1 \
+  generator.inference_engine_tensor_parallel_size=1 \
+  trainer.epochs=1 \
+  trainer.eval_batch_size=32 \
+  trainer.eval_before_train=true \
+  trainer.eval_interval=1 \
+  trainer.update_epochs_per_batch=1 \
+  trainer.train_batch_size=32 \
```
tyler-griggs, line 27: In general, it'd be good to mostly match the settings from `run_gsm8k.sh` because it has reasonable defaults

skyrl-gym/skyrl_gym/envs/reasoning_gym/dataset.py
```diff
@@ -0,0 +1,184 @@
...
+                "reward_spec": {
+                    "method": "rule",
+                    "ground_truth": ground_truth,
+                },
+                "extra_info": {
+                    "index": idx,
+                    "dataset_entry": entry_str,
+                    "question": question,
+                    "solution": entry.get("solution", ""),
+                    "data_source_serialized": base64.b64encode(pickle.dumps(self.data_source)).decode('utf-8'),
```
tyler-griggs, line 92: Ah I see now why you had to serialize. We need to get this object from the trainer to the generator and into the environment to do scoring. There is an ongoing effort to make it easier to use datasets that are "owned" by the generator instead of the trainer, and this falls under that bucket, but it isn't ready yet. 

In the mean time, I would suggest an alternative. I'd suggest you do *not* serialize/pickle the data_source and put it in the dataset. Instead, I'd recommend adding something like the following to `env.py`:

```
from functools import lru_cache
@lru_cache(maxsize=None)
def _get_dataset_for_scoring(dataset_name: str):
    return create_dataset(dataset_name, size=1, seed=1)

def _score_answer(dataset_name: str, answer: str, entry: Any) -> float:
    ds = _get_dataset_for_scoring(dataset_name)
    return float(ds.score_answer(answer, entry=entry))
```

Then, use this to score the answer:
```
reward = _score_answer(self.dataset_name, found_answer, self.original_entry)
```

This is basically a singleton scoring object. I poked through the reasoning gym score_answer method and it seems like it is always independent of the seed and dataset size. So, we can just re-initialize the data_source in the env and use it to score (instead of passing the same data_source object from the trainer). This still isn't entirely clean, but it's slightly better. And we can resolve this when we make it easier for the generator to "own" the datset.

tyler-griggs: Note this will also require you to get the dataset_name out of data_source that's passed through `extras` in the env

skyrl-train/examples/reasoning_gym/generate_datasets.py
```diff
@@ -0,0 +1,64 @@
...
+        size=size,
+        seed=seed
+    )
+    
+    return rg
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--dataset_name", type=str, default="leg_counting")
+    parser.add_argument("--size", type=int, default=200)
```
tyler-griggs, line 57: Let's bump this default up, maybe to 2000

tyler-griggs:
Left a handful of mostly nits, I think it's in pretty good shape!

PR #161, opened 2025-08-19
[FlashRL 2/N] Support list of weights during weight sync for colocated training

SumanthRH:
# What does this PR do?

Supports a list of weights during weight sync for colocated training. During colocated training, we use CUDA IPC for weight syncing. The current impl is syncing weights param by param, which can be pretty inefficient. In this PR, we sycn tensors in batches of a configurable parameter (default 1GB). That is, we collect ipc metadata until the total size of underlying tensors is 1GB and forward to the inference engine. Each TP rank will materialize all tensors in this list  (i.e additional memory usage of 1GB here) and issue a single load_weights call. 

**How much faster is it?**

Even for a 14B model on a 8xH100 node (TP2), the weight sync time can reduce from around 4.4s to 1.6s (60% reduction). This will matter much more for larger models. 

This PR is needed for the FlashRL integration to work well, because we have a custom load weights impl that - long story short - allcoates new storage in each call and also issues some `empty_cache` calls. Without batching, the load weights call will be too slow in such cases. This PR reduces time for weight sync for a 1.5B model with flashrl from 5 mins to < 5s. 

I've tested the PR with our E2E tests for colocated and non-colocated and also tested the remote engine codepath. 

This PR also makes the following changes:
- Fixes bug introduced in #145 for the codepath with trajectory based routing when `response_ids` is not returned by the engine. 
- Fixes bug introduced in #126   for starting remote servers. import of `skyrl_train.utils.ppo_utils` will trigger registering. IN some cases, like with the vllm server init, we will not call `sync_registries` and there will be an error. The solution is to import guard `skyrl_train.utils.ppo_utils` unless the user themselves import it (for custom functions) or they go through the main entrypoint ( main -> `initialize_ray`-> sync)

TODO:
- [x] Verify non-colocated training works
- [x]  Run e2e test

### DIFF
--- a/skyrl-train/skyrl_train/config/ppo_base_config.yaml
+++ b/skyrl-train/skyrl_train/config/ppo_base_config.yaml
@@ -160,7 +160,7 @@ generator:
   backend: "vllm"
   weight_sync_backend: "nccl"
   # if using cuda_ipc, we send in batches of this size in GB
-  weight_transfer_threshold_cuda_ipc_in_GB: 1.0
+  weight_transfer_threshold_cuda_ipc_GB: 1.0
   inference_engine_tensor_parallel_size: 4
   n_samples_per_prompt: 5
   async_engine: true
--- a/skyrl-train/skyrl_train/config/ppo_base_config.yaml
+++ b/skyrl-train/skyrl_train/config/ppo_base_config.yaml
@@ -113,6 +113,7 @@ trainer:
       type: null # filter, replace, or null
       max_sample_batches: 30 # sample at most this many batches before stopping, -1 to sample forever
       min_replace_ratio: 0.3 # minimum proportion of good samples with which to replace bad samples (for replace strategy only)
+    
 
   gradient_checkpointing: true
   gradient_checkpointing_use_reentrant: false
@@ -156,6 +157,8 @@ generator:
   num_inference_engines: 1
   backend: "vllm"
   weight_sync_backend: "nccl"
+  # if using cuda_ipc, we send in batches of this size in GB
+  weight_transfer_threshold_cuda_ipc_in_GB: 0.5
   inference_engine_tensor_parallel_size: 4
   n_samples_per_prompt: 5
   async_engine: true
--- a/skyrl-train/skyrl_train/config/ppo_base_config.yaml
+++ b/skyrl-train/skyrl_train/config/ppo_base_config.yaml
@@ -158,7 +158,7 @@ generator:
   backend: "vllm"
   weight_sync_backend: "nccl"
   # if using cuda_ipc, we send in batches of this size in GB
-  weight_transfer_threshold_cuda_ipc_in_GB: 0.5
+  weight_transfer_threshold_cuda_ipc_in_GB: 1.0
   inference_engine_tensor_parallel_size: 4
   n_samples_per_prompt: 5
   async_engine: true
--- a/skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
+++ b/skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
@@ -103,7 +103,7 @@ class WorkerWrap:
             assert dtype == self.model_config.dtype, f"mismatch dtype: src {dtype}, dst {self.model_config.dtype}"
             weight = torch.empty(shape, dtype=dtype, device="cuda")
             torch.distributed.broadcast(weight, 0, group=self._model_update_group)
-            weight_list.append(((name, weight)))
+            weight_list.append((name, weight))
 
         self.model_runner.model.load_weights(weights=weight_list)
         for weight in weight_list:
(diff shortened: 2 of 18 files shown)


skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
```diff
@@ -95,37 +95,49 @@ def init_weight_update_communicator(
...
-        assert dtype == self.model_config.dtype, f"mismatch dtype: src {dtype}, dst {self.model_config.dtype}"
-        weight = torch.empty(shape, dtype=dtype, device="cuda")
-        torch.distributed.broadcast(weight, 0, group=self._model_update_group)
+        weight_list = []
+        for name, dtype, shape in zip(names, dtypes, shapes):
+            dtype = str_to_torch_dtype(dtype)
+            assert dtype == self.model_config.dtype, f"mismatch dtype: src {dtype}, dst {self.model_config.dtype}"
+            weight = torch.empty(shape, dtype=dtype, device="cuda")
+            torch.distributed.broadcast(weight, 0, group=self._model_update_group)
+            weight_list.append(((name, weight)))
```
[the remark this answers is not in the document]
SumanthRH: oh issue while porting code over. fixed

skyrl-train/skyrl_train/workers/deepspeed/deepspeed_worker.py
```diff
@@ -171,22 +176,31 @@ def gather_and_broadcast(param):
...
-                            )
+                        weights_update_request["names"].append(name)
+                        weights_update_request["dtypes"].append(self.cfg.generator.model_dtype)
+                        weights_update_request["shapes"].append(shape)
+                        weights_update_request["extras"].append(
+                            {
+                                "ipc_handles": ipc_handles,
+                            }
                         )
+                        if current_size / (1024**3) > self.cfg.generator.weight_transfer_threshold_cuda_ipc_in_GB:
```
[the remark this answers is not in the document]
SumanthRH: fixed!

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
```diff
@@ -350,22 +368,39 @@ async def init_weight_update_communicator(
             args=(master_addr, master_port, rank_offset, world_size, group_name, backend, override_existing),
         )
 
-    async def update_named_weight(self, request: NamedWeightUpdateRequest):
+    async def update_named_weights(self, request: NamedWeightsUpdateRequest):
+        if "names" not in request:
+            raise ValueError(f"Expected update weight request with 'names' entry, got keys: {request.keys()}")
+
+        if not len(request["names"]):
```
tyler-griggs, line 378: nit: this check is not added to `VLLMInferenceEngine`'s `update_named_weights` -- add there too?

skyrl-train/skyrl_train/workers/deepspeed/deepspeed_worker.py
```diff
@@ -149,11 +149,16 @@ def gather_and_broadcast(param):
...
                     await update_weight_task
+            torch.distributed.barrier()
+        # CUDA IPC
+        else:
+            from torch.multiprocessing.reductions import reduce_tensor
 
-            # CUDA IPC
-            else:
-                from torch.multiprocessing.reductions import reduce_tensor
+            # we sent in batches of 1GB as an optimization
```
tyler-griggs: nits:

sent --> send

And it is in batches of `weight_transfer_threshold_cuda_ipc_in_GB`, not 1gb

skyrl-train/skyrl_train/workers/fsdp/fsdp_worker.py
```diff
@@ -135,19 +135,23 @@ def gather_and_broadcast(param):
                 await asyncio.to_thread(gather_and_broadcast, param)
                 if torch.distributed.get_rank() == 0:
                     await update_weight_task
+                torch.distributed.barrier()
+        # CUDA IPC
+        else:
+            # we sent in batches of 1GB as an optimization
```
tyler-griggs: nit: same comment as for deepspeed worker above

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -158,6 +159,8 @@ generator:
   num_inference_engines: 1
   backend: "vllm"
   weight_sync_backend: "nccl"
+  # if using cuda_ipc, we send in batches of this size in GB
+  weight_transfer_threshold_cuda_ipc_in_GB: 1.0
```
tyler-griggs: Very nit-y: I like just suffixing with the unit rather than add "in": `cuda_ipc_weight_transfer_threshold_GB`?

tyler-griggs (approved):
Nice!

Issue #139, opened 2025-08-09
Support exporting environment-specific metrics


tyler-griggs:
It is useful to see environment-specific metrics during training runs to understand how the model is progressing. To do so, we should support SkyRL-Gym environments returning metrics that will be reported to wandb (or other logging/reporting tools). 

## TODOs
- [x] add a `metrics` field of type  `Dict[str,Any]` to `BaseTextEnvStepOutput`
- [x] add the logic in `SkyRLGymGenerator` to aggregate these metrics across the trainingbatch and return them to the trainer for logging (under `env/*` on wandb)
- [x] update one of the example environments to report metrics (to serve as an example)

alex-dr:
I just came here to file this feature request :)

tyler-griggs:
@mdawn65 is picking this up

SumanthRH:
Addressed in #386

PR #163, opened 2025-08-19
Support exporting environment-specific metrics

mdawn65:
Issue: #139 

### Summary
This PR adds support for environment-specific metrics collection and logging in SkyRL-Gym environments. 

- Added `metrics` of type `Dict[str,Any]` to `BaseTextEnvStepOutput`
- Added `env_metrics` of type `Optional[List[Dict[str, Any]]]` to `GeneratorOutput`
- Logic added in `concatenate_generator_outputs` in `utils.py`
- Logic added in `generate()` and `generate_batched()` to collect metrics from environment steps
- `eval()` computes aggregation by **mean** across evaluation batches
- Aggregated metrics are logged to wandb under the `env/*` name
- Example environment: `GSM8k`
   - Added `_calculate_metrics()` method for the metrics: response_length, word_count, answer_accuracy
- Added 4 unit tests (non-batched and batched, compatibility with envs that don't report metrics, concatenation logic, edge cases)

### DIFF
--- a/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
+++ b/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
@@ -60,7 +60,7 @@ class SkyRLGymGenerator(GeneratorInterface):
         max_tokens: int,
         max_input_length: int,
         sampling_params: Optional[Dict[str, Any]] = None,
-    ) -> Tuple[List[int], float, str, List[int], List[int], Optional[List[float]]]:
+    ) -> Tuple[List[int], float, str, List[int], List[int], Optional[List[float]], Optional[Dict[str, Any]]]:
         """
         Multi-turn generation loop that executes a single trajectory.
 
@@ -77,6 +77,7 @@ class SkyRLGymGenerator(GeneratorInterface):
             loss_mask: List[int]
             prompt_token_ids: List[int]
             rollout_logprobs: Optional[List[float]]
+            env_metrics: Optional[Dict[str, Any]]
         """
         # Create a new environment instance
         env_extras["max_turns"] = self.max_turns  # TODO(shu): move this to config
--- a/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
+++ b/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
@@ -288,10 +288,10 @@ class SkyRLGymGenerator(GeneratorInterface):
         if self.batched:
             return await self.generate_batched(
                 prompts, env_classes, env_extras, max_tokens, max_input_length, sampling_params
-            )  # Calls for single-turn parallel generation
+            )
 
         # Async agent loop to generate trajectories in parallel.
-        tasks = []  # Use agent_loop() for multi-turn sequential generation
+        tasks = []
         for i in range(len(prompts)):
             tasks.append(
                 self.agent_loop(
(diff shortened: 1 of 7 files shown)


skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -311,6 +316,7 @@ async def generate(self, input_batch: GeneratorInput) -> GeneratorOutput:
         stop_reasons = [output[2] for output in all_outputs]
         loss_masks = [output[3] for output in all_outputs]
         prompt_token_ids = [output[4] for output in all_outputs]
+        env_metrics = [output[6] for output in all_outputs]
```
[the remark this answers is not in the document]
tyler-griggs: Don't worry about this, it is something I will handle separately.

skyrl-train/skyrl_train/trainer.py
```diff
@@ -177,14 +180,38 @@ async def eval(self) -> Dict[str, float]:
 
         # 3. Calculate overall metrics across all datasets
         overall_avg_score, overall_pass_at_n = get_metrics_from_generator_output(concat_generator_outputs, concat_uids)
+
+        # 4. Aggregate environment metrics for logging
```
tyler-griggs: We won't want to handle environment metric aggregation in the training stack because the metrics will be pretty environment specific, and it might surprise env develops when the training stack quietly aggregates its metrics. It should be the responsibility of the environment to product Dict[str, Any] object that can be directly exported by wandb (and other loggers) so that the training stack just directly exports it (see the rollout_metrics as an example). Let me know if I'm misunderstanding something here though!

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -341,11 +347,12 @@ async def generate(self, input_batch: GeneratorInput) -> GeneratorOutput:
             "stop_reasons": stop_reasons,
             "rollout_metrics": rollout_metrics,
             "rollout_logprobs": rollout_logprobs,
+            "env_metrics": env_metrics,
```
tyler-griggs: Can we actually bundle the env_metrics into `rollout_metrics` and only return `rollout_metrics` in `GeneratorOutput`. Long-term, I believe we will actually change this name to just `metrics` or `generation_metrics` to be more generic, but I'd prefer to keep the `GeneratorOutput` interface unchanged.

tyler-griggs: Relatedly, we should probably prefix all the env_metrics keys from the environment with the string "environment/" so that it's well-organized in wandb (or other logging tools)

tyler-griggs (commented):
This is looking great! I left a couple comments. Also, it would be great to launch a simple training run (examples/gsm8k/run_gsm8k.sh with a 0.5B model) and upload metrics to wandb (you can create a free personal account and get wandb api key to use), then attach a screenshot of the reported example metrics (`response_length`, `word_count`, etc.) just to verify that everything looks right

skyrl-train/skyrl_train/trainer.py
mdawn65: Hey Tyler, after we aggregate all of the metrics from generation, the output of `concat_all_env_metrics` is `List[Dict[str, Any]]`. Should I add an aggregate function in env.py (for example, to `gsm8k/env.py`), or should I leave the data as is?

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
mdawn65: Hi Tyler, I've addressed the comments on this new PR: https://github.com/NovaSky-AI/SkyRL/pull/189.

PR #169, opened 2025-08-20
[FlashRL N/N] Add example for FP8 training with FlashRL

SumanthRH:
# What does this PR do?

WIP PR to add FP8 training with FlashRL. 

uses a custom vllm wheel. Found this to be the simplest way to manage the custom vllm patches in flashRL. Wheel is pre-packaged build from branch: https://github.com/SumanthRH/vllm/tree/flashrl . Specifying the git url directly led to uv building  the cpu-only version of vllm for some reason. We'll use this wheel for now

TODO:
- [ ] Verify E2E run on Deepspeed and FSDP
- [ ] Verify training on qwen3 14B and 32B
- [ ] upload wheel to github releases and use the link from Github


skyrl-train/examples/flash_rl/flash_rl_engine.py
```diff
@@ -0,0 +1,126 @@
...
+    sampling_params: Optional[Dict[str, Any]] = None,
+    tokenizer=None,
+    backend="vllm",
+) -> List[InferenceEngineInterface]:
+    """
+    Create a list of RayWrappedInferenceEngine instances wrapping Ray actor handles to InferenceEngineInterface instances.
+    """
+    from skyrl_train.utils import ray_noset_visible_devices, get_all_env_variables, get_ray_pg_ready_with_timeout
+
+    assert not async_engine, "`async_engine` is not supported for FlashRL"
```
erictang000, line 53: just to confirm - we can only use the offline engine for flash-rl, so only single turn rollouts?

skyrl-train/docs/examples/flash_rl.rst
```diff
@@ -0,0 +1,58 @@
...
+
+.. code-block:: bash
+    :caption: Environment variables at ``examples/flash_rl/.env.flashrl``
+
+    FLASHRL_CONFIG=fp8_vllm
+    ...
+
+.. warning::
+
+   FlashRL integration is experimental. While generation times can improve for large models with quantization, we've observed that the time spent in weight syncing is much higher with FlashRL for fp8. This negates most of the benefits of fp8 inference. The slowdown is primarily due to slow weight quantization in vLLM's ``process_weights_after_loading`` function. We are actively working on improving this.
```
SumanthRH: This is an important warning. I've already improved weight syncing with the batching impl + fixes for FlashRL's `patch_load_weights` method, but it is still not good enough. We will revisit the fp8 slowdown, and meanwhile also see if int8 can provide good overall step time improvements

skyrl-train/examples/flash_rl/main_dapo_flashrl.py
```diff
@@ -0,0 +1,189 @@
+"""
```
erictang000, line 1: should we remove https://github.com/NovaSky-AI/SkyRL/tree/main/skyrl-train/examples/tis_correction now that we have this included in the full flash-rl integration? or do you think it's worth keeping them separate

skyrl-train/examples/flash_rl/main_dapo_flashrl.py
```diff
@@ -0,0 +1,189 @@
...
+            inference_engines = create_ray_wrapped_inference_engines_from_config_flashrl(
+                self.cfg, self.colocate_pg, tokenizer
+            )
+        else:
+            inference_engines = create_remote_inference_engines_from_config(self.cfg)
+
+        inference_engine_client = InferenceEngineClient(inference_engines)
+
+        generator: GeneratorInterface = self.get_generator(self.cfg, tokenizer, inference_engine_client)
+        # breakpoint()
```
erictang000: remove

skyrl-train/examples/flash_rl/main_dapo_flashrl.py
```diff
@@ -0,0 +1,189 @@
...
+        else:
+            raise ValueError(f"Unknown strategy type: {self.cfg.trainer.strategy}")
+
+        # NOTE (sumanthrh): Instantiate tracker before trainer init.
+        # We have custom validation before this step to give better error messages.
+        tracker = self.get_tracker()
+
+        tokenizer = self.tokenizer
+        if self.cfg.generator.run_engines_locally:
+            inference_engines = create_ray_wrapped_inference_engines_from_config_flashrl(
```
erictang000, line 145: do you think it's worth pulling this if else block into a helper to something like `get_inference_engines` in the main_base script to simplify the code you need to change here?

skyrl-train/docs/examples/flash_rl.rst
```diff
@@ -0,0 +1,58 @@
...
+
+FlashRL + SkyRL
+---------------
+
+SkyRL now supports an initial integration with FlashRL. Currently, we only support training with `online FP8 quantization <https://docs.vllm.ai/en/v0.9.2/features/quantization/fp8.html#online-dynamic-quantization>`_ in vLLM. You should simply specify ``FLASHRL_CONFIG=fp8_vllm`` in your environment variables and use the ``--extra flashrl`` flag when running the training script.
+
+How does it work?
+~~~~~~~~~~~~~~~~~~
+
+We pass `quantization=fp8`  flag to the vLLM engine at initialization time. This means that the weights are loaded as usual in half precision and then quantized down to fp8. During training, generations are sampled as usual, and in this case, sampled from quantized weights. Since we use online quantization, the scale factor used for quantizing activations are computed on the fly by vLLM internally.
```
erictang000, line 31:
```suggestion
We pass the `quantization=fp8`  flag to the vLLM engine at initialization time. This means that the weights are loaded as usual in half precision and then quantized down to FP8. During training, generations are sampled as usual, but now from the quantized weights. Since vLLM uses online quantization, the scale factors used for quantizing activations are computed dynamically during runtime. 
```

skyrl-train/docs/examples/flash_rl.rst
```diff
@@ -0,0 +1,58 @@
...
+
+SkyRL now supports an initial integration with FlashRL. Currently, we only support training with `online FP8 quantization <https://docs.vllm.ai/en/v0.9.2/features/quantization/fp8.html#online-dynamic-quantization>`_ in vLLM. You should simply specify ``FLASHRL_CONFIG=fp8_vllm`` in your environment variables and use the ``--extra flashrl`` flag when running the training script.
+
+How does it work?
+~~~~~~~~~~~~~~~~~~
+
+We pass `quantization=fp8`  flag to the vLLM engine at initialization time. This means that the weights are loaded as usual in half precision and then quantized down to fp8. During training, generations are sampled as usual, and in this case, sampled from quantized weights. Since we use online quantization, the scale factor used for quantizing activations are computed on the fly by vLLM internally. 
+
+The sampled rollouts are then used to compute the policy loss. We further apply the TIS correction factor to the policy loss and then update the policy model weights. These weights, in half precision, are then synced with the inference engine layer by layer. These are then loaded and quantized down to fp8 similar to how we quantized the weights at initialization.
```
erictang000, line 33:
```suggestion
The sampled rollouts are then used to compute the policy loss. We further apply the TIS correction factor to the policy loss and then update the policy model weights. These weights, in half precision, are then synced with the inference engine layer by layer. These are then loaded and quantized down to FP8 similar to how we quantized the weights at initialization. 
```

skyrl-train/docs/examples/flash_rl.rst
erictang000:
```suggestion
   FlashRL integration is experimental. While generation times can improve for large models with quantization, we've observed that the time spent in weight syncing is much higher with FlashRL for FP8. This negates most of the benefits of FP8 inference. The slowdown is primarily due to slow weight quantization in vLLM's ``process_weights_after_loading`` function. We are actively working on improving this.
```

erictang000: > This negates most of the benefits of fp8 inference.

can we word this less strongly (may negate some benefits due to slow weight quantization but we are actively working on improving this in vLLM)

skyrl-train/examples/flash_rl/run_dapo_flashrl.sh
```diff
@@ -0,0 +1,91 @@
+set -x
+
+# Colocated DAPO training+generation for Qwen2.5-1.5B-Instruct on GSM8K with FP8 rollouts.
+# The configuration is tested on 2 H100 GPUs.
+
+# uv run examples/gsm8k/gsm8k_dataset.py --output_dir $HOME/data/gsm8k
+# export WANDB_API_KEY=<your_key_here>
+# bash examples/flash_rl/run_dapo_flashrl.sh
+
+DATA_DIR="/mnt/user_storage/data/dapo"
```
erictang000: make DATA_DIR consistent with  `uv run examples/gsm8k/gsm8k_dataset.py --output_dir $HOME/data/gsm8k` shown above

skyrl-train/examples/flash_rl/run_dapo_flashrl_32b.sh
```diff
@@ -0,0 +1,92 @@
+set -x
+
+# Colocated DAPO training+generation for Qwen3-32B on GSM8K with FP8 rollouts.
+# The configuration is tested on 2 8xH100 GPUs.
+
+# uv run examples/gsm8k/gsm8k_dataset.py --output_dir $HOME/data/gsm8k
+# export WANDB_API_KEY=<your_key_here>
+# bash examples/flash_rl/run_dapo_flashrl_32b.sh
+
+DATA_DIR="/mnt/user_storage/data/gsm8k"
```
erictang000: same comment here

skyrl-train/skyrl_train/workers/deepspeed/deepspeed_worker.py
```diff
@@ -157,44 +158,60 @@ def gather_and_broadcast(param):
...
-                            weights_update_request = {"names": [], "dtypes": [], "shapes": [], "extras": []}
-                            # force collect any sent tensors if possible to be memory efficient
-                            torch.cuda.ipc_collect()
-
-                    torch.distributed.barrier()
-                    torch.cuda.synchronize()
+            module_to_params: Dict[str, List[str]] = {}
+            params = dict(model.named_parameters())
+            for param_name, param in model.named_parameters():
+                # TODO (sumanthrh): When would this fail? Works for many AutoModelForCausalLM models for now
```
erictang000, line 164: so to sanity check - if there was a module with a weight less than 2 `.`s deep like `lm_head.weight`, this gives `".".join([])`, which is the empty string, so everything with less than 2 `.` separators will be sent together?

could we maybe do a check for if the `len(param_name.split(".")[:-2]) < 2` and handle that case more nicely since this is on the main path for both fsdp and deepspeed.

skyrl-train/skyrl_train/workers/deepspeed/deepspeed_worker.py
```diff
@@ -157,44 +158,60 @@ def gather_and_broadcast(param):
...
-                            current_size = 0
-                            weights_update_request = {"names": [], "dtypes": [], "shapes": [], "extras": []}
-                            # force collect any sent tensors if possible to be memory efficient
-                            torch.cuda.ipc_collect()
-
-                    torch.distributed.barrier()
-                    torch.cuda.synchronize()
+            module_to_params: Dict[str, List[str]] = {}
+            params = dict(model.named_parameters())
+            for param_name, param in model.named_parameters():
```
erictang000, line 163: question: if later we wanted to optimize weight sync perf by bucketing weight updates together, would making this choice to get flash-rl to work now affect it?

skyrl-train/examples/flash_rl/flash_rl_engine.py
erictang000: maybe worth a clarification in the doc, I didn't realize until i hit this line of code

SumanthRH: Yeah let me add a warning

skyrl-train/examples/flash_rl/main_dapo_flashrl.py
SumanthRH: not worth right now :)

skyrl-train/examples/flash_rl/run_dapo_flashrl.sh
SumanthRH: Oops, this got left out while cleaning. I'll make it consistent yep

skyrl-train/skyrl_train/workers/deepspeed/deepspeed_worker.py
SumanthRH: Wdym by bucketing? There is a form of batching here that is already an impr over the previous param by param weight sync. I'm guessing you are referring to placing a batch of weights in a single buffer (currently we have indv. buffers for each weight) and then sending that over as one IPC handle. 

Even in this case, the current module by module approach is a natural strategy. This is currently combined with  the batch until threshold is reached logic, so it's still flexible as in #161 

We can revisit when we impl single buffer per batch and then profile for large models IMO. Wdyt

erictang000: yeah I guess my question was - if later we want to do larger buffer, will it break flash-rl integration if it does param by param allocation? i guess just depends on how we implement it + what gets upstreamed into vllm weight loaders, but agree this is an improvement for now and we revisit later

SumanthRH: Yes let's revisit later.

skyrl-train/skyrl_train/workers/deepspeed/deepspeed_worker.py
SumanthRH: Hmm what is the issue if the module name is empty again? it's just saying that these params all belong to the module at the root level, and the batching strat is still correct?

skyrl-train/examples/flash_rl/main_dapo_flashrl.py
SumanthRH: Unclear actually, i was wondering if tis correction can still be useful with bf16 - esp for long trainng runs so that small divergence dont' accumulate errors. leave it as be for now?

SumanthRH: FlashRL authors haven't studied TIS for long training runs yet

skyrl-train/skyrl_train/workers/deepspeed/deepspeed_worker.py
erictang000: hmm i was thinking OOM if lm head and something like `embedding.weight` or something large were materialized together and we might want to do those separately. but just realized we're only syncing every 1GB. should be fine then i think!

erictang000: follow up question: does flash-rl work with the non colocated case? since i don't see modifications for that weight update strategy

skyrl-train/docs/examples/flash_rl.rst
SumanthRH: haha sure

skyrl-train/skyrl_train/workers/deepspeed/deepspeed_worker.py
SumanthRH: added validation for now. I could make it work, but the speed will be terrible. There is also no use for batching in the remote case - except for flashrl

skyrl-train/examples/flash_rl/main_dapo_flashrl.py
SumanthRH: thanks

PR #174, opened 2025-08-20
Add generic trajectory logging for debugging RL training

### LINKED ISSUE #122: Add generic trajectory logger
It is very useful to dig into the complete generated trajectories (including model outputs and environment observations / feedback) to debug model behavior during training. We should support exporting trajectories in an easily-readable format. 

At minimum, we should support dumping the chat history as a table to wandb. 

A better solution is likely creating a generic TrajectoryLogger class that is given complete trajectories and can dump / export them in a user-customizable way. Some users simply want to read the prompts and responses (e.g., as a table in wandb), others may want to create a dataframe and do some data analysis in a notebook -- a generic TrajectoryLogger class would support this. 

To start, the class could exposes a single method, `log`, that takes as input a trajectory in `ConversationType` data type. But, it is possible (and likely) that users will want more flexibility in what they pass into the class, so this is worth considering.

## TODOs
- [ ] Create generic TrajectoryLogger class
- [ ] Create a WandbTableTrajectoryLogger instantiation of this class, which simply uploads a table of prompts and responses 
- [ ] Use the WandbTableTrajectoryLogger in SkyRLGymGenerator (and cover with flag) to add trajectory logging to wandb for all training runs that use the skyrl gym generator. One tricky detail I anticipate is handling de-tokenization of trajecotries so they are uploaded in plaintext.

adigoel4:
Added a generic TrajectoryLogger system to help debug what's happening during training. You can now export complete trajectories (prompts, responses, rewards, etc.) to see exactly what the model is doing.

## Features:
- Base TrajectoryLogger class with WandB and CSV implementations
- Integrated into SkyRLGymGenerator using dependency injection
- Trainer automatically logs trajectories at configurable intervals during training/eval
- Configurable text truncation and trajectory limits to avoid spamming logs
- Handles tokenization properly so everything exports as readable text

## Config:
```yaml
generator:
   trajectory_logging:
      enabled: true
      type: "wandb"  # or "csv" 
      max_trajectories: 10
      log_interval: 100
      max_text_length: 2000
```

I also added CSV logging and integration tests so I could test this on CPU without needing GPU dependencies - not sure if we need to keep those but they work.

  Fixes #122

### DIFF
--- a/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
+++ b/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
@@ -467,7 +467,7 @@ class SkyRLGymGenerator(GeneratorInterface):
             # Handle both single float and list of floats for rewards
             log_rewards = [sum(r) if isinstance(r, list) else r for r in rewards]
             
-            self.trajectory_logger.log(log_prompts, log_responses, log_rewards, step=0, prefix="generation")
+            self.trajectory_logger.log(log_prompts, log_responses, log_rewards)
 
         rollout_metrics = get_rollout_metrics(responses, rewards)
 
--- a/skyrl-train/skyrl_train/generators/trajectory_logger.py
+++ b/skyrl-train/skyrl_train/generators/trajectory_logger.py
@@ -29,9 +29,7 @@ class TrajectoryLogger(ABC):
         self, 
         prompts: List[str],
         responses: List[str], 
-        rewards: List[float],
-        step: int,
-        prefix: str = "train"
+        rewards: List[float]
     ) -> None:
         pass
 
@@ -47,20 +45,17 @@ class WandbTableTrajectoryLogger(TrajectoryLogger):
         self, 
         prompts: List[str],
         responses: List[str], 
-        rewards: List[float],
-        step: int,
-        prefix: str = "train"
+        rewards: List[float]
     ) -> None:
         if self.max_trajectories > 0:
             prompts = prompts[:self.max_trajectories]
             responses = responses[:self.max_trajectories]
             rewards = rewards[:self.max_trajectories]
         
-        table = self.wandb.Table(columns=["step", "prompt", "response", "reward"])
-        for prompt, response, reward in zip(prompts, responses, rewards):
-            table.add_data(step, prompt, response, reward)
+        data = [[i, prompt, response, reward] for i, (prompt, response, reward) in enumerate(zip(prompts, responses, rewards))]
+        table = self.wandb.Table(columns=["step", "prompt", "response", "reward"], data=data)
         
-        self.wandb.log({f"{prefix}/trajectories": table}, step=step)
+        self.wandb.log({"trajectories": table})
 
 
 class CSVTrajectoryLogger(TrajectoryLogger):
@@ -74,9 +69,7 @@ class CSVTrajectoryLogger(TrajectoryLogger):
         self,
         prompts: List[str],
         responses: List[str], 
-        rewards: List[float],
-        step: int,
-        prefix: str = "train"
+        rewards: List[float]
     ) -> None:
         if self.max_trajectories > 0:
             prompts = prompts[:self.max_trajectories]
@@ -84,14 +77,13 @@ class CSVTrajectoryLogger(TrajectoryLogger):
             rewards = rewards[:self.max_trajectories]
         
         df = pd.DataFrame({
-            "step": step,
-            "prefix": prefix,
+            "step": list(range(len(prompts))),
             "prompt": prompts,
             "response": responses,
             "reward": rewards
         })
         
-        filename = os.path.join(self.output_dir, f"{prefix}_trajectories_step_{step}.csv")
+        filename = os.path.join(self.output_dir, "trajectories.csv")
         df.to_csv(filename, index=False)
 
 
--- a/skyrl-train/tests/cpu/test_trajectory_logging_integration.py
+++ b/skyrl-train/tests/cpu/test_trajectory_logging_integration.py
@@ -213,7 +213,7 @@ class TestTrajectoryLoggingIntegration:
             generator_output = await generator.generate(input_batch)
             
             # Verify trajectory logging worked by checking for CSV file
-            csv_file = os.path.join(tmpdir, "generation_trajectories_step_0.csv")
+            csv_file = os.path.join(tmpdir, "trajectories.csv")
             assert os.path.exists(csv_file), f"Expected trajectory file not found at {csv_file}"
             
             # Verify CSV content contains expected trajectory data
@@ -227,9 +227,10 @@ class TestTrajectoryLoggingIntegration:
             assert "response" in df.columns 
             assert "reward" in df.columns
             assert "step" in df.columns
-            assert "prefix" in df.columns
-            assert all(df["step"] == 0)
-            assert all(df["prefix"] == "generation")
+            
+            # Verify step values are indices (0, 1 for 2 trajectories)
+            expected_steps = list(range(len(df)))
+            assert list(df["step"]) == expected_steps, f"Expected steps {expected_steps}, got {list(df['step'])}"
             assert all(df["reward"] == 1.0)
     
     def test_trainer_trajectory_logging_integration(self, trajectory_logging_config, mock_tokenizer):
--- a/skyrl-train/tests/cpu/utils/test_trajectory_logger.py
+++ b/skyrl-train/tests/cpu/utils/test_trajectory_logger.py
@@ -37,7 +37,7 @@ def sample_data():
 def concrete_logger():
     """Create a concrete implementation of TrajectoryLogger for testing."""
     class ConcreteLogger(TrajectoryLogger):
-        def log(self, prompts, responses, rewards, step, prefix="train"):
+        def log(self, prompts, responses, rewards):
             pass
     return ConcreteLogger()
 
@@ -67,9 +67,7 @@ class TestWandbTableTrajectoryLogger:
             logger.log(
                 sample_data["prompts"],
                 sample_data["responses"], 
-                sample_data["rewards"],
-                step=100,
-                prefix="test"
+                sample_data["rewards"]
             )
             
             # Verify table creation
@@ -80,10 +78,10 @@ class TestWandbTableTrajectoryLogger:
             # Verify trajectory limiting (max_trajectories=2)
             assert mock_table.add_data.call_count == 2
             
-            # Verify wandb.log call
+            # Verify wandb.log call with fixed prefix
             mock_wandb_module.log.assert_called_once()
             log_args = mock_wandb_module.log.call_args[0][0]
-            assert "test/trajectories" in log_args
+            assert "trajectories" in log_args
     
     def test_unlimited_trajectories(self, sample_data):
         """Test logging with unlimited trajectories."""
@@ -97,8 +95,7 @@ class TestWandbTableTrajectoryLogger:
             logger.log(
                 sample_data["prompts"],
                 sample_data["responses"],
-                sample_data["rewards"],
-                step=100
+                sample_data["rewards"]
             )
             
             # Should log all 3 trajectories
@@ -125,20 +122,19 @@ class TestCSVTrajectoryLogger:
             logger.log(
                 sample_data["prompts"],
                 sample_data["responses"],
-                sample_data["rewards"],
-                step=100,
-                prefix="train"
+                sample_data["rewards"]
             )
             
             # Verify CSV file creation
-            csv_file = os.path.join(tmpdir, "train_trajectories_step_100.csv")
+            csv_file = os.path.join(tmpdir, "trajectories.csv")
             assert os.path.exists(csv_file)
             
             # Verify CSV content
             df = pd.read_csv(csv_file)
             assert len(df) == 2  # Limited by max_trajectories
-            assert df["step"].iloc[0] == 100
-            assert df["prefix"].iloc[0] == "train"
+            
+            # Verify step values are indices (0, 1)
+            assert list(df["step"]) == [0, 1]
             assert "What is 2+2?" in df["prompt"].values
             assert "The answer is 4." in df["response"].values
             assert 1.0 in df["reward"].values


skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -60,7 +79,7 @@ async def agent_loop(
         max_tokens: int,
         max_input_length: int,
         sampling_params: Optional[Dict[str, Any]] = None,
-    ) -> Tuple[List[int], float, str, List[int], List[int], Optional[List[float]]]:
+    ) -> Tuple[List[int], float, str, List[int], List[int], Optional[List[float]], Optional[Trajectory]]:
```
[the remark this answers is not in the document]
adigoel4: Created `AgentLoopOutput` NamedTuple and replaced all the references accordingly (ie. `output[0]` -> `output.response_ids`)

skyrl-train/skyrl_train/trainer.py
```diff
@@ -159,6 +159,11 @@ async def eval(self) -> Dict[str, float]:
             )
             generator_output: GeneratorOutput = await self.generate(generator_input)
             generator_outputs.append(generator_output)
+            
+            # Log eval trajectories if enabled (log each batch during eval)
+            if (self.cfg.generator.get('trajectory_logging', {}).get('enabled', False) and 
+                hasattr(self.generator, '_flush_trajectories')):
+                self.generator._flush_trajectories(self.global_step, prefix="eval")
```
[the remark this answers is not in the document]
adigoel4: added public method to `GeneratorInterface` and updated `trainer.py` to remove `hasattr` checks

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -203,6 +203,16 @@ generator:
...
   eval_n_samples_per_prompt: 1
+  
+  # Trajectory logging configuration
+  trajectory_logging:
+    enabled: false  # Set to true to enable trajectory logging
+    type: "wandb"  # Type of logger to use ("wandb" or "csv")
+    max_trajectories: 10  # Maximum number of trajectories to log per batch
+    max_text_length: 2000  # Maximum length of text fields in logged data
+    log_full_history: false  # Whether to log full chat history or just prompt/response
+    log_interval: 100  # Log trajectories every N steps (0 = log every step)
```
tyler-griggs: nit: shouldn't `1` mean log every step? :D

skyrl-train/skyrl_train/generators/base.py
```diff
@@ -33,3 +33,16 @@ async def generate(self, input_batch: GeneratorInput) -> GeneratorOutput:
             GeneratorOutput: Generated trajectories
         """
         raise NotImplementedError()
+    
+    def flush_trajectories(self, step: int, prefix: str = "train") -> None:
```
tyler-griggs: I'd actually prefer not to add this method to the GeneratorInterface, which we'd like to keep very simple. Instead, I expect the TrajectoryLogger to basically be a utility that we provide and users can optionally activate and optionally incorporate into any custom Generators they implement.

I would also recommend not triggering trajectory flushing inside the trainer. Basically, the trainer should not need to know anything about the how/when to flush trajectories, that should just be the responsibility of the Generator.

skyrl-train/skyrl_train/trainer.py
```diff
@@ -159,6 +159,10 @@ async def eval(self) -> Dict[str, float]:
             )
             generator_output: GeneratorOutput = await self.generate(generator_input)
             generator_outputs.append(generator_output)
+            
+            # Log eval trajectories if enabled (log each batch during eval)
```
tyler-griggs: Yeah, like my comment above, I wouldn't have the trainer participate in triggering trajectory logging. Instead, the logging behavior is determined by config, and the Generator handles the logging.

tyler-griggs: That said, I do understand that we'd like to prefix train/eval trajectory logging separately. Long-term we should solve this by passing something like an `is_eval` flag into the Generator so that it can perform differentiated logic based on whether it's training or eval. But I wouldn't do that in this PR, just a note for the future!

skyrl-train/skyrl_train/utils/trajectory_logger.py
tyler-griggs, line 1: Could you actually move this to `skyrl_train/generators/`? I think this should be specific to the generator

skyrl-train/skyrl_train/utils/trajectory_logger.py
```diff
@@ -0,0 +1,351 @@
...
+            Formatted conversation string
+        """
+        formatted = []
+        for msg in conversation:
+            role = msg.get("role", "unknown")
+            content = msg.get("content", "")
+            formatted.append(f"[{role}]: {content}")
+        return "\n".join(formatted)
+    
+    def _truncate_text(self, text: str, max_length: int = 1000) -> str:
```
tyler-griggs: Was there a technical reason you added the `_truncate_text` utility, or is it just to be helpful? I ask because I expect in most cases users will want to see the entirety of the trajectory, otherwise it's not quite as useful. This is especially true for the response, because the answer or final output is technically at the end. Unless there's a technical reason, I would remove the truncation option, and we can add it later if there is a need for it.

skyrl-train/skyrl_train/utils/trajectory_logger.py
```diff
@@ -0,0 +1,351 @@
...
+            max_length: Maximum length
+            
+        Returns:
+            Truncated text
+        """
+        if len(text) <= max_length:
+            return text
+        return text[:max_length - 3] + "..."
+    
+    def _detokenize(self, token_ids: List[int]) -> str:
```
tyler-griggs: Is there a case where `_detokenize()` is called when `self.tokenizer is None`? 

I feel like it's actually fair to assume that if a user wants to log trajectories then they should do the detokenization themselves. As much as possible, we should try to limit the responsibility of the TrajectoryLogger to only logging trajectories. I'd recommend removing the `tokenizer`, removing `response_tokens` as an Optional input, and just require text-based input. This should also allow you to remove some code paths and `if` statements in the following code.

I could be wrong here, and we may want to add support for taking a tokenizer + response_tokens as input, but I'd strongly prefer to start as minimal as possible.

skyrl-train/skyrl_train/utils/trajectory_logger.py
```diff
@@ -0,0 +1,351 @@
...
+        max_trajectories: Optional[int] = 10,
+        max_text_length: int = 2000,
+        log_full_history: bool = False
+    ):
+        """
+        Initialize the WandB table trajectory logger.
+        
+        Args:
+            tokenizer: Optional tokenizer for converting between tokens and text
+            max_trajectories: Maximum number of trajectories to log per batch
```
tyler-griggs: Similar comment to what I wrote above, is there a technical reason to have a parameter for `max_text_length`?

`max_trajectories` makes more sense because the data size can be large. But I think we'd always want a full trajectory rather than truncating it.

skyrl-train/skyrl_train/utils/trajectory_logger.py
```diff
@@ -0,0 +1,351 @@
...
+
+class WandbTableTrajectoryLogger(TrajectoryLogger):
+    """
+    Trajectory logger that uploads trajectories as tables to Weights & Biases.
+    """
+    
+    def __init__(
+        self, 
+        tokenizer=None, 
+        max_trajectories: Optional[int] = 10,
```
tyler-griggs: Can we make this default to infinite? E.g., use `-1` to mean "there is no max".

skyrl-train/skyrl_train/utils/trajectory_logger.py
```diff
@@ -0,0 +1,351 @@
...
+        max_text_length: int = 2000,
+        log_full_history: bool = False
```
tyler-griggs: Can we default this to `True`? In general, we don't want to surprise the user by truncating / removing contents of the log. All defaults should be to include everything.

skyrl-train/skyrl_train/utils/trajectory_logger.py
```diff
@@ -0,0 +1,351 @@
...
+            tokenizer: Optional tokenizer for converting between tokens and text
+            max_trajectories: Maximum number of trajectories to log per batch
+            max_text_length: Maximum length of text fields in the table
+            log_full_history: Whether to log full chat history or just prompt/response
+        """
+        super().__init__(tokenizer, max_trajectories)
+        self.max_text_length = max_text_length
+        self.log_full_history = log_full_history
+        
+        try:
```
tyler-griggs: Oh actually you could probably remove this try/catch. We do include `wandb` in the `toml` file so it is already a strict dependency. If you really wanted to be sure, you could move this check to`utils.py` in `validate_cfg()` where we try to do this kind of validation as soon as possible so people don't waste time waiting for training setup just for everything to crash on an import error.

skyrl-train/skyrl_train/config/ppo_base_config.yaml
adigoel4: yeah, changed documentation and if statement in `skyrl_gym_generator.py`

adigoel4:
Just pushed a much more simplified implementation. Looking back on this, I definitely over engineered this without thinking about what is actually necessary right now.

skyrl-train/skyrl_train/utils/trajectory_logger.py
adigoel4: removed try/catch

skyrl-train/skyrl_train/utils/trajectory_logger.py
adigoel4: changed to True

skyrl-train/skyrl_train/utils/trajectory_logger.py
adigoel4: set default to -1

skyrl-train/skyrl_train/utils/trajectory_logger.py
adigoel4: removed detokenizing, which simplified implementation a lot

skyrl-train/skyrl_train/utils/trajectory_logger.py
adigoel4: As mentioned below, fell into the trap of trying to make this "feature-rich" without analyzing what is actually useful

adigoel4: Removed truncation

skyrl-train/skyrl_train/utils/trajectory_logger.py
adigoel4: moved

skyrl-train/skyrl_train/generators/base.py
adigoel4: removed

skyrl-train/skyrl_train/trainer.py
adigoel4: Removed trainer from the logging process. Generator stores `batch_count` and uses it to adjust logging based on `log_interval` (which is populated from the config). This implementation doesn't differentiate between train / eval. Let a comment regarding this future improvement, but happy to pick it up when this PR wraps up

skyrl-train/skyrl_train/utils/trajectory_logger.py
adigoel4: removed

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -68,6 +82,28 @@ def __init__(
             )
         else:
             self.env_executor = None
+        
+        # Initialize trajectory logging components
+        self.trajectory_batch = []
+        self.batch_count = 0  # Internal counter for logging intervals
+        
+        # Use injected logger or create from config
```
tyler-griggs: Unless there's a current case where we want it passed in, let's just create it from config and remove it from the constructor.

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -68,6 +82,28 @@ def __init__(
...
+        if trajectory_logger:
+            self.trajectory_logger = trajectory_logger
+            self.collect_trajectories = True
+            # Get log interval from config if available
+            self.log_interval = generator_cfg.trajectory_logging.get('log_interval', 100) if hasattr(generator_cfg, 'trajectory_logging') else 100
+        elif hasattr(generator_cfg, 'trajectory_logging'):
```
tyler-griggs: Throughout the codebase, our assumption is that `hasattr` should be true and that all config params should have reasonable defaults, so you don't need to check this. However, I would move the check for `trajectory_logging.enabled` to right here, and then just `assert` it is enabled within the `create_trajectory_logger_from_config` call.

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -68,6 +82,28 @@ def __init__(
...
+            self.trajectory_logger = create_trajectory_logger_from_config(
+                generator_cfg.trajectory_logging
+            )
+            self.collect_trajectories = self.trajectory_logger is not None
```
tyler-griggs: Is `collect_trajectories` necessary? Seems like we can always just check whether the logger is non-None

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -68,6 +82,28 @@ def __init__(
...
+            self.log_interval = generator_cfg.trajectory_logging.get('log_interval', 100)
```
tyler-griggs: Please remove these `get` calls that use a default value. The base config (`ppo_bas_config.yaml`) should have reasonable defaults. You can just use `generator_cfg.trajectory_logging.log_interval`. Please make this change throughout the PR.

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -210,6 +210,15 @@ generator:
 
   # number of samples per prompt for evaluation
   eval_n_samples_per_prompt: 1
+  
+  # Trajectory logging configuration
+  trajectory_logging:
+    enabled: false  # Set to true to enable trajectory logging
+    type: "wandb"  # Type of logger to use ("wandb" or "csv")
+    max_trajectories: -1  # Maximum number of trajectories to log per batch (-1 for unlimited)
+    log_full_history: true  # Whether to log full chat history or just prompt/response
```
tyler-griggs: Can't remember if we've already discussed this, but I'd prefer to just remove this. I don't see a case where you'd want to remove parts of the chat history when inspecting trajectories. We can add back later if needed, but my hunch is most users will want to see the full trajectory.

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -210,6 +210,15 @@ generator:
...
+    log_interval: 100  # Log trajectories every N steps (1 = log every step)
```
tyler-griggs: Actually, now that I think about it, we may as well remove this and log every step. Is there any benefit to setting an interval? This is another field we can add back if needed, or users can implement this themselves in a custom trajectory logger.

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -210,6 +210,15 @@ generator:
 
   # number of samples per prompt for evaluation
   eval_n_samples_per_prompt: 1
+  
+  # Trajectory logging configuration
+  trajectory_logging:
+    enabled: false  # Set to true to enable trajectory logging
+    type: "wandb"  # Type of logger to use ("wandb" or "csv")
```
tyler-griggs, line 235: Please add a TODO here for allowing users to bring a custom trajectory logger. They should be able to define their own class (outside of the skyrl-train package) and add it to some registry (see AdvantageEstimatorRegistry for an example) so that it can be referenced by name in this config. 

Actually, this would be a good follow-up PR to work on :)

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -210,6 +210,15 @@ generator:
...
+    max_trajectories: -1  # Maximum number of trajectories to log per batch (-1 for unlimited)
+    log_full_history: true  # Whether to log full chat history or just prompt/response
+    log_interval: 100  # Log trajectories every N steps (1 = log every step)
+    output_dir: "./trajectory_logs"  # Output directory for CSV logger
```
tyler-griggs: Let's default this to be the export path + "/trajectory_logs" (see the other config param for export path)

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -648,3 +711,40 @@ def _get_next_input_ids_with_single_turn_chat_template(
...
+        Args:
+            all_outputs: List of outputs from agent_loop calls
+        """
+        if not (self.trajectory_logger and self.collect_trajectories):
+            return
+            
+        # Collect new trajectories
+        trajectories = [output.trajectory for output in all_outputs if output.trajectory is not None]
+        if trajectories:
+            self.trajectory_batch.extend(trajectories)
```
tyler-griggs: If we get rid of the interval, you can get rid of `self.trajectory_batch` too. Anyways, I think the logger should manage `trajectory_batch` if a user actually needs something like intervals, etc.

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -648,3 +711,40 @@ def _get_next_input_ids_with_single_turn_chat_template(
                 input_ids += obs_tokens
 
         return loss_mask, input_ids, logprobs
+
+    def _handle_trajectory_logging(self, all_outputs) -> None:
```
tyler-griggs: Once we remove interval and trajectory_batch, you can basically remove this whole method and just collect the trajectories and call `log`

skyrl-train/skyrl_train/generators/trajectory_logger.py
```diff
@@ -0,0 +1,279 @@
...
+from abc import ABC, abstractmethod
+from typing import List, Dict, Any, Optional, Union
+from dataclasses import dataclass
+import pandas as pd
+from omegaconf import DictConfig
+from skyrl_train.inference_engines.base import ConversationType
+
+
+@dataclass
+class Trajectory:
```
tyler-griggs: I like the idea, but I'd prefer to start very very simple.

Let's get rid of `Trajectory` and *not* return any new fields in the agent_loop output. Instead, we can detokenize the prompt_ids and response_ids return values, and pass in only three fields to `log` by name: prompts, responses, and rewards.

I know this is very simplistic and we may want to expand in the future, but I'd prefer to introduce as few new concepts as necessary until we have a better understanding of what's needed.

adigoel4:
Went through and tried to make it as simple as possible!

skyrl-train/skyrl_train/generators/trajectory_logger.py
```diff
@@ -0,0 +1,109 @@
...
+
+class TrajectoryLogger(ABC):
+    """
+    Abstract base class for trajectory logging.
+    
+    TODO: Allow users to bring a custom trajectory logger. They should be able to
+    define their own class (outside of the skyrl-train package) and add it to
+    a registry (see AdvantageEstimatorRegistry for an example) so that it can
+    be referenced by name in the config.
+    """
```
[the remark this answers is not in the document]
adigoel4: lines of duplicated slicing logic are worth it for the much cleaner user experience when implementing custom loggers

tyler-griggs:
Merging this into a dev branch `dev/trajectory_logger` to be picked up after SkyRLGymGenerator refactor and clean-up.

PR #176, opened 2025-08-21
[FlashRL N/N] Support Int-8 Rollouts

SumanthRH:
# What does this PR do?

WIP PR to add an example for Int-8 Rollouts. 

TODO: 
- [ ] Benchmark 32B training with int8 
- [ ] Add docs for int8
- [ ] Cleanup


skyrl-train/skyrl_train/generators/utils.py
```diff
@@ -96,7 +96,7 @@ def concatenate_generator_outputs(generator_outputs: List[GeneratorOutput]) -> G
         "loss_masks": sum([output["loss_masks"] for output in generator_outputs], []),
         "rollout_logprobs": (
             sum([output["rollout_logprobs"] for output in generator_outputs], [])
-            if generator_outputs[0]["rollout_logprobs"]
+            if generator_outputs[0]["rollout_logprobs"] is not None
```
SumanthRH, line 99: missed a corner case before where one of the train batches has no good entries

SumanthRH:
I've tested training with GSM8K for Qwen 0.5B instruct with FlashRL's calibration profile: https://huggingface.co/LiyuanLucasLiu/Qwen2.5-0.5B-Instruct-quantized.w8a8-RedHatAI/blob/main/profile.pt 


Int8 related patching has been verified as follows: 
1. Custom vLLM verification: Micro test on quantization by running evaluation with `lm_eval` following: https://github.com/yaof20/Flash-RL/blob/main/tutorial/verify_flashrl_install.md and https://docs.vllm.ai/en/latest/features/quantization/fp8.html?h=lm_eval#3-evaluating-accuracy 
2.  Custom vLLM verification: Micro test by initializing vLLM instance and then triggering a  `load_weights` call by passing base model weights in BF16 to `model.load_weights`. This is intercepted by the patch and then vLLM's post-processing function will quantize it to Int8. After weights are reloaded, we again run `lm_eval` on gsm8k. (Baseline: 0.3359, After load weights: 0.3404). 
3. SkyRL + FlashRL: Small E2E test, more coming.   The evaluation results for GSM8K (this is without dynamic sampling) seem to match. 

image


There is also a noticeable difference in logprobs between rollout and training now
image

SumanthRH:
Update: Weight syncing with int8 was still pretty slow, not matching the results from FlashRL: I noticed about 140 s during weight sync for running Qwen2.5-32B on 2 8xH100 nodes.  This might be due to the fact that we're on 0.9.2.

I tracked it down to an extremely slow transpose + copy operation in vLLM's default kernel for int8 on H100s: https://docs.vllm.ai/en/v0.9.2/api/vllm/model_executor/layers/quantization/kernels/scaled_mm/cutlass.html#__codelineno-0-36

The TLDR is that, in weight syncing, after loading updated params, we make sure to match the stride of the original vLLM tensors. So the transpose can be skipped. This brings down the time to 20-30 seconds, which is much more acceptable.

Impl here: https://github.com/SumanthRH/vllm/blob/4b04dfcbe8e876e89e4f03859340613d6c766479/vllm/model_executor/layers/patch.py#L503-L512  

There is still some room for optimization, but this is left as is for now. This postprocessing is only skipped for int8 and for `CutlassScaledMMLinearKernel`

skyrl-train/docs/examples/flash_rl.rst
```diff
@@ -28,37 +26,95 @@ SkyRL now supports an initial integration with FlashRL. Currently, we only suppo
...
         ...
-        trainer.algorithm.use_tis=true \
-        trainer.algorithm.tis_imp_ratio_cap=2.0 \
+        trainer.algorithm.use_tis=$USE_TIS \
+        trainer.algorithm.tis_imp_ratio_cap=$TIS_IMP_RATIO_CAP \
+        generator.sampling_params.logprobs=0 \
         ...
 
-Here, we've configured training to use TIS with the importance sampling ratio cap of 2.0. Note that for making sure the FlashRL patches are applied for vLLM, we use the ``FLASHRL_CONFIG`` env var in ``examples/flash_rl/.env.flashrl``:
+Here, we've configured training to use TIS with the importance sampling ratio cap of 8.0. ``generator.sampling_params.logprobs=0`` ensures that logprobs for the chosen tokens are returned by the inference engine, which is required for TIS. Note that for making sure the FlashRL patches are applied for vLLM, we use the ``FLASHRL_CONFIG`` environment variable in ``examples/flash_rl/.env.32b_int8``:
```
tyler-griggs: nit: I think it's just `.env.int8`, not `.env.32b_int8`

skyrl-train/docs/examples/flash_rl.rst
```diff
@@ -28,37 +26,95 @@ SkyRL now supports an initial integration with FlashRL. Currently, we only suppo
...
+        trainer.algorithm.tis_imp_ratio_cap=$TIS_IMP_RATIO_CAP \
+        generator.sampling_params.logprobs=0 \
         ...
 
-Here, we've configured training to use TIS with the importance sampling ratio cap of 2.0. Note that for making sure the FlashRL patches are applied for vLLM, we use the ``FLASHRL_CONFIG`` env var in ``examples/flash_rl/.env.flashrl``:
+Here, we've configured training to use TIS with the importance sampling ratio cap of 8.0. ``generator.sampling_params.logprobs=0`` ensures that logprobs for the chosen tokens are returned by the inference engine, which is required for TIS. Note that for making sure the FlashRL patches are applied for vLLM, we use the ``FLASHRL_CONFIG`` environment variable in ``examples/flash_rl/.env.32b_int8``: 
 
 .. code-block:: bash
-    :caption: Environment variables at ``examples/flash_rl/.env.flashrl``
+    :caption: Environment variables at ``examples/flash_rl/.env.32b_int8``
```
tyler-griggs: ditto here

skyrl-train/skyrl_train/workers/deepspeed/deepspeed_worker.py
```diff
@@ -203,7 +204,10 @@ def gather_and_broadcast(param):
                             current_size += weight.nbytes
                             # We send in batches as an optimization
                             # sync if threshold is reached
-                            if current_size / (1024**3) > self.cfg.generator.weight_transfer_threshold_cuda_ipc_GB:
+                            if (
+                                module_done
```
tyler-griggs, line 208: qq: Is the reason why it's `and` here because you only want to do the weight transfer with entire modules at a time? So if the module is not larger than 1GB, it continues to the next module?

tyler-griggs (approved):
Only nits, great work

skyrl-train/skyrl_train/workers/deepspeed/deepspeed_worker.py
SumanthRH: Yeah both conditions are ideally met:
1. we want to sync entire modules at once 
2. We want to sync only when the threshold is reached, to avoid repeated syncing. 

The additional memory usage during weight syncing can now go a bit higher than the threshold. However, given that a module is a single self attn block or a mlp block, this shouldn't be too bad.

skyrl-train/docs/examples/flash_rl.rst
SumanthRH: Oops, I think I pressed Tab one too many times yesterday

SumanthRH: fixed

skyrl-train/docs/examples/flash_rl.rst
SumanthRH: Fixed!

Issue #156, opened 2025-08-18
Optionally export checkpoints to cloud storage (e.g., S3)


tyler-griggs:
Currently, training checkpointing only saves checkpoints to local storage. Many users prefer to export checkpoints to some centralized storage (e.g., S3). Long-term, we will add a more generic interface for users to add custom logic for checkpoint management. For now, we should support exporting the locally-saved checkpointing to S3. 

## TODOs

- [ ] Add configuration parameter to optionally export to S3
- [ ] Implement logic for exporting each checkpoint to S3 once created (so there is a local copy and S3 copy). The parameter `max_ckpts_to_keep` will only apply to local copies of checkpoints (ie, the training stack should **not** delete checkpoints in S3).
- [ ] Testing:
  - [ ] First, manually test by checkpointing and exporting to S3. Then locally download the S3 checkpoint to the training instance and resume from the checkpoint.
  - [ ] Next, add a GPU test to cover this same behavior.

tyler-griggs:
^^ @LokmaSpeedy is picking this up!

tyler-griggs:
Closed by #248

PR #177, opened 2025-08-21
Optionally export checkpoints to S3

LokmaSpeedy:
This solves #156. User can specify if they want to export their checkpoints to S3 by setting  `s3_export` fields on the config. s3 exporting is disabled by default. S3 checkpoints are not deleted automatically.

### **TODO:**
- [x]  Add configuration parameter to optionally export to S3
- [x] Implement logic for exporting each checkpoint to S3 once created (so there is a local copy and S3 copy). The parameter max_ckpts_to_keep will only apply to local copies of checkpoints (ie, the training stack should not delete checkpoints in S3).

Testing:
- [x] First, manually test by checkpointing and exporting to S3. Then locally download the S3 checkpoint to the training instance and resume from the checkpoint.
- [x] Next, add a GPU test to cover this same behavior.
    `skyrl-train/tests/gpu/test_s3_export.py`


bthecohen:
Rather than using `boto` directly, it might be better to use `fsspec` – then it could support other cloud storage providers (e.g. by installing `s3fs`, `gcsfs`, etc.)

tyler-griggs:
This is great, thanks again. Merging some of these changes into #185 and closing in favor of it.

Issue #104, opened 2025-07-22
Enable configuring custom chat template


tyler-griggs:
Currently, the SkyRL Gym Generator masks out thinking tokens for Qwen3 models using the `get_custom_chat_template` hook, however, some users want to train Qwen3 and keep the thinking tokens. 

Users want the ability to choose whether to train on thinking tokens and, more generally, want to be able to provide a custom chat template without forking the code. 

## TODOs
- [ ] Provide easier configuration / hook for users to provide a custom chat template in the SkyRL Gym Generator.
- [ ] Add configuration for using thought tokens or mask out thought tokens (using different custom templates).

PR #178, opened 2025-08-21
[SkyRLGymGenerator] Add genreator.chat_template and change Qwen3 default behavior on thinking tokens accumulation

devpatelio:
In `ppo_base_config`, add `generator.chat_template` to allow users to define custom chat template. This is particularly helpful for Qwen3 models, where you might want to keep the thinking tokens or remove them. 

```.yaml
  chat_template:
    source: "name"  # "name" or "file"
    name_or_path: null  # e.g., "qwen3_with_thinking" or "/path/to/template.j2"
```

You can either specify a chat template defined in `skyrl_train.generators.utils.CUSTOM_CHAT_TEMPLATES` (we currently have `qwen3_with_thinking` and `qwen3_without_thinking`) or load from a file.

Before this PR, the behavior for Qwen3 models are that, they will not keep the non-last turn thinking tokens and we always retokenize (every turn and at the end of the trajectory).

After this PR, Qwen3 models by default will accumulate in a TI/TO manner, just like any other models. Users can modify this behavior by setting `name_or_path` to `qwen3_without_thinking` to keep the old behavior.

Related to #179 and address #104


devpatelio:
In reference to #104

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -177,6 +177,59 @@ generator:
   remote_inference_engine_urls: ["127.0.0.1:8001"]
   max_turns: 1
 
+  # custom chat template support and thinking token masking
+  enable_thinking_tokens: false
+  custom_chat_templates:
```
tyler-griggs: I don't think we'll want to add the templates directly in the configuration file. It bloats the config file, but also just makes it tricky to use the commandline to launch, etc. The goal is to enable users to a) choose an existing already-implemented template by name, and b) allow users to bring their own custom chat template. 

My proposal is to update this config to be:

```
chat_template:
  source: one of "file" or "name"
  name_or_path: either the name of a pre-made template or path to new template
```
If `source` is `path`, then `name_or_path` should be a path to a file that holds the template (e.g., some `.j2` file). If `source` is `name`, then it should refer to a chat template that we have pre-implemented and is already in the code base (e.g., `qwen3_with_thinking`)

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -39,7 +39,19 @@ def __init__(
         self.use_conversation_multi_turn = generator_cfg.use_conversation_multi_turn
 
         # optionally use custom chat template to get loss masks (i.e. for Qwen3)
-        self.custom_chat_template = get_custom_chat_template(model_name)
```
tyler-griggs, line 67: nit: You can just access with `.` and avoid the `getattr`. We want to keep all default values in the `ppo_base_config.yaml` to reduce surprising users.

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -234,9 +253,40 @@ async def generate_batched(
...
+                loss_masks.append(custom_loss_mask)
+                
+                print(f"BATCHED CUSTOM CHAT TEMPLATE DEBUG:")
+                print(f"Response length: {len(sample_response_ids)}")
+                print(f"Loss mask length: {len(custom_loss_mask)}")
+                print(f"Loss mask: {custom_loss_mask}")
+                print(f"Masked tokens: {sum(custom_loss_mask)}/{len(custom_loss_mask)} ({sum(custom_loss_mask)/len(custom_loss_mask)*100:.1f}%)")
+            else:
+                # keep original default masking behaviour
+                if len(sample_response_ids) > max_tokens:
```
tyler-griggs: nit: it does seem like there's potential for du-duplicating code and pulling this truncation out if the `if`/`else`.

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -531,21 +581,48 @@ def _update_engine_input_token_ids(
...
 
-        if len(new_obs) > 0:
-            for obs in new_obs:
-                obs_tokens = self.tokenizer.encode(obs["content"], add_special_tokens=False)
-                loss_mask += [0] * len(obs_tokens)
-                # logprobs for observation tokens doesn't matter since they will be masked out during loss computation
-                if logprobs:
-                    logprobs += [1] * len(obs_tokens)
-                input_ids += obs_tokens
+        if self.custom_chat_template:
```
tyler-griggs: In somewhat unfortunate timing, a PR was just merged that makes some non-trivial changes to this method, so it would be good to rebase!

skyrl-train/skyrl_train/generators/utils.py
```diff
@@ -4,35 +4,69 @@
 import numpy as np
 from skyrl_train.generators.base import GeneratorOutput
 
-CUSTOM_CHAT_TEMPLATES = {
```
tyler-griggs, line 7: This could actually be a good place to put the pre-made chat templates (instead of in the config) and then just refer to them by name in the config and let `get_custom_chat_template` resolve which chat template is being referred to

tyler-griggs (commented):
Thanks for sending this in! Left a couple comments, primarily around moving chat templates out of the config. There was also a pretty big PR that was just merged that touches some similar code in `skyrl_gym_generator.py` so it'd be good to merge changes!

skyrl-train/skyrl_train/config/ppo_base_config.yaml
devpatelio: Yeah for sure, this makes sense. I actually already started doing this under the templates and existing jinja files already so once I configure the config file to accept paths I'll push the commit!

devpatelio:
Here's the wandb for the updated run with the file and name support: https://wandb.ai/devpatelio/updated_template_system

tyler-griggs:
@devpatelio quick check: is this PR ready for review? I see several debugging `print` statements, and a couple of the files in `examples/gsm8k` look like they are dev-only tests. Should these all be removed?

devpatelio:
@tyler-griggs Sorry about that, I forgot to remove some of the debug print statements + fix some of the rebasing variable names. The PR is ready to review now!

skyagent/skyagent/functional/templates/custom_chat_template.j2
tyler-griggs, line 1: Is this file used somewhere? or can we remove it?

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -177,6 +177,15 @@ generator:
   remote_inference_engine_urls: ["127.0.0.1:8001"]
   max_turns: 1
 
+  # custom chat template support and thinking token masking for backwards compatibility with qwen3 models (can remove later)
+  enable_thinking_tokens: false
```
tyler-griggs: can we remove this, and just use the template name instead? E.g., `qwen3_with_thinking` and `qwen3_without_thinking`)

skyrl-train/examples/gsm8k/run_gsm8k.sh
```diff
@@ -9,8 +9,9 @@ set -x
 # NOTE (sumanthrh): `micro_train_batch_size_per_gpu` and `micro_forward_batch_size_per_gpu` can be tuned
 
 DATA_DIR="$HOME/data/gsm8k"
-NUM_GPUS=4
-LOGGER="wandb"  # change to "console" to print to stdout
+NUM_GPUS=1
+LOGGER="console"  # change to "console" to print to stdout
```
tyler-griggs: nit: please revert this to original defaults

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -295,17 +306,46 @@ async def generate_batched(
...
-            if logprobs is not None:
-                sample_logprobs = logprobs[i][: len(response_ids)]
-                truncated_logprobs.append(sample_logprobs)
+            # NOTE (sumanthrh): We add a guard since response_ids is `None` with remote inference engine
+            if all_response_ids is not None:
+                sample_response_ids = all_response_ids[i]
+            else:
+                sample_response_ids = self.tokenizer.encode(response)
+
+            # apple custom chat template masking if configured for batch mode
```
tyler-griggs: nit: apple --> apply :D

tyler-griggs (commented):
Made a quick pass

devpatelio:
@tyler-griggs Just addressed all the comments - let me know if there's anything else that catches your eye.

devpatelio:
@tyler-griggs Let me know if you'd like me to remove the model name now since we're not using Qwen3 explicitly for the chat template selection (can leave it otherwise). Otherwise the PR should be good to go!

CharlieFRuan:
Hey @devpatelio, thanks for the PR! Is this ready for another round of review?

devpatelio:
Hey @CharlieFRuan, it should be good to go! I made a few small changes for unused variable assignments and modified the existing test scripts to match the new usage.

CharlieFRuan:
Thanks! Will take a look soon

skyrl-train/skyrl_train/generators/utils.py
```diff
@@ -15,24 +28,73 @@
...
         "{% set mycontent = full_content.split('</think>')[-1].lstrip('\n') %}"
         "{% endif %}"
         "{{mycontent + '<|im_end|>'}}"
         "{% endgeneration %}"
-        "{{'\n'}}"
+        "{% endif %}"
+        "{% endfor %}"
+    ),
+    # custom template that removes thinking tokens with Q/A format
+    "custom_thinking_remover": (
```
CharlieFRuan: What is `custom_thinking_remover` used for?

skyrl-train/skyrl_train/generators/utils.py
```diff
@@ -15,24 +28,73 @@
...
+    """
+    Get custom chat template based on the new config structure.
+    
+    Args:
+        chat_template_config: Config dict with 'source' and 'name_or_path' fields.
+        
+    Returns:
+        Chat template string or None
+    """
+    if chat_template_config:
```
CharlieFRuan: Style-wise, this method can be more concise. Currently we reach a depth of 4 nested if-else.

You can first do `if chat_template_config is None: return`, and the rest of the code can be not inside an if branch.

Besides, you've already checked `if not source` and `if not name_or_path`, you would't need `if source and name_or_path` right?

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -327,17 +329,52 @@ async def generate_batched(
...
+
             if logprobs is not None:
-                sample_logprobs = logprobs[i][: len(response_ids)]
+                if self.custom_chat_template:
+                    raise ValueError("Cannot get logprobs when custom chat template is used due to misalignment in tokenization.")
+                sample_logprobs = logprobs[i][:len(sample_response_ids)]
                 truncated_logprobs.append(sample_logprobs)
+                
+            loss_masks.append(loss_mask)
+            truncated_responses.append(sample_response_ids)
```
CharlieFRuan: I think many of the code you have in `generate_batched()` might be legacy code. They might have sticked around due to some rebase. Could you double check which code you are intentionally adding?

I know `custom_chat_template` before this PR is not applied in `generate_batched()`. We can keep it that way. Let's make this PR only about substituting that `self.custom_chat_template = get_custom_chat_template(model_name)` and use `self.custom_chat_template` the same way across the file.

skyrl-train/skyrl_train/generators/utils.py
```diff
@@ -5,8 +5,21 @@
 from skyrl_train.generators.base import GeneratorOutput
 
 CUSTOM_CHAT_TEMPLATES = {
-    # chat template for qwen3 thinking mode to remove think tokens similar to generation phase
-    "qwen3_thinking": (
+    # chat template for qwen3 that preserves thinking tokens (trains on all tokens)
+    "qwen3_with_thinking": (
```
CharlieFRuan, line 10: What is the difference bewteen the current `qwen3_without_thinking` and the `qwen3_thinking` before this PR?

CharlieFRuan (commented):
Thanks for the PR! Sorry for the delay in reviewing this, left some comments. The `chat_template` config looks good, but `skyrl_gym_generator.py` needs more fixes.

In a future PR, we would want to differentiate the flag `cfg.generator.retokenize` and `cfg.genreatr.chat_template`, and apply the custom chat template to non-retokenize codepaths as well.

Currently, prior to this PR, having a `custom_chat_template` means that we always retokenize chat history, rather than performing token-in-token-out style of appending tokens. But these behaviors should be decoupled. Refer to https://skyrl.readthedocs.io/en/latest/tutorials/skyrl_gym_generator.html for more details.

skyrl-train/skyrl_train/generators/utils.py
devpatelio: I'm pretty sure we just added the split lines for the thinking tokens primarily, it should be the same as the old one.

skyrl-train/skyrl_train/generators/utils.py
devpatelio: Gotcha, new commit should handle this.

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
devpatelio: Got it, i refactored with the old code and literally just kept it simple by replacing the input to the old endpoint call with the updated chat config.

skyrl-train/skyrl_train/generators/utils.py
```diff
@@ -15,24 +28,54 @@
...
+    """
+    if chat_template_config is None: 
         return None
+    
+    source = chat_template_config.get("source")
+    if not source:
+        raise ValueError("'source' is required in chat_template_config")
+
+    name_or_path = chat_template_config.get("name_or_path")
+    if not name_or_path:
```
CharlieFRuan, line 63: Should we return `None` or raise error here? The default path is using:

```
  chat_template:
    source: "name"  # "name" or "file"
    name_or_path: null  # e.g., "qwen3_with_thinking" or "/path/to/template.j2"
```
right?

Where `chat_template_config` would be the above, but not None.

Could you check some tests in `test/gpu/gpu_ci` or `run_gsm8k` for a little bit, that they do not break?

skyrl-train/skyrl_train/generators/utils.py
```diff
@@ -15,24 +28,54 @@
         "{% generation %}"
         "{% set full_content = message['content'] %}"
         "{% set mycontent = message['content'] %}"
-        "{% set is_last_message = loop.last and messages[-1]['role'] == 'assistant' %}"
-        "{% if '</think>' in full_content and not is_last_message %}"
+        "{% if '</think>' in full_content %}"
```
CharlieFRuan: Ah, by without thinking, we want to keep the thinking tokens for the last message. Just as the default qwen3 chat template do. For `without_thinking` we should keep the

```
        "{% set is_last_message = loop.last and messages[-1]['role'] == 'assistant' %}"
        "{% if '</think>' in full_content and not is_last_message %}"
```

if I understand correctly.

Could you add a unit test where the default Qwen3 chat template is the same as our `qwen3_without_thinking` after applying to some chat history?

skyrl-train/skyrl_train/generators/utils.py
devpatelio: Cool, can confirm the None is the right thing to return.

CharlieFRuan:
@devpatelio Following up on this -- any updates? Would be good to land this and the following item soon

> differentiate the flag `cfg.generator.retokenize` and `cfg.genreator.chat_template`

devpatelio:
Hey Charlie, thanks for the reminder. I just implemented the last fix for the template to keep the last message's thinking tokens + test with gsm8k works as expected. I'll create a separate PR right now for the subsequent task.

skyrl-train/skyrl_train/generators/utils.py
CharlieFRuan: Hmm if `None` is the right thing to return, are we keeping the `raise ValueError("'name_or_path' is required in chat_template_config")` or should we `return None`? Could you check running `run_gsm8k.sh` locally? I am guessing if we keep `raise ValueError`, running `run_gsm8k.sh` will lead us to error since it is using Qwen2.5, though I could be wrong

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -58,9 +58,10 @@ def __init__(
         self.max_turns = generator_cfg.max_turns
         self.batched = generator_cfg.batched
         self.use_conversation_multi_turn = generator_cfg.use_conversation_multi_turn
+        self.chat_template = generator_cfg.chat_template
```
CharlieFRuan:
```suggestion
self.custom_chat_template = get_custom_chat_template(generator_cfg.chat_template)
```

Having a `self.chat_template` might be confusing. let merge these two lines

skyrl-train/skyrl_train/generators/utils.py
```diff
@@ -21,19 +34,52 @@
         "{% endif %}"
         "{{mycontent + '<|im_end|>'}}"
         "{% endgeneration %}"
-        "{{'\n'}}"
```
CharlieFRuan, line 24: Could you remind me why we deleted the `'\n'` here? Could you add a unit test that tests whether the original qwen3's chat template is identical to our `qwen3_without_thinking`?

Say
```python
tokenizer = AutoTokenizer.from_pretrained("Qwen/Qwen3-4B")
messages = [
  {"content": "hi", "role": "system"},
  {"content": "hi", "role": "user"},
  {"content": "<think>thinking</think>hi", "role": "assistant"},
  {"content": "hi", "role": "user"},
  {"content": "<think>thinking</think>hi", "role": "assistant"}
  {"content": "hi", "role": "user"},
]
original_str = tokenizer.apply_chat_template(messages)
my_str = tokenizer.apply_chat_template(messages, chat_template=OUR_QWEN3_WITHOUT_THINKING)
```

I did not test this, but something along the line.

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
```diff
@@ -72,6 +71,11 @@ def mock_generate(input_batch):
 
     mock_llm.generate = AsyncMock(side_effect=mock_generate)
     # Create a mock generator config
+
+    chat_template_config = None
+    if "Qwen3" in model_name:
+        chat_template_config = {"source": "name", "name_or_path": "qwen3_without_thinking"}
```
CharlieFRuan:
```suggestion
    if "Qwen3" in model_name:
        chat_template_config = {"source": "name", "name_or_path": "qwen3_without_thinking"}
    else:
        chat_template_config = {"source": "name", "name_or_path": None}
```

Would this be a more realistic config? Since for other models we would still have the config, but just the default values (i.e. `None`)

CharlieFRuan (commented):
Thanks again, left more comments, should be close. Just wanted to:
1. Make sure whether to `raise ValueError` or `return None`
2. Ensure run_gsm8k still works
3. Add a unit test to check whether our `qwen3_without_thinking` is the same as the official Qwen3 chat template, hence checking whether the `"{{'\n'}}"` deletion is needed

skyrl-train/examples/gsm8k/run_gsm8k.sh
CharlieFRuan, line 1: please revert these

skyrl-train/skyrl_train/generators/utils.py
```diff
@@ -5,8 +5,20 @@
...
+    "qwen3_with_thinking": (
+        "{% for message in messages %}"
+        "{% if (message['role'] != 'assistant') %}"
+        "{{'<|im_start|>' + message['role'] + '\n' + message['content'] + '<|im_end|>' + '\n'}}"
+        "{% elif (message['role'] == 'assistant')%}"
+        "{{'<|im_start|>' + message['role'] + '\n'}}"
+        "{% generation %}"
+        "{{message['content'] + '<|im_end|>'}}"
+        "{% endgeneration %}"
+        "{% endif %}"
```
CharlieFRuan, line 20:
```suggestion
        "{% endgeneration %}"
        "{{'\n'}}"
        "{% endif %}"
```

Should we have a `"{{'\n'}}"` here? To match both `qwen3_without_thinking` and Qwen3's official template (albeit the thinking tokens), basically should be the same as Qwen2.5's

skyrl-train/skyrl_train/generators/utils.py
```diff
@@ -28,12 +40,46 @@
...
+    if source == "file":
+        try:
+            with open(name_or_path, "r", encoding="utf-8") as f:
+                return f.read()
+        except FileNotFoundError:
+            raise ValueError(f"Template file '{name_or_path}' not found")
+        except Exception as e:
+            raise ValueError(f"Error reading template file '{name_or_path}': {e}")
+
+    raise ValueError(f"Invalid source '{source}'. Must be 'name' or 'file'")
```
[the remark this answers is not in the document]
CharlieFRuan: This might be a good addition

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator.py
```diff
@@ -971,10 +985,8 @@ async def test_agent_loop_token_level_rewards_multi_turn_conversation_format(
     """use_conversation_multi_turn=True; verify rewards placed at ends of assistant segments before observations."""
     mock_tokenizer.eos_token_id = 4
 
-    # Tokenizer: initial prompt -> 2 tokens; observation template -> 2 tokens each call
```
CharlieFRuan, line 985: hmm why are we removing these comments?

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator.py
```diff
@@ -95,6 +102,7 @@ def mock_generator_cfg():
     cfg.max_input_length = 512
     cfg.batched = True
     cfg.max_turns = 1
+    cfg.chat_template = {"source": "name", "name_or_path": "qwen3_without_thinking"}
```
CharlieFRuan: Since this test is not about qwen3 or custom chat template, let's use the default `cfg.chat_template = {"source": "name", "name_or_path": None}`

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator.py
```diff
@@ -103,6 +111,7 @@ def mock_env_cfg():
     cfg = MagicMock()
     cfg.max_env_workers = 0
     cfg.env_class = "gsm8k"
+    cfg.get.return_value = MagicMock()
```
CharlieFRuan: Check if this is still needed if we do `cfg.chat_template = {"source": "name", "name_or_path": None}` for most places

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator.py
```diff
@@ -940,6 +953,7 @@ def close(self):
     cfg.max_turns = 10
     cfg.zero_reward_on_non_stop = False
     cfg.use_conversation_multi_turn = False
+    cfg.chat_template = {"source": "name", "name_or_path": "qwen3_without_thinking"}
```
CharlieFRuan: Same as above, since this test is not about qwen3 or custom chat template, let's use the default `cfg.chat_template = {"source": "name", "name_or_path": None}`

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator.py
```diff
@@ -1119,6 +1132,7 @@ def close(self):
     cfg.max_turns = 10
     cfg.zero_reward_on_non_stop = False
     cfg.use_conversation_multi_turn = True
+    cfg.chat_template = {"source": "name", "name_or_path": "qwen3_without_thinking"}
```
CharlieFRuan: **This one we can keep**, since we have `model_name="Qwen/Qwen3-0.6B",` and wanted to test the re-tokenize codepath specifically. Let's add a TODO comment about it. We should revisit this test once we separate the `retokenize` config from the custom chat template config

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator.py
```diff
@@ -1196,6 +1210,7 @@ def close(self):
     cfg.max_turns = 1
     cfg.zero_reward_on_non_stop = False
     cfg.use_conversation_multi_turn = False
+    cfg.chat_template = {"source": "name", "name_or_path": "qwen3_without_thinking"}
```
CharlieFRuan: Same as above, since this test is not about qwen3 or custom chat template, let's use the default `cfg.chat_template = {"source": "name", "name_or_path": None}`

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
```diff
@@ -187,6 +194,34 @@ def mock_generate(input_batch):
     assert generator_output["loss_masks"][0] == expected_loss_masks
 
 
+def test_qwen3_original_vs_without_thinking_chat_template():
+    """
+    Test comparing original Qwen3 chat template with qwen3_without_thinking.
+    Explains why we deleted the '\\n' - to match the tokenizer's actual behavior.
```
CharlieFRuan: Let's remove this `Explains why we deleted the '\\n' - to match the tokenizer's actual behavior.`, will be out of context for readers.

But just to confirm -- is deleting `\\n` or keeping it allowing us to match the tokenizer's actual behavior?

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator.py
CharlieFRuan, line 1: Could we add another test for `cfg.chat_template = {"source": "path", "name_or_path": "/path/to/file"}` since this codepath is not tested anywhere?

I know in this test we mock the tokenizer, so it'd be a bit dummy. But still good to test.

In a future PR we can improve this test by not mocking the tokenizer. Basically rebasing and improving https://github.com/NovaSky-AI/SkyRL/pull/197

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator.py
devpatelio: I mightve deleted them by accident when i modified the applied_chat_template_side_effect to return dictionaries, sorry for that!

skyrl-train/examples/gsm8k/run_gsm8k.sh
```diff
@@ -19,7 +19,7 @@ uv run --isolated --extra $INFERENCE_BACKEND -m skyrl_train.entrypoints.main_bas
   data.train_data="['$DATA_DIR/train.parquet']" \
   data.val_data="['$DATA_DIR/validation.parquet']" \
   trainer.algorithm.advantage_estimator="grpo" \
-  trainer.policy.model.path="Qwen/Qwen2.5-1.5B-Instruct" \
+  trainer.policy.model.path="Qwen/Qwen2.5-0.5B-Instruct" \
```
CharlieFRuan: revert please... In general, we don't want to have changes that are not needed to complete this PR's task, unless well-motivated

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator.py
```diff
@@ -103,6 +111,7 @@ def mock_env_cfg():
     cfg = MagicMock()
     cfg.max_env_workers = 0
     cfg.env_class = "gsm8k"
+    # cfg.get.return_value = MagicMock()
```
CharlieFRuan: remove this commented code please

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
```diff
@@ -284,3 +322,150 @@ def mock_generate(input_batch):
...
+            "response_ids": [tokenizer.encode(mock_llm_output_text, add_special_tokens=False)] * num_prompts,
+        }
+
+    mock_llm.generate = AsyncMock(side_effect=mock_generate)
+
+    test_file_dir = Path(__file__).parent
+    template_path = (
+        test_file_dir
+        / ".."
+        / ".."
```
CharlieFRuan: you can make a copy of that jinja to `./skyrl-train/tests/cpu/generators`. We want to keep things modularized. We don't want a unit test of skyrl-train depending on a jinja in skyagent

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
```diff
@@ -284,3 +322,150 @@ def mock_generate(input_batch):
         last_token_id_false = out_false["response_ids"][0][-1]
         assert last_token_id_true == tokenizer.eos_token_id
         assert last_token_id_false == tokenizer.encode(mock_text, add_special_tokens=False)[-1]
+
+
+def test_load_custom_jinja2_file_template():
```
CharlieFRuan: this test is not needed since the test below covers it. Remove this test

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
```diff
@@ -284,3 +322,150 @@ def mock_generate(input_batch):
...
+    assert "enable_thinking" in custom_template
+
+    print("✓ Successfully loaded custom jinja2 template")
+    print(f"  Template file: {chat_template_config['name_or_path']}")
+    print(f"  Template length: {len(custom_template)} characters")
+    print(f"  Contains thinking tags: {'<think>' in custom_template and '</think>' in custom_template}")
+
+
+@pytest.mark.asyncio
+async def test_generate_custom_jinja2_file_template():
```
CharlieFRuan: Let's move this test as a parametrized version of `test_skyrl_gym_generator_chat_templating_exact` to cut duplicated code. 

Something like `"model_name", ["Qwen/Qwen2.5-0.5B-Instruct", "unsloth/Llama-3.2-1B-Instruct", "Qwen/Qwen3-0.6B", "Qwen/Qwen3-0.6B-FROM_PATH"]`

and some post-processing of the model name. But essentially:

```python
    chat_template_config = None
    if model_name == "Qwen/Qwen3-0.6B":
        chat_template_config = {"source": "name", "name_or_path": "qwen3_without_thinking"}
    elif model_name == "Qwen/Qwen3-0.6B-FROM_PATH":
        chat_template_config = {"source": "file", "name_or_path": str(template_path)}
    else:
        chat_template_config = {"source": "name", "name_or_path": None}
```

devpatelio: @CharlieFRuan Ready for review! Might need to double check if the way I'm testing the thinking tokens is correct for the parameterized version. Let me know if there's any changes needed there but otherwise all tests are running and edits are made!

skyrl-train/tests/cpu/generators/qwen3_acc_thinking.jinja2
CharlieFRuan, line 1: it'd be good to note where you copied this jinja from. e.g. is it the original qwen3 jinja? Is it with all thinking tokens, or no thinking tokens? what is the difference between this and the ones we have in `CUSTOM_CHAT_TEMPLATES`? We could actually just reuse one from `CUSTOM_CHAT_TEMPLATES` to keep things simple, since we just want to test that the `from_path` codepath works

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
```diff
@@ -108,14 +131,43 @@ def mock_generate(input_batch):
     }
     generator_output: GeneratorOutput = await generator.generate(input_batch)
 
+    if is_custom_jinja_from_file:
```
CharlieFRuan: we can remove this if branch's asserts. since is the checks after this work, these checkings will likely work (i.e. we have harder checkings than those in this if branch)

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
```diff
@@ -183,8 +249,32 @@ def mock_generate(input_batch):
...
+    tokenizer = AutoTokenizer.from_pretrained("Qwen/Qwen3-4B")
+
+    messages = [
+        {"content": "hi", "role": "system"},
+        {"content": "hi", "role": "user"},
+        {"content": "<think>thinking</think>hi", "role": "assistant"},
+        {"content": "hi", "role": "user"},
+        {"content": "<think>thinking</think>hi", "role": "assistant"},
+        {"content": "hi", "role": "user"},
+    ]
```
CharlieFRuan, line 221: this looks good

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
```diff
@@ -166,11 +218,25 @@ def mock_generate(input_batch):
...
-        expected_loss_masks = (
-            expected_assistant_loss_mask  # <|im_start|>assistant\nb<|im_end|>\n
-            + expected_user_loss_mask  # <|im_start|>user\n1<|im_end|>\n
-        ) * 2 + expected_assistant_loss_mask  # last <|im_start|>assistant\nb<|im_end|>\n
+        if is_custom_jinja_from_file:
+            actual_loss_masks = generator_output["loss_masks"][0]
+            actual_response_ids = generator_output["response_ids"][0]
+            resp_str = tokenizer.decode(actual_response_ids)
+
+            print(f"  Thinking template response string: {repr(resp_str)}")
```
CharlieFRuan: hmm why aren't we checking loss mask as rigorously as the other codepaths for `is_custom_jinja_from_file`?

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
```diff
@@ -49,20 +50,31 @@ def _register_test_env_if_needed():
 
 @pytest.mark.asyncio
 @pytest.mark.parametrize(
-    "model_name", ["Qwen/Qwen2.5-0.5B-Instruct", "unsloth/Llama-3.2-1B-Instruct", "Qwen/Qwen3-0.6B"]
+    "model_name",
+    ["Qwen/Qwen2.5-0.5B-Instruct", "unsloth/Llama-3.2-1B-Instruct", "Qwen/Qwen3-0.6B", "Qwen/Qwen3-0.6B-FROM_PATH"],
```
CharlieFRuan, line 54: for the `FROM_PATH`, it can be identical in behavior to the `Qwen/Qwen3-0.6B` codepath. We just want to check that the codepath works. So we don't need to have different llm outputs for `if is_custom_jinja_from_file`. That is, in the jinja path, we can just put the same jinja as `CUSTOM_CHAT_TEMPLATE["qwen3_without_thinking"]`.

Then we can remove many of the if else below.

skyrl-train/tests/cpu/generators/qwen3_acc_thinking.jinja2
devpatelio: I copied the skyrl-agent/skyrl_agent/functional/templates/qwen3_acc_thinking.jinja2 file template so it has the thinking tokens. I can also just make a jinja2 file from the CUSTOM_CHAT_TEMPLATES version of the thinking template if that helps as well. Let me know what you prefer!

CharlieFRuan: let's make `Qwen/Qwen3-0.6B` and `Qwen/Qwen3-0.6B-FROM_PATH` identical in behavior except the latter loading from a jinja file (i.e. loads a jinja identical to that will be used for the former (which is from CUSTOM_CHAT_TEMPLATES)). This will make the unit test cleaner!

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
```diff
@@ -166,11 +187,12 @@ def mock_generate(input_batch):
     expected_user_loss_mask = [0] * len(empty_user) + [0]  # extra 0 for single observation token
 
     if custom_chat_template is not None:
-        # For custom_chat_template, the first generation prompt IDs are part of `resp_str`, hence has corresponding mask
         expected_loss_masks = (
             expected_assistant_loss_mask  # <|im_start|>assistant\nb<|im_end|>\n
             + expected_user_loss_mask  # <|im_start|>user\n1<|im_end|>\n
         ) * 2 + expected_assistant_loss_mask  # last <|im_start|>assistant\nb<|im_end|>\n
+        assert len(expected_loss_masks) == len(generator_output["loss_masks"][0])
```
CharlieFRuan: You can leave this unchanged in the PR. the if an else share the same two lines

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
```diff
@@ -166,11 +187,12 @@ def mock_generate(input_batch):
     expected_user_loss_mask = [0] * len(empty_user) + [0]  # extra 0 for single observation token
 
     if custom_chat_template is not None:
-        # For custom_chat_template, the first generation prompt IDs are part of `resp_str`, hence has corresponding mask
```
CharlieFRuan, line 169: Keep this comment

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
```diff
@@ -183,8 +205,42 @@ def mock_generate(input_batch):
         )
         if "Qwen" in model_name:
             expected_loss_masks = expected_loss_masks[:-1]  # remove the extra 0 for \n
-    assert len(expected_loss_masks) == len(generator_output["loss_masks"][0])
-    assert generator_output["loss_masks"][0] == expected_loss_masks
+        assert len(expected_loss_masks) == len(generator_output["loss_masks"][0])
```
CharlieFRuan: un-indent these two lines to be shared by the if and else

skyrl-train/skyrl_train/generators/utils.py
```diff
@@ -28,12 +41,45 @@
 }
 
 
-def get_custom_chat_template(model_name: str) -> str:
-    if "Qwen3" in model_name:
-        return CUSTOM_CHAT_TEMPLATES["qwen3_thinking"]
-    else:
+def get_custom_chat_template(chat_template_config: Optional[dict] = None) -> Optional[str]:
```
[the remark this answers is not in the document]
CharlieFRuan: @devpatelio this suggestion is actually correct. We should do `from omegaconf import DictConfig` at the top, and change it to `Union[dict, DictConfig]`. This is because `generator_cfg.chat_template` is a DictConfig and not a dict.

Would be good to quickly check that `run_gsm8k.sh` works with:
- Not specifying anything (i.e. using the default `chat_template` values)
- Specifying something (e.g. from name)

PR #184, opened 2025-08-22
get chat_end_index after env init

### LINKED ISSUE #183: When `env.init` changes the prompt length, the `chat_end_index` is wrong
https://github.com/NovaSky-AI/SkyRL/blob/825f2e82e289d6011d80957c48132618fed3d460/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py#L90-L94

In our use case, we are injecting the system prompt in addition to the user message in the `env.init`, so the length changes before and after.
In this case the `chat_end_index` is 1 instead of 2. We think it might be causing some bugs down the line, leading to this assertion error:
```
AssertionError: Response ids and loss masks must have the same length, for sample 0 got 27 and 50
```

etnbrd:
This PR moves the `chat_end_index` to after the `chat_history` has been modified by the `env.init` function, to get the right number.

In our use case, we are injecting the system prompt in addition to the user message in the `env.init`, so the length changes before and after.
In this case the `chat_end_index` is 1 instead of 2. We think it might be causing some bugs down the line, leading to this assertion error:
```
AssertionError: Response ids and loss masks must have the same length, for sample 0 got 27 and 50
```

Resolves #183

### DIFF
--- a/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
+++ b/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
@@ -88,10 +88,10 @@ class SkyRLGymGenerator(GeneratorInterface):
 
         # need copy here since the prompt is a list of messages and we are going to modify it
         chat_history = copy.deepcopy(prompt)
-        chat_end_index = len(chat_history)
 
         # Init() returns the first prompt to be given to the model, and optional metadata dict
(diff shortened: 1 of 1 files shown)


tyler-griggs:
Great catch, thanks

tyler-griggs (approved):

PR #185, opened 2025-08-22
[DRAFT] Enable cloud checkpointing

bthecohen:
This is a draft proposal for how to enable cloud checkpointing. It's a bit invasive; basically wraps all local disk operations with some utils that support cloud storage via `fsspec`.


tyler-griggs:
Couldn't push to this branch, so re-uploaded the final PR here: #248

Thanks so much :)

PR #191, opened 2025-08-24
[docs] Update libnuma installation instructions

tyler-griggs:
Add quick script for installing numactl without using docker. This is relevant for those who don't want to use docker, but don't have sudo access (e.g., RunPod VMs).

Screenshot 2025-08-24 at 4 56 40 PM


CharlieFRuan (approved):
Thanks!

PR #192, opened 2025-08-25
[Fix] Set skip_special_tokens to True in default sampling params

pgasawa:
By default, SkyRL currently calls inference engines with `skip_special_tokens=False`. This causes special tokens (e.g. `<|im_end|>`) to appear in the decoded text output. This means that the decoded output passed in to environments via `env.step(output)` (in the `agent_loop`) may include stop tokens, which likely isn’t expected behavior.

This PR addresses that by setting skip_special_tokens=True when sampling from vLLM and SGLang.

I briefly validated this change functions as intended by comparing two runs w/ and w/o, and am no longer seeing the `<|im_end|>` tokens when I log `output`.


tyler-griggs (approved):
thanks!

PR #194, opened 2025-08-25
[cleanup] Make agent_loop output a dataclass

tyler-griggs:
The tuple return was getting out of control

### DIFF
--- a/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
+++ b/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
@@ -6,6 +6,7 @@ from typing import List, Dict, Any, Optional, Tuple
 import numpy as np
 from concurrent.futures import ThreadPoolExecutor
 from tqdm.asyncio import tqdm
+from dataclasses import dataclass
 
 from skyrl_train.generators.base import GeneratorInterface, GeneratorInput, GeneratorOutput
 from skyrl_train.inference_engines.inference_engine_client import InferenceEngineClient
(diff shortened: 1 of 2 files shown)


skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -15,6 +16,17 @@
 from skyrl_train.generators.utils import get_custom_chat_template, get_generation_prompt_ids, apply_overlong_filtering
 
 
+@dataclass
```
CharlieFRuan, line 19: nit: InferenceEngineInput/Output and GeneratorInput/Output are all `TypedDict`, perhaps we can make this `AgentLoopOutput` a `TypedDict` as well to stay uniformed? Though functionally not much difference.

tyler-griggs: Yeah, what do you think is better here? I actually like the `dataclass` more because of the automatic runtime checks. I believe the reason we went with TypedDict for the other interface definitions originally is because dict-like structures are generally easier to deal with at API boundaries (eg, because they're easier to serialize to json), so I thought that'd be helpful as we separate the different major components (generator, trainer). But admittedly, beyond that, I don't remember all trade-offs considered when we initially went with `TypedDict` for other interface definitions.

CharlieFRuan: Yeah that's fair! @gemini-code-assist what's your take

CharlieFRuan: Makes sense, let's go with dataclass!

PR #195, opened 2025-08-25
1/N GPU CI Migration

tyler-griggs:
Migrating a first batch of tests over to `gpu_ci`, for #199


skyrl-train/examples/search/searchr1_dataset.py
tyler-griggs, line 1: The `test_skyrl_gym_generator` test uses the SearchR1 dataset, but only the test split (which is much smaller than the train split), so this change enables only downloaded the test split

skyrl-train/tests/gpu/gpu_ci/test_skyrl_gym_generator.py
tyler-griggs, line 1: This file is just a copy-paste of the original SkyRL Gym test, but left behind on test (text2sql) that needs to be made more lightweight before transferring over

skyrl-train/tests/gpu/gpu_ci/test_trainer_full_checkpointing.py
SumanthRH, line 1: how much time does this take on L4 now?

tyler-griggs: I measured the full block of migrated tests (the 5 files in this PR), which was originally 41min but is now 23min. Still quite slow but... a start.

SumanthRH: ok sg, let's revisit once all files are moved over

tyler-griggs: Agreed, and added this as step 2 in the issue

PR #206, opened 2025-08-26
SkyPilot Example

lucamanolache:
Adds an example of how to launch a training run on GSM8K with [skypilot](https://github.com/skypilot-org/skypilot).


skyrl-train/docs/tutorials/skypilot.rst
```diff
@@ -0,0 +1,167 @@
...
+-------
+
+`SkyPilot <https://docs.skypilot.co/en/latest>`_ is a system to run, manage,
+and scale AI workloads on any infrastructure. It handles all of the logic of
+provisioning, setting up, and tearing down clusters on any cloud.
+
+Setup Skypilot
+--------------
+
+To setup SkyPilot on AWS
```
lynnliu030: Does this example only work with AWS? we can say skypilot works with all clouds, but here we provide aws as an example

skyrl-train/examples/gsm8k/gsm8k-grpo-skypilot.yaml
```diff
@@ -0,0 +1,125 @@
...
+  uv run -- python3 -c "import wandb; wandb.login(relogin=True, key='$WANDB_API_KEY')"
+
+  export RAY_RUNTIME_ENV_HOOK=ray._private.runtime_env.uv_runtime_env_hook.hook
+  if [ "$SKYPILOT_NODE_RANK" == "0" ]; then
+    ps aux | grep ray | grep 6479 &> /dev/null || ray start --head  --disable-usage-stats --port 6479
+    sleep 15
+    uv run --isolated --extra "$INFERENCE_BACKEND" -m skyrl_train.entrypoints.main_base \
+      data.train_data="['${DATA_DIR}/train.parquet']" \
+      data.val_data="['${DATA_DIR}/validation.parquet']" \
+      trainer.algorithm.advantage_estimator="grpo" \
```
lynnliu030, line 95: rename this file to something like `gsm8k-grpo-skypilot.yaml`?

lynnliu030 (commented):
I think it'd be useful to also have some screenshot similar to what's shown in [here](https://verl.readthedocs.io/en/latest/start/multinode.html#option-2-launch-via-skypilot-on-kubernetes-or-clouds), maybe just one for the skypilot successful setup and one for successfully starting the job?

skyrl-train/docs/tutorials/skypilot.rst
lucamanolache: When testing I used first `nebius` instead of `aws` since it was cheaper. The only changes needed for switching between clouds would be potentially the GPU names (its been a while since i tested it but iirc `nebius` didn't have L4s but had L40s)

lynnliu030: Great! I think then we can just say this instruction works for setting up SkyPilot on any clouds, but we just show one example with AWS

andylizf:
@lucamanolache we are not using docker image right? Be careful that skypilot-org/skypilot/pull/7181 has not released yet (still on nightly-build only). So `erictang000/skyrl-train-ray-2.48.0-py3.12-cu12.8` won't work. We should pin skypilot version to > 0.10.3, or have some comments about this.

skyrl-train/docs/tutorials/skypilot.rst
```diff
@@ -0,0 +1,168 @@
...
+
+.. code-block:: bash
+
+    conda create -y -n sky python=3.10
```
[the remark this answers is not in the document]
andylizf: lets use uv instead. ;)

skyrl-train/docs/tutorials/skypilot.rst
```diff
@@ -0,0 +1,168 @@
...
+      INFERENCE_BACKEND: "vllm"
+      # INFERENCE_BACKEND: "sglang"
+
+
+    # --------------- Environment Setup (setup) ---------------
+    # Commands run on each node of the remote cluster to set up the environment (e.g., install dependencies). These are run directly inside Docker.
+    setup: |
+      sudo apt-get update
+      sudo apt-get install libpython3.12-dev -y
+      sudo apt-get install python-dev python3-dev -y
```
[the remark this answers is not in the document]
andylizf: +1. iiuc we can just remove these 3 lines.

skyrl-train/docs/tutorials/skypilot.rst
```diff
@@ -0,0 +1,167 @@
...
+      cd skyrl-train
+      uv venv --python 3.12
```
andylizf:
```suggestion
      uv venv --python 3.12 --seed
```

skyrl-train/docs/tutorials/skypilot.rst
```diff
@@ -0,0 +1,167 @@
...
+    # check the skypilot documentation for what clouds network tier is supported.
+    #  network_tier: best # when using multiple nodes, communication can become a bottleneck
+
+    num_nodes: 1         # cluster size
+
+    # --------------- Work Directory Synchronization (workdir) ---------------
+    # Defines the local working directory to be synchronized to the remote cluster.
+    # Here, '.' means synchronizing the directory where the sky submit command is currently run.
+    # For this script, we are running the command from SkyRL
+    workdir: .
```
andylizf, line 66: should first say, clone SkyRL, and cd into the dir, and run `sky launch`...

andylizf (changes requested):
Left some discussions.

skyrl-train/docs/tutorials/skypilot.rst
```diff
@@ -0,0 +1,168 @@
...
+    # lines if you have the data locally already.
+
+    #file_mounts:
+    #  # Remote path (relative to remote user's home directory): Local path
+    #  # /remote/dir1/file: /local/dir1/file
+    #  data/gsm8k: ~/data/gsm8k
+
+    # --------------- Enviornment variables ---------------
+    envs:
+      NUM_GPUS: 4
```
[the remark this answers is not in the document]
andylizf: Good point!

skyrl-train/examples/gsm8k/gsm8k-grpo-skypilot.yaml
```diff
@@ -0,0 +1,126 @@
...
+  num_nodes="$(echo "$SKYPILOT_NODE_IPS" | wc -l | awk '{print $1}')"
+
+  DATA_DIR=$HOME/data/gsm8k/
+
+  # login wandb
+  uv run -- python3 -c "import wandb; wandb.login(relogin=True, key='$WANDB_API_KEY')"
+
+  export RAY_RUNTIME_ENV_HOOK=ray._private.runtime_env.uv_runtime_env_hook.hook
+  if [ "$SKYPILOT_NODE_RANK" == "0" ]; then
+    ps aux | grep ray | grep 6479 &> /dev/null || ray start --head  --disable-usage-stats --port 6479
```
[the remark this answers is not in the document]
andylizf: That's a good point. Should fix this.

lucamanolache:
> @lucamanolache we are not using docker image right? Be careful that [skypilot-org/skypilot/pull/7181](https://github.com/skypilot-org/skypilot/pull/7181) has not released yet (still on nightly-build only). So `erictang000/skyrl-train-ray-2.48.0-py3.12-cu12.8` won't work. We should pin skypilot version to > 0.10.3, or have some comments about this.

I think I'm using a different `ray` than `skypilot` is using. Its been a long time since I've last looked at this so not too sure about how I got it to work. I remember having some issues related to `ray` version when working on this.

skyrl-train/docs/tutorials/skypilot.rst
lucamanolache: yes this would probably be better lol

andylizf:
> I think I'm using a different `ray` than `skypilot` is using. Its been a long time since I've last looked at this so not too sure about how I got it to work. I remember having some issues related to `ray` version when working on this.

Yes, but it's worth noting that skypilot cannot directly use image `erictang000/skyrl-train-ray-2.48.0-py3.12-cu12.8` for some reasons.

lynnliu030 (approved):

PR #212, opened 2025-08-28
Fix `torch.autocast` incompatibility with DeepSpeed

tohtana:
The current usage of `torch.autocast` has two issues when combined with DeepSpeed:
- `torch.autocast` is not compatible with DeepSpeed's bf16 training. `torch.autocast` expects FP32 parameters, but DeepSpeed converts all parameters to BF16.
- Starting from v0.17.2, DeepSpeed supports `torch.autocast` with ZeRO and uses it internally. If `torch.autocast` is already enabled outside the DeepSpeed engine, it raises an error.

TThis PR updates the behavior as follows:
- Replace `torch.autocast` with `strategy.autocast`, which is nop when DeepSpeed is enabled
  - `strategy.autocast` becomes a no-op when DeepSpeed is enabled.
  - This allows compatibility with newer DeepSpeed versions.
- Add a config item `torch_autocast` in yaml. 
  - If true: DeepSpeed integrates `torch.autocast` with ZeRO, as described [here](https://deepspeed.readthedocs.io/en/latest/training.html#pytorch-automatic-mixed-precision-amp).
  - If false, DeepSpeed uses its own BF16 mixed precision.

### DIFF
--- a/skyrl-train/pyproject.toml
+++ b/skyrl-train/pyproject.toml
@@ -69,7 +69,7 @@ flashinfer-python = [
 
 [project.optional-dependencies]
 deepspeed = [
-    "deepspeed==0.16.5"
+    "deepspeed==0.17.5"
 ]
 dev = [
     "ruff==0.11.9",
--- a/skyrl-train/skyrl_train/config/deepspeed_config/eval.yaml
+++ b/skyrl-train/skyrl_train/config/deepspeed_config/eval.yaml
@@ -5,6 +5,9 @@ zero_optimization:
   offload_param:
     device: cpu # set to cpu for offload, else none
     pin_memory: true
+# set true to use torch.autocast. If false, deepspeed uses its own mixed precision handling
+torch_autocast:
+  enabled: false
 gradient_clipping: 1.0
 prescale_gradient: false
 wall_clock_breakdown: false
--- a/skyrl-train/skyrl_train/config/deepspeed_config/train.yaml
+++ b/skyrl-train/skyrl_train/config/deepspeed_config/train.yaml
@@ -20,6 +20,9 @@ zero_optimization:
   zero_hpz_partition_size: 1
   zero_quantized_weights: false
   zero_quantized_gradients: false
+# set true to use torch.autocast. If false, deepspeed uses its own mixed precision handling
+torch_autocast:
+  enabled: true
 # set true to save memory
 # disables stage3_prefetch_bucket_size, stage3_max_live_parameters, stage3_max_reuse_distance
 disable_trace_cache: false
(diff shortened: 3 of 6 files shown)


tohtana:
Thank you @erictang000,

The tests pass, but I found a couple of issues that need to be fixed on the DeepSpeed side:
- Gradient aggregation of tied modules can be inaccurate (Submitted a [PR](https://github.com/deepspeedai/DeepSpeed/pull/7534) to fix this)
- `torch.autocast` is not applied to the loss calculation after the model's forward pass

I’ll submit another PR to the DeepSpeed repo to address the second issue. Let’s hold off on this PR until those fixes are merged.

tohtana:
I found that the above approach doesn't apply autocast to loss function, which could be sensitive to precision. So I made some changes on DeepSpeed side to make it better with the use case in this repository. (I reverted the changes in "strategy" classes).

To verify the correctness, I ran gsm8k example with FSDP2, DeepSpeed (builtin mixed precision), and DeepSpeed (torch.autocast). The results show a pretty good match.
image

The performances are also very close.
image

The changes on DeepSpeed side will be included in the next patch release. Once it is merged and I verify it again, we will be ready to merge this PR.

tohtana:
I found that the previous approach didn’t apply autocast to the loss function, which can be sensitive to precision. To address this, I made some [changes on the DeepSpeed side](https://github.com/deepspeedai/DeepSpeed/pull/7543) to improve compatibility with the use case in this repository. (I reverted the earlier changes in the strategy classes.)

To verify correctness, I ran the gsm8k example with: FSDP2, DeepSpeed (builtin mixed precision), and DeepSpeed (torch.autocast). The results match closely:
image

Performance is also very similar:
image

The DeepSpeed-side changes will be included in the next patch release. Once that is merged and I verify again, we’ll be ready to merge this PR. With the new version, we’ll also be able to use ZeRO stages 1 and 2.

erictang000:
@tohtana is this ready to merge after the 0.17.6 release? https://github.com/deepspeedai/DeepSpeed/releases/tag/v0.17.6

tohtana:
@erictang000 Sorry for the delay. I resolved the conflict and verified again.
The loss values look good.
image

The performance is also okay. 
image

DeepSpeed seems slow for evaluation. I will take a look to see if we can improve it. Also, I will try offloading for ZeRO 2/3.

tohtana:
@erictang000 Can we merge this PR now?

I think I can work on these on different PRs.
> DeepSpeed seems slow for evaluation. I will take a look to see if we can improve it. Also, I will try offloading for ZeRO 2/3.

PR #217, opened 2025-08-28
[InfEngines] Lift tokenization up to InferenceEngineClient

tyler-griggs:
## What does this PR do?
Previously, `apply_chat_template` calls were sprinkled throughout the code to tokenize inference engine prompts at the engine (`vllm_engine`, `sglang_engine`) or server (`remote_inference_engine`) level. This is unnecessary. All inference requests go through the `InferenceEngineClient` so this PR moves tokenization up to the client. Now, the engine and server level will expect (and require) token IDs are passed instead of text.

This will also be useful to perform some automated SamplingParam updates, such as reducing the `max_generate_length` if the prompt is getting close to the `max_model_len`

## Testing
Updated GPU tests.

## What's next?
Perform the automated SamplingParam updates discussed above.

### DIFF
--- a/skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
+++ b/skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
@@ -63,7 +63,9 @@ class InferenceEngineClient(InferenceEngineInterface):
         """
         # Group prompts by engine
         engine_groups: dict[int, dict[str, list]] = {}
-        assert len(prompt_token_ids) == len(trajectory_ids), f"Mismatch between number of prompts ({len(prompt_token_ids)}) and trajectory_ids ({len(trajectory_ids)})"
+        assert len(prompt_token_ids) == len(
+            trajectory_ids
+        ), f"Mismatch between number of prompts ({len(prompt_token_ids)}) and trajectory_ids ({len(trajectory_ids)})"
         for i, (token_ids, traj_id) in enumerate(zip(prompt_token_ids, trajectory_ids)):
             engine_idx = abs(hash(str(traj_id))) % len(self.engines)
             group = engine_groups.setdefault(engine_idx, {"token_ids": [], "indices": []})


skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
```diff
@@ -36,39 +38,44 @@ async def generate(self, input_batch: InferenceEngineInput) -> InferenceEngineOu
...
             raise ValueError("Either `prompts` or `prompt_token_ids` must be provided, but not both.")
+        if prompt_token_ids is None:
+            prompt_token_ids = self.tokenizer.apply_chat_template(
+                prompts,
+                add_generation_prompt=True,
+                add_special_tokens=False,
+                return_dict=True,
+                tokenize=True,
+            )["input_ids"]
+            prompts = None
```
CharlieFRuan:
```suggestion
            del prompts
```

Maybe this is more explicit?

CharlieFRuan (approved):
Thanks, much cleaner!

On a side note, I know this PR only focuses on preventing low-level engines from calling `tokenizer.apply_chat_template()`, and that we remove passing `tokenizer` to vllm_engine but not:
- `remote_inference_engine.py` because vllm server only returns text, and sglang only returns tokens
- `sglang_engine.py` because sglang only returns tokens
- So we need a tokenizer in those engines to decode output tokens or encode output string

I think using `return_logprob=True` for sglang might allow `sglang_engine` and `sglang_server` to return both text and token with some overhead.

```python
import sglang as sgl

def main():
    engine = sgl.Engine(
        model_path="Qwen/Qwen2.5-1.5B-Instruct",
        disable_cuda_graph=True,
    )
    sampling_params = {
        "temperature": 0.0,
        "max_new_tokens": 16,
    }
    output = engine.generate(input_ids=[1, 2, 3], sampling_params=sampling_params, return_logprob=True)
    print(output)

if __name__ == "__main__":
    main()
```

Besides, after vLLM's token-out PR is released, vLLM should be able to get both text and token output as well.

Thus, in future PR perhaps we can remove tokenizers completely from engines at a lower level than `InferenceEngineClient`.

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
tyler-griggs: Ah I'll just remove it. I only had it there when `prompts` was later passed into the later methods. But I removed it as an argument to the other methods so we don't need it any more.

tyler-griggs:
Thanks! Your point about removing tokenizer from everything below the `InferenceEngineClient` is a good one -- let's do it.

Issue #203, opened 2025-08-25
Support Megatron training backend for MoE training


tyler-griggs:

erictang000:
- [x] Megatron DP Convergence (see https://github.com/NovaSky-AI/SkyRL/pull/223#issuecomment-3256455433)

benrhodes26:
Is this checklist up-to-date?

erictang000:
hey @benrhodes26, just took a pass, should be up to date now!

PR #223, opened 2025-08-29
[trainer] Initial Megatron TP + PP Support

erictang000:
# Overview
This PR initializes Megatron support for GRPO with TP + PP (and DP) supported. We use [mbridge](https://github.com/ISEEKYAN/mbridge) to translate from huggingface to megatron, and plan to migrate to [Megatron-Bridge](https://github.com/NVIDIA-NeMo/Megatron-Bridge) once the repo is more mature and we can safely upgrade some dependencies.

## What's not Included
For a full set of limitations see the code here: https://github.com/NovaSky-AI/SkyRL/blob/1350d7be6a340587596c91cd1e3854c6900fe3ce/skyrl-train/skyrl_train/utils/utils.py#L117.

Note: there is a known issue with `Qwen2.5-1.5B` and `Qwen2.5-3B` models, which don't specify `lm_head.weight` in the model.safetensors file, causing a weight loading issue with `mbridge` - issue here: https://github.com/ISEEKYAN/mbridge/issues/10.

## Followups
For a full task list needed for full Megatron support, see #203.

## Reproduction runs (gsm8k with Qwen3-0.6B)
image

## TODOs

- [x] Port GPU test to be runnable without locally saved training batch input
- [x] fix entropy calculation
- [x] dependencies
- [x] add any needed doc-strings
- [x] code-attribution if i missed anything


skyrl-train/skyrl_train/workers/megatron/megatron_policy.py
```diff
@@ -0,0 +1,328 @@
...
+            policy_loss, clip_ratio = self.policy_loss_fn(
+                action_log_probs,
+                old_action_log_probs,
+                advantages,
+                config=self.cfg.trainer.algorithm,
+                loss_mask=loss_mask,
+                rollout_logprobs=rollout_action_logprobs,
+            )
+
+            # with torch.no_grad():
```
erictang000: todo: fix

skyrl-train/tests/gpu/test_megatron_worker.py
```diff
@@ -0,0 +1,380 @@
+"""
```
erictang000, line 1: todo: using a batch i saved locally for this at the moment - port over to a smaller training batch sample

skyrl-train/pyproject.toml
```diff
@@ -108,6 +119,22 @@ flashrl = [
     "torchvision",
 ]
 
+[dependency-groups]
+
+# This is a default group so that we install these even with bare `uv sync`
```
erictang000: todo: figure out how to reconcile deps - flash-rl requires torch 2.7.0, but in order to build transformer-engine-pytorch, i need torch 2.7.1. I think I can get this working by asking the user to build transformer-engine first, but need to test out installation in a fresh env.

skyrl-train/skyrl_train/distributed/megatron/megatron_utils.py
SumanthRH, line 1: nit: let's add our references (ex: verl  / nemo rl)  for these files for future reference

SumanthRH: Also helps while reviewing :)

skyrl-train/pyproject.toml
SumanthRH: FlashRL extra is currently standalone, so for the initial integration we can just say that the two dep groups `mcore` and `flashrl` conflict, and that should be enough right?

erictang000: problem was that i needed to include the build dependency-group as a default which clashes torch versions with flash-rl. but this is maybe just needed to build transformer-engine so can give instructions to build it separately include it as a no-build-isolation dep (btw this takes like more than 10 mins :(. )

will see how that goes tho

SumanthRH: Ah I see

skyrl-train/skyrl_train/config/megatron_config/policy.yaml
```diff
@@ -0,0 +1,39 @@
+# @package megatron_config.policy
+tensor_model_parallel_size: 1
+pipeline_model_parallel_size: 1
+context_parallel_size: 1
+expert_model_parallel_size: 1
+expert_tensor_parallel_size: 1
+
+# Allow to override Distributed Data Parallel (DDP) config
+# TODO (erictang000): understand exactly what is going on here
```
erictang000: pending...

skyrl-train/skyrl_train/config/megatron_config/policy.yaml
```diff
@@ -0,0 +1,39 @@
...
+ddp_config:
+  grad_reduce_in_fp32: false
+  overlap_grad_reduce: false
+  overlap_param_gather: false
+  average_in_collective: true
+  data_parallel_sharding_strategy: "optim_grads_params"
+
+seed: 42
```
SumanthRH: Nit: is this seed argument needed here? It looks like we pass `cfg.trainer.seed` to `MegatronStrategy` at init, so we should not have it in the config here

skyrl-train/skyrl_train/distributed/megatron/megatron_utils.py
```diff
@@ -0,0 +1,358 @@
...
+
+    def _iter_opts(opt):
+        if isinstance(opt, ChainedOptimizer):
+            return opt.chained_optimizers
+        return [opt]
+
+    def offload_tensor_to_cpu(tensor):
+        if tensor is None:
+            return
+        tensor.data = tensor.data.to("cpu", non_blocking=True)
```
SumanthRH, line 187:
```suggestion
        tensor = tensor.to("cpu", non_blocking=True)
```

skyrl-train/skyrl_train/distributed/megatron/megatron_utils.py
```diff
@@ -0,0 +1,358 @@
...
+    def _iter_opts(opt):
+        if isinstance(opt, ChainedOptimizer):
+            return opt.chained_optimizers
+        return [opt]
+
+    def load_tensor_to_gpu(tensor):
+        if tensor is None:
+            return
+        device_id = torch.cuda.current_device()
+        tensor.data = tensor.data.to(device_id, non_blocking=True)
```
SumanthRH, line 228:
```suggestion
        tensor = tensor.to(device_id, non_blocking=True)
```

skyrl-train/skyrl_train/utils/ppo_utils.py
```diff
@@ -505,6 +505,11 @@ def sync_registries():
     logger.info("Synced registries to ray actor")
 
 
+def _safe_exp_delta(delta: torch.Tensor, clip: float = 20.0, out_dtype=None) -> torch.Tensor:
```
SumanthRH, line 508: a small comment on overflow/ underflow here would be nice

skyrl-train/skyrl_train/workers/megatron/megatron_worker.py
```diff
@@ -0,0 +1,353 @@
...
+        self.strategy.offload_to_cpu(self.actor_module, self.optimizer, pin_memory, non_blocking)
+
+    def backload_to_gpu(self, non_blocking=True):
+        self.strategy.backload_to_gpu(self.actor_module, self.optimizer, non_blocking)
+
+    def init_worker_process_group(self):
+        if not torch.distributed.is_initialized():
+            torch.distributed.init_process_group(backend="nccl")
+
+        # override the init_process_group method to use megatron distributed setup to create the mesh
```
SumanthRH: comment doesn't match the code? (or maybe is insufficient)

skyrl-train/skyrl_train/workers/megatron/megatron_worker.py
```diff
@@ -0,0 +1,353 @@
...
+        Perform one micro-batch of training, accumulate gradients, and step the optimizer only after `accumulation_steps` micro-batches.
+        """
+        self.model.train()
+        experience.to_device(torch.cuda.current_device())
+
+        sequences = experience.sequences
+        num_actions = experience.num_actions
+        attention_mask = experience.attention_mask
+
+        metrics = self.model.forward_backward_micro_batch(
```
SumanthRH: Within a DP replica, all ranks will end up with the same metrics object, and then we do a simple all reduce over the full world size to get the metrics for the global micro batch in `ppo_train` - impl in Worker. 

This is worth mentioning as a comment here?

skyrl-train/skyrl_train/workers/megatron/megatron_worker.py
```diff
@@ -0,0 +1,353 @@
...
+        self.strategy.offload_to_cpu(self.actor_module, None, pin_memory, non_blocking)
+
+    def backload_to_gpu(self, non_blocking=True):
+        self.strategy.backload_to_gpu(self.actor_module, None, non_blocking)
+
+    def init_worker_process_group(self):
+        if not torch.distributed.is_initialized():
+            torch.distributed.init_process_group(backend="nccl")
+
+        # override the init_process_group method to use megatron distributed setup to create the mesh
```
SumanthRH: same comment as before

skyrl-train/skyrl_train/workers/megatron/megatron_worker.py
```diff
@@ -0,0 +1,353 @@
...
+
+    def init_model(self, model_path, num_training_steps: int = 1e9):
+        # get hf_config and tf_config
+        self.init_configs(
+            model_path,
+            self.cfg.trainer.ref.megatron_config.model_config_kwargs,
+            self.cfg.trainer.ref.megatron_config.transformer_config_kwargs,
+        )
+
+        # wrap with DDP for training
```
SumanthRH: Comment is incorrect - `wrap_with_ddp` is False

skyrl-train/skyrl_train/workers/worker.py
```diff
@@ -805,7 +808,8 @@ def _forward_micro_batch(self, micro_batch: TrainingInputBatch) -> TrainingOutpu
         sequences = micro_batch["sequences"]
         response_length = micro_batch.metadata["response_length"]
         attention_mask = micro_batch["attention_mask"]
-        with torch.no_grad(), torch.autocast(dtype=torch.bfloat16, device_type="cuda"):
+
```
SumanthRH, line 811: Is this change needed?

skyrl-train/skyrl_train/workers/megatron/megatron_worker.py
```diff
@@ -0,0 +1,353 @@
...
+        bridge.set_extra_args(**transformer_config_kwargs)
+        tf_config = bridge.config
+        self.bridge = bridge
+
+        self.hf_config = hf_config
+        self.tf_config = tf_config
+        self.tokenizer = tokenizer
+
+    def make_megatron_module(self, model_config_kwargs, wrap_with_ddp=True):
+        model = self.bridge.get_model(
```
SumanthRH, line 69: Nit: Type hint? 

Looks like this is no longer using the model classes in models.py, good to highlight. (is this just an `AutoBridge` instance, or anything else)

skyrl-train/skyrl_train/workers/megatron/megatron_policy.py
```diff
@@ -0,0 +1,328 @@
...
+        def forward_step(batch_iter, model):
+            batch = next(batch_iter)
+
+            sequences = batch["sequences"]
+            attention_mask = batch["attention_mask"].to(bool)
+            position_ids = batch["position_ids"]
+
+            seq_len = sequences.shape[1]
+
+            new_sequences, new_attention_mask, new_position_ids = remove_left_padding(
```
SumanthRH, line 90: Btw why is only left padding removed here? Is this related to a quirk for Mbridge models? We currently pad on both sides

skyrl-train/skyrl_train/workers/megatron/megatron_policy.py
```diff
@@ -0,0 +1,328 @@
...
+        policy_loss_fn: Optional[Callable] = None,
+    ):
+        self.cfg = config
+        self.hf_config = hf_config
+        self.tf_config = tf_config
+        self.actor_module = actor_module
+        self.actor_optimizer = actor_optimizer
+        self.policy_loss_fn = policy_loss_fn
+
+        # NOTE (erictang000): this is a potentially brittle way to disable the finalize_model_grads_func
```
SumanthRH: `finalize_model_grads_func` is this function : https://github.com/NVIDIA/Megatron-LM/blob/02a1dd02fc72307c1c6327c0667c9e7cecf2eaff/megatron/core/distributed/finalize_model_grads.py#L376

Doing a quick search, it is called at the end of the forward + backward op, ex:  https://github.com/NVIDIA/Megatron-LM/blob/02a1dd02fc72307c1c6327c0667c9e7cecf2eaff/megatron/core/pipeline_parallel/schedules.py#L640-L647 

Looking at the function, it's performing 
1. grad all reduce across dp replicas 
2. grad all reduce across SP ranks for layernorm layers 
3. grad all reduce for embedding layers for PP 
4. optionally for moe models run an `update_expert_token_bias` function
5. scale gradient if `num_tokens` is provided


1, 2, and 3 should be unaffected if we do one `finalize_` call at the end of the mini batch. As far as I understand, there should also not be any effect on memory usage here. 

For 4., with a quick glance the function seems to be compatible again with making this `finalize_` call at the end of the mini batch, but i haven't looked too closely. the `local_token_per_expert` attribute for the model is averaged across ranks. This value should acccumulate over micro batches, since I only see it being updated in the forward pass here: https://github.com/NVIDIA/Megatron-LM/blob/cca55ccdf408d56f97348f844a6aeb8f34d07672/megatron/core/transformer/moe/router.py#L448 , apart from the update on `finalize_`: https://github.com/NVIDIA/Megatron-LM/blob/cca55ccdf408d56f97348f844a6aeb8f34d07672/megatron/core/distributed/finalize_model_grads.py#L294-L295 . 

There are actually some subtle ways this change could have an affect, for example, this `apply_z_loss` function for the router depends on the : https://github.com/NVIDIA/Megatron-LM/blob/cca55ccdf408d56f97348f844a6aeb8f34d07672/megatron/core/transformer/moe/router.py#L333-L366 . This doesn't seem relevant to us AFAIK because we set `calculate_per_token_loss=False`. 

For 5., I again don't think it's relevant since we set `calculate_per_token_loss=False`


Btw, I see another comment on the utility of this `finalize_` call after each micro batch here: https://github.com/NVIDIA/Megatron-LM/blob/cca55ccdf408d56f97348f844a6aeb8f34d07672/megatron/core/models/huggingface/module.py#L28-L34, but I haven't given this much thought. 


Wdyt?

skyrl-train/tests/gpu/test_megatron_worker.py
```diff
@@ -0,0 +1,412 @@
...
+from skyrl_train.utils.utils import print_mem, validate_cfg
+from skyrl_train.entrypoints.main_base import config_dir
+from skyrl_train.distributed.dispatch import concatenate_outputs_after_mesh_dispatch
+from skyrl_train.utils.torch_utils import logprobs_from_logits
+from skyrl_train.training_batch import TrainingInputBatch
+
+
+# NOTE (erictang000): in the normal training flow, the model weights get downloaded in the HF cache by the inference engine
+# we should figure out a way to download the model weights in the test environment (for now trying to do this by
+# ordering the tests so the inference engine is initialized first). Also will be an issue for disaggregated training
```
SumanthRH: Could you explain the exact issue you faced here?

skyrl-train/tests/gpu/test_megatron_worker.py
SumanthRH, line 1: Q: Can we follow our convention to group tests by functionality as much as possible? ex: test_worker_offload.py can contain the `test_megatron_offload_memory_and_correctness` test? 

Differences in backends are also clearly visible then

SumanthRH (commented):
This is great! Leaving some comments after a first pass

skyrl-train/skyrl_train/workers/megatron/megatron_policy.py
erictang000: thanks for helping dig in more - maybe the conclusion from your last few points is that it might be better to try to refactor this to use the built in megatron `forward_backward_func` instead of fighting their abstraction to fit into our previous setup? Especially if there's places we are uncertain if it would affect correctness.

I'll try writing a version of the forward_backward_func that does the whole training step and make sure metrics match with the current implementation.

skyrl-train/skyrl_train/workers/megatron/megatron_policy.py
erictang000: yes this is needed for correctness for Megatron models (should be unrelated to the mbridge library). When I tested without it things were wrong

took the util function from verl where they do this: https://github.com/volcengine/verl/blob/e90f18c40aa639cd25092b78a5ff7e2d2508c088/verl/models/mcore/model_forward.py#L66

skyrl-train/tests/gpu/test_megatron_worker.py
erictang000: it's just nice to have them all in a single file for testing (also for dependencies later we might want to keep megatron separate due to the heavyweight installation deps). Megatron tests will also require more GPUs for all the permutations of TP + PP + DP + CP etc...

but we can reorganize - can I leave a comment to reorganize after the next few megatron PRs get in? just for my sanity during dev work...

skyrl-train/tests/gpu/test_megatron_worker.py
erictang000: for the megatron worker it just does `AutoConfig(Qwen/Qwen3-0.6B)` and inside the mbridge library while loading weights it assumes that the model weights are also there at like `/home/ray/.cache/huggingface/hub/models/Qwen/Qwen3-0.6B`. 

This is true if we are colocated and the inference engine is initialized first (since it'll download the weights and configs to the cache. But not true if we just run a worker test without inference engine init.

skyrl-train/skyrl_train/workers/megatron/megatron_policy.py
```diff
@@ -0,0 +1,328 @@
...
+        output = forward_backward_func(
+            forward_step_func=forward_step,
+            data_iterator=batch_generator,
+            model=self.actor_module,
+            num_microbatches=1,
+            seq_length=seq_len,  # no use when input_shapes was set
+            micro_batch_size=micro_batch_size,  # no use when input_shapes was set
+            forward_only=True,
+        )
+
```
SumanthRH: This seems incorrect, and from what I understand, we will end up with no pipelining for pipeline parallelism (no interleaving). `num_microbatches=1`

I actually assumed that the "microbatch" for pipelining is the same as the "micro batch" in our terminology. Can we modify the code to do that?


We currently have a synchronization point at the end of each `forward_backward_micro_batch` where we materialize loss tensors to CPU, so there's no overlap. For pipelining, it would probably be best to just feed a mini batch directly to megatron and use `num_microbatches=mini_batch_size_per_gpu/micro_batch_size_per_gpu` or so

skyrl-train/tests/gpu/test_megatron_worker.py
SumanthRH: sg, let's leave it as is

skyrl-train/tests/gpu/test_megatron_worker.py
SumanthRH: Hmm okay. 

Ideally we do this download ourselves then before init, it's pretty simple:
```
from huggingface_hub import snapshot_download
if rank == 0:
    snapshot_download(model_name) # will be no-op if already downloaded
torch.distributed.barrier()
```

skyrl-train/skyrl_train/workers/megatron/megatron_policy.py
SumanthRH: Makes sense to me, thanks!

skyrl-train/pyproject.toml
```diff
@@ -87,8 +93,8 @@ docs = [
     "sphinx-autobuild>=2021.3.14"
 ]
 vllm = [
-    "vllm==0.9.2",
-    "torch==2.7.0",
+    "vllm==0.10.0",
+    "torch==2.7.1",
```
SumanthRH: rebase on main?

skyrl-train/skyrl_train/distributed/megatron/megatron_utils.py
erictang000: interestingly using `.data` here seems to actually impact offloaded memory!

using `.data`:
image


not using `.data`:
image

maybe we can leave these as is?

skyrl-train/skyrl_train/config/megatron_config/policy.yaml
erictang000: good catch

skyrl-train/skyrl_train/workers/worker.py
erictang000: reverted

skyrl-train/skyrl_train/workers/megatron/megatron_policy.py
erictang000: resolved by above refactoring

skyrl-train/skyrl_train/workers/megatron/megatron_worker.py
erictang000: added

skyrl-train/tests/gpu/test_megatron_worker.py
erictang000: hmm does the rank==0 check work multi-node? this won't be an issue colocated but for disagg we might need to download to each machine? can leave a note to handle this later

SumanthRH: *local rank == 0

erictang000:
note: after refactoring I'm seeing slightly slower convergence with DP=4, and grad norm is almost 2x higher. Tried tuning all the ddp config knobs, but none seem to be relevant. DP=1 runs seem to converge slightly faster and also have lower grad norm. 

image

skyrl-train/tests/gpu/test_megatron_worker.py
erictang000: oh sick i forgot we had that

SumanthRH (approved):
Approving, let's disable Megatron DDP for now and fix in a future PR

PR #225, opened 2025-08-30
Bump vllm to 0.10.1.1

tyler-griggs:
Bumps vllm to 0.10.1.1, which depends on torch 2.7.1. Only required a couple small fixes.

Ran all GPU tests.

Also ran `run_gsm8k.sh`, green is vllm==0.10.1.1 & brown is vllm==0.9.2:

Screenshot 2025-08-29 at 8 42 24 PM


SumanthRH (approved):
Thanks

Issue #201, opened 2025-08-25
[Clarification request] Per-turn rewards in SKyRLGymGenerator agent_loop


alex-dr:
I'm interested in multi-turn environments with turn-level rewards in the SkyRLGymGenerator.

From my reading of the code, only the final reward from the trajectory is used: https://github.com/NovaSky-AI/SkyRL/blob/main/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py#L187 - the `reward` variable is updated with the most recent reward at each step, and the final value returned at the end of the method.

However, loss functions like GSPO explicitly support token-level advantages.

How am I supposed to give varying token-level rewards when only the final reward is ever returned, or is this not supported?

tyler-griggs:
Hey @alex-dr, a great point! The trainer itself supports token-level rewards ([link](https://github.com/NovaSky-AI/SkyRL/blob/main/skyrl-train/skyrl_train/trainer.py#L752)), and to test this code path I used a simple environment that maintained it's own token-level rewards. But, you're totally right, token-level rewards appear not to be supported by the SkyRLGymGenerator right now and we should update it. 

We'd like to improve support for this! Can you share a little more about your setup/environment and we can help make it happen? For example, would this be helpful: 

Update `BaseTextEnvStepOutput` to change the `reward` field from `float` to `Optional[float]`. On each `step()`, if the `reward` is non-`None`, we will append it to a list that racks each returned `reward` with which token id index it is associated with (ie, the final token id in the assistants response for that turn). Then, after the agent loop has completed, the `SkyRLGymGenerator` converts this to token-level rewards and passes to the trainer. 

As described, this is basically a response-level reward. Do you want to assign even finer-grained rewards from the environment (e.g., token-level)?

alex-dr:
The first setup you described sounds helpful. I do not want finer-grained reward assignment than that.  I'm more interested in things like giving penalties for improperly formatted tool calls / rewards for successful ones, and possibly some NL verification/LLM feedback along the way.

However, directly broadcasting the reward to the tokens may not always be the appropriate behavior. Note for example that in the [GRPO paper](https://arxiv.org/pdf/2402.03300) the process supervision version of the advantage calculation (4.1.3) tracks only a vector of rewards collected along the trajectory, and assigns per token advantage based on the sum of future rewards in the trajectory (after a group normalization). This may be different for other algorithms.

tyler-griggs:
[...]

PR #226, opened 2025-08-30
[Generator] Support turn-level rewards in SkyRLGymGenerator

tyler-griggs:
Addressing issue #201 

## What does this PR do?
Add support for turn-level rewards in SkyRLGymGenerator. The trainer already supports token-level rewards, but previously the SkyRLGymGenerator ignored rewards returned by `step()` except for the final turn. This PR tracks per-`step` rewards and builds a token-level reward list that associates each `step()` reward with the final token in the assistant's response that led to the reward. 

## Tests
Added CPU tests for main functionality and edge cases (e.g., response truncation due to exceeding max length)

e2e test with step-level rewards: **WIP**


skyrl-train/examples/turn_level_rewards/README.md
SumanthRH, line 1: to be filled?

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
SumanthRH, line 1: Should the changes be in the same generator or do we just have a custom generator example for step level rewards? Latter seemed like it's easier to maintain (this is also a niche example at the moment)

tyler-griggs: Hm, I'm partial to keeping it as one generator. I actually think the custom generator example will be harder to maintain. At least, I have a hunch it will quickly become far out of date as we iterate on the skyrl gym generator. I don't actually think this is a very intrusive change, just one list with post-processing. We are currently ignoring most of the returns of `step()`, which we probably shouldn't be doing anyways. 

That said, I do feel that the SkyRL Gym Generator is getting (too) complex, but I think the way to do it is by pruning some of the code paths like `retokenize_chat_history`.

skyrl-train/examples/turn_level_rewards/README.md
tyler-griggs: Yeah it's a WIP

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
SumanthRH: >  I actually think the custom generator example will be harder to maintain. At least, I have a hunch it will quickly become far out of date as we iterate on the skyrl gym generator.

How so? We actually don't have to support all the features we will add to skyrl gym generator in the future. Like step level rewards doesn't have to support `use_conversation_multi_turn`

 I guess we might change the stuff around chat templating and tokenization soon, but a standalone generator should still work okay?
 
 > I do feel that the SkyRL Gym Generator is getting (too) complex, but I think the way to do it is by pruning some of the code paths like retokenize_chat_history.
 
 Hmm yeah this is my main concern: SkyRLGymGenerator is **the** example for how a generator should look like with SkyRL, and it's getting pretty involved (ex: the `use_conversation_multi_turn` codepaths, etc)
 
 
 I'm not too strongly attached to either btw, up to you

tyler-griggs: > I guess we might change the stuff around chat templating and tokenization soon, but a standalone generator should still work okay?
Yeah, this was one pending set of changes I was referring to. And I do agree that the standalone generator would probably work, it just wouldn't be all that useful if it falls out of date with what's supported in the SkyRLGymGenerator. 

Ultimately, the reason I lean towards keeping this in is a combination of the change being relatively small, and I do think we should be making use of `step()` outputs instead of ignoring them. 

Charlie and I spoke recently about ways to simplify the SkyRL Gym Generator and prune some of the bloat, and I think we can shed it that way.

tyler-griggs:
Okay now ready for review! I'll add a wandb screenshot shortly.

skyrl-train/examples/turn_level_rewards/gsm8k_multi_turn_dataset.py
```diff
@@ -0,0 +1,78 @@
...
+    args.output_dir = os.path.expanduser(args.output_dir)
+
+    data_source = "openai/gsm8k"
+
+    dataset = datasets.load_dataset(data_source, "main")
+
+    train_dataset = dataset["train"]
+    val_dataset = dataset["test"]
+
+    instruction_following = 'Let\'s think step by step and output the final answer after "####".'
```
SumanthRH: we can assume max_turns is > 1 for this example and thus the prompt can be :

```suggestion
    instruction_following = 'Let\'s think step by step and output a tentative numeric answer after "####".'
```

tyler-griggs: Good call, thanks

PR #229, opened 2025-08-30
Add config validation for train batch size vs dp size

tyler-griggs:
## What does this PR do?
If `dp_size > train_batch_size`, the trainer will call `_remove_tail_data()` and truncate the training batch to size 0, which won't throw an error until it enters the generator. This PR adds a check in `validate_cfg` that `asserts` the train batch size is equal to or larger than the data parallel size.

## Testing
Added a CPU test for this.


CharlieFRuan:
I'll review today

PR #230, opened 2025-08-31
[Generator][HTTP] Add OpenAI API inference HTTP server for generator, support /chat/completions

CharlieFRuan:
### Motivation and changes

Resolves issue https://github.com/NovaSky-AI/SkyRL/issues/96

Almost identical to the previous PR that was merged to a separate branch `http-inf_engine-openai` https://github.com/NovaSky-AI/SkyRL/pull/101

This PR adds a FastAPI server wrapper around `InferenceEngineClient` such that users can perform rollout by posting `/chat/completions` request to the server's URL, rather than using `InferenceEngineClient.generate()`. This is helpful for users that want to train existing agentic frameworks. They can substitute the existing URL to SkyRL's URL.

To use this feature, simply set the following in the config:
```yaml
generator
  enable_http_endpoint: true
  http_endpoint_host: "127.0.0.1"
  http_endpoint_port: 8000
```

Then you can post requests to `http://127.0.0.1/800/v1/chat/completions`.

Please refer to `examples/inference_http_endpoint` for example usage.

Note that this PR only supports `/chat/completions`, which inherently cannot ensure token-in-token-out and re-tokenizes the chat history every turn and at the very end. We will add `/completions` as a follow up PR to allow token-in-token-out. We also do not support `tools` as of now, since the parsing of tools may lose the original model's output string. Users are responsible for parsing their own tool calls from the model's response.

### Tests
- Ensured all GPU CI tests passed locally
- Added inference http server GPU tests, and OAI protocol CPU tests
- Compared `run_gsm8k.sh` (uses `SkyRLGymGenerator`) with `run_gsm8k_with_inference_http_endpoint.sh` (uses `SkyRLGymHTTPGenerator`, only for demo purpose)

Reward:
image

Timing:
image

For `generate()`'s timing, there is a 20-30% slowdown, though gsm8k is a non-agentic workload. Will further profile as a follow up.

### Future TODOs
- https://github.com/NovaSky-AI/SkyRL/issues/231


skyrl-train/uv.lock
```diff
@@ -1384,13 +1384,13 @@ name = "mlx-lm"
...
-    { name = "transformers", version = "4.56.0", source = { registry = "https://pypi.org/simple" }, marker = "(platform_machine != 'aarch64' and extra == 'extra-11-skyrl-train-flashrl') or (platform_machine != 'aarch64' and extra == 'extra-11-skyrl-train-vllm') or (platform_python_implementation != 'CPython' and extra == 'extra-11-skyrl-train-flashrl') or (platform_python_implementation != 'CPython' and extra == 'extra-11-skyrl-train-vllm') or (sys_platform != 'linux' and extra == 'extra-11-skyrl-train-flashrl') or (sys_platform != 'linux' and extra == 'extra-11-skyrl-train-vllm') or (extra == 'extra-11-skyrl-train-flashrl' and extra == 'extra-11-skyrl-train-sglang') or (extra == 'extra-11-skyrl-train-flashrl' and extra == 'extra-11-skyrl-train-vllm') or (extra == 'extra-11-skyrl-train-sglang' and extra == 'extra-11-skyrl-train-vllm')" },
+    { name = "jinja2", marker = "sys_platform != 'linux'" },
```
CharlieFRuan, line 1460: are all these changes expected? I only added `fastapi` and `uvicorn` to `dev` @gemini-code-assist

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
```diff
@@ -16,11 +18,95 @@ class InferenceEngineClient(InferenceEngineInterface):
...
+
+                shutdown_server(
+                    host=self.inference_http_server_host,
+                    port=self.inference_http_server_port,
+                    max_wait_seconds=10,
+                )
+                if hasattr(self, "_server_thread") and self._server_thread.is_alive():
+                    self._server_thread.join(timeout=10)
+            except Exception as e:
+                print(f"Error shutting down HTTP server: {e}")
```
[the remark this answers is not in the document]
CharlieFRuan: Yes, agreed. Will be addressed when the TODO is addressed in a future PR.

skyrl-train/uv.lock
CharlieFRuan: @gemini-code-assist

skyrl-train/examples/inference_http_server/rollout_with_http_server_litellm.py
tyler-griggs, line 1: Should we move this under `tests/gpu`?

skyrl-train/examples/inference_http_server/skyrl_gym_inference_http_generator.py
```diff
@@ -0,0 +1,167 @@
...
+            "loss_masks": loss_masks,
+            "stop_reasons": stop_reasons,
+            "rollout_metrics": rollout_metrics,
+            "rollout_logprobs": truncated_logprobs,
+        }
+
+        return generator_output
+
+
+async def _generate_with_http_server(
```
tyler-griggs: Should we simplify this and just also call litellm? My hunch is that most people who want to call into an HTTP endpoint will be using litellm or something similar (rather than aiohttp), so it would probably be a more realistic example to reference.

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
```diff
@@ -16,11 +18,95 @@ class InferenceEngineClient(InferenceEngineInterface):
     Note that InferenceEngineClient sub-classes InferenceEngineInterface so it can be used as if talking to a single engine.
     """
 
-    def __init__(self, engines: List[InferenceEngineInterface], tokenizer: PreTrainedTokenizerBase):
+    def __init__(self, engines: List[InferenceEngineInterface], tokenizer: PreTrainedTokenizerBase, config: DictConfig):
```
tyler-griggs: Can you update this to `training_cfg`? Sometimes just the `generator` cfg is passed in places, so I've been naming that `generator_cfg`.

tyler-griggs: It just helps to distinguish what config is being passed around

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
```diff
@@ -16,11 +18,95 @@ class InferenceEngineClient(InferenceEngineInterface):
...
+        self.backend = config.generator.backend
+        self.use_inference_http_server = config.generator.use_inference_http_server
+        self.inference_http_server_host = config.generator.inference_http_server_host
+        self.inference_http_server_port = config.generator.inference_http_server_port
+        if self.use_inference_http_server:
+            self._spin_up_http_server()
+
         print(f"InferenceEngineClient initialized with {len(engines)} engines.")
 
+    # ----------------------------
```
tyler-griggs: nit: can we move these to the bottom of the file? I'd prefer a user first sees the methods in the `InferenceEngine` interface if they scroll through here.

skyrl-train/skyrl_train/inference_engines/openai_api_protocol.py
```diff
@@ -0,0 +1,252 @@
...
+
+    result = {}
+    if request.response_format and request.response_format.type == "json_schema":
+        result["json_schema"] = _convert_json_schema_to_str(request.response_format.json_schema.json_schema)
+    elif request.response_format and request.response_format.type == "json_object":
+        result["json_schema"] = '{"type": "object"}'
+    return result
+
+
+# TODO(Charlie): consolidate sampling params building logics across the repo.
```
tyler-griggs, line 204: +1 :D

tyler-griggs: Actually would be good to add this to an Issue, maybe in the TODO list you already have?

skyrl-train/skyrl_train/utils/utils.py
```diff
@@ -294,6 +294,11 @@ def validate_cfg(cfg: DictConfig):
                 "to `True` to append tokenizer.eos_token_id to the assistant-generated response to match the chat template."
             )
 
+    if cfg.generator.use_inference_http_server and not cfg.generator.async_engine:
+        raise ValueError(
+            "In your config, generator.async_engine must be True when using http server inference engine client."
```
tyler-griggs: nit: can probably remove "In your config". I like to just specify the fields exactly, like:

generator.async_engine must be True when `generator.use_inference_http_server==True`.

skyrl-train/examples/inference_http_server/rollout_with_http_server_litellm.py
tyler-griggs: I actually feel like we should fast follow this PR with the PR on custom generator (e.g., mini-swe-agent integration) and use that as the key example of using the http server. This file is nice, but it looks too much like a test to give a good understanding of how you'd build a custom generator around it.

tyler-griggs: cc: @SumanthRH because he's working on mini-swe-agent

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -175,6 +175,9 @@ generator:
   gpu_memory_utilization: 0.8
   max_num_seqs: 1024
   remote_inference_engine_urls: ["127.0.0.1:8001"]
+  use_inference_http_server: false
```
tyler-griggs: nit: What do you think of:

`enable_http_endpoint`
`http_endpoint_host`
`http_endpoint_port`

Somehow this feels slightly more canonical than `inference_http_server`. Also I would like to make a pass over the config and do some re-organization (which sadly will include a couple breaking changes). I think we should have an inference engine client sub-config (and, generally, more sub-configs for better organization), and these flags will move to that sub-config. In that case, it probably isn't necessary to specify "inference".

skyrl-train/examples/inference_http_server/rollout_with_http_server_litellm.py
CharlieFRuan: > I actually feel like we should fast follow this PR with the PR on custom generator (e.g., mini-swe-agent integration) and use that as the key example of using the http server. This file is nice, but it looks too much like a test to give a good understanding of how you'd build a custom generator around it.

Yes agreed. I can move this to test

skyrl-train/skyrl_train/inference_engines/openai_api_protocol.py
CharlieFRuan: > maybe in the TODO list you already have

Yep it is, tracked in https://github.com/NovaSky-AI/SkyRL/issues/231

skyrl-train/skyrl_train/inference_engines/launch_inference_http_server.py
tyler-griggs, line 1: nit: can we change this file name since it does more than `launch`?

CharlieFRuan: fixed.

skyrl-train/skyrl_train/config/ppo_base_config.yaml
CharlieFRuan: Yep, removing `inference` makes sense. There doesn't seem to be much difference between `endpoint` and `server`. Will go with `endpoint` to distinguish with say `vllm_server.py`

skyrl-train/skyrl_train/utils/utils.py
CharlieFRuan: fixed.

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
CharlieFRuan: Good point, fixed.

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
CharlieFRuan: I'll do `full_config` perhaps? maybe `training_config` could be confused with `config.trainer`

skyrl-train/skyrl_train/trainer.py
```diff
@@ -117,7 +117,8 @@ def build_dataloader(self, dataset: PromptDataset, is_train=True):
             batch_size=batch_size,
             shuffle=True if is_train else False,
             collate_fn=dataset.collate_fn,
-            num_workers=8,
+            # TODO(Charlie): debug why inference http endpoint is slow when num_workers is 8
+            num_workers=0 if self.cfg.generator.enable_http_endpoint else 8,
```
CharlieFRuan, line 121: for some reason the first `enumerate()` would take 30 seconds for gsm8k if `num_workers` is 8. could be due to how there is an extra thread spinning when we launch the http endpoint. Will revisit

skyrl-train/examples/inference_http_server/rollout_with_http_server_litellm.py
CharlieFRuan: > Should we move this under tests/gpu?

Done.

skyrl-train/examples/inference_http_server/skyrl_gym_inference_http_generator.py
CharlieFRuan: As mentioned offline, LiteLLM's max connection of 100 is hard to bypass. Will keep this example as-is; will likely be replaced by another agentic example soon anyway.

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
```diff
@@ -16,9 +18,25 @@ class InferenceEngineClient(InferenceEngineInterface):
...
 
-    def __init__(self, engines: List[InferenceEngineInterface], tokenizer: PreTrainedTokenizerBase):
+    def __init__(
+        self, engines: List[InferenceEngineInterface], tokenizer: PreTrainedTokenizerBase, full_config: DictConfig
+    ):
+        """
+        Args:
+            engines: List[InferenceEngineInterface] - The inference engines, remote or local.
+            tokenizer: PreTrainedTokenizerBase - The tokenizer to use.
+            config: DictConfig - See ppo_base_config.yaml
```
tyler-griggs: nit: config --> full_config

skyrl-train/skyrl_train/inference_engines/inference_http_endpoint.py
tyler-griggs, line 1: nit: can we name this file something a little more specific? The word `client` should probably be in the name since this is not a generic inference HTTP endpoint, but rather an endpoint sitting in front of the inference engine `client`.

tyler-griggs: E.g., `client_http_endpoint.py`

skyrl-train/tests/gpu/gpu_ci/test_inference_http_endpoint.py
tyler-griggs, line 1: nit: if you update the name of `inference_http_endpoint.py`, please update this as well.

tyler-griggs (approved):
Just a few nits, otherwise LGTM

skyrl-train/skyrl_train/inference_engines/inference_http_endpoint.py
CharlieFRuan: I think `client_http_endpoint.py` kind of has no content. I'll name the file to `inference_engine_client_http_endpoint` (though is a mouthful) and refer to it as "inference http endpoint" in docs/comments

PR #238, opened 2025-09-03
[HTTP][Generator] Let vllm python engine handle OAI request, remove openai_api_protocol

CharlieFRuan:
### Motivation

Before this PR, we wrote `openai_api_protocol.py` that re-defines OpenAI's `ChatCompletionRequest`, etc. and do conversino between OpenAI API and our own InferenceEngineInput/Output. This is unnecessary and makes the `InferenceEngineInput/Output` a narrow waste.

### What this PR does

This PR removes the perviously self-defined `openai_api_protocol.py` by enabling vLLM's Python engine to directly handle OpenAI request. This way, we support whatever field vLLM supports without converting the OpenAI API request/response.

So far we only handled vLLM's Python engine but not SGLang's and leave that as a TODO.

- `vllm_server.py`: supported by simply rerouting the request
- `vllm_engine.py`: supported by enabling Python vllm engine to handle OAI request (reuse some vLLM's internal code, referred to verl's `vllm_async_server.py`)
- `sglang_server.py`: not supported yet due to we do `skip_tokenizer_init=True` by default. Can be fixed by enabling token-in-token-out with SGLang using return logprob.
- `sglang_engine.py`: not supported yet.

### Tested:
- `test_inference_engine_client_http_endpoint.py`: error handling, with local `vllm_engine.py`, with remote `vllm_server.py`
- `./examples/inference_http_endpoint/run_gsm8k_with_inference_http_endpoint.sh`

### Wandb run compared to pr230:
Reward and Eval:
image

Timing (1.15x speedup for generation compared to pr230, but there is still a non-trivial slowdown compared to non-http. though this is on gsm8k, and for agentic the slowdown should be less significant):
image

### Future TODO:
- [ ] Add error handling test for `vllm_server.py` with HTTP endpoint
- [ ] Support `/completion`
- [ ] Support SGLang backend with HTTP endpoint
- [ ] Toggle vLLM server log easily
- [ ] Consider removing `xxx_engine.py` and always use `xxx_server.py`
- [ ] Add a doc on `enable_http_endpoint`

### Reference
- https://github.com/volcengine/verl/blob/e90f18c40aa639cd25092b78a5ff7e2d2508c088/verl/workers/rollout/vllm_rollout/vllm_async_server.py#L327


skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -204,6 +205,7 @@ generator:
   # sampling params for evaluation
   eval_sampling_params:
     max_generate_length: ${generator.sampling_params.max_generate_length}
+    repetition_penalty: 1.0
```
CharlieFRuan, line 213: otherwise we would use model's default generation config due to vLLM's logic, which is 1.1 for Qwen2.5. We should consolidate the various sources of sampling params soon.

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
```diff
@@ -311,7 +316,27 @@ class AsyncVLLMInferenceEngine(BaseVLLMInferenceEngine):
     def _create_engine(self, *args, **kwargs):
         # TODO (erictang000): potentially enable log requests for a debugging mode
         engine_args = vllm.AsyncEngineArgs(enable_log_requests=False, **kwargs)
-        return vllm.AsyncLLMEngine.from_engine_args(engine_args)
+        engine = vllm.AsyncLLMEngine.from_engine_args(engine_args)
+
+        # Adapted from https://github.com/volcengine/verl/blob/e90f18c40aa639cd25092b78a5ff7e2d2508c088/verl/workers/rollout/vllm_rollout/vllm_async_server.py#L327
+        model_config = engine.model_config
+        model_path = kwargs.get("model")
+        model_name = "/".join(model_path.split("/")[-2:])
```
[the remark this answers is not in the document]
CharlieFRuan: `kwargs` should always have `model`. This code is internal

skyrl-train/skyrl_train/inference_engines/inference_engine_client_http_endpoint.py
```diff
@@ -55,70 +57,81 @@ def set_global_state(inference_engine_client: "InferenceEngineClient", uvicorn_s
...
-        logger.error(f"Error in chat completion: {str(e)}\n{traceback.format_exc()}")
-        raise e
+        error_response = ErrorResponse(
+            error=ErrorInfo(
+                message=f"Error in chat completion: {str(e)}",
+                type=HTTPStatus.INTERNAL_SERVER_ERROR.phrase,
+                code=HTTPStatus.INTERNAL_SERVER_ERROR.value,
+            ),
+        )
+        return JSONResponse(content=error_response.model_dump(), status_code=HTTPStatus.INTERNAL_SERVER_ERROR.value)
```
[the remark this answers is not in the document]
CharlieFRuan: Resolved.

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
```diff
@@ -73,6 +73,11 @@ async def generate(self, input_batch: InferenceEngineInput) -> InferenceEngineOu
             # Split evenly across engines
             return await self._generate_batched(prompt_token_ids, sampling_params)
 
+    async def chat_completion(self, request_payload: Dict[str, Any]) -> Dict[str, Any]:
+        trajectory_id = request_payload["json"].pop("trajectory_id", 0)
```
tyler-griggs: If `trajectory_id` is not provided, won't all requests land on the same inference engine if we default the value to `0`? I think we should provide simple load balancing at a minimum (e.g., random, round robin).

skyrl-train/skyrl_train/inference_engines/inference_engine_client_http_endpoint.py
```diff
@@ -202,9 +226,25 @@ async def lifespan(app: fastapi.FastAPI):
...
-        return await handle_chat_completion(request, raw_request)
+    @app.post("/v1/chat/completions")
+    async def chat_completion(raw_request: Request):
+        """
+        Takes in OpenAI's `ChatCompletionRequest` and returns OpenAI's `ChatCompletionResponse`.
+
+        Note that the specific fields inside the request and response depend on the backend you use.
+        If `config.generator.backend` is `vllm`, then the request and response will be vLLM's.
+        Same for SGLang. SkyRL does not do any field checkings besides `model` and `trajectory_id`
+        but instead offload everything to the underlying engines.
```
tyler-griggs: nit: slightly awkward language :D maybe:

```
SkyRL does not perform field validation beyond `model` and `trajectory_id`, and otherwise depends on the underlying engines' validation.
```

skyrl-train/skyrl_train/inference_engines/inference_engine_client_http_endpoint.py
```diff
@@ -202,9 +226,25 @@ async def lifespan(app: fastapi.FastAPI):
...
+
+        Make sure you add in `trajectory_id` to ensure load balancing and sticky routing. The same
```
tyler-griggs, line 241: Related to the above: could we provide at least very naive load balancing without `trajectory_id`?

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
```diff
@@ -311,7 +320,27 @@ class AsyncVLLMInferenceEngine(BaseVLLMInferenceEngine):
...
+            model_config,
+            models,
+            "assistant",
+            request_logger=RequestLogger(max_log_len=4096),
+            chat_template=None,
+            chat_template_content_format="auto",
+            # enable_auto_tools=config.multi_turn.tool_config_path is not None,
+            # tool_parser=config.multi_turn.format,  # hermes, llama3_json, ...
+        )
+        return engine
```
tyler-griggs:
```suggestion
        base_model_paths = [BaseModelPath(name=model_name, model_path=model_path)]
        models = OpenAIServingModels(engine, model_config, base_model_paths)
        self.openai_serving_chat = OpenAIServingChat(
            engine_client=engine,
            model_config=model_config,
            models=models,
            response_role="assistant",
            chat_template=None,
            chat_template_content_format="auto",
        )
        return engine
```

tyler-griggs: Just cleaning up a bit!

For `request_logger`, I have found this to be extremely noisy and rapidly grows the logs, so I removed it.

For `enable_auto_tools` and `tool_parser`, any reason to keep those around as comments?

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
```diff
@@ -311,7 +320,27 @@ class AsyncVLLMInferenceEngine(BaseVLLMInferenceEngine):
     def _create_engine(self, *args, **kwargs):
         # TODO (erictang000): potentially enable log requests for a debugging mode
         engine_args = vllm.AsyncEngineArgs(enable_log_requests=False, **kwargs)
-        return vllm.AsyncLLMEngine.from_engine_args(engine_args)
+        engine = vllm.AsyncLLMEngine.from_engine_args(engine_args)
+
+        # Adapted from https://github.com/volcengine/verl/blob/e90f18c40aa639cd25092b78a5ff7e2d2508c088/verl/workers/rollout/vllm_rollout/vllm_async_server.py#L327
+        model_config = engine.model_config
+        model_path = kwargs.get("model")
+        model_name = "/".join(model_path.split("/")[-2:])
```
tyler-griggs: Is this... stable? :D I know verl uses this but oof.

tyler-griggs (commented):
Looking great to me!

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
CharlieFRuan: > For request_logger, I have found this to be extremely noisy and rapidly grows the logs, so I removed it.

Oh yea this was related to my TODO

>  Toggle vLLM server log easily

Makes sense by default we might not want this. But this was crucial for helping me realize we are using repetition penalty 1.1. We can add some config in the future. For now I'll comment it out and add a note if you're fine with it

> For enable_auto_tools and tool_parser, any reason to keep those around as comments?

If we want to support OAI-style tool calling in future these fields are needed, hence I wanted to keep them around. I can add a note

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
CharlieFRuan: one workaround is to ask users provide both model path and their own model name, kind of similar to vLLM's `served_model_name`: https://github.com/vllm-project/vllm/issues/13257

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
SumanthRH: +1

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
CharlieFRuan: I just realized this code is currently erroneous. It will only work for `HFUser/ModelName`. Rationale as follows:

The only source of model name/path in SkyRL is the `trainer.policy.model.path`. This can be say `/my/path/to/Qwen2.5-1.5B-Instruct` or `Qwen/Qwen2.5-1.5B-Instruct`.

We require users, in their ChatCompletionRequest, to have the `model` field to be exactly `trainer.policy.model.path`, as I check here:
https://github.com/NovaSky-AI/SkyRL/blob/dcea0286e16a0218afef11438cfdc0cdff66aa06/skyrl-train/skyrl_train/inference_engines/inference_engine_client_http_endpoint.py#L100

That is the current behavior, both before and after this PR.

In addition, the `model_name` field you highlighted, will be checked against `request.model` by vLLM.

Therefore, `model_name` should simply be `model_path` (i.e. `trainer.policy.model.path`).

If we want to have a model name on top of `trainer.policy.model.path`, we need to add a config like the vllm issue above.

For this PR maybe I'll just do `model_path = model_name`?

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
CharlieFRuan: yea makes sense. I am routing to a random engine in the latest commit. Also as a side note, without trajectory_id, user will be bombarded with warning in `_validate_chat_completion()`

skyrl-train/skyrl_train/inference_engines/inference_engine_client_http_endpoint.py
CharlieFRuan: Fixed, thanks!

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
tyler-griggs: Re: logging, Got it, sounds good. Yeah we should probably provide more optionality in enabling/disabling debug logs but makes sense to keep them off by default. 

Re: tools, can we at least delete the commented args, and just move the TODO above the `OpenAIServingChat(` call to say something like `# TODO(Charlie): revisit "enable_auto_tools" and "tool_parser" when we need to support OAI-style tool calling`. I know this is nit-picky, I just don't love that we have commented-out assignment to config paths that don't exist in our config.

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
tyler-griggs: Good catch, I think `model_path = model_name` makes sense. To check my understanding, this basically means that if `trainer.policy.model.path == /my/path/to/Qwen2.5-1.5B-Instruct`, then `/my/path/to/Qwen2.5-1.5B-Instruct` should be used everywhere (in vLLM model name initialization, in `litellm` calls by any agent, etc.), right? And similarly for `trainer.policy.model.path == Qwen/Qwen2.5-1.5B-Instruct`.

I do think that makes it less likely for any surprises to come up with different naming.

skyrl-train/tests/gpu/gpu_ci/test_inference_engine_client_http_endpoint.py
```diff
@@ -207,82 +242,88 @@ async def generate_output(prompt):
...
+        # TODO(Charlie): add TP > 1 tests for sglang when we support it
+        # TODO(Charlie): sglang remote server not supported for /chat/completion
+        # yet because we have skip_tokenizer_init=True. Fix by getting tokens
+        # via return logprobs instead.
+        # pytest.param("sglang", 1, marks=pytest.mark.sglang),
     ],
+    # ids=["vllm", "sglang"],
+    ids=["vllm"],
 )
-def test_full_build_sampling_params(backend: str):
```
tyler-griggs, line 249: qq: not sure how important this test was, but any reason to get rid of it?

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
CharlieFRuan: >  To check my understanding, this basically means that if trainer.policy.model.path == /my/path/to/Qwen2.5-1.5B-Instruct, then /my/path/to/Qwen2.5-1.5B-Instruct should be used everywhere (in vLLM model name initialization, in litellm calls by any agent, etc.), right? And similarly for trainer.policy.model.path == Qwen/Qwen2.5-1.5B-Instruct.

Yes, correct.

> I do think that makes it less likely for any surprises to come up with different naming.

Sounds good, we can only do `model_path` everywhere for now

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
CharlieFRuan: Yes completely makes sense, should be addressed in latest commit

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
tyler-griggs: Okay cool, thanks! And long term, if we have a method of "sanitizing" the model name in a way we feel is consistent, we could update this.

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
tyler-griggs: Thank you!

skyrl-train/tests/gpu/gpu_ci/test_inference_engine_client_http_endpoint.py
CharlieFRuan: oh yea this was to test a util `build_sampling_params()` in the previous `openai_api_protocol.py`, which converts from our re-defined OAI request to vllm/sglang sampling params. No longer relevant so removed

tyler-griggs: Got it, thanks

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
CharlieFRuan: addressed in https://github.com/NovaSky-AI/SkyRL/pull/970

PR #239, opened 2025-09-03
1/N Terminal Bench Integration

tyler-griggs:
This PR is a WIP implementation of the Terminal Bench integration. We are using the `sandboxes` repo instead of `terminal-bench` because `sandboxes` is the new and improved version. The primary limitation of the current integration is that we have hardcoded training to only use one TBench task (`hello-world`). As `sandboxes` migrates more tasks from TBench, we will add support for specifying the training set of tbench tasks to use.


skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -154,6 +155,7 @@ trainer:
 
 
 generator:
+  model_name: ${trainer.policy.model.path}
```
CharlieFRuan, line 161: ~~this is not needed anymore, since I pass the full config to inference_engine_client.py~~
https://github.com/NovaSky-AI/SkyRL/blob/ff1c8ba44d1b580bfd13568e04f404901773706e/skyrl-train/skyrl_train/inference_engines/inference_engine_client.py#L32

skyrl-train/examples/terminal_bench/terminal_bench_generator.py
```diff
@@ -0,0 +1,171 @@
...
+        tasks = []
+        for i in range(len(prompts)):
+            tasks.append(
+                self.terminal_bench_agent_loop(
+                    agent_num=i,
+                    prompt="",
+                )
+            )
+
+        all_outputs = await asyncio.gather(*tasks)
```
CharlieFRuan: minor comment, perhaps we can extract this into utils in future too, essentially standardizing AgentLoopOutput, and its conversion to GeneratorOutput. A solid cut of 20 lines of code

skyrl-train/examples/terminal_bench/terminal_bench_generator.py
```diff
@@ -0,0 +1,171 @@
...
+
+        # Process response messages (everything after the first message)
+        response_messages = chat_history[1:]
+
+        response_ids = []
+        loss_mask = []
+
+        for message in response_messages:
+            # Apply chat template and tokenize each message
+            msg_encoding = self.tokenizer.apply_chat_template([message], add_generation_prompt=False, tokenize=True)
```
CharlieFRuan: will this repeatedly apply system prompt? perhaps we need a fixed-base method similar to `skyrl_gym_generator.py`?
https://github.com/NovaSky-AI/SkyRL/blob/ff1c8ba44d1b580bfd13568e04f404901773706e/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py#L579-L585

Besides, currently it seems that tokens like `<|user|>` is not masked (i.e. generation prompt should be masked).

Overall the encoding does not seem elegant, we should fix (either in this PR or in followup PR)

skyrl-train/skyrl_train/config/ppo_base_config.yaml
CharlieFRuan: As discussed offline, adding this here may make things cleaner. TBenchGenerator needs the model name in generator_config

skyrl-train/examples/terminal_bench/terminal_bench_generator.py
tyler-griggs: Yeah interesting. I could image the agent outputs could be diverse, but I suppose it does basically boil down to the fields needed to determine `GeneratorOutput`.

skyrl-train/examples/terminal_bench/terminal_bench_generator.py
tyler-griggs: Yes agreed, this is not good. I believe this was a workaround for the issue that Etash observed regarding assistant masks not working for Qwen2.5. I have started addressing this in a follow-up PR, will send out shortly

CharlieFRuan: > assistant masks not working for Qwen2.5

I think that is because the official masks do not have the generation tag:
https://github.com/NovaSky-AI/SkyRL/blob/ea9794370487ee78d0d0d4eb94a2ec6f25595d28/skyrl-train/skyrl_train/generators/utils.py#L15

> I have started addressing this in a follow-up PR, will send out shortly

Great!

PR #244, opened 2025-09-06
[Generator][HTTP] Support /completion endpoint for token-in-token-out

CharlieFRuan:


CharlieFRuan:
move to https://github.com/NovaSky-AI/SkyRL/pull/260

PR #248, opened 2025-09-07
Cloud Checkpointing

tyler-griggs:
[Note: primarily authored by @bthecohen, and borrowed code from @LokmaSpeedy, I mostly just tested :)]

## What does this PR do?
Enables cloud checkpointing (e.g., to S3, GCP). This PR wraps most local disk operations with utils that support cloud storage via `fsspec`. These utilities allow you to treat files as local, and transparently reads/write cloud files when enabled.

Now, an S3 or GCP file path can be passed as the checkpointing path in the training config, such as:
```
trainer.ckpt_path="s3://path/to/ckpts/"
```

This PR also fixes the broken `save_hf_model` for `deepspeed_strategy` which did not work in multi-node settings. The `save_hf_model` implementation was updated to gather model shards on rank 0 (similar to FSDP implementation)

## Testing
* Updated GPU tests, all passing.
* Manually tested save/load checkpoints and save/load HF weights to S3 with deepspeed and fsdp2.

### DIFF
--- a/skyrl-train/docs/checkpointing-logging/checkpointing.rst
+++ b/skyrl-train/docs/checkpointing-logging/checkpointing.rst
@@ -3,6 +3,8 @@ Checkpointing
 
 SkyRL provides checkpointing features to resume training from a previous state. Training state is saved at regular intervals and provides flexible configuration options for checkpoint management.
 
+Checkpointed state can be stored in the local file system or uploaded to cloud storage (S3, GCP).
+
 What State is Saved
 -------------------
 
@@ -88,6 +90,9 @@ Checkpointing behavior is controlled by several parameters in the YAML configura
 ``ckpt_path``
   - **Default**: ``"${oc.env:HOME}/ckpts/"``
   - **Purpose**: Base directory where all checkpoints are stored
+  - **Options**:
+    - Local file system path (e.g., ``/path/to/ckpts/``)
+    - Cloud storage path (S3, GCP) (e.g., ``s3://path/to/ckpts/``)
 
 **Checkpoint Cleanup**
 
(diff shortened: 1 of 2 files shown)


skyrl-train/skyrl_train/utils/io.py
```diff
@@ -0,0 +1,153 @@
...
+def isdir(path: str) -> bool:
+    """Check if path is a directory."""
+    fs = _get_filesystem(path)
+    return fs.isdir(path)
+
+
+def list_dir(path: str) -> list[str]:
+    """List contents of a directory."""
+    fs = _get_filesystem(path)
+    return fs.ls(path)
```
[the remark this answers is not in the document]
tyler-griggs: Hm fair point, the docs for fsspec do say it returns a dict, but it seems to be returning lists just fine. Added a CPU test to cover this just in case.

skyrl-train/tests/cpu/utils/test_io.py
```diff
@@ -0,0 +1,419 @@
...
+import os
+import tempfile
+import pytest
+from unittest.mock import patch, Mock
+import torch
+
+# Import our io module
+import sys
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "../../.."))
```
SumanthRH: Nit; is this needed?

tyler-griggs: It is not -- thanks!

tyler-griggs:
Thanks for the review! Added notes to the configuration and checkpointing docs (added to the top because seems very relevant)

SumanthRH (approved):
Great

PR #250, opened 2025-09-07
[fix] Bring back pretty log formatting

tyler-griggs:
Brings back nicely-formatted logs (ie, coloring, bold, etc.) that was lost when the training loop moved off the head node.

Old:
Screenshot 2025-09-07 at 12 52 00 PM

New: 
Screenshot 2025-09-07 at 4 30 04 PM

### DIFF
--- a/skyrl-train/skyrl_train/utils/utils.py
+++ b/skyrl-train/skyrl_train/utils/utils.py
@@ -500,8 +500,8 @@ def configure_worker_logging() -> None:
 
     logging.root.handlers = [_InterceptHandler()]
     level_name = os.getenv("LOG_LEVEL", "INFO").upper()
-    level_no = getattr(logging, level_name, logging.INFO)
-    logging.root.setLevel(level_no)
+    level = getattr(logging, level_name, logging.INFO)
+    logging.root.setLevel(level)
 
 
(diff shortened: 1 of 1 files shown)


skyrl-train/skyrl_train/utils/utils.py
```diff
@@ -465,6 +467,41 @@ def prepare_runtime_environment(cfg: DictConfig) -> dict[str, str]:
...
+    class _InterceptHandler(logging.Handler):
+        def emit(self, record: logging.LogRecord) -> None:
+            try:
+                level = logger.level(record.levelname).name
+            except ValueError:
+                level = record.levelno
+            logger.opt(depth=6, exception=record.exc_info).log(level, record.getMessage())
+
+    logging.root.handlers = [_InterceptHandler()]
+    logging.root.setLevel(logging.INFO)
```
SumanthRH: Hmm if I'm debugging and I set `LOG_LEVEL=DEBUG` in my env file, this will override the root logger to use INFO?
Maybe set it only if `LOG_LEVEL` is not set?

tyler-griggs: Good call!

SumanthRH (approved):

PR #257, opened 2025-09-08
Reuse mlflow if exists

sergeypastukhov-ddog:
With this PR mlflow logging works in two regimes:
* If no mlflow active run exists - we create mlflow and we close it when training is completed (`_MlflowLoggingAdapter` follows RAII)
* If mlflow active run exists - we don't create new mlflow experiment run and we don't close it when training is completed (the caller follows RAII)

### DIFF
--- a/skyrl-train/skyrl_train/utils/tracking.py
+++ b/skyrl-train/skyrl_train/utils/tracking.py
@@ -45,20 +45,7 @@ class Tracking:
             self.logger["wandb"] = wandb
 
         if "mlflow" in default_backend:
-            import os
-
-            import mlflow
-
-            MLFLOW_TRACKING_URI = os.environ.get("MLFLOW_TRACKING_URI", None)
-            if MLFLOW_TRACKING_URI:
-                mlflow.set_tracking_uri(MLFLOW_TRACKING_URI)
-
-            # Project_name is actually experiment_name in MLFlow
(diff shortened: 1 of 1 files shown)


skyrl-train/skyrl_train/utils/tracking.py
```diff
@@ -176,16 +163,34 @@ def finish(self):
 
 
 class _MlflowLoggingAdapter:
-    def log(self, data, step):
+    def __init__(self, project_name, experiment_name, config):
+        import os
+
         import mlflow
 
+        if mlflow.active_run() is None:
```
SumanthRH, line 171: Could you clarify when this will be not be `None` for context? Is this when appropriate env vars for `MLFLOW_TRACKING_URI` and `MLFLOW_EXPERIMENT_NAME` are set and thsu I can add to an existing run (say when I resume)?

sergeypastukhov-ddog: Hello, Sumanth, it is not None if someone else executed `mlflow.start_run` before. In this case we are already in some mlflow experiment. Main branch code now would just throw an error (because you can't start new experiment while active run exists)

SumanthRH: Got it. thanks

PR #260, opened 2025-09-08
[Generator][HTTP] Support /completion endpoint for token-in-token-out

CharlieFRuan:
This PR implements the `/completions` endpoint for our HTTP inference engine client endpoint, as a follow up to the `/chat/completions`. This is a necessity if user wants to do token-in-token-out while using an HTTP endpoint.

The main tricky part of `/completions` compared to `/chat/completions` is that it can be either batched or single-request. We handle routing using the util method introduced in https://github.com/NovaSky-AI/SkyRL/pull/267, making `handle_completions()` very similar to `generate()`.

### Tested
- Added a `_generate_with_chat_completions_http_endpoint()` to the dummy `examples/inference_http_endpoint/skyrl_gym_http_generator.py`, which offers a good sanity check. The GSM8K result (with 4xa100 sxm) is as follow:
image

- Legend:
  - `charlie-pr260-...`: this PR running `skyrl-train/examples/inference_http_endpoint/skyrl_gym_http_generator.py` with `/completions`
  - `charlie-pr267-4xa100`: regular `run_gsm8k.py` with Python `InferenceEngineClient.generate()` at commit https://github.com/NovaSky-AI/SkyRL/commit/4750548aed6001aa5d692946338008be197461cc
  - `charlie-pr238-http`: `skyrl-train/examples/inference_http_endpoint/skyrl_gym_http_generator.py` with  `/chat/completions`
- Correctness is good. Performance-wise, it is better than `/chat/completions` (since `/completions` is batched, so only one request per engine), but still slower than Python counterpart.

- Added GPU unit test (did not parameterize the test, but use a for loop within the test, hence saving time by avoiding spinning up engine multiple times; tried using fixture by it makes server/ray shutdown tricky). Verified locally

### DIFF
--- a/skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
+++ b/skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
@@ -12,6 +12,7 @@ from skyrl_train.inference_engines.utils import (
     route_prompts_to_engines,
     hash_with_sha256,
     postprocess_completion_request,
+    aggregate_completion_usage_info,
 )
 from omegaconf import DictConfig
 import threading
@@ -206,9 +207,18 @@ class InferenceEngineClient(InferenceEngineInterface):
                 ).model_dump()
 
         # 4. Combine choices and preserve original order.
+        # If there is only one result, we return it directly.
+        if len(results) == 1:
+            return results[0]
+
+        # Use the first result as base response. There are some fields that cannot be shared
+        # across sub-requests. For now it is just the usage field.
+        final_response = dict(results[0])
+        final_response["usage"] = aggregate_completion_usage_info(results, self.backend)
+
+        # Aggregate choices. TODO(Charlie): improve logic when we need to support n > 1
         # vLLM sets index positions per sub-batch, so we reset indices to be 0..n-1 for the combined response.
         combined_choices: list[Dict[str, Any]] = [None] * num_prompts
-        base_response = results[0]  # Use the first result as base response.
         for indices, result in zip(indices_list, results):
             # indices are the original prompt indices that the task's response corresponds to
             for local_idx, original_idx in enumerate(indices):
@@ -220,7 +230,6 @@ class InferenceEngineClient(InferenceEngineInterface):
         for new_idx in range(len(combined_choices)):
             assert combined_choices[new_idx]["index"] == new_idx
 
-        final_response = dict(base_response)
         final_response["choices"] = combined_choices
         return final_response
 
--- a/skyrl-train/skyrl_train/inference_engines/utils.py
+++ b/skyrl-train/skyrl_train/inference_engines/utils.py
@@ -173,3 +173,40 @@ def postprocess_completion_request(
         )
 
     return trajectory_id_value, prompt
+
+
+def aggregate_completion_usage_info(
+    results: List[Dict[str, Any]],
+    backend: str,
+) -> Dict[str, Any]:
+    """
+    Aggregate the completion usage info and return the final usage info, since our
+    inference engine client breaks down a batched request into sub-requests and routes to engines,
+    where each engine only returns its sub-request's usage info. When we return the final response,
+    we need to aggregate the usage info.
+
+    NOTE(Charlie): we don't explicitly import vllm here for ease of CPU test. Whether these fields
+    are still compatible with newer vllm versions can be checked in our GPU tests, where we explicitly
+    check the dictionary with `CompletionResponse.model_validate()`.
+    """
+    if backend == "vllm":
+        # required fields
+        usage_info = {
+            "prompt_tokens": sum(result["usage"]["prompt_tokens"] for result in results),
+            "total_tokens": sum(result["usage"]["total_tokens"] for result in results),
+        }
+        # optional fields
+        if results[0]["usage"].get("completion_tokens") is not None:
+            usage_info["completion_tokens"] = sum(result["usage"]["completion_tokens"] for result in results)
+        if results[0]["usage"].get("prompt_tokens_details") is not None:
+            if results[0]["usage"]["prompt_tokens_details"].get("cached_tokens") is not None:
+                usage_info["prompt_tokens_details"] = {
+                    "cached_tokens": sum(
+                        result["usage"]["prompt_tokens_details"]["cached_tokens"] for result in results
+                    )
+                }
+        return usage_info
+    elif backend == "sglang":
+        raise NotImplementedError("SGLang is not supported yet")
+    else:
+        raise ValueError(f"Unsupported backend: {backend}")
--- a/skyrl-train/tests/cpu/inf_engines/test_inference_engine_client.py
+++ b/skyrl-train/tests/cpu/inf_engines/test_inference_engine_client.py
@@ -166,10 +166,10 @@ def test_completion_batched_routing_and_order_preservation(num_prompts, with_tra
             indices 0, 1, 2, 3, ...
             """
             body = request_payload["json"]
-            prompts = body["prompt"]
+            my_prompts = body["prompt"]
             # Return per-sub-batch indices 0..len-1; client is expected to remap to global order
             choices = []
-            for i, p in enumerate(prompts):
+            for i, p in enumerate(my_prompts):
                 choices.append(
                     {
                         "index": i,
@@ -177,11 +177,21 @@ def test_completion_batched_routing_and_order_preservation(num_prompts, with_tra
                         "finish_reason": "stop",
                     }
                 )
+            num_prompt_tokens = sum(len(p) for p in my_prompts)
+            num_completion_tokens = num_prompt_tokens * 2  # since we doubled the prompts
             return {
                 "id": "cmpl-mock",
                 "object": "text_completion",
                 "model": body.get("model", "dummy-model"),
                 "choices": choices,
+                "usage": {
+                    "prompt_tokens": num_prompt_tokens,
+                    "total_tokens": num_prompt_tokens + num_completion_tokens,
+                    "completion_tokens": num_completion_tokens,
+                    "prompt_tokens_details": {
+                        "cached_tokens": num_prompt_tokens,
+                    },
+                },
             }
 
     # Create a minimal config to avoid spinning up HTTP endpoint
@@ -227,6 +237,18 @@ def test_completion_batched_routing_and_order_preservation(num_prompts, with_tra
     for i, choice in enumerate(resp["choices"]):
         assert choice["index"] == i
         assert choice["text"] == expected_texts[i]
+
+    # also check usage aggregation here
+    global_num_prompt_tokens = sum(len(p) for p in prompts)
+    global_num_completion_tokens = global_num_prompt_tokens * 2  # since we doubled the prompts
+    assert resp["usage"] == {
+        "prompt_tokens": global_num_prompt_tokens,
+        "total_tokens": global_num_prompt_tokens + global_num_completion_tokens,
+        "completion_tokens": global_num_completion_tokens,
+        "prompt_tokens_details": {
+            "cached_tokens": global_num_prompt_tokens,
+        },
+    }
 
 
 # -------------------------------------------


skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
```diff
@@ -124,9 +129,97 @@ async def chat_completion(self, request_payload: Dict[str, Any]) -> Dict[str, An
...
+                # former is vllm format, latter is sglang format
+                error_code = result["error"]["code"] if "error" in result else result["code"]
+                error_type = result["error"]["type"] if "error" in result else result["type"]
+                return ErrorResponse(
+                    error=ErrorInfo(
+                        message=f"In one of the engines that SkyRL manages, an error occurred: {result.error.message}",
+                        type=error_code,
+                        code=error_type,
+                    ),
+                ).model_dump()
```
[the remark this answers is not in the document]
CharlieFRuan: Fixed.

skyrl-train/tests/gpu/gpu_ci/test_inference_engine_client_http_endpoint.py
```diff
@@ -491,5 +726,33 @@ def run_server():
...
+            f"{base_url}/v1/completions",
+            data="some invalid json",
+            headers={"Content-Type": "application/json"},
+        )
+        assert response.status_code == HTTPStatus.BAD_REQUEST
+
+        # When batched and trajectory_id wrong length -> 400 from server or client-side error
+        bad_payload = {"model": MODEL, "prompt": ["hi", "hello", "ok"], "trajectory_id": [0, 1]}
+        r = requests.post(f"{base_url}/v1/completions", json=bad_payload)
+        assert r.status_code in (HTTPStatus.BAD_REQUEST, HTTPStatus.INTERNAL_SERVER_ERROR)
```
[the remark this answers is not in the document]
CharlieFRuan: Fixed.

tyler-griggs:
Can you briefly clarify what the 4 colors correspond to in the wandb plots?

skyrl-train/examples/inference_http_endpoint/skyrl_gym_http_generator.py
```diff
@@ -69,7 +70,10 @@ async def generate_batched(
         )
 
         # The only line different from SkyRLGymGenerator.generate_batched
-        engine_output = await _generate_with_http_endpoint(self.base_url, self.model_name, engine_input)
+        # engine_output = await _generate_with_chat_completions_http_endpoint(self.base_url, self.model_name, engine_input)
```
tyler-griggs, line 79: nit: delete commented line?

skyrl-train/examples/inference_http_endpoint/skyrl_gym_http_generator.py
tyler-griggs, line 1: I'm being nit-picky here, but do you think this examples sub-folder is needed? Basically, users just need to enable `enable_http_endpoint` (which, long-term, I feel we should just enable by default) and then build the url from `http_endpoint_host` and `http_endpoint_port`.  Since OpenAI APIs are so popular, I don't know if we need to demonstrate how a user can generate with a v1/completions or v1/chat/completions endpoint, which seems to be most of the LoC in this file. Further, I doubt most users of the http endpoint are going to use InferenceEngineInput, so I also don't think it will be helpful to them to see this conversion.

Please push back if I'm missing something, but I'd consider removing this directory.

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
```diff
@@ -124,9 +129,97 @@ async def chat_completion(self, request_payload: Dict[str, Any]) -> Dict[str, An
...
+        """
+        Handles an OpenAI /completions request.
+
+        Since `request.prompt` can be `Union[list[int], list[list[int]], str, list[str]]`,
+        (i.e. {batched, single} x {string, token IDs}), we need to route the request to engines
+        differently, based on whether it's a single or batched request, and whether trajectory_id
+        is provided. This is similar to `generate()` method.
+        For single, we do the same routing logic as `chat_completion()`. For batched, we route by
+        trajectory_id if present, and if not we split evenly across engines.
+        """
```
tyler-griggs, line 152: nit: could be nice to add a note that this maintains the same order?

skyrl-train/skyrl_train/inference_engines/inference_engine_client_http_endpoint.py
```diff
@@ -234,16 +239,41 @@ async def chat_completion(raw_request: Request):
...
+        they get routed to the same engine for better prefix caching. If unprovided, we will route
+        to a random engine which is not performant.
+
+        API reference:
+        - https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html
+        - https://docs.sglang.ai/basic_usage/openai_api_completions.html
+        """
+        return await handle_openai_request(raw_request, endpoint="/chat/completions")
+
+    # Text completion endpoint
```
tyler-griggs: nit: what do you mean by this? It supports tokens, right?

skyrl-train/skyrl_train/inference_engines/inference_engine_client_http_endpoint.py
```diff
@@ -234,16 +239,41 @@ async def chat_completion(raw_request: Request):
...
+    @app.post("/v1/completions")
+    async def completions(raw_request: Request):
+        """
+        Takes in OpenAI's `CompletionRequest` and returns OpenAI's `CompletionResponse`.
+
+        Note that the specific fields inside the request and response depend on the backend you use.
+        If `config.generator.backend` is `vllm`, then the request and response will be vLLM's.
+        SkyRL does not do any field checkings besides `model` and `trajectory_id` but instead offload
+        everything to the underlying engines.
```
tyler-griggs:
```suggestion
        SkyRL only validates the fields `model` and `trajectory_id`, and otherwise offloads
        field validation to the underlying engines.
```

skyrl-train/skyrl_train/inference_engines/inference_engine_client_http_endpoint.py
```diff
@@ -234,16 +239,41 @@ async def chat_completion(raw_request: Request):
...
+
+        Make sure you add in `trajectory_id` to ensure load balancing and sticky routing. Since
+        `request.prompt` can be `Union[list[int], list[list[int]], str, list[str]]`, i.e.
+        {batched, single} x {string, token IDs}, we follow the following logic for request routing:
+        - If batched request, `trajectory_id`, if provided, must have the same length as `request.prompt`
+          so that each request.prompt[i] will be routed based on `trajectory_id[i]`.
+        - If single request, `trajectory_id`, if provided, must be a single integer or a singleton list.
+        Where each trajectory_id is a string or an integer.
```
tyler-griggs:
```suggestion
        - For batched request: `trajectory_id`, if provided, must have the same length as `request.prompt`
          so that each request.prompt[i] is routed based on `trajectory_id[i]`.
        - For single request: `trajectory_id`, if provided, must be a single integer or a singleton list, where
          each trajectory_id is a string or an integer.
```

skyrl-train/skyrl_train/inference_engines/inference_engine_client_http_endpoint.py
tyler-griggs: this and the below are just nits on wording :)

skyrl-train/tests/cpu/inf_engines/test_inf_engine_utils.py
```diff
@@ -0,0 +1,128 @@
+"""
+Test for postprocess_completion_request function that postprocesses completion requests in inference engine client.
```
tyler-griggs: nit: I assume this file can / will be used to test other methods in the inference_engines/utils.py file. You could probably remove this comment, or make it more generic?

skyrl-train/tests/gpu/gpu_ci/test_inference_engine_client_http_endpoint.py
```diff
@@ -45,36 +47,46 @@
...
+
+def _check_chat_completions_outputs(outputs, test_type, num_samples, backend):
+    # check error
+    for output in outputs:
+        assert not ("error" in output or output.get("object", "") == "error"), f"Error in output: {output}"
     print_n = 5
     assert len(outputs) == num_samples
     print(f"First {print_n} generated responses out of {num_samples} using {test_type}:")
-    print(f"outputs[0]: {outputs[0]}")
+    # print(f"outputs[0]: {outputs[0]}")
```
tyler-griggs: nit: uncomment?

skyrl-train/tests/gpu/gpu_ci/test_inference_engine_client_http_endpoint.py
```diff
@@ -98,35 +110,175 @@ def _check_outputs(outputs, test_type, num_samples, backend):
...
+    for output in outputs:
+        assert not ("error" in output or output.get("object", "") == "error"), f"Error in output: {output}"
+
+    num_outputs = sum(len(output["choices"]) for output in outputs)
+    assert num_outputs == len(prompts)
+
+    print_n = 5
+    # Content checks
+    print(f"First {print_n} generated responses out of {num_outputs} using completions:")
+    # print(f"outputs[0]: {outputs[0]}")
```
tyler-griggs: nit: uncomment here as well?

skyrl-train/tests/gpu/gpu_ci/test_inference_engine_client_http_endpoint.py
```diff
@@ -98,35 +110,175 @@ def _check_outputs(outputs, test_type, num_samples, backend):
...
+@pytest.mark.vllm
+def test_http_endpoint_completions_routing_and_batching():
+    """
+    Since /completions endpoint supports both single and batched requests, and we support
+    either using trajectory_id or not, we test all combinations.
+
+    We test with 2 engines of TP=1 to check if routing works.
+
+    Ideally we want to test that for the batched case, despite routing, the choices should have
+    the same order as the input prompt (i.e. order presreved), but not sure how to test this. We
```
tyler-griggs: Yeah good point, it is a little tricky. This could be a pretty subtle and nasty bug, so it would be great to test for it. But I assumed that your CPU test for `test_route_prompts_to_engine` has pretty good coverage for this since it's mostly CPU behavior we're checking for?

I guess what's missing in the CPU test is whether the underlying inference engine (eg, vllm) is correctly maintaining the order, right?

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
```diff
@@ -124,9 +129,97 @@ async def chat_completion(self, request_payload: Dict[str, Any]) -> Dict[str, An
...
+        # return an error response. That is, there is no partial success, following vLLM and SGLang's behavior.
+        for result in results:
+            if "error" in result or result.get("object", "") == "error":
+                # former is vllm format, latter is sglang format
+                error_code = result["error"]["code"] if "error" in result else result["code"]
+                error_type = result["error"]["type"] if "error" in result else result["type"]
+                return ErrorResponse(
+                    error=ErrorInfo(
+                        message=f"In one of the engines that SkyRL manages, an error occurred: {result['error']['message']}",
+                        type=error_code,
```
tyler-griggs: I think these two fields (code and type) are swapped

skyrl-train/examples/inference_http_endpoint/skyrl_gym_http_generator.py
```diff
@@ -172,3 +176,52 @@ async def _generate_with_http_endpoint(
...
+    Generate responses using direct ClientSession.post calls with the InferenceEngineClient HTTP endpoint.
+
+    Equivalent to running `self.inference_engine_client.generate()`, but with the HTTP endpoint.
+    """
+    prompts = input_batch.get("prompts")
+    # Since we are using /completions, we need to template the prompts ourselves
+    prompts = [tokenizer.apply_chat_template(prompt, add_generation_prompt=True, tokenize=False) for prompt in prompts]
+    trajectory_ids = input_batch.get("trajectory_ids")
+    sampling_params = input_batch.get("sampling_params")
+    if trajectory_ids is not None:
```
tyler-griggs, line 203: I think these are not actually used -- is the intent to pass `trajectory_ids` into the `payload` below?

CharlieFRuan:
> Can you briefly clarify what the 4 colors correspond to in the wandb plots?

Just added in the PR description; also removed a line that doesn't contribute much

skyrl-train/examples/inference_http_endpoint/skyrl_gym_http_generator.py
CharlieFRuan: Yea that is totally fair. Can we remove at a separate PR? This example was helpful in confirming correctness as shown in the PR description. There wasn't another easy way to ensure correctness (reward is normal, hence showing batched input/output order is maintained in `/completions`, etc.)

skyrl-train/examples/inference_http_endpoint/skyrl_gym_http_generator.py
CharlieFRuan: Ah I kind of wanted to keep this since it shows how to use `/chat/completions` and `/completions` respectively.

Though we'll remove this file soon regardless

skyrl-train/tests/gpu/gpu_ci/test_inference_engine_client_http_endpoint.py
CharlieFRuan: I guess `test_route_prompts_to_engine` is not quite enough. Because the main thing that is preserving the order is this snippet of code in `InferenceEngineClient.completion()`:

```python
        combined_choices: list[Dict[str, Any]] = [None] * num_prompts
        base_response = results[0]  # Use the first result as base response.
        for indices, result in zip(indices_list, results):
            # indices are the original prompt indices that the task's response corresponds to
            for local_idx, original_idx in enumerate(indices):
                choice = result["choices"][local_idx]
                choice["index"] = original_idx  # overwrite index with the global position
                combined_choices[original_idx] = choice
```

the logic of which is pretty much shared by `.generate()`.

I printed out both the input and output in `test_http_endpoint_completions_routing_and_batching()`, so at least we could eyeball when debugging whether the output matches the prompt (i.e. order preserved).

I think these lines of code are rarely touched, so I am relatively comfortable

skyrl-train/examples/inference_http_endpoint/skyrl_gym_http_generator.py
CharlieFRuan: ohh right, I was mainly trying to test the behavior of "splitting prompts evenly across engines". I guess since we're removing the file soon, it should be fine.

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
CharlieFRuan: Added!

skyrl-train/skyrl_train/inference_engines/inference_engine_client_http_endpoint.py
CharlieFRuan: Some Cursor tabbed comment... Removed.

skyrl-train/skyrl_train/inference_engines/inference_engine_client_http_endpoint.py
CharlieFRuan: thanks, adopted!

skyrl-train/tests/cpu/inf_engines/test_inf_engine_utils.py
CharlieFRuan: Yep, made it more generic.

skyrl-train/tests/gpu/gpu_ci/test_inference_engine_client_http_endpoint.py
CharlieFRuan: With logprobs turned on, the output will blast the log. Removed both instead.

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
CharlieFRuan: good catch, thanks!

skyrl-train/examples/inference_http_endpoint/skyrl_gym_http_generator.py
tyler-griggs: Ooh I see, makes sense!

skyrl-train/examples/inference_http_endpoint/skyrl_gym_http_generator.py
tyler-griggs: That makes sense to me! If the plan is to remove it, could we add a TODO in-code just to keep track? Maybe also motivation to remove the `engine.py` paths, switch the SkyRLGymGenerator over to HTTP, and use that more easily for correctness checks :D Pending some perf profiling for http, perhaps.

skyrl-train/tests/gpu/gpu_ci/test_inference_engine_client_http_endpoint.py
tyler-griggs: Got it!

skyrl-train/tests/gpu/gpu_ci/test_inference_engine_client_http_endpoint.py
tyler-griggs: I guess we could extract this logic into a utils method and then test it on CPU tests, and could even combine it wiht `route_prompts_to_engine` to test the behavior "end to end". 

I don't have strong feelings here, so I'll leave it to you whether you think that's necessary or not!

skyrl-train/examples/inference_http_endpoint/skyrl_gym_http_generator.py
```diff
@@ -172,3 +178,52 @@ async def _generate_with_http_endpoint(
...
+
+    # Use aiohttp session for direct HTTP requests
+    conn = aiohttp.TCPConnector(limit=0, limit_per_host=0)  # 0 = no limit; without conn, has 100
+    async with aiohttp.ClientSession(connector=conn, timeout=aiohttp.ClientTimeout(total=None)) as session:
+        headers = {"Content-Type": "application/json"}
+        payload = {
+            "model": model_name,
+            "prompt": prompts,
+            **(sampling_params or {}),
+        }
```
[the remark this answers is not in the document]
CharlieFRuan: As discussed in one of the threads above, this file will be removed soon, so it is fine. It was left out in purpose to test "split input evenly across engines"

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
```diff
@@ -124,9 +129,100 @@ async def chat_completion(self, request_payload: Dict[str, Any]) -> Dict[str, An
...
+                # former is vllm format, latter is sglang format
+                error_code = result["error"]["code"] if "error" in result else result["code"]
+                error_type = result["error"]["type"] if "error" in result else result["type"]
+                return ErrorResponse(
+                    error=ErrorInfo(
+                        message=f"In one of the engines that SkyRL manages, an error occurred: {result['error']['message']}",
+                        type=error_type,
+                        code=error_code,
+                    ),
+                ).model_dump()
```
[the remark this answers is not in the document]
CharlieFRuan: Good one, fixed.

skyrl-train/tests/gpu/gpu_ci/test_inference_engine_client_http_endpoint.py
CharlieFRuan: Actually I was able to mock the `engine.completion()` and `engine.generate()`. So E2E was able to be tested. See `test_completion_batched_routing_and_order_preservation()` in `test_inference_engine_client.py`. There is also an equivalent one for `InferenceEngineClient.generate()`. I think these tests are pretty solid (since vllm's APIs won't change, can be perfectly mocked).

I also moved `route_prompts_to_engines()` to the same `test_inference_engine_client.py`, which tests both the utils and some mockable logics

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
```diff
@@ -124,9 +129,100 @@ async def chat_completion(self, request_payload: Dict[str, Any]) -> Dict[str, An
...
+
+        # 4. Combine choices and preserve original order.
+        # vLLM sets index positions per sub-batch, so we reset indices to be 0..n-1 for the combined response.
+        combined_choices: list[Dict[str, Any]] = [None] * num_prompts
+        base_response = results[0]  # Use the first result as base response.
```
tyler-griggs: I'm not knowledgeable enough about the structure of the responses, but I assume this isn't going to break anything? For example, I know there's often a `usage` field and maybe that's something that's typically surfaced to the base response as a sum across the `usage` fields in all requests. I'm not sure, just wanted to ask whether this base response adheres to the APIs

skyrl-train/tests/gpu/gpu_ci/test_inference_engine_client_http_endpoint.py
CharlieFRuan: In addition, I realized that `n > 1` won't work because that would require a different way of aggregating the results. I added an error throwing and a test that checks that error. We can support when needed

skyrl-train/examples/inference_http_endpoint/skyrl_gym_http_generator.py
CharlieFRuan: Added a TODO comment, will consolidate next steps for HTTP endpoint / rollout stack in general (removing `engine.py`, profiling, in-house DP routing vs. inference frameworks' routing, sticky session, etc.)

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
CharlieFRuan: Ohh you are right, good catch. I'll be more explicit about what to keep on the highest level

skyrl-train/examples/inference_http_endpoint/skyrl_gym_http_generator.py
```diff
@@ -172,3 +182,52 @@ async def _generate_with_http_endpoint(
...
+    # Since we are using /completions, we need to template the prompts ourselves
+    prompts = [tokenizer.apply_chat_template(prompt, add_generation_prompt=True, tokenize=False) for prompt in prompts]
+    trajectory_ids = input_batch.get("trajectory_ids")
+    sampling_params = input_batch.get("sampling_params")
+    if trajectory_ids is not None:
+        assert len(prompts) == len(trajectory_ids), "prompts and trajectory_ids must have the same length"
+
+    # Use aiohttp session for direct HTTP requests
+    conn = aiohttp.TCPConnector(limit=0, limit_per_host=0)  # 0 = no limit; without conn, has 100
+    async with aiohttp.ClientSession(connector=conn, timeout=aiohttp.ClientTimeout(total=None)) as session:
```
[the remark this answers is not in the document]
CharlieFRuan: The file will be removed soon, it is fine.

skyrl-train/skyrl_train/inference_engines/remote_inference_engine.py
```diff
@@ -117,6 +117,17 @@ async def chat_completion(self, request_payload: Dict[str, Any]) -> Dict[str, An
...
+    async def completion(self, request_payload: Dict[str, Any]) -> Dict[str, Any]:
+        body = request_payload.get("json", {})
+        headers = {"Content-Type": "application/json"}
+        response = None
+        async with aiohttp.ClientSession(timeout=aiohttp.ClientTimeout(total=None)) as session:
+            request_url = f"{self.url}/v1/completions"
+            async with session.post(request_url, json=body, headers=headers) as resp:
+                response = await resp.json()
+
+        return response
```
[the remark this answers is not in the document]
CharlieFRuan: Kind of prefer the response = None being on top, which is used across the file.

On session-reuse, this is a good TODO to add. Perhaps as part of some performance profiling when we want to strip `xxx_engine.py`

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
CharlieFRuan: Addressed. For now it only seems that `UsageInfo` cannot be reused from a sub-request. I added a helper that aggregate `UsageInfo` and tested in CPU test (along with the routing and order).

I guess this is a minor Con of us maintaining DP router

CharlieFRuan:
/gemini review

Please pay closer attention to the most recent commit, which added `aggregate_completion_usage_info()`

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
```diff
@@ -124,9 +130,109 @@ async def chat_completion(self, request_payload: Dict[str, Any]) -> Dict[str, An
...
+        # return an error response. That is, there is no partial success, following vLLM and SGLang's behavior.
+        for result in results:
+            if "error" in result or result.get("object", "") == "error":
+                # former is vllm format, latter is sglang format
+                error_details = result.get("error", result)  # resolves vllm/sglang format difference
+                error_code = error_details["code"]
+                error_type = error_details["type"]
+                return ErrorResponse(
+                    error=ErrorInfo(
+                        message=f"In one of the engines that SkyRL manages, an error occurred: {error_details['message']}",
```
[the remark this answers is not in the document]
CharlieFRuan: @gemini-code-assist This is fine. Can you review code around `aggregate_completion_usage_info()` and see if they are ok?

CharlieFRuan:
Tested `skyrl-train/tests/gpu/gpu_ci/test_inference_engine_client_http_endpoint.py` locally after all the changes

CharlieFRuan:
Thanks for the detailed review! @tyler-griggs

PR #265, opened 2025-09-08
[trainer][megatron] Override fused attention with flash-attn when flash_attn=True for Megatron

erictang000:
# Overview

This PR makes sure that the flash-attn backend is chosen when flash_attn=True for the Megatron backend. Previously the backend for TransformerEngine was "auto" and was being overridden in the [`get_attention_backend` function ](https://github.com/NVIDIA/TransformerEngine/blob/release_v2.5/transformer_engine/pytorch/attention/dot_product_attention/utils.py#L916).

### Speedups (~10x for PP, 5x for TP on the first step)
image

### Correctness:

avg raw reward is slightly lower with flash-attn
image

Grad norm is a little lower with flash-attn 
image

But overall evals seem ok
image

### DIFF
--- a/skyrl-train/pyproject.toml
+++ b/skyrl-train/pyproject.toml
@@ -63,7 +63,7 @@ conflicts = [
 skyrl-gym = { path = "./skyrl-gym" , editable = true }
 torch = { index = "pytorch-cu128" }
 torchvision = { index = "pytorch-cu128" }
-flash-attn =  { url = "https://github.com/Dao-AILab/flash-attention/releases/download/v2.8.0.post2/flash_attn-2.8.0.post2+cu12torch2.7cxx11abiFALSE-cp312-cp312-linux_x86_64.whl" }
+flash-attn = { url = "https://github.com/Dao-AILab/flash-attention/releases/download/v2.8.0.post2/flash_attn-2.8.0.post2+cu12torch2.7cxx11abiFALSE-cp312-cp312-linux_x86_64.whl" }
 # NOTE (sumanthrh): We explictly use a flashinfer wheel from their index. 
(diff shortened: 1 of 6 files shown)


skyrl-train/pyproject.toml
```diff
@@ -63,7 +63,10 @@ conflicts = [
 skyrl-gym = { path = "./skyrl-gym" , editable = true }
 torch = { index = "pytorch-cu128" }
 torchvision = { index = "pytorch-cu128" }
-flash-attn = { url = "https://github.com/Dao-AILab/flash-attention/releases/download/v2.8.0.post2/flash_attn-2.8.0.post2+cu12torch2.7cxx11abiFALSE-cp312-cp312-linux_x86_64.whl" }
+flash-attn = [
+    { url = "https://github.com/Dao-AILab/flash-attention/releases/download/v2.8.0.post2/flash_attn-2.8.0.post2+cu12torch2.7cxx11abiFALSE-cp312-cp312-linux_x86_64.whl", marker = "extra != 'mcore'" },
+    { url = "https://github.com/Dao-AILab/flash-attention/releases/download/v2.7.4.post1/flash_attn-2.7.4.post1+cu12torch2.7cxx11abiFALSE-cp312-cp312-linux_x86_64.whl", marker = "extra == 'mcore'" }
+]
```
SumanthRH: btw is the dep change needed when you do an override to use Tri Dao's flash attn anyway? Does mcore break ?

erictang000: yeah, the max supported flash-attn version for the pinned transformerengine version is 2.7.4.post1, so it will break if you use 2.8.0 (it will fall back to no flash-attention)

PR #267, opened 2025-09-09
[InfEngineClient] Extract out routing logic to a helper

CharlieFRuan:
This PR removes `_generate_batched()` and `_generate_with_trajectory_routing()` in `InferenceEngineClient` by adding a helper method that:

```python
def route_prompts_to_engines(
    num_prompts: int, num_inference_engines: int, trajectory_ids: Optional[Union[list[int], list[str]]]
) -> dict[int, list[int]]:
```

returning a mapping from engine ID to the prompt IDs they work on.

This removes the duplicated code when gathering the output.

In addition, for the `/completions` PR, this will be highly beneficial. Without this PR we would need to duplicate `_generate_with_trajectory_routing()` and `_generate_batched()`, which are for `InferenceEngineInput` and make another pair for `CompletionRequest`.

We also change `hash()` to `int.from_bytes(hashlib.sha256(str(x).encode()).digest(), "big")`, since the former is seeded by a random seed that is per-python-process. The latter is deterministic.

### Testing
- Added new unit test
- GPU CI passed
- GSM8K, i.e. single-turn code path (ran both batched and non-batched). Same generation time, reward, eval. Compared against PR #238 
image

- search-r1 (compared against the TITO PR, with single-turn and 4turns). Reward and eval are same -- meaning keeping the original order worked as expected; timing is roughly 7% slower for some reason. Will assume it to be variance.
image

### DIFF
--- a/skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
+++ b/skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
@@ -7,6 +7,7 @@ from skyrl_train.inference_engines.base import (
 from transformers import PreTrainedTokenizerBase
 import asyncio
 from typing import List, Any, Optional, Dict
+from skyrl_train.inference_engines.utils import route_prompts_to_engines, hash_with_sha256
 from omegaconf import DictConfig
 import threading
 import random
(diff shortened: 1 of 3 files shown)


SumanthRH (approved):
Thanks

tyler-griggs (approved):
Much better, thank you

PR #269, opened 2025-09-09
[Fix][CI] Address generator CI test fails when model stop reason is length

CharlieFRuan:
Our unit test checks whether, for turns=3, the final conversation generates 3 EOS token. This will not be the case when the model's generation stop due to length. To address, we do the following:
- In the dummy environment, change observation `"turn {i}"` to `"give me another solution {i}"`, which might make more sense for the model
- Increase max generate length from 1000 to 3000
- Final guard is to check when the stop reason is not `"stop"`, we don't check the number of EOS token IDs

### DIFF
--- a/skyrl-train/tests/gpu/gpu_ci/test_skyrl_gym_generator.py
+++ b/skyrl-train/tests/gpu/gpu_ci/test_skyrl_gym_generator.py
@@ -46,7 +46,7 @@ class TestEnv(BaseTextEnv):
         self.turns += 1
         done = self.turns >= self.max_turns
         return BaseTextEnvStepOutput(
-            observations=[{"role": "user", "content": f" {self.turns}"}] if not done else [],
+            observations=[{"role": "user", "content": f"{OBSERVATION_PROMPT} {self.turns}"}] if not done else [],
             reward=0,
             done=done,
             metadata={},
--- a/skyrl-train/tests/gpu/gpu_ci/test_skyrl_gym_generator.py
+++ b/skyrl-train/tests/gpu/gpu_ci/test_skyrl_gym_generator.py
@@ -57,9 +57,9 @@ register(
 )
 
 MODEL_TO_GENERATION_PROMPT = {
-    "Qwen/Qwen2.5-3B-Instruct": "<|im_start|>assistant\n",
-    "unsloth/Llama-3.2-3B-Instruct": "<|start_header_id|>assistant<|end_header_id|>\n\n",
-    "Qwen/Qwen3-1.7B": "<|im_start|>assistant\n",
+    "Qwen/Qwen2.5-1.5B-Instruct": "<|im_start|>assistant\n",
+    "unsloth/Llama-3.2-1B-Instruct": "<|start_header_id|>assistant<|end_header_id|>\n\n",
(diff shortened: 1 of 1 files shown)


skyrl-train/tests/gpu/gpu_ci/test_skyrl_gym_generator.py
```diff
@@ -302,26 +301,39 @@ async def test_generator_formatting_use_conversation_multi_turn(model_name):
...
+                # total eos in full response: 2 user eos + 3 assistant eos
+                assert sum(1 for _ in resp_ids if _ == tokenizer.eos_token_id) == 5
+            else:
+                # On length stops, the model may not produce EOS at the end of each assistant turn.
+                # Only check that generation prompts are masked out.
+                assert (
+                    MODEL_TO_GENERATION_PROMPT[model_name] in masked_out_resp_str
+                    and MODEL_TO_GENERATION_PROMPT[model_name] not in masked_in_resp_str
+                ), "generation prompts should be loss masked out"
+                logger.warning(f"Got stop reason {stop_reason}, so we did not fully checked the response")
```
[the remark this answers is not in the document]
CharlieFRuan: addressed.

skyrl-train/tests/gpu/gpu_ci/test_skyrl_gym_generator.py
```diff
@@ -56,9 +57,9 @@ def step(self, action: str):
 )
 
 MODEL_TO_GENERATION_PROMPT = {
-    "Qwen/Qwen2.5-1.5B-Instruct": "<|im_start|>assistant\n",
-    "unsloth/Llama-3.2-1B-Instruct": "<|start_header_id|>assistant<|end_header_id|>\n\n",
-    "Qwen/Qwen3-0.6B": "<|im_start|>assistant\n",
+    "Qwen/Qwen2.5-3B-Instruct": "<|im_start|>assistant\n",
+    "unsloth/Llama-3.2-3B-Instruct": "<|start_header_id|>assistant<|end_header_id|>\n\n",
+    "Qwen/Qwen3-1.7B": "<|im_start|>assistant\n",
```
SumanthRH: Hmm this doesn't seem like the best fix. Idealy we keep models as small as possible


This is motivated by small models repeating themselves and hitting hte `length` limit? 

Intuitively the agent loop exits based on three criteria:
1. step limit: `max_turns` in our case
2. budget: cost for openai models, in our case it's just max length
3. model-initiated: model wants to exit. 

It looks like we should just add 2. and this is fixed?

CharlieFRuan: Yea I included solution 2, but in that case not much can be checked and hence makes the test a bit lenient.

But since this doesn't seem to happen frequently we can do it with a warning.

CharlieFRuan: I will revert this model size change, but keep the max generate length increase

SumanthRH: Got it. 

I also feel this test itself is not great given that we are just checking for total eos tokens and thus need to deal with these edge cases. We should revisit the test later.

SumanthRH (approved):

PR #274, opened 2025-09-09
[trainer][megatron] Sequence packing + Context Parallel for Megatron

erictang000:
# Overview
Adds sequence packing + context parallel support for megatron backend. Note that context parallel without sequence packing is not supported.


SumanthRH (commented):
Looks good! As a sanity check, let's just benchmark seq packing time with the current packing impl

erictang000:
seems to just be faster in the forward pass without sequence packing for this micro batch size

without seq pack (~78ms)
image

with seq pack (~105 ms)
image

this is consistent with the ~25% slowdown in overall time
image

erictang000:
narrowing it down more, seems like the gap is in the application of rotary positional embeddings:

with seq pack:
image

without seq pack:
image

PR #276, opened 2025-09-10
Add configurable timeout for placement groups

SumanthRH:
# What does this PR do?

Adds a configurable timeout variable `SKYRL_RAY_PG_TIMEOUT` so that users can specify a custom timeout for allocating different placement groups in SkyRL.

### DIFF
--- a/skyrl-train/skyrl_train/entrypoints/main_base.py
+++ b/skyrl-train/skyrl_train/entrypoints/main_base.py
@@ -14,6 +14,7 @@ from skyrl_train.trainer import RayPPOTrainer
 from skyrl_train.inference_engines.inference_engine_client import InferenceEngineClient
 from skyrl_train.inference_engines.remote_inference_engine import create_remote_inference_engines
 from skyrl_train.utils.utils import initialize_ray, get_ray_pg_ready_with_timeout
+from skyrl_train.utils.constants import SKYRL_RAY_PG_TIMEOUT
 from skyrl_train.generators.base import GeneratorInterface
 from omegaconf import OmegaConf, DictConfig
 from pathlib import Path
@@ -134,7 +135,7 @@ class BasePPOExp:
             return prompts_dataset
         return None
 
-    def get_colocate_pg(self, timeout: int = 180) -> PlacementGroup:
+    def get_colocate_pg(self, timeout: int = SKYRL_RAY_PG_TIMEOUT) -> PlacementGroup:
         """Initializes a placement group for colocated training.
 
         A single placement group that packs all the inference engines together is created.
--- a/skyrl-train/skyrl_train/utils/constants.py
+++ b/skyrl-train/skyrl_train/utils/constants.py
@@ -1,0 +1,8 @@
+import os
+
+# Timeout for allocating the placement group for different actors in SkyRL
+SKYRL_RAY_PG_TIMEOUT = int(os.environ.get("SKYRL_RAY_PG_TIMEOUT", 30))
+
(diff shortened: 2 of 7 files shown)


skyrl-train/skyrl_train/entrypoints/main_base.py
```diff
@@ -134,7 +135,7 @@ def get_eval_dataset(self):
             return prompts_dataset
         return None
 
-    def get_colocate_pg(self, timeout: int = 180) -> PlacementGroup:
+    def get_colocate_pg(self, timeout: int = SKYRL_RAY_PG_TIMEOUT_IN_S) -> PlacementGroup:
```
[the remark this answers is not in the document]
SumanthRH: Umm I actually don't think the previous high timeout value was neeeded. Will check anyways

skyrl-train/skyrl_train/utils/constants.py
```diff
@@ -0,0 +1,15 @@
+import os
+
+#
+SKYRL_RAY_PG_TIMEOUT_IN_S = int(os.environ.get("SKYRL_RAY_PG_TIMEOUT_IN_S", 60))
+"""
+Timeout for allocating the placement group for different actors in SkyRL
+"""
+
+# For some reason the `LD_LIBRARY_PATH` is not exported to the worker with .env file.
+SKYRL_LD_LIBRARY_PATH_EXPORT = bool(os.environ.get("SKYRL_LD_LIBRARY_PATH_EXPORT", False))
```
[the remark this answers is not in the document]
SumanthRH: The downsides of vibecoding

skyrl-train/skyrl_train/utils/utils.py
Swynfel, line 1: `peer_access_supported` (around line 558) also contained a hard-coded placement group timeout but doesn't use `get_ray_pg_ready_with_timeout` (in line
`ray.get(placement_group([{"CPU": 1, "GPU": 2}], strategy="PACK").ready(), timeout=120)`)
I think it would make sense to skip this function (`peer_access_supported`) if the environment variables have already been set explicitely, but that's a question out-of-scope for this PR. For the current flow, I think using the same environment variable for the timeout is fine

SumanthRH: Hmm good point. I updated the PR to use the same timeout even for the `peer_access_supported` check! 

>  think it would make sense to skip this function (peer_access_supported) if the environment variables have already been set explicitely, but that's a question out-of-scope for this PR.

I might have misunderstood your message, but as such we definitely want to run `peer_access_supported` all the time so that people get a good experience on L4s and H100s out of the box. This function also runs in the beginning before the training run starts, and will be the first place where timeout errors could occur if the cluster is autoscaling.

erictang000 (approved):

skyrl-train/skyrl_train/utils/utils.py
Swynfel: `peer_access_supported` is used to automatically set "NCCL_P2P_DISABLE" and "NCCL_SHM_DISABLE" to the best value
My comment was that, if the user already manually set these values themself, the check is not "needed" and could be skipped

SumanthRH: Ah, good point! Agreed that it is leave it out of scope for this PR, but will be nice to have!


I hope the changes in the current PR make sense in the context of timeouts?

Swynfel: Yes it does! For our use-case, it will completely fix the auto-scaling issue

PR #289, opened 2025-09-12
Wandb same-step logging

tyler-griggs:
Previously, wandb used to log step `x`'s metrics after step `x+1` completed.

This PR ensures wandb metrics for step `x` are pushed to wandb immediately after step `x` has completed.

### DIFF
--- a/skyagent/examples/run_verl/verl_trainer.py
+++ b/skyagent/examples/run_verl/verl_trainer.py
@@ -357,7 +357,7 @@ class SkyAgentPPOTrainer(RayPPOTrainer):
         logger = Tracking(
             project_name=self.config.trainer.project_name,
             experiment_name=self.config.trainer.experiment_name,
-            backends=self.config.trainer.logger,
+            default_backend=self.config.trainer.logger,
             config=OmegaConf.to_container(self.config, resolve=True),
         )
 


skyrl-train/skyrl_train/trainer.py
```diff
@@ -253,7 +253,7 @@ def train(self):
             with self.eval_weights_manager:
                 with Timer("eval", self.all_timings):
                     eval_metrics = asyncio.run(self.eval())
-                    self.tracker.log(eval_metrics, step=self.global_step)
+                    self.tracker.log(eval_metrics, step=self.global_step, commit=True)
```
SumanthRH, line 256: This is the behaviour after this PR:

1. start from scratch: when `eval_before_train=True`, we log evaluation metrics at step 0. all training metrics will appear as step 1 before the next model update - this is because we start from step 1: 
https://github.com/NovaSky-AI/SkyRL/blob/1edc0e3db9568e1c051ef682659157efd9783f2c/skyrl-train/skyrl_train/trainer.py#L270

If `eval_interval=1`, then we evaluate after the policy model has been updated in `train_critic_and_policy`, but in the inference engine, we are still using the old weights - https://github.com/NovaSky-AI/SkyRL/blob/1edc0e3db9568e1c051ef682659157efd9783f2c/skyrl-train/skyrl_train/trainer.py#L231-L237 . This evaluation result will appear as a metric for step 1 now. 
2. resume from step k: Pretty similar - when `eval_before_train=True`, we log evaluation metrics at step k before the training step. all training metrics will appear as step k+1. If we evaluate at this step now based on `eval_interval`, the evaluation uses the old weights (so same weights as before training step) but this will appear as the result for step k+1. 


This seems a bit incorrect? Do we just not `commit` here so that the evaluation metrics and training metrics will always appear as step k+1 (when training starts at step k)?

SumanthRH: After offline discussion with @tyler-griggs , we realized this behavoiur is actually the same as existing behavoiur, PR doesn't affect it. Marking as resolved!

skyagent/examples/run_verl/verl_trainer.py
```diff
@@ -357,7 +357,7 @@ def fit(self):
         logger = Tracking(
             project_name=self.config.trainer.project_name,
             experiment_name=self.config.trainer.experiment_name,
-            default_backend=self.config.trainer.logger,
+            backends=self.config.trainer.logger,
```
SumanthRH: @tyler-griggs can you not change this file? it's modifying the VERL backend in SkyAgent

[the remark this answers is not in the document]
SumanthRH: no longer relevant. this change is reverted

Issue #292, opened 2025-09-12
Ray worker OOM


EtashGuha:
https://gist.github.com/EtashGuha/ba51fcc02e8ae21f645d471a6c639829
This is Gist detailing my OOM's.

My Config: 
data.cache_dir: /scratch/08134/negin/dc-agent-shared/data_cache
data.train_data:
- /scratch/08134/negin/dc-agent-shared/data_cache/datasets--mlfoundations-dev--sandboxes-tasks/snapshots/6fdf67053a80836ab1d5007104baded9f3513733
data.val_data:
- /scratch/08134/negin/dc-agent-shared/data_cache/datasets--mlfoundations-dev--sandboxes-tasks-hello-world/snapshots/d8ea629733b7e4ba8471196d71a2ccf6a966ef60
generator.agent_name: terminus
generator.async_engine: true
generator.backend: vllm
generator.batched: true
generator.gpu_memory_utilization: 0.8
generator.http_server_inference_engine_client_host: 127.0.0.1
generator.http_server_inference_engine_client_port: 8000
generator.inference_engine_tensor_parallel_size: 1
generator.n_samples_per_prompt: 4
generator.num_inference_engines: 4
generator.run_engines_locally: true
generator.sampling_params.max_generate_length: 16000
generator.sandboxes_dir: /scratch/08134/negin/dc-agent-shared/sandboxes
generator.trial_runs_dir: /scratch/08134/negin/dc-agent-shared/sandboxes/run
generator.use_http_server_inference_engine_client: true
generator.weight_sync_backend: nccl
trainer.algorithm.advantage_estimator: grpo
trainer.algorithm.use_kl_loss: true
trainer.ckpt_interval: 10
trainer.ckpt_path: /scratch/08134/negin/dc-agent-shared/sandboxes/exports/checkpoints/tacc_sandboxes_rl_full_v2
trainer.epochs: 5
trainer.eval_batch_size: 1
trainer.eval_interval: 5
trainer.export_path: /scratch/08134/negin/dc-agent-shared/sandboxes/exports/tacc_sandboxes_rl_full_v2
trainer.hf_save_interval: 1000000
trainer.logger: wandb
trainer.max_prompt_length: 16000
trainer.micro_forward_batch_size_per_gpu: 1
trainer.micro_train_batch_size_per_gpu: 1
trainer.placement.colocate_all: true
trainer.placement.critic_num_gpus_per_node: 1
trainer.placement.critic_num_nodes: 4
[...]

PR #297, opened 2025-09-15
[trainer] Discard inference engine weights after sleeping

erictang000:
# Overview
Uses sleep level 2 (which discards inference engine weights when sleeping) for VLLM. SGLang previously had this behavior by default.

Changes the `InferenceWeightsManager` to remove the `no_sync` flag, since we always have to sync before eval now due to no longer saving the weights on CPU. Adds a `sleep_on_exit` flag, which toggles whether or not to sleep on exiting the `InferenceWeightsManager` context. For the eval weights manager, we set `sleep_on_exit` to `false`, to avoid an unnecessary sleep/wake_up/sync cycle prior to the next batch of training rollouts.

This also helps avoid some unnecessary CPU memory storage of the inference engine weights, which should help partially address #292. 

A summary diagram of improved memory management and weight syncing behavior is shown below.

Before:
image

After:
image

### DIFF
--- a/skyrl-train/skyrl_train/trainer.py
+++ b/skyrl-train/skyrl_train/trainer.py
@@ -224,16 +224,16 @@ class RayPPOTrainer:
 
         self.global_step = 0
         self.weights_manager = InferenceWeightsManager(
-            self.policy_model, self.inference_engine_client, self.cfg.trainer.placement.colocate_all
+            self.policy_model,
+            self.inference_engine_client,
+            self.cfg.trainer.placement.colocate_all,
+            sleep_on_exit=True,
         )
-        # NOTE(Charlie): sglang's engine needs to sync weights after wake up. see https://github.com/sgl-project/sglang/issues/7939
-        # Change it to True after sglang fixes the issue.
-        sync_before_eval = self.cfg.trainer.placement.colocate_all and self.cfg.generator.backend == "sglang"
         self.eval_weights_manager = InferenceWeightsManager(
             self.policy_model,
             self.inference_engine_client,
             self.cfg.trainer.placement.colocate_all,
-            no_sync=not sync_before_eval,
+            sleep_on_exit=False,
         )
 
         # Load checkpoint state if resumption is enabled
@@ -242,24 +242,20 @@ class RayPPOTrainer:
                 self.load_checkpoints()
                 logger.info(f"Resumed training from global_step {self.global_step}")
 
-        # create rank0 policy model and inference_engines groups, then broadcast weights to inference_engines
+        # create rank0 policy model and inference_engines groups
         with Timer("setup_policy_and_generator"):
             self.setup_policy_and_generator()
-            if self.cfg.trainer.placement.colocate_all:
-                self.policy_model.backload_to_gpu()
 
         # Eval before training
+        inference_engine_is_active = False
         if self.cfg.trainer.eval_interval > 0 and self.cfg.trainer.eval_before_train:
+            if self.cfg.trainer.placement.colocate_all:
+                self.policy_model.backload_to_gpu()
             with self.eval_weights_manager:
                 with Timer("eval", self.all_timings):
                     eval_metrics = asyncio.run(self.eval())
                     self.tracker.log(eval_metrics, step=self.global_step)
-            # Policy model is backloaded to GPU after eval
-            if self.cfg.trainer.placement.colocate_all:
-                self.policy_model.backload_to_gpu()
-
-        # setup for dynamic sampling
-        keep_sampling = False
+            inference_engine_is_active = True
 
         # initialize kl controller
         if self.cfg.trainer.algorithm.use_kl_in_reward:
@@ -278,8 +274,10 @@ class RayPPOTrainer:
                         self.cfg.generator.n_samples_per_prompt, rand_prompts, self.cfg.generator.sampling_params
                     )
 
-                    # if we are continuing sampling, we don't want to trigger weight management
-                    weights_manager = ConditionalWeightsManager(self.weights_manager, not keep_sampling)
+                    # if we inference engine is already active due to continuing sampling or eval, we don't want to trigger weight management
+                    weights_manager = ConditionalWeightsManager(
+                        self.weights_manager, condition=not inference_engine_is_active
+                    )
 
                     # NOTE: Policy model is on GPU at the beginning of each training step
                     # After exiting the context manager, policy model is on CPU with `colocate_all` enabled.
@@ -293,11 +291,15 @@ class RayPPOTrainer:
                         if self.cfg.trainer.algorithm.dynamic_sampling.type is not None:
                             generator_output, uids, keep_sampling = self.handle_dynamic_sampling(generator_output, uids)
                             # update weights manager condition to ensure we trigger sleep only when we are not continuing sampling
-                            weights_manager.update_condition(not keep_sampling)
+                            weights_manager.update_condition(condition=not keep_sampling)
+                            inference_engine_is_active = keep_sampling
                             if keep_sampling:  # continue sampling
                                 # update progress bar for current batch (but not global step)
                                 pbar.update(1)
                                 continue
+
+                    # if we exit the generation context manager, then the inference engine is asleep
+                    inference_engine_is_active = False
 
                     # 1.2 postprocess rewards
                     with Timer("postprocess_generator_output", self.all_timings):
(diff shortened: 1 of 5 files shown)


skyrl-train/skyrl_train/trainer.py
SumanthRH, line 1: QQ: What happens with `update_ref_every_epoch` at the end of the epoch now?

Until now, the convention was that at the end of the training step, the policy model is backloaded to GPU. This is the assumption with which the current `update_ref_with_policy` function is written. 

There could be issues here now because the inference engine can be on GPU with policy model offloaded if `self.global_step % eval_interval == 0` and then we update the ref model after this step. For ex: we end up calling `policy_model.offload_to_cpu` before updating ref model and also finally call a `policy_model.backload_to_gpu` at the end of `update_ref_with_policy`

https://github.com/NovaSky-AI/SkyRL/blob/e5e88bdc4504930e0a3d8d38be35dfdcf036367c/skyrl-train/skyrl_train/trainer.py#L1419-L1425

skyrl-train/skyrl_train/trainer.py
```diff
@@ -278,8 +274,10 @@ def train(self):
...
 
-                    # if we are continuing sampling, we don't want to trigger weight management
-                    weights_manager = ConditionalWeightsManager(self.weights_manager, not keep_sampling)
+                    # if the inference engine is already active due to continuing sampling or eval, we don't want to trigger weight management
+                    weights_manager = ConditionalWeightsManager(
+                        self.weights_manager, condition=not inference_engine_is_active
+                    )
 
                     # NOTE: Policy model is on GPU at the beginning of each training step
                     # After exiting the context manager, policy model is on CPU with `colocate_all` enabled.
```
SumanthRH: This comment is inaccurate now? 

It looks like we don't have such a fixed condition anymore in the training step - i.e policy model may or may not be on GPU at the beginning depending on whether `eval` ran before that step

skyrl-train/skyrl_train/trainer.py
erictang000: good catch

skyrl-train/skyrl_train/trainer.py
SumanthRH: actually why can't we have a redundant offloading after eval so that there is sucha fixed condition? Would be easier to reason about the trainig loop and modify

skyrl-train/skyrl_train/trainer.py
SumanthRH, line 1: let's update the async trainer as well?

skyrl-train/skyrl_train/trainer.py
erictang000: redundant offloading includes weight sync though, which can be non-trivial

erictang000: especially if we do eval before train, i think this is bad user experience on step 0 - you have to sync, eval, sync, train

instead of just sync, eval, train

erictang000: @CharlieFRuan thoughts?

erictang000: i think we could consider killing the inferenceweightsmanager abstraction - I think it's clunky, and removing it could actually simplify the code

looking at slime's training loop: https://github.com/THUDM/slime/blob/1785a8d43da88c3e758f111fe7ae42d4304f6191/train.py#L49

they just do
```python
for i in range(n):
  rollout 
  offload rollout
  train
  onload rollout weights
  sync weights
  eval
```

which uses the convention where the inference engine is on gpu at the beginning of the loop

we previously had
```python
for i in range(n):
  sync weights
  rollout 
  offload rollout
  train
  offload train # this was missing a sync weights in our impl actually!
  eval
```

this PR modifies it to look like the slime training loop for the batch where we run eval - I think this is more natural if you want to run evals every n steps, because you actually want the ordering of `train -> sync -> eval`, whereas before we were missing a sync before eval

skyrl-train/skyrl_train/trainer.py
erictang000: updated the weights manager class to add back no_sync to avoid unnecessary syncing for the async trainer evals, but actually needed no other changes for async trainer! (since sleeping/wake up is all for colocation)
image

skyrl-train/skyrl_train/trainer.py
SumanthRH: > especially if we do eval before train, i think this is bad user experience on step 0 - you have to sync, eval, sync, train

Hmm yeah agreed we don't want this 

But we should be able to have a convention for what is on gpu at the end of training step

SumanthRH: > i think we could consider killing the inferenceweightsmanager abstraction - I think it's clunky, and removing it could actually simplify the code

Agreed with all the conditional checks etc it is not worth it...the context manager was good when it was just doing one thing at a time, but with additional args it is pretty unclear what is happening at enter/exit

let's do this in a follow-up PR

skyrl-train/skyrl_train/weights_manager.py
```diff
@@ -32,19 +32,21 @@ class InferenceWeightsManager:
     This class is used to synchronize the weights of the policy model to the InferenceEngines.
     It also wakes up the inference engine if `colocate_all` is enabled.
 
-    If `no_sync` is enabled, the weights will not be synchronized, but offloading/backloading will still happen.
+    Optionally puts the inference engine to sleep on exit
```
SumanthRH: So now we have `no_sync` and `sleep_on_exit`?

erictang000: yeah lol... that's what made me think the `InferenceWeightsManager` abstraction is probably not the right one

erictang000: leaving the comment deletion, since we always do sync for the colocated case

SumanthRH: let the comment exist as long as the flag exists :)

skyrl-train/skyrl_train/trainer.py
```diff
@@ -242,24 +242,20 @@ def train(self):
                 self.load_checkpoints()
                 logger.info(f"Resumed training from global_step {self.global_step}")
 
-        # create rank0 policy model and inference_engines groups, then broadcast weights to inference_engines
+        # create rank0 policy model and inference_engines groups
```
SumanthRH, line 245: QQ: At init - and at the beginning of the `train` function, everything is on CPU for the colocated case

https://github.com/NovaSky-AI/SkyRL/blob/7406b96a3aa29ba9fe44a3679effcbd9c2c1f909/skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py#L221-L223

we call sleep here, and then we offload all models to CPU in colocated case at the end of `trainer.build_models`. 

But with the changes in this PR, `.sleep` will end up discarding engine weights , so the CPU memory usage reduction will happen even at initialization

SumanthRH (commented):
overall looks fine, just need to handle the `update_ref_every_epoch` case!

skyrl-train/skyrl_train/trainer.py
erictang000: yeah that's right

skyrl-train/skyrl_train/weights_manager.py
erictang000: the comment is actually misleading b/c the only case where it's used there is no offloading or backloading...

SumanthRH:
```suggestion
    If `no_sync` is enabled, the weights will not be synchronized.
    Optionally puts the inference engine to sleep on exit if `sleep_on_exit` is `True`
```

SumanthRH: Ok I updated the comment, and removed the part that could be misleading.

skyrl-train/skyrl_train/trainer.py
erictang000: fixed!

SumanthRH, line 1: Have you run to completion with the new changes? 

It looks like there is a case where `inference_engine_is_active=True` at the end of the training loop (i.e policy model on CPU) and then when we save the final training checkpoints and weights:

https://github.com/NovaSky-AI/SkyRL/blob/46ad0f587f614553bf5091553d4f1f980442477e/skyrl-train/skyrl_train/trainer.py#L394-L401

we can have an error because the policy model is not on the GPU

SumanthRH: Could we just ensure that at the end of an epoch, we check if `inference_engine_is_active` and if so, we backload policy model to GPU?

SumanthRH: That way we support both the `update_ref_every_epoch` case and this particular case?

SumanthRH (changes requested):
Seems like there can be a case of ckpting failure at the end of training if we run eval at the final step?

skyrl-train/skyrl_train/trainer.py
erictang000: good point, done!

erictang000: i checked that the ref update works, but yeah didn't run to completion

skyrl-train/skyrl_train/trainer.py
```diff
@@ -382,12 +382,12 @@ def train(self):
 
                 del training_input, generator_output
 
+            if inference_engine_is_active and self.cfg.trainer.placement.colocate_all:
```
SumanthRH, line 386:
```suggestion
            # Ensure that the policy model is on GPU at the end of every epoch
            if inference_engine_is_active and self.cfg.trainer.placement.colocate_all:
```

PR #299, opened 2025-09-16
[SkyRL-Gym] Make SQL and Search envs return None for intermediate steps

erictang000:
# Overview
After #226, the SkyRLGymGenerator expects turn level rewards to be None if the env uses trajectory level rewards. After #271, this causes issues computing metrics for SQL and Search envs. Setting intermediate reward to None and making it an optional type fixes this.

### DIFF
--- a/skyrl-gym/skyrl_gym/envs/base_text_env.py
+++ b/skyrl-gym/skyrl_gym/envs/base_text_env.py
@@ -8,7 +8,7 @@ ConversationType = List[MessageType]
 
 class BaseTextEnvStepOutput(TypedDict):
     observations: ConversationType  # OpenAI API Messages Format
-    reward: float
+    reward: Optional[float]  # None if intermediate steps have no reward
     done: bool
     metadata: Dict[str, Any]
     postprocessed_action: Optional[str] = None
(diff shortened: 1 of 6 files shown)


CharlieFRuan:
As discussed offline, will revert https://github.com/NovaSky-AI/SkyRL/pull/271 first to avoid the error that Eric is observing from happening.

The error is due to this line:
https://github.com/NovaSky-AI/SkyRL/blob/2e7aba951a6577ceccd21540648c7481c295ad62/skyrl-train/skyrl_train/generators/utils.py#L65

Where it was that `sum(seq_rewards)` doesn't work bc `seq_rewards` was a float, but rewards[0] is a list, which means some were lists and some were floats. This is because intermediate reward for sql and search are 0, so some trajectory's reward is [0, 1], and some trajectory's reward can be [1]. The first goes into else, the second goes into if:

https://github.com/NovaSky-AI/SkyRL/blob/2e7aba951a6577ceccd21540648c7481c295ad62/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py#L292-L298

Reverting #271 will prevent the error at the cost of not seeing pass_at_n metric. Which we will follow up with a fix.

erictang000:
closing and tracking actual fix in #311

Issue #279, opened 2025-09-10
[SkyRLGymGenerator] Cleaner generation length handling


tyler-griggs:
In SkyRLGymGenerator, we post-process the trajectories to truncate them to the configured maximum trajectory length ([code link](https://github.com/NovaSky-AI/SkyRL/blob/cbd0680d7572f3045f68e0de68d78eb1f947364d/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py#L285-L286)). This is a bit messy.

Instead, we should just update the sampling parameter's `max_generate_length` parameter to be something like `max generate length = min(max model len - len(chat_history), max_generate_length)`. Basically, we truncate the `max_generate_length` if it exceeds the maximum context length. Then, we can remove the messy post-processing.

PR #304, opened 2025-09-16
Improved max generation length for trajectories

devpatelio:
Addresses #279

### DIFF
--- a/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
+++ b/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
@@ -183,27 +183,23 @@ class SkyRLGymGenerator(GeneratorInterface):
         # Accumulate per-step rewards. Format: (reward, response_end_token_idx)
         per_step_rewards: List[Tuple[Optional[float], Optional[int]]] = []
         
-        cumulative_response_tokens = 0
     
         while not done:
-            max_generate_length = min(max_generate_length, max_input_length + max_tokens - len(input_ids))
-            if max_generate_length <= 0:
+
+            if len(input_ids) > max_input_length:
                 stop_reason = "length"
                 break
-                
-            updated_sampling_params = sampling_params.copy() if sampling_params is not None else {}
-            updated_sampling_params["max_generate_length"] = max_generate_length
-
+            
             if retokenize_chat_history:
                 engine_input = InferenceEngineInput(
-                    prompts=[chat_history], trajectory_ids=[trajectory_id], sampling_params=updated_sampling_params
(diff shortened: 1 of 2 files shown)


skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -175,16 +174,23 @@ async def agent_loop(
         per_step_rewards: List[Tuple[Optional[float], Optional[int]]] = []
 
         while not done:
-            # 1. Generate output
+            updated_sampling_params = sampling_params.copy() if sampling_params is not None else {}
+            updated_sampling_params["max_generate_length"] = min(
+                max_tokens, max_tokens + max_input_length - len(input_ids)
+            )
+            max_response_tokens = updated_sampling_params["max_generate_length"]
```
tyler-griggs: Why give `max_response_tokens` a value, when it will be overwritten below before use?

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -267,23 +273,22 @@ async def agent_loop(
             per_step_rewards = [(reward, idx - initial_prompt_length) for reward, idx in per_step_rewards]
         assert len(loss_mask) == len(response_ids), "loss_mask and response_ids should have the same length"
 
+        if self.max_turns > 1:
```
tyler-griggs: The goal of this PR was to remove this code entirely. Ideally, we don't need to recompute the `max_response_tokens` and then truncate the response_ids and loss_mask below. Ideally, they are already guaranteed to be within the `max_response_tokens` since we updated the sampling params.

Please test that this is the case (both manually, and with an actual test).

skyrl-train/tests/gpu/gpu_ci/test_policy_local_engines_e2e.py
tyler-griggs, line 1: Can we move these test migrations into a different PR?

skyrl-train/tests/gpu/test_training_step.py
```diff
@@ -1,6 +1,8 @@
 """
 Run with:
-uv run --isolated --extra dev -- pytest tests/gpu/test_training_step.py
+uv run --isolated --extra dev --extra deepspeed pytest tests/gpu/gpu_ci/test_training_step.py
+
+Ensure you set WANDB_API_KEY in your environment.
```
tyler-griggs, line 5: Please update the test to use `trainer.logger="console"` instead of requiring `WANDB_API_KEY`. We don't want to use wandb in the CI pipeline.

Issue #311, opened 2025-09-17
Pass at N not computed for agent loop code path with SkyRL-Gym


erictang000:
Pass at N is not computed for the multi turn agent loop code path after #226 added support for token level rewards 

https://github.com/NovaSky-AI/SkyRL/blob/e9c37b8063649ce3c48776a44716333415c79bb4/skyrl-train/skyrl_train/generators/utils.py#L66

Per offline discussion, we should update this to always treat rewards[i][-1] as the per-trajectory final reward, denoting whether it "passed" or not.

PR #317, opened 2025-09-18
[Metrics] Add back pass_at_n computation for (per-token) rewards

CharlieFRuan:
Fixes https://github.com/NovaSky-AI/SkyRL/issues/311.

### PRs around this issue
- `pass_at_n` no longer computed for multi-turn rollouts after https://github.com/NovaSky-AI/SkyRL/pull/226
- This PR fixed it by introducing `None` reward, which is ill-defined and later reverted: https://github.com/NovaSky-AI/SkyRL/pull/271

### This PR
- Deeming the last turn's reward as the entire trajectory's reward (and being > 0 signifies a "pass") for the purpose of computing `pass@N`
- Adding documentation about (per-turn) rewards, metrics, and per-token rewards conversion (for better intuition) in `Creating a New Environment or Task` for lack of a better place to put it at
- Add unit test and more documentation to the metric util
- Remove the `Optional[float]` annotation in `skyrl_gym_generator.py`, since our `BaseTextEnvStepOutput.reward` is `float` and not `Optional[float]`. Also added corresponding documentation, saying returning `0.0` as reward for intermediate turns if not using turn-level reward
- Also, added a minor fix to pass_at_n computation, where negative reward is taken into account. See this comment for more: https://github.com/NovaSky-AI/SkyRL/pull/317#discussion_r2357824428

### Test
- Ran `run_gsm8k.sh` with:
  - orange: `batched=True` (previously working already, since it was not the `agent_loop()` codepath that does not convert to per-token rewards)
  - green: `batched=False` (the codepath where `pass_at_n` is not computed prior to this PR)
  - grey: baseline from a previous stable PR's run
image

### Rendered doc
image

### DIFF
--- a/skyrl-train/docs/tutorials/new_env.rst
+++ b/skyrl-train/docs/tutorials/new_env.rst
@@ -147,7 +147,7 @@ The final implementation is available in `examples/multiply/env.py <https://gith
 --------------------------------
 
 In the example above, unless ``done=True``, the reward is ``0.0``. That is, the model only receives a single reward for the entire trajectory.
-You can experiment with turn-level rewards by returning a non-zero in any turn. Otherwise, if you only want to use outcome rewards, you can simply return ``reward=0.0`` for all intermediate turns.
+You can experiment with turn-level rewards by returning a non-zero reward in any turn. Otherwise, if you only want to use outcome rewards, you can simply return ``reward=0.0`` for all intermediate turns.
 
 SkyRL automatically computes the following metrics for logging purposes:
 


skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -172,7 +172,7 @@ async def agent_loop(
         loss_mask = []  # this excludes the prompt
         rollout_logprobs = None
         # Accumulate per-step rewards. Format: (reward, response_end_token_idx)
-        per_step_rewards: List[Tuple[Optional[float], Optional[int]]] = []
+        per_step_rewards: List[Tuple[float, Optional[int]]] = []
```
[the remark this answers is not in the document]
CharlieFRuan: Removed.

skyrl-train/skyrl_train/generators/utils.py
```diff
@@ -54,25 +54,44 @@ def get_metrics_from_generator_output(
...
+
+    # TODO: We should make metrics customizable by the environment
+    # map from the example's uid to each trajectory's reward on that same example
+    uid_to_trajectory_rewards = defaultdict(list)
+    if is_per_token_rewards:
+        # For each trajectory, we sum over the token rewards for `mean_raw_reward` computation
+        mean_raw_reward = float(np.mean([sum(trajectory_rewards) for trajectory_rewards in rewards]))
+        # Assume the last token's reward signifies the trajectory's reward for `pass_at_n` computation
+        for i, cur_trajectory_rewards in enumerate(rewards):
+            uid_to_trajectory_rewards[uids[i]].append(cur_trajectory_rewards[-1])
```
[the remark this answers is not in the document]
CharlieFRuan: Raised error instead.

skyrl-train/skyrl_train/generators/utils.py
```diff
@@ -49,30 +49,49 @@ def get_generation_prompt_ids(tokenizer) -> List[int]:
...
 
+    # TODO: We should make metrics customizable by the environment
+    # map from the example's uid to each trajectory's reward on that same example
+    uid_to_trajectory_rewards = defaultdict(list)
     if isinstance(rewards[0], list):
-        # We just compute mean over sequence reward.
-        # TODO: We should make metrics customizable by the environment
-        mean_raw_reward = float(np.mean([sum(seq_rewards) for seq_rewards in rewards]))
-        pass_at_n = None  # not computed for token-level rewards since it's ill-defined
+        assert all(
```
erictang000: nit: can we group this assertion here in `validate_generator_output`? https://github.com/NovaSky-AI/SkyRL/blob/f0015a705af0af82a0012538ff1b62c3b158cbea/skyrl-train/skyrl_train/utils/trainer_utils.py#L579

just to have things in one place

skyrl-train/skyrl_train/generators/utils.py
```diff
@@ -49,30 +49,49 @@ def get_generation_prompt_ids(tokenizer) -> List[int]:
...
+            isinstance(reward, list) for reward in rewards
+        ), "rewards must be List[float], or List[List[float]], but received a mixture of List[float] and float"
+        # Token-level rewards: rewards is List[List[float]]
+        # For each trajectory, we sum over the token rewards for `mean_raw_reward` computation
+        mean_raw_reward = float(np.mean([sum(trajectory_rewards) for trajectory_rewards in rewards]))
+        # Assume the last token's reward signifies the trajectory's reward for `pass_at_n` computation
+        for i, cur_trajectory_rewards in enumerate(rewards):
+            if len(cur_trajectory_rewards) == 0:
+                raise ValueError("Token-level rewards must be a non-empty list.")
+            uid_to_trajectory_rewards[uids[i]].append(cur_trajectory_rewards[-1])
```
erictang000, line 77: btw i'm testing this for search with Qwen3-30B right now, and it appears the actual reward is in `cur_trajectory_rewards[-2]` rather than [-1] when `use_conversation_multi_turn=false` 

this is because of appending the eos here

 https://github.com/NovaSky-AI/SkyRL/blob/f0015a705af0af82a0012538ff1b62c3b158cbea/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py#L275

maybe the correct change is to leave this as is here, but update the logic in the generator to make sure that the last reward value is always at the end...

CharlieFRuan: Wow, nice catch! Thanks!

Yes agreed, we should move that last turn's reward to the manually appended eos token. This is consistent with when we did not have turn-level reward:

https://github.com/NovaSky-AI/SkyRL/blob/f0015a705af0af82a0012538ff1b62c3b158cbea/skyrl-train/skyrl_train/trainer.py#L804-L809

cc @tyler-griggs for cross-checking. I changed your unit test `test_agent_loop_token_level_rewards_multi_turn()` as a result

skyrl-train/skyrl_train/generators/utils.py
CharlieFRuan: Good point! Addressed.

skyrl-train/docs/tutorials/new_env.rst
```diff
@@ -143,6 +143,40 @@ The multi-turn version gives partial credit for formatting the answer correctly,
...
+
+When computing metrics for logging purposes:
+
+- ``pass_at_n``: for each trajectory, we assume that the last turn's reward signifies the entire trajectory's reward, and being positive signifies a "pass".
+- ``mean_raw_reward``: for each trajectory, we sum over all the turns' rewards. We then take the average over all the trajectories.
+
+If you do not want to provide turn-level rewards, you can simply return ``0.0`` for all intermediate turns.
+
+Whether you provide turn-level rewards or per-trajectory rewards, the final rewards used to train the model will be translated to per-token rewards.
+Say we have 3 turns and each turn has 4 response tokens. If the turn-level rewards are ``[1.0, 2.0, 3.0]``, then the per-token rewards will be:
```
tyler-griggs:
```suggestion
In the example above, unless ``done=True``, the reward is ``0.0``. That is, the model only receives a single reward for the entire trajectory.
You can experiment with turn-level rewards by returning a non-zero in any turn. Otherwise, if you only want to use outcome rewards, you can simply return ``reward=0.0`` for all intermediate turns.

SkyRL automatically computes the following metrics for logging purposes:

- ``pass_at_n``: for each trajectory, we assume that the last turn's reward signifies the entire trajectory's reward, and any positive value is considered a "pass".
- ``mean_raw_reward``: for each trajectory, we sum over all the turns' rewards. We then take the average over all the trajectories.

Whether you use turn-level rewards or outcome rewards, the rewards used to train the model will be translated to per-token rewards. For example, if there are 3 turns with 4 response tokens each and the turn-level rewards are ``[1.0, 2.0, 3.0]``, the resulting per-token rewards are:
```

skyrl-train/docs/tutorials/new_env.rst
```diff
@@ -143,6 +143,40 @@ The multi-turn version gives partial credit for formatting the answer correctly,
...
+
+.. code-block:: python
+
+   [
+      0.0, 0.0, 0.0, 1.0,
+      0.0, 0.0, 0.0, 2.0,
+      0.0, 0.0, 0.0, 3.0,
+   ]
+
+If we only have a per-trajectory reward of ``1.0`` (i.e. intermediate turns' rewards are all ``0.0``), the per-token rewards will be:
```
tyler-griggs:
```suggestion
If there is only an outcome reward of ``1.0`` (i.e. intermediate turns' rewards are all ``0.0``), the per-token rewards are:
```

skyrl-train/docs/tutorials/new_env.rst
tyler-griggs: nitpicky edits :)

skyrl-train/skyrl_train/utils/trainer_utils.py
```diff
@@ -622,3 +622,14 @@ def validate_generator_output(input_batch: GeneratorInput, generator_output: Gen
     # loss masks should be non-zero for at least one element for trainer
     if np.concatenate(generator_output["loss_masks"]).sum() == 0:
         logger.warning("All outputs are loss masked, which may lead to NaN loss, please check your generation logic!!")
+
+    # check that ther rewards is either List[float-like] or List[List[float-like]]
```
tyler-griggs:
```suggestion
    # check that the rewards are either List[float-like] or List[List[float-like]]
```

skyrl-train/skyrl_train/utils/trainer_utils.py
```diff
@@ -622,3 +622,14 @@ def validate_generator_output(input_batch: GeneratorInput, generator_output: Gen
...
+    rewards = generator_output["rewards"]
+    if isinstance(rewards[0], list):
+        assert all(
+            isinstance(reward, list) for reward in rewards
+        ), "rewards must be List[float], or List[List[float]], but received a mixture of List[float] and float"
```
tyler-griggs: Regarding these two comments, I guess we don't actually know that a mixture of list[float] or floats were passed in, we just know that some were list and some were not a list. Someone really could have messed up and passed a dict in :D Should we just truncate the error message to:
```
rewards must be List[float], or List[List[float]]
```

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -287,16 +289,25 @@ async def agent_loop(
 
         # Build reward output
         if retokenize_chat_history:
+            # TODO(Charlie): Currently, the possible response truncation will not affect the reward
+            # in if branch, but some final rewards may be lost in the else branch. Fix this
```
tyler-griggs:
```suggestion
            # in the if branch, but some final rewards may be lost in the else branch. Fix this
```

skyrl-train/skyrl_train/generators/utils.py
```diff
@@ -49,30 +49,43 @@ def get_generation_prompt_ids(tokenizer) -> List[int]:
...
+    for each example.
+
+    Rewards can be either per-trajectory or per-token, and metrics are computed correspondingly.
     """
     rewards: Union[List[float], List[List[float]]] = generator_output["rewards"]
     if not len(rewards):
         raise ValueError(f"`rewards` must be a non-empty list, got {rewards}")
 
+    # TODO: We should make metrics customizable by the environment
+    # map from the example's uid to each trajectory's reward on that same example
```
tyler-griggs:
```suggestion
    # TODO: We should make metrics customizable by the environment.
    # Map from the example's uid to each trajectory's reward on that same example
```

tyler-griggs: It read as if this was a two-line sentence

skyrl-train/skyrl_train/generators/utils.py
```diff
@@ -49,30 +49,43 @@ def get_generation_prompt_ids(tokenizer) -> List[int]:
...
 
 @torch.no_grad()
-def get_metrics_from_generator_output(
-    generator_output: GeneratorOutput, uids: List[str]
-) -> Tuple[float, Optional[float]]:
+def get_metrics_from_generator_output(generator_output: GeneratorOutput, uids: List[str]) -> Tuple[float, float]:
     """
     Get `mean_raw_reward` (or avg_score), `pass_at_n` from generator output.
+
+    m: `len(uids)` is the number of unique examples.
```
tyler-griggs: Why define `m`?

skyrl-train/skyrl_train/generators/utils.py
```diff
@@ -49,30 +49,43 @@ def get_generation_prompt_ids(tokenizer) -> List[int]:
...
+
+    n: `len(generator_output["rewards"]) / len(uids)` is the number of trajectories we generate
```
tyler-griggs: This kinda reads like `n` is an argument.  You can probably just merge this comment with the one above the `pass_at_n` calculation to explain what `n` is referring to

tyler-griggs (commented):
thanks!

skyrl-train/skyrl_train/utils/trainer_utils.py
CharlieFRuan: Ah fair, I'll fix. Was kind of hard to check for `float` too because it could be int, np.float, etc.

Followed your suggestion here.

skyrl-train/skyrl_train/generators/utils.py
CharlieFRuan: Good point, addressed!

CharlieFRuan:
@tyler-griggs All addressed, appreciate the detailed review!

skyrl-train/docs/tutorials/new_env.rst
```diff
@@ -143,6 +143,37 @@ The multi-turn version gives partial credit for formatting the answer correctly,
 
 The final implementation is available in `examples/multiply/env.py <https://github.com/NovaSky-AI/SkyRL/blob/main/skyrl-train/examples/multiply/env.py>`_.
 
+(Turn-level) Rewards And Metrics
+--------------------------------
+
+In the example above, unless ``done=True``, the reward is ``0.0``. That is, the model only receives a single reward for the entire trajectory.
+You can experiment with turn-level rewards by returning a non-zero in any turn. Otherwise, if you only want to use outcome rewards, you can simply return ``reward=0.0`` for all intermediate turns.
```
tyler-griggs:
```suggestion
You can experiment with turn-level rewards by returning a non-zero reward in any turn. Otherwise, if you only want to use outcome rewards, you can simply return ``reward=0.0`` for all intermediate turns.
```

tyler-griggs: Oops, I must have dropped "reward" in my prior suggestion

skyrl-train/docs/tutorials/new_env.rst
```diff
@@ -143,6 +143,37 @@ The multi-turn version gives partial credit for formatting the answer correctly,
...
+
+SkyRL automatically computes the following metrics for logging purposes:
+
+- ``pass_at_n``: for each trajectory, we assume that the last turn's reward signifies the entire trajectory's reward, and any positive value is considered a "pass".
```
tyler-griggs:
```suggestion
- ``pass_at_n``: The ``n`` in ``pass_at_n`` is the number of trajectories we generate for each example. ``pass_at_n`` is 1 if any trajectory succeeded, and 0 otherwise. For each trajectory, we assume that the last turn's reward signifies the entire trajectory's reward, and any positive value is considered a "pass".
```

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -287,16 +289,25 @@ async def agent_loop(
...
                     break
-                token_level_rewards[idx] += step_reward
+                if appended_eos_token and i == len(per_step_rewards) - 1:
+                    # NOTE(Charlie): If we appended the eos token, we need to place
+                    # the reward at the last token (the manually appended eos token)
+                    # rather than the last turn's assistant-generated token. This matches
+                    # the logic in trainer.py::postprocess_generator_output when rewards are List[float].
+                    token_level_rewards[-1] = step_reward
+                else:
+                    token_level_rewards[idx] += step_reward
```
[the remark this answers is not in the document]
CharlieFRuan: if idx >= len(respnose_ids), the appended eos token will be out of bound as well. So the reward is supposed to be lost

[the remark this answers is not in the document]
CharlieFRuan: However it is possible that:
- `max_response_tokens` is say 4
- we have response_ids as [1, 1, 1, 1, eos]
- we have step reward (step_reward=2, idx=3)
- We will truncate the response ids to `[1, 1, 1, 1]`
- But the code logic we have here will assign the reward 2 to the last token 1, rather than truncating the reward
- Not sure what should be the correct behavior though

i.e. `token_level_rewards[-1]` is the same as `token_level_rewards[idx]` due to truncation (the only truncated token is the eos

PR #322, opened 2025-09-18
Plumb `engine_init_kwargs` from CLI to inference engine

### LINKED ISSUE #318: Allow passing vllm engine kwargs during inference engine init
Can we have a way to pass some vllm engine arguments to SkyRL train's vllm inference engine? Arguments like  `--tool-call-parser` and `--enable-auto-tool-choice` would help use tools during rollout generation, which is essential to train tool-use and coding agent models. 

```bash
generator.vllm_engine_kwargs.enable_auto_tool_choice=true \
generator.vllm_engine_kwargs.tool_call_parser="hermes"
```

tyler-griggs:
Resolves #318 

## What does this PR do?
Adds support for specifying inference engine args in the training config and CLI override arguments, which are passed directly to the vLLM or SGLang Engine init. For example, to specify vLLM's tool params, you can add the following to the training script (note the `+`, which is required):
```
+generator.engine_init_kwargs.enable_auto_tool_choice=true \
+generator.engine_init_kwargs.tool_call_parser="hermes"
```

### DIFF
--- a/skyrl-train/docs/configuration/config.rst
+++ b/skyrl-train/docs/configuration/config.rst
@@ -532,7 +532,7 @@ Generation Parameters
 - ``generator.eval_n_samples_per_prompt``: Number of samples to generate per prompt for evaluation.
 - ``generator.max_turns``: Maximum number of turns for generation with multi-turn RL.
 - ``generator.use_conversation_multi_turn``: Whether to use conversation format for multi-turn generation. If set to ``true`` then observations are appended to the chat history as a new turn. If set to ``false`` then observations are appended as-is to the assistant response in token space and generation is continued  (after removing any EOS token in the response).  We've observed some cases where model can be sensitive to chat history format (ex: in SkyRL-SQL), and thus ``false`` can be used for full control over the exact tokens added after environment interaction.
-- ``generator.engine_init_kwargs``: Inference engine arguments passed directly to the vLLM or SGLang engine. To specify an engine arg in the CLI override, use the format: +generator.engine_init_kwargs.[arg_name]=value
+- ``generator.engine_init_kwargs``: Inference engine arguments passed directly to the vLLM or SGLang engine. To specify an engine arg in the CLI override, use the format: +generator.engine_init_kwargs.[arg_name]=value. If duplicate kwargs are passed or kwargs clash with existing generator arguments (e.g., ``tensor_parallel_size``), an error is raised.
 
 Misc Configuration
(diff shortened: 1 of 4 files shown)


skyrl-train/docs/configuration/config.rst
```diff
@@ -451,6 +451,7 @@ Generator Configuration
     max_num_seqs: 1024
     remote_inference_engine_urls: ["127.0.0.1:8001"]
     max_turns: 1
+    vllm_engine_kwargs: {}
```
SumanthRH: Q: Can we just do `engine_init_kwargs` so that this can be used with both sglang and vllm?

tyler-griggs: Done.

skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
```diff
@@ -167,6 +168,7 @@ def create_ray_wrapped_inference_engines(
                 max_num_seqs=max_num_seqs,
                 # only need the logprobs for the chosen token if any
                 max_logprobs=1,
+                **engine_init_kwargs,
```
CharlieFRuan, line 171: nit: If `engine_init_kwargs` have say `tensor_parallel_size`, it might raise an error. This error may be a bit unclear to users. Should we do a try catch, or check clashing kwarg beforehand? Not sure how unclear the error will be.

If it is clear-ish then I think the current code is fine

mohit-rag:
@SumanthRH  @tyler-griggs 

This might need further edits to work correctly on vllm at least. 

Passing enable_auto_tool_choice and other kwargs to AsyncEngineArgs will raise an error - they need to be stripped first. 

Something like 

```python
+enable_auto_tools = kwargs.pop("enable_auto_tools", kwargs.pop("enable_auto_tool_choice", False))
+tool_parser = kwargs.pop("tool_parser", kwargs.pop("tool_call_parser", None))

+engine_args = vllm.AsyncEngineArgs(enable_log_requests=False, **kwargs)
+engine = vllm.AsyncLLMEngine.from_engine_args(engine_args)
....
        self.openai_serving_chat = OpenAIServingChat(
            engine_client=engine,
            model_config=model_config,
            models=models,
            response_role="assistant",
            request_logger=None,
            chat_template=None,
            chat_template_content_format="auto",
            +enable_auto_tools=enable_auto_tools,
            +tool_parser=tool_parser,
        )
```

PR #298, opened 2025-09-15
[Megatron] Add checkpointing support

tyler-griggs:
## What does this PR do?
This PR implements support for `save_checkpoint` and `load_checkpoint` for the Megatron training backend. We use Megatron's `dist_checkpointing` library to perform checkpointing in parallel across ranks, which also allows for reloading the checkpoints in a different parallelism scheme. 

Other minor changes: 
* Rename `save/load_ckpt` to `save/load_checkpoint`
* Removed unused arguments to `save/load_checkpoint`, primarily the inclusion of non-backend specific state (`tag`, `client_state`, `global_step`). This change keeps training backend checkpointing logic focused on the training backend's state.

## Testing
* Extended two GPU checkpointing tests to cover Megatron
  * Also moves `test_save_load_checkpoint.py` into `gpu_ci`. Note, however, that Megatron tests are disabled in CI because they currently require a different `flash-attn` install.
* Manually save-and-resumed several times:
Screenshot 2025-09-15 at 3 30 17 PM

## What's next?
- [ ]  Test multi-node checkpointing
- [ ]  Implement `save_hf_model`

PR #323, opened 2025-09-19
[Tests] Fix GPU CI failures after #298

SumanthRH:
# What does this PR do?

#298 broke GPU CI a bit: 
1. Megatron related dependencies have not been resolved properly on `main` yet, and this test should be skipped.
2. We use a simple 4xL4 instance, but then the test was modified in #298 to request 8 GPUs (non-colocated training)

### DIFF
--- a/skyrl-train/tests/gpu/gpu_ci/test_trainer_full_checkpointing.py
+++ b/skyrl-train/tests/gpu/gpu_ci/test_trainer_full_checkpointing.py
@@ -30,7 +30,7 @@ from tests.gpu.utils import import_worker, ray_init_for_tests
 from skyrl_train.entrypoints.main_base import config_dir
 
 MODEL_NAME = "Qwen/Qwen3-0.6B"
-NUM_GPUS = 4
+NUM_GPUS = 2
 
 
 class DummyDataset(Dataset):
--- a/skyrl-train/tests/gpu/gpu_ci/test_trainer_full_checkpointing.py
+++ b/skyrl-train/tests/gpu/gpu_ci/test_trainer_full_checkpointing.py
@@ -138,7 +138,9 @@ def create_minimal_trainer(cfg: DictConfig):
         ("fsdp2", False),
         ("fsdp2", True),
         # FIXME (erictang000): Add megatron test when dependencies work
-        # pytest.param("megatron", False, marks=pytest.mark.megatron),
+        pytest.param(
+            "megatron", False, marks=pytest.mark.skip(reason="Skipping megatron test until dependency issues are fixed")
+        ),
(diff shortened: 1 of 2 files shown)


skyrl-train/tests/gpu/gpu_ci/test_trainer_full_checkpointing.py
```diff
@@ -137,7 +138,10 @@ def create_minimal_trainer(cfg: DictConfig):
         ("fsdp", False),
         ("fsdp2", False),
         ("fsdp2", True),
-        pytest.param("megatron", False, marks=pytest.mark.megatron),
+        # FIXME (erictang000): Add megatron test when dependencies work
+        pytest.param(
```
tyler-griggs: Ah actually I meant to modify `ci/run_gpu_ci.sh` to update this:
```
-m "not (sglang or integrations)"
``` 
to this:
```
-m "not (sglang or integrations or megatron)"
```
Could we do this and move the todo here instead? Then, once megatron is ready, we don't need to change each file, just the shell script.

skyrl-train/tests/gpu/gpu_ci/test_trainer_full_checkpointing.py
```diff
@@ -30,7 +30,7 @@
 from skyrl_train.entrypoints.main_base import config_dir
 
 MODEL_NAME = "Qwen/Qwen3-0.6B"
-NUM_GPUS = 4
```
tyler-griggs, line 33: Could we just set num policy and num critic GPUs to `NUM_GPUS // 2` instead? If you want to get this quick CI fix in, I can do that in a follow up. Basically I want the test to be runnable for megatron and non-megatron (related to the comment below).

SumanthRH: How will megatron run if we do this:

```python

    # Megatron-specific
    if strategy == "megatron":
        cfg.trainer.policy.megatron_config.tensor_model_parallel_size = 2
        cfg.trainer.policy.megatron_config.pipeline_model_parallel_size = 2
        cfg.trainer.placement.policy_num_gpus_per_node = 4
```

with non colocation? 

(requests 6 GPUs)

SumanthRH: if you wanted to test TP > 1 and PP > 1 , then we should fix the test in a follow-up where we set these individually so that we can run it with 4 GPUs total.

SumanthRH:
confirmed locally that tests pass with 4 GPUs for checkpointing! should be good to merge

PR #325, opened 2025-09-19
[NCCL] Add stricter checks for P2P and set NCCL_P2P/SHM_DISABLE if needed, fix watchdog timeout

CharlieFRuan:
### Before this PR

Some users (including myself) have observed that the experiments that require more than 1 GPU occasionally run into the following error after hanging for 10 minutes

```
   (FSDPRefRayActorBase pid=31857) [rank1]:[E918 23:18:00.308397639 ProcessGroupNCCL.cpp:632] [Rank 1] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=1, OpType=BROADCAST, NumelIn=233373696, NumelOut=233373696, Timeout(ms)=600000) ran for 600023 milliseconds before timing out.
```

Example full log: https://gist.github.com/CharlieFRuan/64ecd3003933b94ba2906d64800edf76

This is likely due to some NCCL P2P issues on the machine's hardware/driver. A simple torch.distributed test can illustrate the problem. That is, for this test: https://gist.github.com/CharlieFRuan/316bee4dfe1137c4da89582624058ea0, if running ``torchrun --standalone --nproc_per_node=NUM_GPUS test_distributed.py`` hangs and ``NCCL_P2P_DISABLE=1 torchrun --standalone --nproc_per_node=4 test_distributed.py`` works, it means that the NCCL P2P access is not supported on your machine.

### This PR

While we have a util `run_p2p_access_check()`, which when fails add `NCCL_P2P_DISABLE=1` and `NCCL_SHM_DISABLE=1` when initializing ray, it did not cover these cases.

This PR adds a `run_nccl_torch_distributed_check()` that does a similar basic torch.distributed check. Besides, instead of having a boolean `peer_access_supported()` check, we use `get_env_vars_for_p2p_access()` that returns one of the following, progressively when the previous fails:
- `{}`
- `{"NCCL_SHM_DISABLE": "1"}`
- `{"NCCL_P2P_DISABLE": "1", "NCCL_SHM_DISABLE": "1"}`

Notably, if we do not run the `_get_env_vars_for_p2p_access_helper()` in a ray placement group, it may pass the test despite later on we still run into the same old hanging issue.

### Tests

- The previously hanging machine (4xL40s, 4xA40 on runpod) now can run `run_gsm8k.sh`.
- The previously working 4xA100 on runpod still works
- GPU CI running
- Need some multinode test?

### References
- https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html
- https://docs.vllm.ai/en/latest/usage/troubleshooting.html#incorrect-hardwaredriver

### DIFF
--- a/skyrl-train/skyrl_train/utils/utils.py
+++ b/skyrl-train/skyrl_train/utils/utils.py
@@ -661,9 +661,7 @@ def _get_env_vars_for_p2p_access_helper() -> dict[str, str]:
         logger.info("NCCL P2P/SHM checks passed with NCCL_P2P_DISABLE AND NCCL_SHM_DISABLE")
         return {"NCCL_P2P_DISABLE": "1", "NCCL_SHM_DISABLE": "1"}
 
-    raise RuntimeError(
-        "NCCL P2P/SHM checks failed even with fallbacks. Please verify drivers and interconnect."
-    )
+    raise RuntimeError("NCCL P2P/SHM checks failed even with fallbacks. Please verify drivers and interconnect.")
 
 
 def get_env_vars_for_p2p_access(max_num_gpus_per_node: int) -> dict[str, str]:
--- a/skyrl-train/skyrl_train/utils/utils.py
+++ b/skyrl-train/skyrl_train/utils/utils.py
@@ -483,7 +483,7 @@ def prepare_runtime_environment(cfg: DictConfig) -> dict[str, str]:
             ]
         )
     max_num_gpus_per_node = max(gpu_counts) if gpu_counts else 1
-    overrides = get_env_vars_override_for_peer_access(max_num_gpus_per_node=max_num_gpus_per_node)
+    overrides = get_env_vars_for_p2p_access(max_num_gpus_per_node=max_num_gpus_per_node)
     if overrides != {}:
         log_str = f"Peer access is not supported on this node type, setting the following env vars: {overrides}. "
         if max_num_gpus_per_node > 1:
@@ -628,12 +628,9 @@ def print_mem(tag: str, mem: dict):
     )
 
 
-def run_p2p_access_check():
+def _get_env_vars_for_p2p_access_helper() -> dict[str, str]:
     """
-    Three levels of disabling P2P access:
-    - No disabling
-    - Disabling P2P
-    - Disabling P2P and SHM
+    Use basic torch tests to determine if NCCL P2P/SHM disabling is needed for peer access.
     """
     device_count = torch.cuda.device_count()
     if device_count < 2:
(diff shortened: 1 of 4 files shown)


skyrl-train/skyrl_train/utils/utils.py
```diff
@@ -624,43 +628,71 @@ def print_mem(tag: str, mem: dict):
...
+
+    # NOTE(Charlie): For some reason, if we directly run `_get_env_vars_for_p2p_access_helper()`
+    # without a ray placement group, it might pass even when it shouldn't.
+    ray.init()
+    pg = placement_group([{"CPU": 1, "GPU": max_num_gpus_per_node}], strategy="PACK")
+    get_ray_pg_ready_with_timeout(pg, timeout=SKYRL_RAY_PG_TIMEOUT_IN_S)
+    result = ray.get(
+        ray.remote(num_gpus=2, scheduling_strategy=PlacementGroupSchedulingStrategy(pg))(
+            _get_env_vars_for_p2p_access_helper
+        ).remote()
```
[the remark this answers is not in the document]
CharlieFRuan: Addressed.

skyrl-train/skyrl_train/utils/utils.py
```diff
@@ -624,43 +628,69 @@ def print_mem(tag: str, mem: dict):
...
-                run_p2p_access_check
-            ).remote()
-        )
-        ray.shutdown()
-        return result
-    else:
-        return run_p2p_access_check()
+        return {"NCCL_P2P_DISABLE": "1", "NCCL_SHM_DISABLE": "1"}
+
+    # NOTE(Charlie): For some reason, if we directly run `_get_env_vars_for_p2p_access_helper()`
```
CharlieFRuan, line 681: Highlighting this note here

skyrl-train/skyrl_train/utils/utils.py
```diff
@@ -482,10 +483,13 @@ def prepare_runtime_environment(cfg: DictConfig) -> dict[str, str]:
             ]
         )
     max_num_gpus_per_node = max(gpu_counts) if gpu_counts else 1
-    if not peer_access_supported(max_num_gpus_per_node=max_num_gpus_per_node):
-        logger.info("Peer access is not supported on this node type, disabling NCCL P2P and SHM")
-        env_vars["NCCL_P2P_DISABLE"] = "1"
-        env_vars["NCCL_SHM_DISABLE"] = "1"
+    overrides = get_env_vars_for_p2p_access(max_num_gpus_per_node=max_num_gpus_per_node)
```
SumanthRH, line 486: Hmm what's the additional time now for the p2p check with this PR? 

Overall our initialization times are already pretty bad, so I'm wondering if we should have such an elaborate check before ray init (placement group + running an all reduce) or if we should just allow the user to override `NCCL_P2P_DISABLE` and `NCCL_SHM_DISABLE`

CharlieFRuan: yeah... I would prefer a troubleshooting guide for the users instead as well, like the vLLM one linked in the PR. Especially this doesn't seem to happen that frequently, and might be more case-by-case. So more fail-safe for users themselves to test

CharlieFRuan: I'll close this and make another PR if you agree

SumanthRH: > yeah... I would prefer a troubleshooting guide for the users instead as well, like the vLLM one linked in the PR. Especially this doesn't seem to happen that frequently, and might be more case-by-case. So more fail-safe for users themselves to test

perfect. let's just do this

SumanthRH:
Closing in favour of adding more documentation

PR #326, opened 2025-09-19
[Sleep][Fix] For unit tests, sleep at level 1 if we do not explicitly sync weights

CharlieFRuan:
After https://github.com/NovaSky-AI/SkyRL/pull/297, we discard weights (i.e. not kept in CPU either) when sleeping the engines (namely level 2 sleeping).

This causes various GPU tests to fail, as some tests sleep the engine but do not explicitly sync weights, leading to gibberish output.

One might think we should just make those tests that do not sync weights to have `colocate_all` as False, but having it as True makes it easier to pass the config validation (otherwise the test might require double the number of GPUs).

This PR fixes it by adding a `sleep_level` kwarg to `create_ray_wrapped_inference_engines()` and GPU test helper `init_inference_engines()`.

We do not want to always sleep at level 1 in unit tests, because we also want to test level 2 sleeping behavior.

Therefore, in unit tests, we sleep at level 1 when the test does not have weight sync, but sleep at level 2 when they do.

It is worth noting that SGLang's engine sleep always requires weight syncing (i.e. they always discards weights), so sleep level 1 vs. 2 is not applicable.

### Test
- GPU CI passed, printed output not gibberish anymore: https://github.com/NovaSky-AI/SkyRL/actions/runs/17851699567
- Failing tests are not due to sleep but are addressed in a separate PR
image

### Side notes
- Unrelated to #297 and this PR, our remote server's sleep method follows vLLM's API, not SGLang's. When we want to sleep with remote servers, we should fix it

### DIFF
--- a/skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
+++ b/skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
@@ -227,7 +227,7 @@ def create_ray_wrapped_inference_engines(
     if inference_engine_enable_sleep:
         if backend == "vllm":
             sleep_refs = [engine.inference_engine_actor.sleep.remote(level=sleep_level) for engine in engines]
-        else:
+        elif backend == "sglang":
             # TODO(Charlie): since our ray-wrapped SGLang's sleep API does not discard weights, it
             # can just sleep at the default level, whether testing or not. Revisit when changed.
             sleep_refs = [engine.inference_engine_actor.sleep.remote() for engine in engines]
--- a/skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
+++ b/skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
@@ -228,8 +228,8 @@ def create_ray_wrapped_inference_engines(
         if backend == "vllm":
             sleep_refs = [engine.inference_engine_actor.sleep.remote(level=sleep_level) for engine in engines]
         elif backend == "sglang":
-            # TODO(Charlie): since our ray-wrapped SGLang's sleep API does not discard weights, it
-            # can just sleep at the default level, whether testing or not. Revisit when changed.
+            # NOTE(Charlie): we always need to sync weights after waking up: https://github.com/sgl-project/sglang/issues/7939
+            assert sleep_level == 2, "SGLang always discards weights, so sleep_level is not applicable."
             sleep_refs = [engine.inference_engine_actor.sleep.remote() for engine in engines]
         ray.get(sleep_refs)
 
--- a/skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
+++ b/skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
@@ -79,6 +79,7 @@ def create_ray_wrapped_inference_engines(
     max_num_seqs=1024,
     tokenizer=None,
     backend="vllm",
+    sleep_level=2,  # we only set to 1 for unit tests that do not explicitly sync weights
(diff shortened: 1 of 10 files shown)


skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
```diff
@@ -224,7 +225,12 @@ def get_sglang_engine():
     engines = [RayWrappedInferenceEngine(actor_handle) for actor_handle in inference_engine_actors]
 
     if inference_engine_enable_sleep:
-        sleep_refs = [engine.inference_engine_actor.sleep.remote() for engine in engines]
+        if backend == "vllm":
+            sleep_refs = [engine.inference_engine_actor.sleep.remote(level=sleep_level) for engine in engines]
+        else:
+            # TODO(Charlie): since our ray-wrapped SGLang's sleep API does not discard weights, it
+            # can just sleep at the default level, whether testing or not. Revisit when changed.
+            sleep_refs = [engine.inference_engine_actor.sleep.remote() for engine in engines]
```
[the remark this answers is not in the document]
CharlieFRuan: Fixed.

skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
```diff
@@ -224,7 +225,12 @@ def get_sglang_engine():
     engines = [RayWrappedInferenceEngine(actor_handle) for actor_handle in inference_engine_actors]
 
     if inference_engine_enable_sleep:
-        sleep_refs = [engine.inference_engine_actor.sleep.remote() for engine in engines]
+        if backend == "vllm":
+            sleep_refs = [engine.inference_engine_actor.sleep.remote(level=sleep_level) for engine in engines]
+        elif backend == "sglang":
+            # TODO(Charlie): since our ray-wrapped SGLang's sleep API does not discard weights, it
```
erictang000: i think it always discards weights IIRC?

skyrl-train/skyrl_train/inference_engines/remote_inference_engine.py
```diff
@@ -143,6 +143,8 @@ async def wake_up(self, *args: Any, **kwargs: Any):
 
     async def sleep(self, *args: Any, **kwargs: Any):
         async with aiohttp.ClientSession() as session:
+            # TODO(Charlie): this is vLLM's API, not SGLang (which uses tags). Fix when need to
```
erictang000, line 146: QQ: why would we want to sleep for remote engines?

erictang000: is this case colocated but not managed by us?

erictang000 (commented):
lgtm, just some questions on the comments you left

skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
CharlieFRuan: Oh you are right, thanks for the catch. I even created an issue for it before... https://github.com/sgl-project/sglang/issues/7939

skyrl-train/skyrl_train/inference_engines/remote_inference_engine.py
CharlieFRuan: Yes, when say we want to remove the `xxx_engine.py` codepaths and only use `xxx_server.py`

skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
CharlieFRuan: Fixed!

SumanthRH (approved):

PR #329, opened 2025-09-21
[Trainer] Seperate Eval from Training

benji-cannot-code:
As a continuation of creating an eval-only entry point, this PR: 
1. Further decouples evaluation logic from training. It does so by creating a separate eval file and pulling out functions shared by eval and training, including `prepare_generator_input` and `build_dataloader`. 
2. Provides minor fixes such as making `prepare_generator_input` use consistent `n_samples_per_prompt` and changing `print` statements to use `logger` for some parts of the codebase. 
3. Adds documentation to indicate that we now have an eval-only entrypoint. 

Testing

- [x] E2E gsm8k training 
- [x] E2E gsm8k eval-only
- [x] Eval-only entrypoint GPU test
- [x] Local view of docs to make sure it looks fine

PR #336, opened 2025-09-22
[fix] Fix generation entrypoint

tyler-griggs:
The `InferenceEngineClient` interface was updated but the update was not propagated here.


benji-cannot-code:
I think I have the same fix in #329

SumanthRH (approved):

PR #338, opened 2025-09-23
[trainer] Force multi-node ranks to be allocated in continuous order

### LINKED ISSUE #324: Ensure consecutive worker ranks are grouped together on nodes
Ray placement groups do not return bundle indices in a guaranteed order: https://github.com/ray-project/ray/issues/51117

Currently for the colocated case we naively create a large bundle and [iterate through it](https://github.com/NovaSky-AI/SkyRL/blob/f0015a705af0af82a0012538ff1b62c3b158cbea/skyrl-train/skyrl_train/workers/worker.py#L425) to assign ranks, which can result in adjacent worker ranks (i.e. worker rank 0 and rank 1) being placed on different GPUs, causing issues for multi-node training when using various parallelism strategies.

We should adopt a solution similar to slime: https://github.com/THUDM/slime/blob/3e043e4d69eee9732618f1e6a8ea56aa2472353c/slime/ray/placement_group.py#L37

where they first initialize a placement group, place dummy actors to get the actual node and gpu id for each bundle index, then kill those actors, re-sort the bundle indices by node ip and gpu id, and then are able to use that to assign actors with consecutive ranks to the same node.

erictang000:
# Overview
Closes #324

We need to create an actor to place on each node and kill it (rather than just using `ray.util.placement_group_table` to look up node id) since we need to get gpu id for each bundle index, which can only be grabbed from inside a ray actor. 

ran GSM8k on 2 nodes of 4xL40S colocated and non-colocated
image


skyrl-train/tests/gpu/test_multi_node_pg.py
```diff
@@ -0,0 +1,86 @@
...
+    return get_test_actor_config()
+
+
+def test_multi_node_pg_init(ray_init_fixture, cfg):
+    try:
+        cfg = get_test_actor_config()
+        cfg.trainer.placement.colocate_all = True
+        cfg.generator.weight_sync_backend = "nccl"
+        cfg.trainer.strategy = "fsdp"
+        cfg.trainer.placement.policy_num_nodes = 5
```
erictang000: note that I only saw this fail consistently once I had above 16 gpus. But we can scale this down when we include multi-node tests in CI later

skyrl-train/skyrl_train/utils/utils.py
```diff
@@ -586,6 +591,41 @@ def get_ray_pg_ready_with_timeout(pg: PlacementGroup, timeout: int = 60):
...
+    info_actors = []
+    for i in range(num_bundles):
+        info_actors.append(
+            InfoActor.options(
+                scheduling_strategy=PlacementGroupSchedulingStrategy(
+                    placement_group=pg,
+                    placement_group_bundle_index=i,
+                )
+            ).remote()
+        )
```
SumanthRH, line 620: QQ: does this work fine for if a bundle has more than 1 gpu? 

Looks like it does, wondering if you confirmed this.


IN the colocated case, each bundle just has 1 gpu:
 https://github.com/NovaSky-AI/SkyRL/blob/b19a9d32a6a35229693f4dd35f91c01f469314af/skyrl-train/skyrl_train/entrypoints/main_base.py#L153-L159 


But that is not true in the non-colocated case:

https://github.com/NovaSky-AI/SkyRL/blob/8427aa3d33f8b5a194f1db6889d565dceaf7f20c/skyrl-train/skyrl_train/trainer.py#L566-L574


However, finally the mapping of actors to ranks in `_initiate_actors` will assign them to a bundle index in order, so nearby ranks will be guaranteed to be placed close by:

https://github.com/NovaSky-AI/SkyRL/blob/8427aa3d33f8b5a194f1db6889d565dceaf7f20c/skyrl-train/skyrl_train/workers/worker.py#L446-L450


So this function itself seems fine. I do think we're missing a case in `PPORayActorGroup` though where `pg` is not passed in

skyrl-train/skyrl_train/workers/worker.py
```diff
@@ -391,7 +403,7 @@ def _initiate_actors(self, pg: Optional[PlacementGroup], num_gpus_per_actor: flo
                 num_gpus=num_gpus_per_actor,
                 resources=self._resources,
                 scheduling_strategy=PlacementGroupSchedulingStrategy(
-                    placement_group=pg, placement_group_bundle_index=0
+                    placement_group=pg, placement_group_bundle_index=reordered_bundle_indices[0]
```
SumanthRH: This will raise an `UnboundLocalError`  when `pg` is not passed in to `_initiate_actors`. This is the case when we use a critic model with `colocate_all` as `False` or when `colocate_policy_ref` is `False` in non-colocated case. 

Code for the critic model: 
https://github.com/NovaSky-AI/SkyRL/blob/8427aa3d33f8b5a194f1db6889d565dceaf7f20c/skyrl-train/skyrl_train/trainer.py#L601-L630


We create `pg` inside the function in this case;

https://github.com/NovaSky-AI/SkyRL/blob/8427aa3d33f8b5a194f1db6889d565dceaf7f20c/skyrl-train/skyrl_train/workers/worker.py#L391-L400

so both the if conditions will be true but then the first code block is supposed to also obtain `reordered_bundle_indices` .

skyrl-train/tests/gpu/test_multi_node_pg.py
```diff
@@ -0,0 +1,86 @@
+"""
+Run with:
+uv run --isolated --extra dev -- pytest tests/gpu/test_multi_node_pg.py
```
SumanthRH, line 3:
```suggestion
uv run --isolated --extra dev -- pytest tests/gpu/test_multi_node_pg.py
NOTE: Currently this requires 20 GPUs to run
```

skyrl-train/skyrl_train/utils/utils.py
erictang000: yeah, we only need to reorder bundle indices in the case where you have 1 gpu per bundle. The case where we have num_gpus_per_node per bundle should not require the reordering logic

skyrl-train/skyrl_train/workers/worker.py
erictang000: good catch

erictang000: let me update the test to cover these cases

SumanthRH: Thanks

PR #340, opened 2025-09-24
LoRA support 

lynnliu030:


skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -29,6 +29,9 @@ trainer:
   policy:
     model:
       path: "Qwen/Qwen2.5-1.5B-Instruct"
+      lora_rank: 0
+      lora_alpha: 16
+      lora_dropout: 0
```
tyler-griggs:
```suggestion
      lora:
          rank: 0
          alpha: 16
          dropout: 0
```

tyler-griggs: Can we make the config more modular and put the config args under a single `lora` config object?

skyrl-train/skyrl_train/distributed/fsdp_strategy.py
```diff
@@ -354,6 +361,39 @@ def _fix_fsdp_config(self, config):
...
+        if self.is_rank_0():
+            io.makedirs(lora_save_path, exist_ok=True)
+            peft_config = asdict(model.peft_config.get("default", {}))
+            if peft_config:
+                peft_config["task_type"] = peft_config["task_type"].value
+                peft_config["peft_type"] = peft_config["peft_type"].value
+                peft_config["target_modules"] = list(peft_config["target_modules"])
+
+        try:
+            if fsdp_version(model) > 0:
```
tyler-griggs: Is this check necessary? If I understand correct, version that is not 1 or 2 is not supported. Or am I missing something?

skyrl-train/skyrl_train/distributed/fsdp_strategy.py
```diff
@@ -354,6 +361,39 @@ def _fix_fsdp_config(self, config):
...
+
+        if self.is_rank_0():
+            io.makedirs(lora_save_path, exist_ok=True)
+            peft_config = asdict(model.peft_config.get("default", {}))
+            if peft_config:
+                peft_config["task_type"] = peft_config["task_type"].value
+                peft_config["peft_type"] = peft_config["peft_type"].value
+                peft_config["target_modules"] = list(peft_config["target_modules"])
+
+        try:
```
tyler-griggs: Can we remove the try/except? If we can't successfully save the LoRA adapter, this seems like a pretty big issue and I'd rather it throw an error than catch it here.

skyrl-train/skyrl_train/distributed/fsdp_strategy.py
```diff
@@ -354,6 +361,39 @@ def _fix_fsdp_config(self, config):
...
+            if fsdp_version(model) > 0:
+                model = model.to(torch.cuda.current_device())
```
tyler-griggs: Is this `model.to()` call necessary? Will the model not already be on GPU when this is called? Also, in general, if a model needs to be loaded we use `backload_to_gpu` ([link](https://github.com/NovaSky-AI/SkyRL/blob/8ae7d2f69439079e86cf287f0cb2c7e3caa71b5b/skyrl-train/skyrl_train/distributed/fsdp_strategy.py#L131-L144)) because it's sometimes a little more complicated (e.g., [link](https://github.com/NovaSky-AI/SkyRL/blob/8ae7d2f69439079e86cf287f0cb2c7e3caa71b5b/skyrl-train/skyrl_train/distributed/fsdp_utils.py#L168-L184))

skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
```diff
@@ -223,6 +233,8 @@ def get_sglang_engine():
 
     if inference_engine_enable_sleep:
         if backend == "vllm":
+            # NOTE(shu): set to 1 for LoRA
+            sleep_level = 1 if enable_lora else sleep_level
```
tyler-griggs, line 268: Can you explain why we set to 1 for LoRA?

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
```diff
@@ -258,10 +269,24 @@ def _create_engine(self, *args, **kwargs):
...
+        lora_requests = None
+        if self._is_lora:
+            lora_int_ids = list(self.llm.list_loras())
+            if len(lora_int_ids) > 0:
+                lora_int_id = lora_int_ids[0]
+                from vllm.lora.request import LoRARequest
+
+                batch_size = len(prompt_token_ids)
+                lora_requests = [
+                    LoRARequest(lora_name=f"{lora_int_id}", lora_int_id=lora_int_id, lora_path="/dummy_lora_path")
```
tyler-griggs, line 284: `dummy_lora_path` seems suspicious :D What does this mean?

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
```diff
@@ -290,13 +316,31 @@ async def init_weight_update_communicator(
...
         )
 
+    async def _load_lora_from_disk(self, lora_path: str):
+        """Load LoRA adapters from disk using vLLM's native add_lora method."""
+        from vllm.lora.request import LoRARequest
+        import time
+
+        lora_id = int(time.time_ns() % 0x7FFFFFFF)
+        lora_request = LoRARequest(lora_name=f"{lora_id}", lora_int_id=lora_id, lora_path=lora_path)
+        try:
```
tyler-griggs: This is another try/except that might silence important errors. We probably want an error thrown if we can't save or load lora modules. Can we remove it?

skyrl-train/skyrl_train/utils/utils.py
tyler-griggs, line 1: Can we do some config validation in this file as well (under `validate_cfg`)? For example, if lora is enabled, the generator backend needs to be `vllm` and the training backend needs to be `fsdp` or `fsdp2`. If there are other sanity checks for the lora config parameters, this would be a good time to do it so that the code fails early.

skyrl-train/skyrl_train/utils/utils.py
```diff
@@ -453,6 +448,9 @@ def prepare_runtime_environment(cfg: DictConfig) -> dict[str, str]:
             env_vars["NVTE_FUSED_ATTN"] = "0"
 
     if cfg.generator.backend == "vllm":
+        # NOTE(shu): for LoRA
+        env_vars["VLLM_ALLOW_RUNTIME_LORA_UPDATING"] = "true"
```
tyler-griggs:
```suggestion
        env_vars["VLLM_ALLOW_RUNTIME_LORA_UPDATING"] = "true"
```

tyler-griggs: I think they'll know it's for lora :)

skyrl-train/skyrl_train/workers/fsdp/fsdp_worker.py
```diff
@@ -49,6 +50,9 @@ def init_model(self, model_path, num_training_steps: int = None):
         strategy.setup_distributed()
         self.strategy = strategy
 
+        # LoRA specific configs
+        self._is_lora = self.cfg.trainer.policy.model.lora_rank > 0
```
tyler-griggs:
```suggestion
        self._is_lora = self.cfg.trainer.policy.model.lora_rank > 0
```

skyrl-train/skyrl_train/workers/fsdp/fsdp_worker.py
```diff
@@ -64,7 +68,13 @@ def init_model(self, model_path, num_training_steps: int = None):
                 # NOTE (sumanthrh): Model initialization should always be in fp32
                 # during training
                 bf16=False,
+                # LoRA specific configs
```
tyler-griggs:
```suggestion
```

tyler-griggs: Probably do not need to specify this -- the config args have "lora" in the name so it is pretty clear.

skyrl-train/skyrl_train/workers/fsdp/fsdp_worker.py
```diff
@@ -64,7 +68,13 @@ def init_model(self, model_path, num_training_steps: int = None):
...
+                lora_rank=self.cfg.trainer.policy.model.lora_rank,
+                lora_alpha=self.cfg.trainer.policy.model.lora_alpha,
+                lora_dropout=self.cfg.trainer.policy.model.lora_dropout,
                 target_modules=self.cfg.trainer.target_modules,
+                exclude_modules=self.cfg.trainer.exclude_modules,
+                #
```
tyler-griggs:
```suggestion
```

skyrl-train/skyrl_train/workers/fsdp/fsdp_worker.py
```diff
@@ -90,6 +100,40 @@ def init_model(self, model_path, num_training_steps: int = None):
...
+            peft_config = asdict(peft_model.peft_config.get("default", {}))
+            peft_config["task_type"] = peft_config["task_type"].value
+            peft_config["peft_type"] = peft_config["peft_type"].value
+            peft_config["target_modules"] = list(peft_config["target_modules"])
+
+            # Save LoRA parameters and config
+            try:
+                save_file(lora_params, os.path.join(lora_sync_path, "adapter_model.safetensors"))
+                with open(os.path.join(lora_sync_path, "adapter_config.json"), "w", encoding="utf-8") as f:
+                    json.dump(peft_config, f, ensure_ascii=False, indent=4)
```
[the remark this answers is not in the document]
tyler-griggs: ^^

skyrl-train/skyrl_train/workers/fsdp/fsdp_worker.py
```diff
@@ -90,6 +100,40 @@ def init_model(self, model_path, num_training_steps: int = None):
...
+        if torch.distributed.get_rank() == 0:
+            os.makedirs(lora_sync_path, exist_ok=True)
+
+            peft_config = asdict(peft_model.peft_config.get("default", {}))
+            peft_config["task_type"] = peft_config["task_type"].value
+            peft_config["peft_type"] = peft_config["peft_type"].value
+            peft_config["target_modules"] = list(peft_config["target_modules"])
+
+            # Save LoRA parameters and config
+            try:
```
tyler-griggs: Remove try/except here as well?

skyrl-train/skyrl_train/workers/fsdp/fsdp_worker.py
```diff
@@ -105,7 +149,23 @@ async def broadcast_to_inference_engines(self, inference_engine_client):
                 state_dict_type=StateDictType.SHARDED_STATE_DICT,
                 state_dict_config=ShardedStateDictConfig(),
             )
-        params = self.model.model.state_dict()
+
+        # Check if this is a LoRA model
+        peft_model = getattr(self.model.model, "_fsdp_wrapped_module", self.model.model)
+
+        if self._is_lora and hasattr(peft_model, "peft_config"):
```
tyler-griggs: Is it an error if `self._is_lora==True` but peft_model does not have the `peft_config` attribute? Basically, if this is a lora model, do we expect to always have a peft_config?

I'm wondering whether we should not have these two conditions in the `if` condition, and instead should be asserting that peft_model has the `peft_config` much sooner in the code to detect issues earlier

skyrl-train/skyrl_train/workers/fsdp/fsdp_worker.py
```diff
@@ -105,7 +149,23 @@ async def broadcast_to_inference_engines(self, inference_engine_client):
...
+            from skyrl_train.distributed.fsdp_utils import collect_lora_params
+
+            # assume base model is already synced, sync LoRA adapters
+            # TODO(shu): add a lora disk path to the config?
+
+            params = collect_lora_params(module=self.model.model)
+            lora_sync_path = getattr(self.cfg.generator, "lora_sync_path", "/tmp/skyrl_lora_sync")
```
tyler-griggs: Let's not add a default here (`"/tmp/skyrl_lora_sync"`). We should assume that the base config has reasonable defaults for all parameters, so we can just do:
```
lora_sync_path = self.cfg.generator.lora_sync_path
```

skyrl-train/skyrl_train/workers/fsdp/fsdp_worker.py
```diff
@@ -105,7 +149,23 @@ async def broadcast_to_inference_engines(self, inference_engine_client):
...
+            await self._save_lora_adapters_and_sync(params, peft_model, lora_sync_path, inference_engine_client)
```
tyler-griggs: Actually can we just pull this whole block into it's own method? Basically if `self._is_lora()`, we directly call `_save_lora_adapters_and_sync` and move the calls to `collect_lora_params` into that method.

skyrl-train/tests/gpu/test_lora.py
```diff
@@ -0,0 +1,104 @@
+"""
+# Run only vllm tests (requires vllm extra):
+uv run --isolated --extra dev --extra vllm --extra deepspeed pytest tests/gpu/test_policy_local_engines_e2e.py -m "vllm"
```
tyler-griggs:
```suggestion
uv run --isolated --extra dev --extra vllm pytest tests/gpu/test_policy_local_engines_e2e.py -m "vllm"
```

tyler-griggs: deepspeed is not used

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
```diff
@@ -244,6 +245,16 @@ def _get_engine(self):
...
         return self.llm.engine if hasattr(self.llm, "engine") else self.llm
 
+    def _is_lora_disk_loading_request(self, request: NamedWeightsUpdateRequest) -> bool:
+        """Check if this is a LoRA disk loading request."""
+        is_lora_disk = (
+            request.get("extras")
+            and len(request["extras"]) > 0
+            and "lora_disk_path" in request["extras"][0]
+            and request["names"][0].startswith("lora_disk_load")
+        )
```
tyler-griggs: Can we simplify this to something like this:
```
is_lora = request["names"][0] == "lora_disk_load"
if is_lora:
    assert request.get("extras")
            and len(request["extras"]) > 0
            and "lora_disk_path" in request["extras"][0], "vLLM LoRA weight update requests must contain the disk load path under key `lora_disk_load`"
return is_lora
```

In this case, the unique identifier that indicates this is a LoRA load is `lora_disk_load`, and the rest is just a sanity check.

skyrl-train/skyrl_train/workers/fsdp/fsdp_worker.py
```diff
@@ -90,6 +100,40 @@ def init_model(self, model_path, num_training_steps: int = None):
...
+                    json.dump(peft_config, f, ensure_ascii=False, indent=4)
+            except Exception as e:
+                raise Exception(
+                    f"[LoRA-TRACK][rank-{torch.distributed.get_rank()}]: Error saving LoRA adapters to disk: {e}"
+                )
+
+            # Send LoRA disk loading request to inference engine
+            lora_request = {
+                "names": ["lora_disk_load"],  # hard code this...
+                "extras": [{"lora_disk_path": lora_sync_path}],
```
tyler-griggs:
```suggestion
            # Send LoRA disk loading request to inference engine. `lora_disk_load` is a specific identifier 
            # to tell the inference engine to extract the `lora_disk_path`. 
            lora_request = {
                "names": ["lora_disk_load"], 
                "extras": [{"lora_disk_path": lora_sync_path}],
```

skyrl-train/skyrl_train/distributed/fsdp_strategy.py
lynnliu030: Yes! I can remove this check

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
lynnliu030: `LoRARequest` requires a non-empty `lora_path` so we just pass in a placeholder; it's not used for loading anything in this case (this is done in `add_lora`!)

skyrl-train/skyrl_train/utils/utils.py
lynnliu030: Good point, thanks!

skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
lynnliu030: Setting sleep level to 1 as a simple first-version implementation. In this mode only the LoRA weights are saved and synchronized, while the base model weights remain on CPU during the training step in colocated training. The sync is pretty straightforward here (essentiall just checkpointing and reloading lora weights) and not needing to move or sync the full base model; cc @SumanthRH

SumanthRH: Yeah this is pretty important for lora because we don't want to discard the base weights ever to avoid wasted syncs from the policy  - these are never updated so the inference engine can retain them

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
SumanthRH: maybe a good comment here for now?

skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
lynnliu030: I think verl can deal with sleep_level=2 too by also syncing the base weights together,  do we want to support that if needed?

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -57,6 +62,9 @@ trainer:
   critic:
     model:
       path: null
+      lora_rank: 0
```
tyler-griggs: Oh can we do the same hierarchical config update here, too?

skyrl-train/skyrl_train/distributed/fsdp_utils.py
```diff
@@ -481,3 +482,71 @@ def to_str(precision):
...
+                    lora_params.update(sub_lora_params)
+                    submodule._is_root = False
+                torch.cuda.empty_cache()
+    return lora_params
+
+
+def collect_lora_params(module: FSDP) -> OrderedDict:
+    """
+    collect lora params or full params if base model is not ready in vllm
+    work with if isinstance(self.module._fsdp_wrapped_module, PeftModel)
```
tyler-griggs:
```suggestion
    Requires `module._fsdp_wrapped_module` is a `PeftModel`
```

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
```diff
@@ -367,13 +407,40 @@ def _create_engine(self, *args, **kwargs):
...
         return engine
 
+    async def _load_lora_from_disk(self, lora_path: str):
+        """Load LoRA adapters from disk using vLLM's native add_lora method."""
+        from vllm.lora.request import LoRARequest
+        import time
+
+        lora_id = int(time.time_ns() % 0x7FFFFFFF)
+        lora_request = LoRARequest(lora_name=f"{lora_id}", lora_int_id=lora_id, lora_path=lora_path)
+        try:
```
tyler-griggs: Can we remove this try/except so it doesn't silence LoRA load failures? If we fail to load LoRA, training should probably be stopped instead of catching the Exception.

skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
tyler-griggs: Makes sense! This could be a helpful note to add to docs.

skyrl-train/skyrl_train/workers/fsdp/fsdp_worker.py
```diff
@@ -49,6 +51,8 @@ def init_model(self, model_path, num_training_steps: int = None):
         strategy.setup_distributed()
         self.strategy = strategy
 
+        self._is_lora = self.cfg.trainer.policy.model.lora.rank > 0
```
tyler-griggs, line 53: (Note that this is currently unused, but my comment below suggests that we use it)

skyrl-train/skyrl_train/workers/fsdp/fsdp_worker.py
```diff
@@ -105,7 +146,18 @@ async def broadcast_to_inference_engines(self, inference_engine_client):
                 state_dict_type=StateDictType.SHARDED_STATE_DICT,
                 state_dict_config=ShardedStateDictConfig(),
             )
-        params = self.model.model.state_dict()
+
+        # Check if this is a LoRA model
+        peft_model = getattr(self.model.model, "_fsdp_wrapped_module", self.model.model)
+
+        if hasattr(peft_model, "peft_config"):
```
tyler-griggs: Should we be using `self._is_lora` here instead? 

If `self._is_lora==True`, then `hasattr(peft_model, "peft_config")` must be `True`, otherwise it's an error, right? 

If that's the case, can we just use `if self._is_lora` here, and then `assert hasattr(peft_model, "peft_config")`?

skyrl-train/tests/gpu/test_lora.py
```diff
@@ -0,0 +1,104 @@
...
+        return cfg
+
+
+@pytest.mark.parametrize(
+    ("colocate_all", "weight_sync_backend", "strategy", "backend", "tp_size"),
+    [
+        pytest.param(False, "nccl", "fsdp", "vllm", 2, marks=pytest.mark.vllm),
+        pytest.param(True, "nccl", "fsdp", "vllm", 2, marks=pytest.mark.vllm),
+        pytest.param(False, "gloo", "fsdp", "vllm", 2, marks=pytest.mark.vllm),
+        pytest.param(True, "gloo", "fsdp", "vllm", 2, marks=pytest.mark.vllm),
```
tyler-griggs: We can remove the `gloo` tests here. The LoRA logic doesn't change based on gloo vs nccl backend, and we have the other inference engine tests for nccl.

skyrl-train/examples/lora/run_qwen2_5_0.5b_gsm8k_grpo_lora.sh
tyler-griggs, line 1: One missing piece is docs. It would be great to have a docs page on LoRA (how to use it, how to configure it, and the current limitations). Also, the config docs should be updated to include the new params in `ppo_base_config.yaml`.

skyrl-train/examples/lora/run_qwen2_5_0.5b_gsm8k_grpo_lora.sh
```diff
@@ -0,0 +1,60 @@
...
+  generator.weight_sync_backend=nccl \
+  generator.async_engine=true \
+  generator.batched=true \
+  environment.env_class=gsm8k \
+  generator.n_samples_per_prompt=5 \
+  generator.gpu_memory_utilization=0.8 \
+  trainer.logger="$LOGGER" \
+  trainer.project_name="gsm8k_0.5b_lora" \
+  trainer.run_name="gsm8k_0.5b_lora_test" \
+  trainer.resume_mode=null \
```
SumanthRH, line 58: Can we test resuming from a checkpoint with LoRA? I assume we still use the same `trainer.policy.model.path`, and resumption shoudl just work because we have a weight sync before generation - where we'll end up syncing the lora weights. Would be good to test and confirm

skyrl-train/docs/examples/lora.rst
```diff
@@ -0,0 +1,166 @@
...
+Target Modules
+~~~~~~~~~~~~~~
+
+By default, LoRA is applied to all linear layers in the model. You can customize which modules to target:
+
+.. code-block:: yaml
+
+    trainer:
+      target_modules: "all-linear"  # Apply to all linear layers
+      # OR specify specific modules:
```
SumanthRH:
```suggestion
      target_modules: "all-linear"  # Apply to all linear layers OR
      # specify specific modules as a list
```

minor edit

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -144,6 +153,8 @@ trainer:
   flash_attn: true
   disable_fast_tokenizer: false
   target_modules: "all-linear"
+  exclude_modules: null
+  use_orm_score: false
```
tyler-griggs: Is this used somewhere? Should we remove it?

skyrl-train/docs/examples/lora.rst
```diff
@@ -0,0 +1,166 @@
...
+These limitations are being addressed in future releases, with plans for in-memory synchronization and improved adapter management.
+
+What's Next?
+------------
+
+Now that you understand LoRA training in SkyRL, you might want to explore:
+
+- :doc:`../configuration/config`: Complete configuration reference
+- :doc:`ppo`: Standard PPO training without LoRA
+- :doc:`../tutorials/async`: Asynchronous training patterns
```
tyler-griggs, line 166: Can remove this.

tyler-griggs (approved):
great!

skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
```diff
@@ -162,34 +171,35 @@ def create_ray_wrapped_inference_engines(
...
-                    num_gpus=0.2 if use_hybrid_engine else 1,
-                    enable_sleep_mode=inference_engine_enable_sleep,
-                    noset_visible_devices=noset_visible_devices,
-                    max_num_batched_tokens=max_num_batched_tokens,
-                    max_num_seqs=max_num_seqs,
-                    max_logprobs=1,  # only need chosen-token logprobs
-                    **dp_kwargs,
-                    **engine_init_kwargs,
-                )
-                inference_engine_actors.append(engine)
```
SumanthRH, line 192: This change has also changed the indentation, so now only one engine is being allocated for the DP group

SumanthRH: Looks like this was fixed in https://github.com/NovaSky-AI/SkyRL/pull/483

PR #343, opened 2025-09-24
[trainer] Changed logging inside of skyrl_train folder from print() statement to use logger.log from the loguru package (UPDATE)

zhenga1:
Changed all the `print()` statements to logger from the `loguru` package.


tyler-griggs:
Thanks!

PR #347, opened 2025-09-25
Added conversion pipeline of FSDP2 model checkpoint shards to Huggingface .safetensors

zhenga1:
Wrote script for converting FSDP2 model shard into huggingface `.safetensors`.

Filepath:
`~/skyrl-train/scripts/convert_fsdp_to_hf.py`

Tested on Qwen0.5B models trained with `run_gsm8k.sh` with the fsdp2 backend.

### DIFF
--- a/skyrl-train/scripts/convert_fsdp_to_hf.py
+++ b/skyrl-train/scripts/convert_fsdp_to_hf.py
@@ -1,0 +1,219 @@
+"""
+Helper Script to convert FSDP shards to safetensor model files, compatible with Huggingface API
+
+The main purpose is to be able to enable users who choose not to enable HF model saves during training, such as enable the `hf_save_interval` parameter, to
+also be able to benefit from a way to create a HF safetensors model.
+
+Example usage:
+uv run --isolated --frozen --extra vllm scripts/convert_fsdp_to_hf.py --ckpt-dir /home/ubuntu/ckpts/gsm8k_0.5B_ckpt/global_step_10 --out-dir /home/ubuntu/hf/glob_step_10
+"""
+import argparse
+import json
+import os
+import re
+import sys
+from pathlib import Path
+import shutil
+from safetensors.torch import save_file, save_model
+from typing import Dict, Any, Optional
+#from loguru import logger
+
+import torch
+from safetensors.torch import save_file as save_safetensors
+
+def find_policy_dir(chkpt_dir):
+    pol = chkpt_dir / "policy"
+    if not pol.exists():
+        #logger.error(f"Expected 'policy/' under {chkpt_dir}")
+        print(f"[error] Expected 'policy/' under {chkpt_dir}")
+        raise FileNotFoundError(f"Expected 'policy/' under {chkpt_dir}")
+    return pol
+
+def get_model_shards(policy_dir: Path):
+    # policy_dir.glob returns a generator iterating over all matches 
+    shards = sorted(policy_dir.glob("model_world_size_*_rank_*.pt"))
+    if not shards:
+        # check for possibility that model is saved as follows as well
+        shards = sorted(policy_dir.glob("model*.pt"))
+    if not shards:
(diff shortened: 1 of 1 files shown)


skyrl-train/scripts/convert_fsdp_to_hf.py
```diff
@@ -0,0 +1,219 @@
+"""
+Helper Script to convert FSDP shards to safetensor model files, compatible with Huggingface API
+
```
tyler-griggs, line 3: Can you add a brief comment about what the expected output format is? Ie, what will the directory `out_dir` look like after this is run?

skyrl-train/scripts/convert_fsdp_to_hf.py
```diff
@@ -0,0 +1,219 @@
...
+def find_policy_dir(chkpt_dir):
+    pol = chkpt_dir / "policy"
+    if not pol.exists():
+        #logger.error(f"Expected 'policy/' under {chkpt_dir}")
+        print(f"[error] Expected 'policy/' under {chkpt_dir}")
+        raise FileNotFoundError(f"Expected 'policy/' under {chkpt_dir}")
+    return pol
+
+def get_model_shards(policy_dir: Path):
```
tyler-griggs: Can we simplify this method to assume it matches exactly the directory structure we expect? Then we don't have to check multiple formats. If we save a checkpoint with fsdp or fsdp2,  I believe the directory under `policy` right now has a bunch of `model_world_size_*_rank_*.pt` files. Should we just load from that and not check "model*.pt"?

By the way, our docs might be wrong here, because they say the FSDP structure is `model_state.pt`.[docs link](https://skyrl.readthedocs.io/en/latest/checkpointing-logging/checkpointing.html#directory-structure). We can update that next

tyler-griggs: Can we add a return type to this. What is `shards` type? Is it a `List[Path]`?

skyrl-train/scripts/convert_fsdp_to_hf.py
```diff
@@ -0,0 +1,219 @@
...
+    ap.add_argument("--out-dir", type=str, required=True, 
+                    help="Output for HF model folder")
+    ap.add_argument("--validate-load", action="store_true", help="Try loading with transformers after saving")
+    args = ap.parse_args()
+
+    ckpt_dir = Path(args.ckpt_dir).resolve()
+    output_dir = Path(args.out_dir).resolve()
+
+    policy_dir = find_policy_dir(ckpt_dir)
+    shards = get_model_shards(policy_dir)
```
tyler-griggs: Are these `shards` actually file paths? If so, can we call this `shard_paths` just so it's more clear.

tyler-griggs (commented):
Hey @zhenga1 , thanks for sending this in! Very cool that it's working for the 0.5B model.

There seem to be some incomplete state in the PR, such as commented lines and debugging statements, and a call to `validate_load` which doesn't exist. Could you please clean up this PR, and ensure it adheres to the formatter (you can `cd` into `SkyRL/skyrl-train` and run `uv run bash ../format.sh`  for auto-formatting), and then I'll make another pass?

CharlieFRuan:
Tried it out on my own; I don't think this script considers when the training is done multiple GPUs, could you double check when training with, say 2 GPUs?

I got the following error when validating the converted HF (I used 8 GPUs to train):
```
RuntimeError: Error(s) in loading state_dict for Linear:
        size mismatch for weight: copying a param with shape torch.Size([18992, 4096]) from checkpoint, the shape in current model is torch.Size([151936, 4096]).
```

CharlieFRuan:
You can likely refer to the `save_model()` in `trainer.py`, but without the constraint of "if I got the FSDP checkpoint with 8 GPUs, I have to convert to HF checkpoint with 8 GPUs"

PR #359, opened 2025-09-28
Rename classes with overloaded / incorrect names

tyler-griggs:
This PR renames the following:
* `FSDPRefRayActorBase` -> `FSDPRefRayWorkerBase`: The other training backends use `Worker`, not `Actor`. Also, `Actor` is overloaded with Ray actors, so worker is preferred.
* In models.py, `Actor` -> `HFWrappedModel`: Here also `Actor` is overloaded with Ray actors, and it really just refers to a wrapping of the underlying HF model. Also, it is not used specifically for the policy/actor model in RL, but rather used as the base for all models. 
* `MegatronPolicy` -> `MegatronWrappedModel`: For Megatron, we use Megatron models rather than HF models, so this is the Megatron analog to `HFWrappedModel`. Also, this class is used for non-Policy roles (e.g., Reference), so `Policy` is misleading

### DIFF
--- a/skyrl-train/skyrl_train/workers/fsdp/fsdp_worker.py
+++ b/skyrl-train/skyrl_train/workers/fsdp/fsdp_worker.py
@@ -304,7 +304,6 @@ class FSDPCriticWorkerBase(CriticWorkerBase):
         return output
 
 
-
 class FSDPRefWorkerBase(RefWorkerBase):
     def offload_to_cpu(self, pin_memory=True, non_blocking=True):
         self._set_numa_affinity(torch.distributed.get_rank() % torch.cuda.device_count())


skyrl-train/skyrl_train/workers/fsdp/fsdp_worker.py
```diff
@@ -27,7 +27,7 @@
 )
 
 
-class FSDPPolicyRayActorBase(PolicyWorkerBase):
+class FSDPPolicyRayWorkerBase(PolicyWorkerBase):
```
erictang000: The Deepspeed version is just: `DeepSpeedPolicyWorkerBase` - Maybe we could remove the `Ray` specifier from this class as well for consistency?

skyrl-train/skyrl_train/model_wrapper.py
```diff
@@ -19,11 +19,11 @@
 from flash_attn.bert_padding import pad_input, unpad_input
```
erictang000, line 18: should this file be renamed to `model_wrapper.py` for consistency?

skyrl-train/skyrl_train/workers/fsdp/fsdp_worker.py
tyler-griggs: Good call!

skyrl-train/skyrl_train/model_wrapper.py
tyler-griggs: yes good call.

erictang000 (approved):

erictang000:
@tyler-griggs sorry for the delay - lgtm, just need to merge main + resolve conflicts. Thanks!

PR #363, opened 2025-09-30
[megatron] Docs for Megatron

erictang000:


skyrl-train/docs/examples/images/parallelism.svg
SumanthRH, line 1: Could you label what the arrows mean ( like black arrow is input data and purple is model outputs)

skyrl-train/docs/examples/megatron.rst
```diff
@@ -0,0 +1,151 @@
...
+    export PYTHONPATH="/home/ray/anaconda3/lib/python3.12/site-packages"
+
+Flash Attention
+~~~~~~~~~~~~~~~
+Next, in order to use flash attention with the megatron backend, you must use ``flash_attn`` version ``2.7.4.post1`` or lower for compatibility with ``TransformerEngine==2.5.0``.
+You can replace the ``flash-attn`` wheel in the ``pyproject.toml`` file with the following to use the ``2.7.4.post1`` release, and you can find wheels for other versions `here <https://github.com/Dao-AILab/flash-attention/releases>`_.
+
+.. code-block:: bash
+
+    flash-attn = { url = "https://github.com/Dao-AILab/flash-attention/releases/download/v2.7.4.post1/flash_attn-2.7.4.post1+cu12torch2.7cxx11abiFALSE-cp312-cp312-linux_x86_64.whl" }
```
SumanthRH, line 106: Hmm could we revisit this ? is there no way to have a different flash attn source for mcore?

I think there are some issues with the extra marker with `uv` , not sure if we can work around that

skyrl-train/docs/examples/megatron.rst
```diff
@@ -0,0 +1,151 @@
...
+
+    # @package megatron_config.policy
+    tensor_model_parallel_size: 1
+    pipeline_model_parallel_size: 1
+    context_parallel_size: 1
+    expert_model_parallel_size: 1
+    expert_tensor_parallel_size: null
+
+    ddp_config:
+        # pass through kwargs to configure the Megatron DistributedDataParallelConfig object
```
SumanthRH:
```suggestion
        # pass-through kwargs to configure the Megatron DistributedDataParallelConfig object
```

skyrl-train/docs/examples/megatron.rst
```diff
@@ -0,0 +1,151 @@
...
+        # https://github.com/NVIDIA/Megatron-LM/blob/core_r0.13.0/megatron/core/distributed/distributed_data_parallel_config.py#L8
+        ...
+    
+    optimizer_config_kwargs:
+        # pass through kwargs to configure the Megatron OptimizerConfig object
```
SumanthRH:
```suggestion
        # pass-through kwargs to configure the Megatron OptimizerConfig object
```

skyrl-train/docs/examples/megatron.rst
```diff
@@ -0,0 +1,151 @@
...
+        # https://github.com/NVIDIA/Megatron-LM/blob/core_r0.13.0/megatron/core/optimizer/optimizer_config.py#L12
+        ...
+    
+    transformer_config_kwargs:
+        # pass through kwargs to configure the Megatron TransformerConfig object
```
SumanthRH:
```suggestion
        # pass-through kwargs to configure the Megatron TransformerConfig object
```

skyrl-train/docs/configuration/config.rst
```diff
@@ -158,6 +158,47 @@ We use the same configuration group for FSDP1 and FSDP2
...
+.. code-block:: yaml
+
+    megatron_config:
+      tensor_model_parallel_size: 1 
+      pipeline_model_parallel_size: 1
+      context_parallel_size: 1
+      expert_model_parallel_size: 1
+      expert_tensor_parallel_size: null
+
+      ddp_config: # pass through config to Megatron DDPConfig object
```
SumanthRH:
```suggestion
      ddp_config: # pass-through config to Megatron's `DDPConfig` object
```

skyrl-train/docs/configuration/config.rst
```diff
@@ -158,6 +158,47 @@ We use the same configuration group for FSDP1 and FSDP2
...
+        # https://github.com/NVIDIA/Megatron-LM/blob/core_r0.13.0/megatron/core/distributed/distributed_data_parallel_config.py#L8
+        ...
+      optimizer_config_kwargs: # pass through kwargs to Megatron OptimizerConfig object
```
SumanthRH:
```suggestion
      optimizer_config_kwargs: # pass-through kwargs to Megatron's `OptimizerConfig` object
```

skyrl-train/docs/configuration/config.rst
```diff
@@ -158,6 +158,47 @@ We use the same configuration group for FSDP1 and FSDP2
...
+        # any overlapping arguments with those we attempt to resolve in trainer.policy.optimizer_config will be overridden by the values here
+        # https://github.com/NVIDIA/Megatron-LM/blob/core_r0.13.0/megatron/core/optimizer/optimizer_config.py#L12
+        ...
+      model_config_kwargs: # pass through kwargs to the Huggingface model config (i.e. for overriding vocab size, etc)
```
SumanthRH:
```suggestion
      model_config_kwargs: # pass-through kwargs to the HuggingFace model config (i.e. for overriding vocab size, etc)
```

skyrl-train/docs/configuration/config.rst
```diff
@@ -158,6 +158,47 @@ We use the same configuration group for FSDP1 and FSDP2
...
+        ...
+      transformer_config_kwargs: # pass through kwargs to the Megatron TransformerConfig object
```
SumanthRH:
```suggestion
      transformer_config_kwargs: # pass-through kwargs to Megatron's `TransformerConfig` object
```

skyrl-train/docs/examples/megatron.rst
erictang000: i think we tried for a while last time and it seems not possible at the moment. maybe i can try opening an issue on uv?

skyrl-train/docs/examples/megatron.rst
```diff
@@ -0,0 +1,151 @@
...
+   :header-rows: 0
+
+   * - .. image:: images/search-r1-3b.svg
+         :width: 400px
+         :align: center
+     - .. image:: images/search-r1-30b.svg
+         :width: 400px
+         :align: center
+
+.. centered:: Left: Matching Qwen2.5-3B-Instruct reward curves for Megatron and FSDP2. Right: Qwen3-30B-A3B reward curve for Megatron (330 steps on 4 H100 nodes over 4 days) vs Qwen2.5-3B FSDP baseline.
```
SumanthRH: The right curve is bit weird - why are we comparing Qwen-3-30B on Megatron with Qwen 2.5 3B on FSDP again?

skyrl-train/docs/examples/megatron.rst
erictang000: i think the main blocking issue was that we need flash-attn for the base trainer with no extras, and then with markers there's no way to express that you should use a different version only if the mcore extra is used (because you can't express "mcore not in extras" in a concise way)

skyrl-train/docs/examples/megatron.rst
erictang000: we don't have baseline for 30B with FSDP, but I wanted a reference to show that the curve is normal

but i'm indifferent, if you think it's weird we could change it to just the 30B curve

SumanthRH: yeah let's just have the 30B reward curve without the 3B+ FSDP curve

erictang000:
image
image
image
image
image

erictang000:
image
image

skyrl-train/docs/examples/megatron.rst
SumanthRH: > i think the main blocking issue was that we need flash-attn for the base trainer with no extras, and then with markers there's no way to express that you should use a different version only if the mcore extra is used (because you can't express "mcore not in extras" in a concise way)

Yeah frankly i don't think this is permitted with PEP 508: https://peps.python.org/pep-0508/#extras.  Extras are meant to specify additional dependencies only. 


I was thinking more like moving `flash-attn` to all the existing extras we have - and things will mostly work ok because a user will always choose one extra like vllm to run their stuff . 

But this still doesn't work right now with uv. I might try to get a min repro later. 

I created an issue for now: https://github.com/NovaSky-AI/SkyRL/issues/369

PR #381, opened 2025-10-03
[trainer] Remove weights manager usage from the training loop

### LINKED ISSUE #312: Remove the `InferenceWeightsManager` for the training loop
Per discussion: https://github.com/NovaSky-AI/SkyRL/pull/297#discussion_r2350193321

erictang000:
Closes #312.

Changes the training loop standard to have the inference engine weights on gpu at the beginning of the training loop, and removes the weight manager logic for simplicity. Now the flow for colocated will be:

```python
for i in range(n):
  rollout 
  offload rollout
  train
  onload rollout weights
  sync weights
  eval
```

Makes the same change for AsyncTrainer. I'd like to delete the weights_manager.py file, but it's currently in use by the SkyRLAgent code - leaving a TODO for that. 

Also makes `colocate_all` an instance variable because we were accessing `self.cfg.trainer.placement.colocate_all` like 20 times in the file.

For colocated, gpu memory is the same as before: 
image
image

Async works: 
image


skyrl-train/skyrl_train/trainer.py
```diff
@@ -295,15 +291,6 @@ def train(self):
...
-            if inference_engine_is_active and self.cfg.trainer.placement.colocate_all:
-                asyncio.run(self.inference_engine_client.sleep())
-                self.policy_model.backload_to_gpu()
-                inference_engine_is_active = False
-            if self.cfg.trainer.update_ref_every_epoch and self.ref_model is not None:
-                with Timer("update_ref_with_policy", self.all_timings):
-                    self.update_ref_with_policy()
-
         pbar.close()
         if self.cfg.trainer.ckpt_interval > 0:
```
SumanthRH, line 300: With colocation, policy model weights are now on CPU at the end of the training step. 

The checkpointing code will break here after the end of training - we assume weights on GPU afaik. 

Can you fix and test final checkpointing?

skyrl-train/skyrl_train/trainer.py
```diff
@@ -258,27 +234,47 @@ def train(self):
...
-                if self.cfg.trainer.hf_save_interval > 0 and self.global_step % self.cfg.trainer.hf_save_interval == 0:
-                    with Timer("save_hf_model", self.all_timings):
-                        self.save_models()
-
-                # 6. set logs
+                    # 5. conditionally save checkpoints and hf model
+                    if self.cfg.trainer.ckpt_interval > 0 and self.global_step % self.cfg.trainer.ckpt_interval == 0:
+                        with Timer("save_checkpoints", self.all_timings):
+                            self.save_checkpoints()
+                    if (
```
SumanthRH, line 242: Wait, after this change, checkpointing will be included in step time? 

Shouldn't this be outside the `Timer("step")` context manager like before? I notice more indentation

erictang000: hmm yeah so this was on purpose, because now we need these to happen before we sync weights at the end of the loop

i think preserving this order is the best in order to not unnecessarily offload/backload the policy model. I think the best solution might just be to remove the Timer("step") context manager, and just report overall step times by summing everything in `self.all_timings` that isn't checkpointing.

SumanthRH: seems ok then

skyrl-train/skyrl_train/trainer.py
erictang000: Tested!

image

skyrl-train/skyrl_train/trainer.py
```diff
@@ -258,27 +235,48 @@ def train(self):
...
+
+                    # 7. sync weights to inference engines
+                    if self.colocate_all:
+                        asyncio.run(self.inference_engine_client.wake_up(tags=["weights"]))
+                    with Timer("sync_weights", self.all_timings):
+                        ray.get(self.sync_policy_weights_to_inference_engines())
+                    if self.colocate_all:
+                        with Timer("offload_policy_model_to_cpu"):
+                            self.policy_model.offload_to_cpu()
+                        asyncio.run(self.inference_engine_client.wake_up(tags=["kv_cache"]))
```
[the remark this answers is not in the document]
erictang000: eh maybe in next pr after figuring out optimizer offload

SumanthRH (approved):

SumanthRH:
We should do a follow-up PR to migrate the trainer in SkyRL-Agent!

PR #382, opened 2025-10-03
[fix] Refresh s3 credentials when expired

tyler-griggs:
When exporting checkpoints to s3fs, credentials can expire and checkpoints fail to upload. This PR refreshes s3 credentials when near expiration.

Also, update a couple `_strategy.py` files to correctly use the `io` library.


skyrl-train/s3_upload.py
SumanthRH, line 1: is this file meant to be removed? 

if this is helpful to have for a sanity check? maybe move to `scripts/`? If we're keeping it, would need a note in the docs then, atleast a short section in Troubleshoooting

SumanthRH: hmm tis seems like a basic test for credentials, can probably remove?

tyler-griggs: Oops, yes, this wasn't supposed to be pushed

PR #385, opened 2025-10-03
[trainer] Add option to separately offload optimizer and model weights (WIP)

erictang000:
Allows for more fine grained memory control, which helps reduce peak memory while syncing weights and doing a fwd pass

With Qwen 2.5-7B and the FSDP2 backend:

Before:
image
image

After:
image
image


SumanthRH (approved):
Can you add a test in `skyrl-train/tests/gpu/test_worker_offload.py` with the new flag?

erictang000:
added tests + verified all passing!

PR #389, opened 2025-10-04
[skyrl-tx] Add MultiLoRA support and apply it to the MLP layer

pcmoritz:
This PR implements MultiLoRA support for the models and applies it to the MLP layer.

### DIFF
--- a/skyrl-tx/tx/layers/lora.py
+++ b/skyrl-tx/tx/layers/lora.py
@@ -1,0 +1,100 @@
+from flax import nnx
+import jax
+from jax import numpy as jnp
+
+
+def Param(*shape: int, dtype: jnp.dtype, kernel_init: nnx.Initializer, rngs: nnx.Rngs):
+    return nnx.Param(kernel_init(rngs.param(), shape, dtype))
+
+
+class LoRAMixin:
+    """A mixin for flax NNX modules to add multi-adapter LoRA support.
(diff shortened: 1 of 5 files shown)


skyrl-tx/tx/layers/lora.py
```diff
@@ -0,0 +1,106 @@
...
+            self.lora_B = None
+        else:
+            self.lora_scaling = Param(
+                max_lora_adapters, dtype=dtype,
+                kernel_init=nnx.initializers.constant(1.0), rngs=rngs,
+            )
+            self.lora_A = Param(
+                max_lora_adapters, in_features, max_lora_rank, dtype=dtype,
+                kernel_init=nnx.with_partitioning(
+                    nnx.initializers.he_uniform(),
```
pcmoritz, line 41: It is worth checking this. The blog https://thinkingmachines.ai/blog/lora/ mentions they use a uniform distribution for A with scale 1 / sqrt(in_features) and there is also https://github.com/microsoft/LoRA/blob/a0a92e0f26c067cf94747bdbf1ce73793fa44d19/loralib/layers.py#L124

I believe this should be right and the same as pytorch though, see https://discuss.pytorch.org/t/clarity-on-default-initialization-in-pytorch/84696 (namely a = sqrt(5) corresponds to uniform distribution with scale 1 / sqrt(in_features), corresponds to he_uniform).

PR #393, opened 2025-10-05
Fix Megatron batch validation

Harryllh:
When using Megatron, if `cfg.trainer.policy.sequence_parallel_size` is set to 1, `policy_dp_size` will always be the world size. However, if any of PP, TP, CP is positive, this does not hold. This PR added the correct way to calculate `policy_dp_size` and `ref_dp_size` when using Megatron.

### DIFF
--- a/skyrl-train/skyrl_train/utils/utils.py
+++ b/skyrl-train/skyrl_train/utils/utils.py
@@ -65,7 +65,22 @@ def validate_batch_sizes(cfg: DictConfig):
 
     # Validate policy mini batch size
     policy_world_size = cfg.trainer.placement.policy_num_nodes * cfg.trainer.placement.policy_num_gpus_per_node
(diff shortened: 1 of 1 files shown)


erictang000:
@Harryllh Actually, looks like 2 cpu tests are failing (because of issues with the tests):

```bash
FAILED tests/cpu/test_trainer.py::test_validate_batch_sizes - omegaconf.errors.ConfigAttributeError: Missing key strategy
    full_key: trainer.strategy
    object_type=dict
FAILED tests/cpu/test_trainer.py::test_validate_batch_sizes_lcm_dp_requirement - omegaconf.errors.ConfigAttributeError: Missing key strategy
    full_key: trainer.strategy
    object_type=dict
```

could you take a look/maybe add a test for the megatron case specifically? Thanks!

PR #404, opened 2025-10-06
[Fix] Fix chat templating in Mini-SWE-Agent and Terminal-Bench examples

SumanthRH:
# What does this PR do?

Fixes chat templating in the Mini-swe-agent and the terminal bench examples. 

Previously, we were naively callling `.apply_chat_template` to encode response messages turn by turn - but this can append system messages for each turn depending on the model. 

We use the fixed base approach similar to what we do in the SkyRLGymGenerator.

### DIFF
--- a/skyrl-train/examples/mini_swe_agent/mini_swe_generator.py
+++ b/skyrl-train/examples/mini_swe_agent/mini_swe_generator.py
@@ -19,6 +19,7 @@ from skyrl_train.inference_engines.inference_engine_client import InferenceEngin
 from skyrl_train.inference_engines.utils import get_sampling_params_for_backend
 from skyrl_train.generators.utils import (
     get_rollout_metrics,
+    encode_messages_subset,
 )
 
 
@@ -128,6 +129,9 @@ class MiniSweAgentGenerator(SkyRLGymGenerator):
         self.model_name = model_name
         self.litellm_model_name = "openai/" + self.model_name
 
+        if self.generator_cfg.chat_template.name_or_path is not None:
+            raise NotImplementedError("MiniSWEAgentGenerator doesn't support custom chat template")
+
     async def minisweagent_agent_loop(
         self,
         prompt: ConversationType,
@@ -182,7 +186,7 @@ class MiniSweAgentGenerator(SkyRLGymGenerator):
 
         for message in response_messages:
             # Apply chat template and tokenize each message
-            msg_encoding = self.tokenizer.apply_chat_template([message], add_generation_prompt=False, tokenize=True)
+            msg_encoding = encode_messages_subset([message], self.tokenizer)
 
             # Extend response_ids with the tokens
             response_ids.extend(msg_encoding)
--- a/skyrl-train/skyrl_train/generators/utils.py
+++ b/skyrl-train/skyrl_train/generators/utils.py
@@ -3,6 +3,7 @@ from typing import List, Tuple, Union, Optional, Dict, Any
 from collections import defaultdict
 import numpy as np
 from skyrl_train.generators.base import GeneratorOutput, GeneratorInput, TrajectoryID, BatchMetadata, TrainingPhase
+from skyrl_train.inference_engines.base import ConversationType
 from omegaconf import DictConfig
 
 CUSTOM_CHAT_TEMPLATES = {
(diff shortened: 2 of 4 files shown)


skyrl-train/skyrl_train/generators/utils.py
```diff
@@ -267,3 +268,69 @@ def prepare_generator_input(
...
+        tokenize=True,
+    )
+
+    # We remove tokens after the last EOS token so that it can be captured in `observation_ids`.
+    # For details, see https://skyrl.readthedocs.io/en/latest/tutorials/skyrl_gym_generator.html#multi-turn-tokenization-and-ti-to
+    if messages[0]["role"] != "assistant" and tokenizer.eos_token_id in base_conversation_token_ids:
+        last_eos_token_index = (
+            len(base_conversation_token_ids) - 1 - base_conversation_token_ids[::-1].index(tokenizer.eos_token_id)
+        )
+        base_conversation_token_ids = base_conversation_token_ids[: last_eos_token_index + 1]
```
CharlieFRuan: I am guessing we do not need these tricks if we are re-tokenizing an entire chat history.

```python
if messages[0]["role"] != "assistant" and tokenizer.eos_token_id in base_conversation_token_ids:
```

```python
if messages[-1]["role"] == "assistant" and tokenizer.eos_token_id in conversation_token_ids:
```

Could you try without these two clauses trick and see if the unit tests still pass? The unit tests' expected tokens need to be changed as well to reflect this.

I needed this trick in `skyrl_gym_generator` because the assistant token IDs are generated by the LLM (hence always ends with eos_token, and never the `\n`). Here the assistant tokens are from encoding, so there will be an `\n`.

skyrl-train/examples/mini_swe_agent/mini_swe_generator.py
```diff
@@ -182,7 +186,7 @@ async def minisweagent_agent_loop(
 
         for message in response_messages:
             # Apply chat template and tokenize each message
-            msg_encoding = self.tokenizer.apply_chat_template([message], add_generation_prompt=False, tokenize=True)
+            msg_encoding = encode_messages_subset([message], self.tokenizer)
 
             # Extend response_ids with the tokens
             response_ids.extend(msg_encoding)
```
CharlieFRuan, line 192: Can we add an assertion of 

`initial_input_ids + respons_ids == self.tokenizer.apply_chat_template(messages, add_generation_prompt=False, tokenize=True)`

And same for terminal bench?

At least a warning perhaps

CharlieFRuan (commented):
Thank you for the fix!

skyrl-train/skyrl_train/generators/utils.py
SumanthRH: Hmm I assumed this is our convention for loss masking?

SumanthRH: Should be okay either way ig, but weird to not be consistent

CharlieFRuan: As discussed offline, loss masking is indeed different from SkyRLGymGenerator but it should be fine.

We'll go with no special `\n` handling for now to make the helper more robust.

In the long term these examples should all be TI/TO, so will model after SkyRLGymGenerator's handlings.

skyrl-train/examples/mini_swe_agent/mini_swe_generator.py
SumanthRH: Hmm such an  assertion or warning can be misleading or incorrect because applying the chat template message by message can be pretty different from full conversation. 

For models like Qwen 3 - the thinking tokens for previous messages in the history are discarded by default. Now, if we call `encode_messages_subset` on each message , we end up preserving hte think tokens for each message (even with base convo present it is okay). 


But then with the RHS - the think tokens for previous messages are removed. 

Now, I don't think either is the correct behaviour we want for on policy trainig, but in any case we shouldn't have this assertion.

SumanthRH: As such, for Qwen3 8B, I re-ran the mini swe agent example and it is okay - actually the previous expression was also correct, because for qwen 3 8B there is no default system prompt added;

```python
print(self.tokenizer.apply_chat_template([{"role": "assistant", "content": "What is 1+1?"}], tokenize=False))
# '<|im_start|>assistant\nWhat is 1+1?<|im_end|>\n'
```

SumanthRH: I believe the tests should be sufficient

CharlieFRuan: > For models like Qwen 3 - the thinking tokens for previous messages in the history are discarded by default. Now, if we call encode_messages_subset on each message , we end up preserving hte think tokens for each message (even with base convo present it is okay).

Good point... so the current behavior becomes, during inference we discard thinking tokens, and for training, we keep all thinking tokens.

Made an issue for this: https://github.com/NovaSky-AI/SkyRL/issues/410

skyrl-train/examples/mini_swe_agent/mini_swe_generator.py
```diff
@@ -128,6 +129,9 @@ def __init__(
         self.model_name = model_name
         self.litellm_model_name = "openai/" + self.model_name
 
+        if self.generator_cfg.chat_template.name_or_path is not None:
```
CharlieFRuan, line 132: nit: can we add it for terminal bench generator as well?

CharlieFRuan (approved):
LGTM! only one nit left, after which feel free to merge

PR #405, opened 2025-10-06
[tx] [WIP] Support different LoRA rank for each adapter

pcmoritz:
Propagate the variables introduced in https://github.com/NovaSky-AI/SkyRL/pull/395 to support different LoRA ranks per adapter.

### DIFF
--- a/skyrl-tx/tests/models/test_qwen3_lora_training.py
+++ b/skyrl-tx/tests/models/test_qwen3_lora_training.py
@@ -7,6 +7,7 @@ from huggingface_hub import snapshot_download
 
 from tx.models import Qwen3ForCausalLM
 from tx.utils.models import get_dtype, load_checkpoint
+from tx.layers.lora import update_adapter_config
 
 
 def test_lora_training():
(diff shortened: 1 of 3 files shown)


skyrl-tx/tx/layers/lora.py
```diff
@@ -104,3 +104,30 @@ def __call__(self, x: jax.Array, adapter_indices: jax.Array | None = None) -> ja
...
+    def update_lora_config(path, value):
+        path_str = str(path)
+        if 'lora_ranks' in path_str:
+            return value.at[adapter_index].set(lora_rank)
+        if 'lora_scaling' in path_str:
+            return value.at[adapter_index].set(scaling)
+        if 'lora_A' in path_str:
+            # Zero out columns beyond the rank for this adapter
+            return value.at[adapter_index, :, lora_rank:].set(0.0)
+        return value
```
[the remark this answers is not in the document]
pcmoritz: The `lora_B` parameters are already initialize to zero so this shouldn't be needed

skyrl-tx/tx/tinker/engine.py
```diff
@@ -111,12 +112,24 @@ def create_model(self, model_id: str, lora_config: dict | None = None):
         if adapter_index >= self.max_lora_adapters:
             raise ValueError(f"Maximum number of LoRA adapters ({self.max_lora_adapters}) reached")
 
+        # Extract LoRA rank and alpha from config
+        lora_rank = lora_config["r"]
+        lora_alpha = lora_config["lora_alpha"]
+
+        # Validate rank doesn't exceed max
+        if lora_rank > self.max_lora_rank:
```
tyler-griggs: Small nit: maybe also check for >0? It would currently cause a divide-by-0 error in `update_adapter_config`

skyrl-tx/tx/layers/lora.py
```diff
@@ -104,3 +104,29 @@ def __call__(self, x: jax.Array, adapter_indices: jax.Array | None = None) -> ja
         base_out = super().__call__(x)
         return self.apply_lora(x, base_out, adapter_indices)
 
+
+def update_adapter_config(model: nnx.Module, adapter_index: int, lora_rank: int, lora_alpha: float):
+    """Update lora_ranks and lora_scaling for a specific adapter across all LoRA layers.
```
tyler-griggs, line 109: nit: This might be obvious, but the expectation is that this is called _before_ training. That is, you should not update the same config for the same adapter index multiple times throughout training, which, eg breaks the `lora_B is already zero` below. . It could be worth mentioning this expectation, but not a big deal

skyrl-tx/tests/models/test_qwen3_lora_training.py
```diff
@@ -44,11 +49,24 @@ def loss_fn(model, input_ids, target_ids):
...
-        def get_adapter_2_params(params):
-            return jax.tree.map(lambda p: p[2].copy(), params)
-
-        initial_adapter_2_params = get_adapter_2_params(lora_params)
+        # Helper to extract adapter params at specific index
+        def get_adapter_params(params, adapter_idx):
+            return jax.tree.map(lambda p: p[adapter_idx].copy(), params)
+
+        # Helper to extract out-of-rank params for an adapter
+        def get_out_of_rank_params(params, adapter_idx, rank):
```
tyler-griggs, line 57: If the multi-lora setup requires more extensively using param tree walking, we might consider some lora-specific helper methods for getting/setting various metadata, but probably not needed now

tyler-griggs (approved):
lgtm

skyrl-tx/tx/layers/lora.py
pcmoritz: Great point, I added a note covering this in the docstring :)

PR #415, opened 2025-10-07
Rename db models file

### LINKED ISSUE #396: Rename tx/tinker/models.py to tx/tinker/db_models.py
We currently have two files called models.py -- one time models is referring to machine learning models and one time it is referring to database models. We should rename the DB models file to `db_models.py` to avoid confusion.

KranthiGV:
Closes #396

Confirmed it works by running the tests.


pcmoritz:
Great, thank you so much for your contribution!

PR #423, opened 2025-10-07
[tx] Setup CI workflow for tx

tyler-griggs:
* Setup CI workflow for tx tests. These currently run on CPU-only, but we will add GPU tests as well.
* Adds the formatter that SkyRL uses elsewhere


tyler-griggs:
Thanks! It's using `black`: https://github.com/psf/black

PR #432, opened 2025-10-08
[tx] Add MultiLoRA MoE layer

pcmoritz:
This PR implements MultiLoRA for the MoE layer. It is implemented by routing tokens to the (num_experts, num_adapters) dimensions of the lora_A and lora_B matrices, much in the same way that tokens are routed to the num_experts dimension of the experts matrices for normal MoE layers. We also add a test against merging each LoRA adapter individually into the base model to ensure numerical correctness of the implementation.

### DIFF
--- a/skyrl-tx/tx/layers/lora.py
+++ b/skyrl-tx/tx/layers/lora.py
@@ -183,16 +183,16 @@ class LoRAExpert(LoRAMixin, nnx.Module):
         # Reconstruct expert indices from group_sizes
         expert_indices = jnp.repeat(jnp.arange(self.num_experts), group_sizes, total_repeat_length=x.shape[0])
 
-        # Combine (adapter, expert) into a single routing dimension.
-        combined_indices = adapter_indices_sorted * self.num_experts + expert_indices
-        max_combined = self.max_lora_adapters * self.num_experts
+        # Flatten (adapter, expert) into a single routing dimension.
+        flattened_indices = adapter_indices_sorted * self.num_experts + expert_indices
+        num_flattened_groups = self.max_lora_adapters * self.num_experts
 
         # Reshape lora_A and lora_B to merge (max_lora_adapters, num_experts) dimensions
-        lora_A_reshaped = self.lora_A.value.reshape(max_combined, self.in_features, self.max_lora_rank)
-        lora_B_reshaped = self.lora_B.value.reshape(max_combined, self.max_lora_rank, self.out_features)
+        lora_A_reshaped = self.lora_A.value.reshape(num_flattened_groups, self.in_features, self.max_lora_rank)
+        lora_B_reshaped = self.lora_B.value.reshape(num_flattened_groups, self.max_lora_rank, self.out_features)
 
         # Sort tokens by combined index
-        x_sorted, combined_group_sizes, unsort_indices, _ = prepare_routing(x, combined_indices, max_combined)
+        x_sorted, combined_group_sizes, unsort_indices, _ = prepare_routing(x, flattened_indices, num_flattened_groups)
 
         # Apply LoRA using ragged_dot: x @ A @ B
         intermediate = jax.lax.ragged_dot(x_sorted, lora_A_reshaped, combined_group_sizes)


skyrl-tx/tests/models/test_qwen3.py
```diff
@@ -98,6 +104,65 @@ def load_lora_weights(
     jax_module.lora_ranks.value = jax_module.lora_ranks.value.at[adapter_idx].set(rank)
 
 
+def test_qwen3_moe_layer_lora():
```
tyler-griggs, line 107: nice test!

skyrl-tx/tests/models/test_qwen3.py
```diff
@@ -52,6 +53,15 @@ def test_qwen3(tp: int):
         assert np.allclose(hf_outputs.hidden_states[-1], outputs["hidden_states"][-1], rtol=1e-3, atol=1e-3)
 
 
+def load_moe_base_weights(jax_moe_layer: Qwen3MoeSparseMoeBlock, hf_moe_layer: HFQwen3MoeSparseMoeBlock) -> None:
+    """Load base weights from HF MoE layer to JAX MoE layer."""
+    jax_moe_layer.gate.kernel[:] = hf_moe_layer.gate.weight[:].detach().numpy().T
```
sarahdong: do you need the `[:]` to the weight?

skyrl-tx/tests/models/test_qwen3.py
```diff
@@ -98,6 +104,65 @@ def load_lora_weights(
...
+    config = AutoConfig.from_pretrained(model_name)
+
+    # Enable LoRA
+    config.max_lora_adapters = 3
+    config.max_lora_rank = 4
+
+    hf_moe_layer = hf_model.model.layers[0].mlp
+    x = torch.randn(3, 4, config.hidden_size)
+
+    mesh = jax.make_mesh((1, 1), ("dp", "tp"))
```
sarahdong, line 120: what about tp > 1?

skyrl-tx/tx/layers/lora.py
```diff
@@ -135,6 +135,76 @@ def __call__(self, x: jax.Array, adapter_indices: jax.Array | None = None) -> ja
...
+        self.num_experts = num_experts
+        self.in_features = in_features
+        self.out_features = out_features
+
+        self.weight = Param(num_experts, in_features, out_features, dtype=dtype, kernel_init=kernel_init, rngs=rngs)
+
+        assert self.weight.value.sharding is not None, "LoRAExpert layer needs sharding"
+        sharding = self.weight.value.sharding.spec
+        self.init_lora(
+            max_lora_adapters=max_lora_adapters,
```
sarahdong, line 162: do you want to leave an empty adapter(e.g. all zeros) to have a safe path for cases where you don't match a lora idx, or just simply want the base model?

pcmoritz: Thanks, that makes sense! We could reserve lora index 0 for a lora adapter that is initialized with all zeros so it basically acts as the base model. It probably doesn't make sense to do it as part of this PR since this PR only implements the MoE LoRA, I'll keep it in mind for the future when the need arises and then can do it in a separate PR.

Slightly related but also different, you can also currently set `adapter_indices=None` so no LoRA adapters are used, this will likely be used for implementing full parameter finetuning in the future (but obviously every full parameter finetuning job will need a dedicated full model).

skyrl-tx/tx/layers/lora.py
```diff
@@ -135,6 +135,76 @@ def __call__(self, x: jax.Array, adapter_indices: jax.Array | None = None) -> ja
...
+        base_out = jax.lax.ragged_dot(x, self.weight.value, group_sizes)
+
+        if self.max_lora_adapters == 0 or adapter_indices_sorted is None:
+            return base_out
+
+        # Reconstruct expert indices from group_sizes
+        expert_indices = jnp.repeat(jnp.arange(self.num_experts), group_sizes, total_repeat_length=x.shape[0])
+
+        # Combine (adapter, expert) into a single routing dimension.
+        combined_indices = adapter_indices_sorted * self.num_experts + expert_indices
```
sarahdong: what about flattened_idxes?

skyrl-tx/tests/models/test_qwen3.py
pcmoritz: It is actually not needed, I removed it :)

skyrl-tx/tx/layers/lora.py
pcmoritz: I renamed it now :)

Issue #398, opened 2025-10-06
Implement more LoRA layers


pcmoritz:
Currently only LoRA for the MLP is supported. We should also support it for other layers like

- [x] attention layers
- [ ] embedding layers
- [x] MoE Layers

This can be done in a similar way to how it is done for the MLP using the LoRAMixin, see https://github.com/NovaSky-AI/SkyRL/blob/f8ffa57d5186b624098b4141e0988a5fc2f50576/skyrl-tx/tx/layers/lora.py

The MoE will be a little more interesting to implement, it could for example be done with two invocations of `ragged_dot`, one for the lora_A and one for the lora_B parameter. It might also be worth studying the vLLM MoE LoRA implementation https://github.com/vllm-project/vllm/pull/21229

StreptoBordos:
Hello, I would like to try and work on this issue.

pcmoritz:
That's great, looking forward to the contribution! It is probably best to start with one of the layer types, one that you think is simpler to implement and create a PR for it.  In order to check whether it is correct, you can add it here https://github.com/NovaSky-AI/SkyRL/blob/main/skyrl-tx/tests/models/test_qwen3.py#L111 and the test will compare the numerics against hugging face transformers :)

va1shn9v:
Hi, can I simultaneously work on implementing it for another layer in this case `embedding layer`

pcmoritz:
@va1shn9v 

Yes, that would be great, contributions welcome :)

I'm currently implementing the MultiLoRA MoE layer.

PR #435, opened 2025-10-08
[tx] Add LoRA to attention layers

tyler-griggs:
Partially addresses #398

* Remove `MultiHeadProj` in favor of using `LoRALinear` for Q/K/V/O projections
* Shard projections along `num_heads * head_dim` axis -- to reconsider as we work on multi-node. vLLM also flattens to `num_heads * head_dim` but also asserts that `num_heads % tp_size == 0`. To start, we will make the same assumption, but to be reconsidered with multi-node.

### DIFF
--- a/skyrl-tx/tx/models/qwen3.py
+++ b/skyrl-tx/tx/models/qwen3.py
@@ -1,6 +1,7 @@
 from flax import nnx
 import jax
 from jax import numpy as jnp
+from jax.sharding import get_abstract_mesh
 from transformers import Qwen3Config
 
 from tx.layers.lora import LoRALinear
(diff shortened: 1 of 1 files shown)


skyrl-tx/tx/models/qwen3.py
```diff
@@ -47,41 +35,51 @@ def __init__(self, config: Qwen3Config, *, dtype: jnp.dtype, rngs: nnx.Rngs) ->
...
         self.num_kv_heads = config.num_key_value_heads
         self.head_dim = getattr(config, "head_dim", None) or config.hidden_size // self.num_heads
-
-        self.q_proj = MultiHeadProj(
-            "BMK,KNH->BMNH",
-            config.hidden_size,
-            self.num_heads,
-            self.head_dim,
-            sharding=jax.P(None, "tp", None),
+        self.max_lora_adapters = getattr(config, "max_lora_adapters", 0)
```
pcmoritz: We don't need these in the forward, right? So no need to assign them to `self`, we can just keep them normal variables -- that's also consistent with the Qwen3MLP layer

PR #436, opened 2025-10-09
[Examples] Add an example for step-wise training

SumanthRH:
# What does this PR do?

Adds an example for step-wise training where each turn is represented as an individual sample in the batch. 

Currently, the example still assumes outcome rewards. 

Implements : 
- A custom generator for providing inputs and outputs at each step as an individual sample
- A custom trainer that can handle step-wise generator output 
- A custom advantage estimation function that will compute advantages for the last step and broadcast it to the other steps in that trajectory
- A custom evaluation function to calculate metrics correctly 

Currently this uses TITO with multi-turn chat templating for a simple demonstration. We simply append responses and observations to a running list of input ids. The generator is not yet compatible with qwen3 or gpt-oss like chat templating - where think tokens are removed - this will be added as a follow-up

There are many bits that can be cleaned up (for example, padding logic is brittle at the moment given special handling for the tensor `is_last_step`) but it works as  an initial example. 

I've tested convergence with SkyRL2SQL for the first 20 steps and it seems to match the original wandb curve

Screenshot 2025-10-08 at 4 39 13 PM
Screenshot 2025-10-08 at 4 39 25 PM

Original curve for reference: https://wandb.ai/sky-posttraining-uc-berkeley/skyrl-sql/reports/SkyRL-SQL---VmlldzoxMzM0MTAyMw?accessToken=vrqncoa32qcobvvpuo672yji4gweguk6tjxvaflk1zh73fn70j6l5rj8j619uvry

### DIFF
--- a/skyrl-train/examples/step_wise/step_wise_trainer.py
+++ b/skyrl-train/examples/step_wise/step_wise_trainer.py
@@ -280,7 +280,12 @@ class StepWiseTrainer(RayPPOTrainer):
         if generator_output["rollout_metrics"] is not None:
             self.all_metrics.update(generator_output["rollout_metrics"])
 
-        # don't validate - will error out
+        # Skip validation.
+        # NOTE (sumanthrh): `validate_generator_output` checks if the number of
+        # rows in the `input_batch` is the same as that in `generator_output`
+        # In step wise training, since each step of a row / trajectory
+        # in `input_batch` is represented as a row in `generator_output`,
+        # this check doesn't apply and it is skipped.
         # validate_generator_output(input_batch, generator_output)
 
         return generator_output


skyrl-train/examples/step_wise/step_wise_trainer.py
```diff
@@ -0,0 +1,328 @@
...
+            dp_size = math.lcm(dp_size, self.critic_model.actor_infos[0].rank.dp_size)
+        if self.ref_model is not None:
+            dp_size = math.lcm(dp_size, self.ref_model.actor_infos[0].rank.dp_size)
+
+        pad_size = math.ceil(training_input.batch_size / dp_size) * dp_size - training_input.batch_size
+        new_tensors = {}
+        for key, tensor in training_input.items():
+            if tensor is not None:
+                additional_dims = tuple(tensor.shape[1:]) if len(tensor.shape) > 1 else ()
+                if key in ["attention_mask", "is_last_step"]:
```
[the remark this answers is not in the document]
SumanthRH: Making the attention mask fully zero is not compatible with our sequence packing logic and will lead to an error.


There's definitely a performance improvement possible by supporting this, but this can be done later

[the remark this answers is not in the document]
SumanthRH: is_last_step should not be zeros for padding, that is completely incorrect

skyrl-train/examples/step_wise/step_wise_evaluate.py
```diff
@@ -0,0 +1,138 @@
...
+    )
+    # 2.1 only use the final step metris
+    generator_output_last_step = defaultdict(list)
+    for key in concat_generator_outputs:
+        if isinstance(concat_generator_outputs[key], list):
+            generator_output_last_step[key] = [
+                concat_generator_outputs[key][i]
+                for i in range(len(concat_generator_outputs[key]))
+                if concat_generator_outputs["is_last_step"][i]
+            ]
```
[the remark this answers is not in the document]
SumanthRH: Done!

skyrl-train/examples/step_wise/step_wise_trainer.py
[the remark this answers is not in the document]
SumanthRH: Will keep this as unresolved for context, but I've improved the attention mask related handling by making sure to use attention entries from the input tensor itself for the padding tensor. loss mask is ensured to be all zeros, and is_last_step still requires special handling.  @gemini-code-assist

skyrl-train/examples/step_wise/step_wise_trainer.py
```diff
@@ -0,0 +1,340 @@
...
+
+        # Check if rewards are already token-level (List[List[float]]) or response-level (List[float])
+        if rewards and isinstance(rewards[0], list):
+            # Token-level rewards: rewards is List[List[float]]
+            per_token_rewards = rewards
+        else:
+            # Response-level rewards: rewards is List[float], convert to per-token rewards
+            for reward, response in zip(rewards, responses):
+                per_token_reward = [0.0] * len(response)
+                per_token_reward[-1] = float(reward)
```
tyler-griggs, line 177: We turn response-level rewards to token-level rewards by placing the reward in the final token. But, what if the final turn returned observations from the environment. Then, I believe the generator will place these observations at the very end of the response, which will be included in the response_ids here. But, we mask out the observations, and so this reward may not actually be used. Is this possible?

If so, this is probably an error that might already exist in the codebase. I'm not sure.

skyrl-train/examples/step_wise/step_wise_trainer.py
```diff
@@ -0,0 +1,340 @@
...
+                    padding_tensor = tensor[:pad_size].clone()
+                new_tensors[key] = torch.cat([tensor, padding_tensor], dim=0)
+
+        new_training_input = TrainingInputBatch(new_tensors)
+        new_training_input.metadata = {}
+        new_training_input.metadata["uids"] = training_input.metadata["uids"] + [f"pad{i}" for i in range(pad_size)]
+        new_training_input.metadata["trajectory_ids"] = training_input.metadata["trajectory_ids"] + [
+            f"pad{i}" for i in range(pad_size)
+        ]
+        for key, value in training_input.metadata:
```
tyler-griggs: Wait, how does this work? `metadata` is a dict, right? So then it should require `metadata.items()`to iterate correctly

skyrl-train/examples/step_wise/step_wise_trainer.py
```diff
@@ -0,0 +1,340 @@
...
+        pad_size = math.ceil(training_input.batch_size / dp_size) * dp_size - training_input.batch_size
+        new_tensors = {}
+        if pad_size == 0:
+            return training_input
+        for key, tensor in training_input.items():
+            if tensor is not None:
+                additional_dims = tuple(tensor.shape[1:]) if len(tensor.shape) > 1 else ()
+
+                if key == "is_last_step":
+                    padding_tensor = torch.ones(pad_size, *additional_dims, dtype=tensor.dtype, device=tensor.device)
```
tyler-griggs, line 313: I didn't follow why this should be `ones` instead of `zeros` -- can you explain?

skyrl-train/examples/step_wise/step_wise_trainer.py
SumanthRH: oops adding this bit while cleaning up today, fixing

skyrl-train/examples/step_wise/step_wise_trainer.py
SumanthRH: > But, we mask out the observations, and so this reward may not actually be used. Is this possible?

Hmm I don't think it matters because of how we calculate advantages and returns right now? We ignore loss mask, and just sum over rewards for all the tokens;

https://github.com/NovaSky-AI/SkyRL/blob/722c90c06f72d20bace2b2e957c1d4198c1a74ed/skyrl-train/skyrl_train/utils/ppo_utils.py#L909

SumanthRH: rewards tensor basically doesn't matter after advantages have been computed,  but agreed that this can break depending on the advantage estimator function used and the environment implementation

SumanthRH: Will probably just move this inside the generator for now

skyrl-train/examples/step_wise/step_wise_trainer.py
SumanthRH: is_last_step needs to have a one non-zero entry for each trajectory - each trajctory has atleast one last step... if you add all zeros, that means the padding trajectories have no last step at all.

SumanthRH: Ideally `pad_batch` is very generic and we can use it in other examples as well, but `is_last_step` is special and I'd rather do the padding correctly here

skyrl-train/examples/step_wise/step_wise_trainer.py
SumanthRH: This should be fixed now!

skyrl-train/examples/step_wise/step_wise_generator.py
```diff
@@ -0,0 +1,363 @@
...
+        env_extras["max_turns"] = self.max_turns  # TODO(shu): move this to config
+        env_config = self.skyrl_gym_cfg.get(env_class, DictConfig({}))
+        env = skyrl_gym.make(env_class, env_config=env_config, extras=env_extras)
+
+        session_id = (
+            f"{trajectory_id.instance_id}_{trajectory_id.repetition_id}" if trajectory_id is not None else uuid4().hex
+        )
+        done = False
+
+        # Instantiate chat_history and chat_end_index, which are only used if `retokenize_chat_history==True`.
```
tyler-griggs:
```suggestion
        # Instantiate chat_history, which is only used if `retokenize_chat_history==True`.
```

skyrl-train/examples/step_wise/step_wise_generator.py
```diff
@@ -0,0 +1,363 @@
...
+        # Need copy here since the prompt is a list of messages and we are going to modify it.
+        chat_history = copy.deepcopy(prompt)
+
+        # init() returns the first prompt to be given to the model, and optional metadata dict
+        chat_history, _ = await self._run_in_executor_if_available(env.init, chat_history)
+        # initial_chat_history_length = len(chat_history)
+        # chat_end_index = len(chat_history)
```
tyler-griggs:
```suggestion
```

skyrl-train/examples/step_wise/step_wise_generator.py
```diff
@@ -0,0 +1,363 @@
...
+        input_ids = self.tokenizer.apply_chat_template(
+            chat_history,
+            add_generation_prompt=True,
+            chat_template=self.custom_chat_template,
+            tokenize=True,
+        )
+
+        initial_prompt_length = len(input_ids)
+        loss_mask = []  # this excludes the prompt
+        # rollout_logprobs = None
```
tyler-griggs:
```suggestion
```

skyrl-train/examples/step_wise/step_wise_generator.py
```diff
@@ -0,0 +1,363 @@
...
+        # Accumulate per-step rewards. Format: (reward, response_end_token_idx)
+        per_step_rewards: List[Tuple[float, int]] = []
+        step_id = 0
+        # full_trajectory = {"q": input_ids}
```
tyler-griggs:
```suggestion
```

skyrl-train/examples/step_wise/step_wise_trainer.py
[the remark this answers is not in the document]
SumanthRH: Oh....  padding with all zeros will make an invalid `is_last_step` array.... because you'll have something like `0, 0, 0, 0` which doesn't make sense (trajectory never has a last step).  `is_last_step` is later used to filter rows for adv calculation.... you'll end up with the wrong number of trajectories there (you need at least one trajectory representing the padding rows). If your filter of `is_last_step` is wrong, then you won't be able to broadcast the results back to the original tensor. 

So a valid padding tensor for `is_last_step` must at-least have one entry of 1, so `0, 0, 0, 1` is valid, and also `1, 1, 1, 1` is valid. for simplicity, i'm just using all ones. 

@gemini-code-assist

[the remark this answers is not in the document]
SumanthRH: Fixed now @gemini-code-assist

skyrl-train/examples/step_wise/step_wise_trainer.py
```diff
@@ -0,0 +1,347 @@
...
+        - after calling this method, the same model placement still holds.
+        """
+        # NOTE: we assume that .generate returns samples in the same order as passed in
+        generator_output: GeneratorOutput = await self.generator.generate(input_batch)
+
+        # add rollout metrics to self.all_metrics
+        if generator_output["rollout_metrics"] is not None:
+            self.all_metrics.update(generator_output["rollout_metrics"])
+
+        # don't validate - will error out
```
tyler-griggs: Could you add just a little more detail on why it will error out, just for posterity :)

PR #438, opened 2025-10-09
[train][qwen3][SkyRLGymGen] Fix qwen3 default codepath with token in token out

CharlieFRuan:
This is a follow-up PR to https://github.com/NovaSky-AI/SkyRL/pull/178

In that PR, we changed the default codepath of Qwen3 in `SkyRLGymGenerator` to always keep all the thinking tokens for multi-turn generation and subsequent turn generation. Users can return to the previous behavior (strip thinking tokens for both inference and training) by specifying `config.generator.chat_template` introduced in that PR.

This PR does a couple of things

### 1. Doc update on Qwen3 in SkyRLGymGenerator

First we update the doc to explain the current behavior with Qwen3 in SkyRLGymGenerator.

### 2. Bug fix with fixed base method motivated by Qwen3

Then I noticed a bug when using Qwen3 with codepath 1 (TI/TO with strictly appending token sequence) due to how we have  `<|im_start|>assistant\nI am an assistant.<|im_end|>\n` in our fixed-base method's `base_conversation_token_ids`.

This, when templated by Qwen3's chat template, will add dummy empty thinking tokens `<think>\n\n</think>\n\n`, making the delta not what we expect. So I exclude assistant in the base conversation for both SkyRLGymGenerator, and our helper `encode_messages_subset ()` for the custom generator. As a result, we will have consecutive user messages when constructing observation token IDs, which should be fine.

That is, before this PR, we use `base_conversation_token_ids`:

```
In [7]: tokenizer.decode(base_conversation_token_ids)
Out[7]: '<|im_start|>system\nYou are a helpful assistant.<|im_end|>\n<|im_start|>user\nI am a user.<|im_end|>\n<|im_start|>assistant\n<think>\n\n</think>\n\nI am an assistant.<|im_end|>\n'
```

but adding any message right after it will remove the thinking tokens by default, making the indexing incorrect

### 3. Add more unit tests.

In `test_skyrl_gym_generator_chat_templating_exact()`, we used to test `Qwen3` with our custom chat template that strips thinking tokens, I now also test the codepath 1's Qwen3 (accumulate thinking tokens). Did the same thing for `test_append_eos_after_stop_multi_turn()`.

Then since the `_exact()` test does not check thinking token handling with Qwen3, I added `test_skyrl_gym_generator_chat_templating_qwen3_thinking` that is essentially the same test but just with Qwen3 model, and with mock llm generation to be `"<think>\nmock thinking\n</think>\n\nb"` instead of just `"b"`, making the test more realistic.

### DIFF
--- a/skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
+++ b/skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
@@ -328,7 +328,6 @@ async def test_append_eos_after_stop_multi_turn(model_name, tokenization_codepat
     `skyrl_gym_generator.rst`. For Qwen3, we also test `generator.chat_template` being defined.
     """
     _register_test_env_if_needed()
-    model_name = model_name.replace("-TITO", "")
     tokenizer = AutoTokenizer.from_pretrained(model_name)
 
     stop_tag = "</solution>"


skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
```diff
@@ -210,6 +228,159 @@ def mock_generate(input_batch):
     assert generator_output["loss_masks"][0] == expected_loss_masks
 
 
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "mode",
+    ["retokenize_without_thinking", "default_tito"],
+)
+async def test_skyrl_gym_generator_chat_templating_qwen3_thinking(mode):
```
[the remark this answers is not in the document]
CharlieFRuan: I will do this as a follow up to keep the code change clear.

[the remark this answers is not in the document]
CharlieFRuan: nvm, deduplicated some code

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
```diff
@@ -51,15 +51,71 @@ def _register_test_env_if_needed():
...
 @pytest.mark.asyncio
 @pytest.mark.parametrize(
     "model_name",
-    ["Qwen/Qwen2.5-0.5B-Instruct", "unsloth/Llama-3.2-1B-Instruct", "Qwen/Qwen3-0.6B", "Qwen/Qwen3-0.6B-FROM_PATH"],
+    [
+        "Qwen/Qwen2.5-0.5B-Instruct",
+        "unsloth/Llama-3.2-1B-Instruct",
+        "Qwen/Qwen3-0.6B",
+        "Qwen/Qwen3-0.6B-FROM_PATH",  # tests custom chat template from file
+        "Qwen/Qwen3-0.6B-TITO",  # tests default qwen3 codepath with ti/to
```
SumanthRH: Nit: you could also just parameterize the function with additional args `custom_chat_template`, `is_tito`

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
```diff
@@ -158,6 +183,11 @@ def mock_generate(input_batch):
     else:
         generator_output_str = prompt_str + resp_str
         expected_str = tokenizer.apply_chat_template(expected_chat_history, tokenize=False)
+        if "Qwen3" in model_name and is_tito:
```
SumanthRH: While we're at it , could we improve this test a bit? 

I'd prefer if we simply hardcode the full expected string `expected_str` for different tokenizers  and then use that as an input argument in `parametrize`. It makes the test a lot more obvious, and also avoids duplicating logic from the code we're testing.

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
```diff
@@ -246,7 +398,8 @@ def test_qwen3_original_vs_without_thinking_chat_template():
 
 @pytest.mark.asyncio
 @pytest.mark.parametrize(
-    "model_name", ["Qwen/Qwen2.5-0.5B-Instruct", "unsloth/Llama-3.2-1B-Instruct", "Qwen/Qwen3-0.6B"]
+    "model_name",
+    ["Qwen/Qwen2.5-0.5B-Instruct", "unsloth/Llama-3.2-1B-Instruct", "Qwen/Qwen3-0.6B", "Qwen/Qwen3-0.6B-TITO"],
```
SumanthRH: Similar comment as before: `is_tito` as a separate parameter

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
```diff
@@ -210,6 +240,128 @@ def mock_generate(input_batch):
...
+    ["retokenize_without_thinking", "default_tito"],
+)
+async def test_skyrl_gym_generator_chat_templating_qwen3_thinking(mode):
+    """
+    While test_skyrl_gym_generator_chat_templating_exact tests Qwen3, it does not include the
+    thinking tokens in mock response, making the test not as realistic. We test with thinking tokens
+    in this test. The setup of the two tests are largely the same.
+    `retokenize_without_thinking` is the 3rd codepath described in `skyrl_gym_generator.rst` for Qwen3, and
+    `default_tito` is the 1st (default) codepath.
+    """
```
SumanthRH: Umm shouldn't this really be a part of `test_skyrl_gym_generator_chat_templating_exact` itself? We can parameterize `chat_template` and `expected_str` and that should be good?

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
CharlieFRuan: addressed

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
CharlieFRuan: addressed

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator_chat_templating.py
CharlieFRuan: I added hard-coded expected strings into a constant file, so that it is easier to check what is the expectation.

Two things:
1. I kept `assert expected_str == tokenizer.apply_chat_template(expected_chat_history, tokenize=False)` so that we check `expected_str` is indeed the hard coded constants, just to check that the constants are produced in the way we expect
2. I kept loss mask as is. I feel like the re-construction logic makes the logic clearer and the test more rigorous, compared to some hard-coded 0's and 1's.

PR #440, opened 2025-10-09
Guard against Divide 0 error for NUMA nodes count

zhenga1:
When running remotely on cloud VMs (and most Docker setups), the NUMA nodes count may not be accessible to the container.

So this is a guarding feature that sets NUMA to 1 (default) if the `numa_nodes` count is found to be zero, so calculating `num_gpu_pre_numa_node` is not undefined


SumanthRH:
Thanks!

PR #443, opened 2025-10-09
[Fix][GPU CI] Fix cleanup logic in LoRA test

SumanthRH:
# What does this PR do?

fixes the cleanup logic in the LoRA test. 

While executing tests in the same pytest session, we should not call `xxxRegistry.reset()` unnecessarily because that will clear all registered functions.

This can affect other tests that execute later that depend on the registries. 

This is the root cause for test failures for `test_save_load_checkpoint` yesterday and today

```bash
RuntimeError: Engine core initialization failed. See root cause above. Failed core proc(s): {}
FAILED tests/gpu/gpu_ci/test_save_load_checkpoint.py::test_save_load_checkpoint[deepspeed] - AssertionError: invalid policy_loss_type: regular. Must be one of []
FAILED tests/gpu/gpu_ci/test_save_load_checkpoint.py::test_save_load_checkpoint[fsdp] - AssertionError: invalid policy_loss_type: regular. Must be one of []
FAILED tests/gpu/gpu_ci/test_save_load_checkpoint.py::test_save_load_checkpoint[fsdp2] - AssertionError: invalid policy_loss_type: regular. Must be one of []
```

### DIFF
--- a/skyrl-train/tests/gpu/gpu_ci/conftest.py
+++ b/skyrl-train/tests/gpu/gpu_ci/conftest.py
@@ -15,10 +15,14 @@ def log_once(msg):
 def ray_init_fixture():
     if ray.is_initialized():
         ray.shutdown()
-    env_vars = {}
+    # TODO (team): maybe we should use the default config and use prepare_runtime_environment in some way
+    env_vars = {"VLLM_USE_V1": "1", "VLLM_ENABLE_V1_MULTIPROCESSING": "0", "VLLM_ALLOW_INSECURE_SERIALIZATION": "1"}
     if not peer_access_supported(max_num_gpus_per_node=2):
         log_once("Disabling NCCL P2P for CI environment")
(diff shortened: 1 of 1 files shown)


skyrl-train/tests/gpu/gpu_ci/conftest.py
```diff
@@ -15,10 +15,14 @@ def log_once(msg):
 def ray_init_fixture():
     if ray.is_initialized():
         ray.shutdown()
-    env_vars = {}
+    # TODO (team): maybe we should use the default config and use prepare_runtime_environment in some way
```
erictang000, line 18: yes agree... but need to manually add to all tests bc we would need to pass the cfg object through and you can't via fixture

SumanthRH: agreed

PR #445, opened 2025-10-09
[tx] [WIP] Add jax jitting to speed up forward pass

pcmoritz:
This is not quite working yet, but something like this will be needed. There is a similar code path in https://github.com/NovaSky-AI/SkyRL/blob/main/skyrl-tx/tx/run/train.py#L25 that already works.


pcmoritz:
Replaced by https://github.com/NovaSky-AI/SkyRL/pull/452

PR #452, opened 2025-10-10
[tx] Add jit compilation to engine

pcmoritz:
This PR adds jit compilation, which quite significantly improves performance. We also add a flag to deactivate it, since the warmup time can be quite long so it is good to be able to deactivate it e.g. for development.

### DIFF
--- a/skyrl-tx/tx/tinker/engine.py
+++ b/skyrl-tx/tx/tinker/engine.py
@@ -164,7 +164,9 @@ class TinkerEngine:
             scalar = jax.NamedSharding(self.mesh, jax.P())
             self._loss_and_grad_fn = jax.jit(
                 loss_and_grad_fn,
+                # One input sharding parameter for each argument of loss_for_lora
                 in_shardings=(state_shardings,) + (replicated,) * 5,
+                # One output sharding parameter for each return value of loss_for_lora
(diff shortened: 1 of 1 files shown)


skyrl-tx/tx/tinker/engine.py
```diff
@@ -105,11 +132,59 @@ def is_lora_param(path, value):
...
+            # Extract state once to get the pytree structure and compute the partition
+            state = nnx.state(self.lora_params)
+            state_partition_spec = nnx.get_partition_spec(state)
+            # Create NamedSharding objects that tell us how models parameters and inputs should be sharded
+            state_shardings = jax.tree.map(lambda spec: jax.NamedSharding(self.mesh, spec), state_partition_spec)
+            replicated = jax.NamedSharding(self.mesh, jax.P(None))
+            scalar = jax.NamedSharding(self.mesh, jax.P())
+            self._loss_and_grad_fn = jax.jit(
+                loss_and_grad_fn,
+                in_shardings=(state_shardings,) + (replicated,) * 5,
```
tyler-griggs, line 168: Perhaps add a comment specifying where this "5" came from? It wasn't immediately clear to me that it was referring to the input parameters for `loss_and_grad_fn`

pcmoritz: Fixed now :)

PR #456, opened 2025-10-11
[train][CI] Fix flaky GPU skyrlgymgenerator test due to stop_reason=length

CharlieFRuan:
We frequently see the test `test_generator_formatting_use_conversation_multi_turn` fails, e.g.:

```
FAILED tests/gpu/gpu_ci/test_skyrl_gym_generator.py::test_generator_formatting_use_conversation_multi_turn[Qwen/Qwen3-0.6B] - assert 2 == 3
```

This is because in some of the turns, the stop_reason can be length and hence failing to generate an eos token.

I changed the assertion to a warning for now.

We should fix it properly in the future, either by returning stop_reason for each turn, or change the way we manage max generation length (e.g. set `max_tokens` based on a shared max generation length across all turns) so that only the last turn can be stop_reason='length'.

Unrelatedly, fix Llama CPU test's chat template's date


CharlieFRuan:
GPU CI running here: https://github.com/NovaSky-AI/SkyRL/actions/runs/18422120093

PR #459, opened 2025-10-11
[tx] Support gradient accumulation across multiple fwd/bwd calls to the same model

tyler-griggs:


tyler-griggs:
Closing in favor of #465

PR #460, opened 2025-10-11
[tx] Support gradient accumulation and micro batches

tyler-griggs:
This PR adds support for gradient accumulation, which can be used in two ways: 
* Multiple fwd/bwd requests to the same model (ie, same LoRA adapter) without stepping the optimizer
* Configuring a micro batch size, which splits a batch of fwd/bwd inputs into micro batches and accumulates gradients over the global batch

### DIFF
--- a/skyrl-tx/tx/tinker/engine.py
+++ b/skyrl-tx/tx/tinker/engine.py
@@ -122,17 +122,19 @@ class TinkerEngine:
 
         def loss_for_lora(lora_params):
             merged = nnx.merge(self.graphdef, lora_params, self.non_lora_params)
-            logits = merged(input_ids, attention_mask=attention_mask, adapter_indices=adapter_indices)["logits"]
+            logits = merged(input_ids, attention_mask=attention_mask, adapter_indices=adapter_indices)[
+                "logits"
+            ]  # [B, T, V]
             per_token_losses = optax.softmax_cross_entropy_with_integer_labels(
                 logits=logits, labels=target_ids, where=loss_mask
-            )
+            )  # [B, T]
             # Return sum of losses (we'll divide gradients by per-adapter batch size later)
             return per_token_losses.mean(axis=-1).sum(), (logits, per_token_losses)
 
         loss_and_grad_fn = nnx.value_and_grad(loss_for_lora, has_aux=True)
         (_, (logits, per_token_losses)), lora_grads = loss_and_grad_fn(self.lora_params)
-        logprobs = jax.nn.log_softmax(logits, axis=-1)
-        target_logprobs = jnp.take_along_axis(logprobs, target_ids[..., None], axis=-1).squeeze(-1)
+        logprobs = jax.nn.log_softmax(logits, axis=-1)  # [B, T, V]
+        target_logprobs = jnp.take_along_axis(logprobs, target_ids[..., None], axis=-1).squeeze(-1)  # [B, T]
         return per_token_losses, target_logprobs, lora_grads
 
     def _accumulate_grads(self, lora_grads: nnx.State, example_model_ids: list[str]) -> None:


skyrl-tx/tx/tinker/engine.py
```diff
@@ -75,6 +76,50 @@ def is_lora_param(path, value):
             f"Initialized base model {self.base_model_name} with max_lora_adapters={max_lora_adapters}, max_lora_rank={max_lora_rank}"
         )
 
+    def _micro_batch_size(self, total: int) -> int:
+        """Return effective micro-batch size; 0/absent => disabled (use full fused batch)."""
+        try:
+            mb = int(os.environ.get("TX_MICRO_BATCH_SIZE", "0"))
```
pcmoritz: (As you pointed out we can make this part of the EngineConfig, it also gives me the idea that it will be best to just store the engine config in self.config, so we don't need to create separate variables for all these configs)

skyrl-tx/tx/tinker/engine.py
```diff
@@ -75,6 +76,50 @@ def is_lora_param(path, value):
...
+        except ValueError:
+            mb = 0
+        return total if mb <= 0 else max(1, min(mb, total))
+
+    def _fwd_bwd(self, input_ids, attention_mask, adapter_indices, target_ids, loss_mask):
```
pcmoritz: We should add types for these arguments, and the return types :)

skyrl-tx/tx/tinker/engine.py
```diff
@@ -313,7 +360,7 @@ def expand_adapter_grads(lora_param, adapter_grad):
         self.optimizer.update(self.lora_params, full_lora_grads)
 
         # Clear accumulated gradients
-        self.accumulated_grads[model_id] = None
+        self.accumulated_grads[model_id] = {"grad_sum": None, "denominator": 0}
```
pcmoritz: I think it would be great to add a proper type for this in `types.py` instead of using a dict, it can also have type annotations (e.g. the grad_sum attribute is of type jax.Array | None and denominator is of type float). And then you can also get rid of the all `.get` calls above, since these attribute will always be there :)

skyrl-tx/tx/tinker/engine.py
pcmoritz: Also small nit, since we already use `forward_backward` a lot in this function, it might be better for consistency to just spell it out here too :)

skyrl-tx/tx/tinker/engine.py
```diff
@@ -75,6 +76,50 @@ def is_lora_param(path, value):
...
+        """
+        Accumulate adapter-wise gradient sums and example counts.
+        Scale gradient mean by batch size to get gradient sum.
+        """
+        for model_id, count in example_counts.items():
+            idx = self.models[model_id].adapter_index
+            grad_mean = jax.tree.map(lambda g: g[idx], lora_grads)
+            grad_sum = jax.tree.map(lambda x: x * jnp.asarray(batch_size, dtype=x.dtype), grad_mean)
+            grad_sum_accumulator = self.accumulated_grads[model_id]
+            self.accumulated_grads[model_id] = (
```
pcmoritz: This might be slightly easier to read if you do the following:
```python
accumulator = self.accumulated_grads[model_id]
if accumulator["grad_sum"] is None:
    accumulator["grad_sum"] = grad_sum
    accumulator["denominator"] = count
else:
    accumulator["grad_sum"] = jax.tree.map(lambda a, b: a + b, accumulator["grad_sum"], grad_sum)
    accumulator["denominator"] = accumulator["denominator"] + count
```

skyrl-tx/tx/tinker/engine.py
```diff
@@ -220,60 +267,54 @@ def process_forward_backward_batch(
...
-            adapter_index = self.models[model_id].adapter_index
-
-            # Extract gradients for this specific adapter index.
-            adapter_grads_all_mean = jax.tree.map(lambda g: g[adapter_index], lora_grads)
-            # Gradient is the mean across the global batch. Scale to mean over the adapter's samples.
-            grad_scale = num_total_examples / num_adapter_examples
-            adapter_grads = jax.tree.map(
-                lambda x: x * jnp.asarray(grad_scale, dtype=x.dtype),
-                adapter_grads_all_mean,
+        B_total = int(input_ids.shape[0])
```
pcmoritz: Maybe call this `total_bs` to be consistent with the other names?

skyrl-tx/tx/tinker/engine.py
```diff
@@ -220,60 +267,54 @@ def process_forward_backward_batch(
...
-        for request_id, model_id, start_idx, end_idx in request_batch_slices:
+            for i_local, i_global in enumerate(range(mb_start, mb_end)):
+                L = seq_lens[i_global]
+                token_losses_out[i_global] = per_token_losses[i_local, :L].astype(jnp.float32)
+                logprobs_out[i_global] = target_logprobs[i_local, :L].astype(jnp.float32)
+            # Build per-model example counts for this micro-batch
+            example_counts_mb = {}
+            for mid in example_model_ids[mb_start:mb_end]:
+                example_counts_mb[mid] = example_counts_mb.get(mid, 0) + 1
+            self._accumulate_grads(lora_grads_mb, example_counts_mb, B_mb)
```
pcmoritz: You can just pass `example_model_ids[mb_start:mb_end]` into accumulate_grads, and then inside there you do (use example_model_ids instead of hello world of course)
```Python
In [6]: for model_id, count in collections.Counter(["hello", "world", "hello"]).items():
   ...:     print("model_id", model_id, "count", count)
```
Otherwise this is kind of introducing this new dict datastructure which becomes the "public" interface of accumulate_grads which is not really needed :)

skyrl-tx/tx/tinker/engine.py
tyler-griggs: Good call, done!

skyrl-tx/tx/tinker/engine.py
tyler-griggs: Done!

skyrl-tx/tx/tinker/engine.py
tyler-griggs: good call

skyrl-tx/tx/tinker/engine.py
tyler-griggs: Yeah, I started thinking that using the canonical capital-letter dimension characters would be helpful (B for batch sizes, L for sequence lengths, etc.), but didn't apply them consistently. Maybe we'll use these later, but I switched to `total_bs` for now

skyrl-tx/tx/tinker/engine.py
tyler-griggs: Will address after #464

skyrl-tx/tx/tinker/types.py
```diff
@@ -72,3 +73,10 @@ class SaveWeightsForSamplerOutput(BaseModel):
 class ModelMetadata(BaseModel):
     adapter_index: int
     lora_config: LoraConfig
+
+
+class AccumulatedGradients(BaseModel):
+    model_config = {"arbitrary_types_allowed": True}  # Needed for flax nnx.State
```
pcmoritz: Thinking about this a little more, since this is not a serializable type (and e.g. we are not planning to store this in the database etc), it might be better to just make this a normal class at the top of engine.py, but either way is fine :)

skyrl-tx/tx/tinker/engine.py
tyler-griggs: done!

skyrl-tx/tx/tinker/engine.py
```diff
@@ -71,6 +105,47 @@ def is_lora_param(path, value):
...
+            logits = merged(input_ids, attention_mask=attention_mask, adapter_indices=adapter_indices)["logits"]
+            per_token_losses = optax.softmax_cross_entropy_with_integer_labels(
+                logits=logits, labels=target_ids, where=loss_mask
+            )
+            # Return sum of losses (we'll divide gradients by per-adapter batch size later)
+            return per_token_losses.mean(axis=-1).sum(), (logits, per_token_losses)
+
+        loss_and_grad_fn = nnx.value_and_grad(loss_for_lora, has_aux=True)
+        (_, (logits, per_token_losses)), lora_grads = loss_and_grad_fn(self.lora_params)
+        logprobs = jax.nn.log_softmax(logits, axis=-1)
```
pcmoritz: It would be good to bring the comments with the shapes back here and in the following lines, they really helped. If you don't want to do it as part of this PR, I can do it in the JIT PR afterwards :)

pcmoritz:
Congrats, this PR is awesome!

skyrl-tx/tx/tinker/engine.py
tyler-griggs: Oh I didn't mean to remove those, added back!

PR #463, opened 2025-10-11
[tx] [WIP] Centralized way to parse arguments

pcmoritz:
This PR makes it possible to handle all the argument parsing in a more centralized way, so it can be shared between the api server and the engine.


pcmoritz:
This is now replaced with https://github.com/NovaSky-AI/SkyRL/pull/464

PR #464, opened 2025-10-12
[tx] Define engine options and allow to pass them from API server

pcmoritz:
Going forward (e.g. once we support multiple engines), we will probably store these configs in the database, but for now it is ok to just define them as pydantic models that get parsed from the command line.

### DIFF
--- a/skyrl-tx/tx/tinker/api.py
+++ b/skyrl-tx/tx/tinker/api.py
@@ -364,9 +364,8 @@ async def download_checkpoint_archive(
     if not model:
         raise HTTPException(status_code=404, detail="Model not found")
 
-    # Files are saved at checkpoints_base_path/{model_id}/{checkpoint_id}/
-    checkpoints_base_path = Path(request.app.state.engine_config.checkpoints_base_path)
-    checkpoint_dir = checkpoints_base_path / unique_id / checkpoint_id
+    # Files are saved at checkpoints_base/{model_id}/{checkpoint_id}/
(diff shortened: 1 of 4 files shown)


skyrl-tx/tx/tinker/api.py
```diff
@@ -35,24 +32,16 @@ async def lifespan(app: FastAPI):
...
-            "--base-model",
-            base_model,
-            "--checkpoints-base-path",
-            CHECKPOINTS_BASE_PATH,
-        ]
+    # Build subprocess command with engine config parameters
+    cmd = ["uv", "run", "--extra", "tinker", "-m", "tx.tinker.engine"]
+    for field_name, value in app.state.engine_config.model_dump().items():
+        cmd.append(f"--{field_name.replace('_', '-')}")
+        cmd.append(str(value))
```
tyler-griggs: Do all cli args have a value? E.g., consider a boolean flag like vllm's `--enable-expert-parallel`. In this case, would `str(value)` be `True`/`False` and it would just work? Or would this break?

tyler-griggs: We could also consider moving this logic to `config.py` in a helper like `def config_to_argv(cfg: BaseModel) -> list[str]:`, but not strictly necessary

pcmoritz: Yeah you are right, it currently doesn't support boolean flags. I'll add `--enforce-eager` in my next PR, how about I add it then (currently it just breaks). I followed your suggestion to add a `config_to_argv` now, it is a great idea! That also makes it clearer which arguments are and which arguments aren't supported :)

PR #461, opened 2025-10-11
[tx] Adding save/load state routes

Lucas-Fernandes-Martins:
Hi,

This PR addresses #450 by implementing the save/load state (weights, optimizer state, and adapter_config) routes in a checkpoint, according to the Tinker docs [here](https://tinker-docs.thinkingmachines.ai/save-load). 

For now, I'm using the same base path as the path used for save_weights_for_sampler. Please let me know if we should save training checkpoints elswhere, in order to completely isolate the save/load training state and the weights used for sampling.

Thanks in advance and let me know if you'd like any changes :)

The following script can be used for testing:

```
import tinker
import numpy as np
import tarfile
import os

service_client = tinker.ServiceClient(base_url="http://localhost:8000", api_key="dummy")

base_model = "Qwen/Qwen3-0.6B"
training_client = service_client.create_lora_training_client(
    base_model=base_model
)

# Create some training examples
examples = [
    {
        "input": "banana split",
        "output": "anana-bay plit-say"
    },
    {
        "input": "quantum physics",
        "output": "uantum-qay ysics-phay"
    },
    {
        "input": "donut shop",
        "output": "onut-day op-shay"
    },
    {
        "input": "pickle jar",
        "output": "ickle-pay ar-jay"
    },
    {
        "input": "space exploration",
        "output": "ace-spay exploration-way"
    },
    {
        "input": "rubber duck",
        "output": "ubber-ray uck-day"
    },
    {
        "input": "coding wizard",
        "output": "oding-cay izard-way"
    },
]
 
# Convert examples into the format expected by the training client
from tinker import types
 
# Get the tokenizer from the training client
tokenizer = training_client.get_tokenizer()
 
def process_example(example: dict, tokenizer) -> types.Datum:
    # Format the input with Input/Output template
    prompt = f"English: {example['input']}\nPig Latin:"
 
    prompt_tokens = tokenizer.encode(prompt, add_special_tokens=True)
    prompt_weights = [0] * len(prompt_tokens)
    # Add a space before the output string, and finish with double newline
    completion_tokens = tokenizer.encode(f" {example['output']}\n\n", add_special_tokens=False)
    completion_weights = [1] * len(completion_tokens)
 
    tokens = prompt_tokens + completion_tokens
    weights = prompt_weights + completion_weights
 
    input_tokens = tokens[:-1]
    target_tokens = tokens[1:] # We're predicting the next token, so targets need to be shifted.
    weights = weights[1:]
 
    # A datum is a single training example for the loss function.
[...]

PR #466, opened 2025-10-12
[tx] feat: let tinker checkpoints target cloud paths

### LINKED ISSUE #437: Allow storing model checkpoints on cloud storage
Currently, model checkpoints are stored on a POSIX filesystem (by default on a local filesystem), which is very suboptimal -- ideally we can support different cloud storage options like S3, GCS or Azure Blob Store, so the storage will be independent of the local filesystem that tx is running on, and appropriate policies on access control or backups can be set.

Probably this can be implemented in a vendor agnostic way by using a library such as `smart_open`.

cwwarren:
*fixes #437*

Adds support for storing tinker checkpoints in cloud object stores (AWS/GCP/Azure) using [cloudpathlib](https://github.com/drivendataorg/cloudpathlib).

---

Tested with a modified pig latin training script against CloudFlare R2 (S3 compatible).

```py
...

save_path = training_client.save_weights_for_sampler(name="final").result().path
print(save_path)

with open("model-checkpoint.tar.gz", "wb") as f:
    f.write(rest_client.download_checkpoint_archive_from_tinker_path(f"tinker://{training_client.model_id}/sampler_weights/final").result())
```

```
AWS_ENDPOINT_URL=https://<...>.r2.cloudflarestorage.com \
    AWS_ACCESS_KEY_ID=<...> \
    AWS_SECRET_ACCESS_KEY=<...> \
    CHECKPOINTS_BASE_PATH=s3://<...>/checkpoints \
    uv run --extra tinker --extra gpu -m tx.tinker.api
```

```
SkyRL/skyrl-tx main  ? ❯ uv run python piglatin.py 
Loss per token: 6.7933
Loss per token: 6.4766
Loss per token: 5.8127
Loss per token: 5.0436
Loss per token: 4.2612
Loss per token: 3.5939
tinker://model_78a3f9c5/final

SkyRL/skyrl-tx main ? ❯ tar tvf model-checkpoint.tar.gz
-rw-r--r-- 0/0             804 2025-10-11 18:48 adapter_config.json
-rw-r--r-- 0/0        40415864 2025-10-11 18:48 adapter_model.safetensors
```

Screenshot 2025-10-11 at 7 23 12 PM

---

Also tested locally

```
CHECKPOINTS_BASE_PATH=/tmp/tx_checkpoints \
    uv run --extra tinker --extra gpu -m tx.tinker.api
```

```
SkyRL/skyrl-tx feat/tinker-cloud-checkpoints ? ❯ uv run python piglatin.py 
Loss per token: 6.8108
Loss per token: 6.4627
Loss per token: 5.8067
Loss per token: 5.0007
Loss per token: 4.2140
Loss per token: 3.5998
tinker://model_d237eecf/final

SkyRL/skyrl-tx feat/tinker-cloud-checkpoints ? ❯ tar tvf model-checkpoint.tar.gz
-rw-r--r-- 0/0             804 2025-10-11 19:35 adapter_config.json
-rw-r--r-- 0/0        40415864 2025-10-11 19:35 adapter_model.safetensors

SkyRL/skyrl-tx feat/tinker-cloud-checkpoints ? ❯ ls -al /tmp/tx_checkpoints/model_d237eecf/final/
Permissions Size User  Date Modified Name
.rw-r--r--   804 colin 11 Oct 19:35   adapter_config.json
.rw-r--r--   40M colin 11 Oct 19:35   adapter_model.safetensors
```

### DIFF
--- a/skyrl-tx/tx/tinker/engine.py
+++ b/skyrl-tx/tx/tinker/engine.py
@@ -22,7 +22,7 @@ from huggingface_hub import snapshot_download
 from tx.tinker.db_models import FutureDB, DB_PATH, RequestStatus
 from tx.tinker import types
 from tx.tinker.config import EngineConfig, add_model
-from tx.utils.storage import pack_and_upload
+from tx.utils.storage import download_and_unpack, pack_and_upload
 from tx.utils.models import (
     get_dtype,
     get_model_class,
@@ -486,9 +486,12 @@ class TinkerEngine:
             raise ValueError("Model not loaded. Create the model before loading a checkpoint.")
 
         adapter_index = self.models[model_id].adapter_index
-        checkpoint_dir = self.config.checkpoints_base / model_id / Path(request_data.path).name
+        checkpoint_id = Path(request_data.path).name
+        checkpoint_dir = self.config.checkpoints_base / model_id / f"{checkpoint_id}.tar.gz"
 
-        restored_data = checkpoints.restore_checkpoint(ckpt_dir=checkpoint_dir, target=None, prefix="checkpoint_")
+        with download_and_unpack(checkpoint_dir) as temp_dir:
+            restored_data = checkpoints.restore_checkpoint(ckpt_dir=temp_dir, target=None, prefix="checkpoint_")
+
         if restored_data is None:
             raise FileNotFoundError(f"Training checkpoint not found in {checkpoint_dir}")
 


cwwarren:
I rebased and integrated this into EngineConfig then tested with the pig latin script against R2 again.

pcmoritz:
Thanks a lot for your patience, I'm now finally getting back to this (I wanted to first get https://github.com/NovaSky-AI/SkyRL/pull/461 merged) :)

cwwarren:
No worries at all! Let me know if this makes sense!

pcmoritz:
It makes sense to me, I wasn't aware of https://github.com/drivendataorg/cloudpathlib but it seems nice. Your code looks great, one thing I'm thinking about and would like to change is how can we best unify the code paths so there is only one code path. Maybe structure things with a context manager, and inside of that context manager, the code can write to a temporary directory that the context manager gives, and then when the context exits, the files either get moved to the right location (for local file systems) or uploaded. That way future contributors will only need to write one piece of code (like they would for the local version) and it will automatically work with a cloud path.

I'll play around with that a little and do some pushes to your branch, hope that's ok :)

pcmoritz:
I'm going to work on this more tomorrow, but I think for the checkpoints it makes the most sense to rename `{save, load}_checkpoint` to `{save, load}_safetensors` (these only work with local paths) and define a new function in `util/models.py` that does
```python
def store_checkpoint(config: PretrainedConfig, adapter_config: LoraConfig, model: nnx.Module, output_dir: AnyPath):
    peft_config = PEFTLoraConfig(r=adapter_config.rank, lora_alpha=adapter_config.alpha)
    with staged_upload(output_dir) as temp_dir:
        save_checkpoint(config, model, temp_dir / "model.safetensors")
        peft_config.save_pretrained(temp_dir)
```
and works both with a local path and a cloud path :)

cwwarren:
I like it! I'd be happy to make any changes as well.

pcmoritz:
Thanks a lot @cwwarren, the PR is probably close enough already that I can finish it quickly -- if you have some spare cycles, I wonder if you would like to have a look at implementing custom loss functions https://tinker-docs.thinkingmachines.ai/losses (forward_backward_custom) -- it is an important feature to run RL. I created an issue here: https://github.com/NovaSky-AI/SkyRL/issues/482

pcmoritz:
@cwwarren Ok, I ended up making more changes than initially planned -- it also made sense to store the checkpoints as tar.gz archives in blob store (since that's the format we eventually ship to the user, and also storing less files in blob store is good). I also cleaned up some stuff and make the checkpoints more in line with what peft expects and added a test for it.

Let me know if you have any suggestions before we merge this :)

skyrl-tx/tx/utils/storage.py
```diff
@@ -0,0 +1,67 @@
...
+import io
+from pathlib import Path
+import tarfile
+from tempfile import TemporaryDirectory
+from typing import Generator
+from cloudpathlib import AnyPath
+
+
+@contextmanager
+def pack_and_upload(dest: AnyPath) -> Generator[Path, None, None]:
```
tyler-griggs, line 11: Love it! This is similar to how we handle cloud checkpointing in skyrl-train, it's a nice pattern

cwwarren:
This looks really good! I'd tried to keep my original change set minimal while I better understood the project but I like this better. Fully agreed with storing .tar.gz artifacts that work with the HF ecosystem. 🚀 

> if you have some spare cycles, I wonder if you would like to have a look at implementing custom loss functions https://tinker-docs.thinkingmachines.ai/losses (forward_backward_custom) -- it is an important feature to run RL. I created an issue here: https://github.com/NovaSky-AI/SkyRL/issues/482

That would be an interesting task to take on, although I want to get https://github.com/NovaSky-AI/SkyRL/pull/470 into a slightly less hacky state first.

I read through the docs and client to understand how forward_backward_custom works and it appears relatively straightforward. I'd need to dig in more but I believe the only missing piece in the engine is supporting `/forward` in addition to `/forward_backward`. 

- Docs: https://tinker-docs.thinkingmachines.ai/losses#how-forward_backward_custom-works
- Client: https://github.com/thinking-machines-lab/tinker/blob/1a4665f050217fca757242041d1cd7e9fa0a8816/src/tinker/lib/public_interfaces/training_client.py#L260-L307

tyler-griggs (approved):

pcmoritz (approved):

PR #470, opened 2025-10-13
[tx] allow per-request optimizer hyperparameter overrides

cwwarren:
I sketched out an initial attempt at per-request optimizer hyperparams for optim_step. It wraps `optax.adamw` with [`optax.inject_hyperparams`](https://optax.readthedocs.io/en/latest/api/optimizer_schedules.html#optax.schedules.inject_hyperparams) to allow mutating the hyperparams via the optimizer state then sets them to either the request value or a default before running `optimizer.update` in each optim_step.

I'm not sure if this makes sense architecturally given the global shared optim state, although this should work fine as long as the engine remains single threaded and we always reset the hyperparameters. I'd love to hear thoughts on this (or scrap the PR)-- I was honestly bored on a Sunday and wanted to refresh my brain on the jax ecosystem.

*The default b1/b2/eps are from [thinking-machines-lab/tinker](https://github.com/thinking-machines-lab/tinker/blob/9ba155a34d2eaa5ad8d39b3f50a7769fd173da4e/src/tinker/types/optim_step_request.py#L13).*

### DIFF
--- a/skyrl-tx/tests/tinker/test_engine.py
+++ b/skyrl-tx/tests/tinker/test_engine.py
@@ -7,7 +7,7 @@ import pytest
 from flax import nnx
 
 from tx.tinker.engine import TinkerEngine
-from tx.tinker.config import EngineConfig, ADAM_BETA1, ADAM_BETA2, ADAM_EPS
+from tx.tinker.config import EngineConfig, ADAM_BETA1, ADAM_BETA2, ADAM_EPS, LEARNING_RATE
 from tx.tinker import types
 
 
@@ -98,11 +98,11 @@ def test_adapter_gradient_calculation():
     # Process round 1 batch
     engine.process_forward_backward_batch(reqs_round1)
 
-    grads_A1_round1 = jax.tree.map(lambda x: x.copy(), engine.accumulated_grads[adapter1_id].grad_sum)
+    grads_A1_round1 = jax.tree.map(lambda x: x[0], engine.accumulated_grads.grad_sum)
 
     # Clear stored grads so we can run another fwd/bwd without optimizer update.
-    engine.accumulated_grads[adapter1_id].reset()
-    engine.accumulated_grads[adapter2_id].reset()
+    engine.accumulated_grads.reset(jax.nn.one_hot(0, engine.config.max_lora_adapters, dtype=jnp.bool_))
+    engine.accumulated_grads.reset(jax.nn.one_hot(1, engine.config.max_lora_adapters, dtype=jnp.bool_))
 
     a1_input = make_fwd_bwd_input([[1, 2, 3, 4], [5, 6, 7, 8]])
     a2_input2 = make_fwd_bwd_input([[9, 10, 11, 12], [13, 14, 15, 16], [17, 18, 19, 20], [21, 22, 23, 24]])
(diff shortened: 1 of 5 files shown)


cwwarren:
Thinking more about the architecture, the global shared optimizer state may need to be fixed before custom hyperparams.

Unless I'm missing something, even though the grads are zeroed for the other adapters, if m or v are non-zero they would decay and t would increment leading to param updates for all adapters at each step from drift.

pcmoritz:
Thanks a lot for making the PR and thinking about this! If you have any suggestions on how to best split up the optimizer, let us know :) We are currently looking into the sharding of large MoE models and getting the other PRs merged, but plan to look into this immediately after. If you have more suggestions or code on how to best do it, will be happy to incorporate them :)

cwwarren:
Thanks! Please if this is going the right direction, my jax is a bit rusty.

I threw up a (minimally tested) sketch of masked AdamW which operates independently on a subset of adapters. Right now it should keep optim_step requests independent but with a few additional changes in the engine it allows batching optim_step calls across adapters. I'll hack on it a bit more tomorrow.

pcmoritz:
This looks amazing, thanks a lot for working this out! Sorry for the delay in reviewing this, I'll get to merging it tonight :)

cwwarren:
I'd hold off on merging it. I've been hacking on a few follow on experiments and at minimum this needs some new tests. I have a test that more or less runs a few steps comparing the final weights against optax that caught a few subtle issues.

I implemented optimizer step batching, the scheduler was rough but more notably it seemed slower than before.

After I profiled that gathering the grads into a set of dense tensors took 10x the time as the actual optimizer (+jit), I tried converting the AccumulatedGradients into a dense representation I could just mask and optimize. That brought the times down significantly (to 4-5ms collecting grads, 8ms opt step for a single adapter on a 5090) but the memory is significant and the dense tensor introduces some very subtle numeric errors that add up.

I was debating whether to profile the other approach here — keeping sparse active per-model optimizers, since if the runtime is not notably higher, would be much simpler and lower vram. Would love any thoughts on this.

pcmoritz:
Yeah, for the first iteration of this, I think we should lean on the side of making it simpler (and correct of course), and not try to squeeze every last bit of efficiency out of it. But of course it shouldn't be completely inefficient. In my experience it is much easier to optimize simple code than it is to deal with more complex code :)

Separate optimizers seems pretty reasonable to me, do you see any reason not to do that?

cwwarren:
Agreed, when all the cookbook examples work micro-optimizing can be iterative. This was a fun adventure but I'm going to sweep it back up to a dict[str, nnx.Optimizer] and call it a day for now.

pcmoritz:
Sounds great, maybe keep this code around in another branch (or open another PR), so we can come back to it later if it makes sense :)

pcmoritz:
Btw @cwwarren small heads up, while working on https://github.com/NovaSky-AI/SkyRL/pull/514 I noticed that I broke the qwen3 model numerics in https://github.com/NovaSky-AI/SkyRL/pull/494 (I don't know yet by how much and whether it makes a difference for your PR, just wanted to let you know). I'll try to fix it as soon as possible. Sorry about the disruption!

The symptom is that `uv run --extra dev pytest -s tests/models/test_qwen3.py` passes before https://github.com/NovaSky-AI/SkyRL/pull/494 but not after.

cwwarren:
No worries at all! I was working off an older commit for a while so I don't think I even had that change.

I'm just going to close this PR to preserve the code and open a new one with a more minimal `dict[str, nnx.Optimizer]` approach for you folks.

pcmoritz:
Awesome, sounds great! Thank you so much for working on this, this is really important!

PR #475, opened 2025-10-14
[tx] Improve sharding efficiency by not capturing over model params

pcmoritz:
Currently we capture over the whole non lora params in `loss_for_lora`, instead we need to trace it. See also https://github.com/jax-ml/jax/pull/28157.


tyler-griggs (approved):
Looks great

PR #486, opened 2025-10-15
[tx] Adds a  CheckpointsDB model to track checkpoints and their status

atemaguer:
This PR implements a way to keep track of checkpoints in the database. Additionally, the implementation ensures that checkpoint IDs are unique and an error is throws  when a checkpoint ID is duplicated. Let me know if we need to handle this edge case differently. Next steps after this gets approval will be to add endpoints to support for listing and filtering checkpoints as well as access controls once an auth support is in place.

### DIFF
--- a/skyrl-tx/tests/tinker/test_api.py
+++ b/skyrl-tx/tests/tinker/test_api.py
@@ -100,6 +100,9 @@ def test_training_workflow(service_client):
 
     # Save the optimizer state
     resume_path = training_client.save_state(name="0000").result().path
+    # Get the training run ID from the first save
+    parsed_resume = urlparse(resume_path)
+    original_training_run_id = parsed_resume.netloc
 
     # Run training step
     fwdbwd_future = training_client.forward_backward(processed_examples, "cross_entropy")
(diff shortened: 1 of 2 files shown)


pcmoritz:
@atemaguer Thanks a lot for working on this, I'll get to merging it now. Sorry for the delay, I first wanted to see how some of the other user facing PRs would play out to see if any requirements on this PR would change. This will be super useful to unify our DB schemas and also make it so the API server doesn't have blocking calls to the checkpoint storage system any more :D

atemaguer:
Yeah. I'm planning to push more features to support this functionality. Just create issues describing what needs implementing and I'll take them up

pcmoritz:
@atemaguer That sounds great, I just created a few more issues you might find interesting. I think https://github.com/NovaSky-AI/SkyRL/issues/519 would be particularly good, because anybody who actually wants to run SkyRL tx as a service will need it (and it is probably fine to do that with sqlite to start off with, but it is not fine if there are no DB migrations, so we should start getting into the habit of providing them out of the box).

PR #453, opened 2025-10-10
[megatron] upgrade from mbridge -> Megatron-Bridge (breaking change)

erictang000:
Upgrading from [mbridge](https://github.com/ISEEKYAN/mbridge) to [Megatron-bridge](https://github.com/NVIDIA-NeMo/Megatron-Bridge), the (now) official nvidia supported library for converting from HF to megatron format. The changes are pretty minimal from a user perspective, but do require updating to a new megatron version to use megatron-bridge, making this a breaking change (megatron >= 14.0 is required).

TODOs:
- [x] [Fixed, waiting for this to be included in a release/tagged RC] test HF model saving and checkpointing - currently tie_word_embeddings has issues in megatron-bridge (https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/919)
- [x] pp > 2 is broken in Megatron-Bridge for models with tie_word_embeddings=True: https://github.com/NVIDIA-NeMo/Megatron-Bridge/issues/954

Known test failures:`test_save_and_load_model.py` with `-m "megatron"` will fail for now.

PR #487, opened 2025-10-15
[megatron] improving weight syncing - bucketed param gather + cuda ipc flattening

erictang000:
Blocked on #453.

Post Megatron-Bridge migration, we can now do a bucketed gathering of parameters, with precomputed size metadata, instead of iterating through parameters one by one. We then flatten the bucket into a single tensor, send over the metadata, and recover the weights/shapes on the inference engine side.

For Qwen3-30B-A3B, tp=2, ep=8, etp=1, 8xh100:

Before (48s):
image

After bucketing gather (45s):
image

After flattening cuda ipc (and removing redundant torch.device() calls) - 6s:
image

### DIFF
--- a/skyrl-train/skyrl_train/inference_engines/base.py
+++ b/skyrl-train/skyrl_train/inference_engines/base.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from typing import List, Dict, TypedDict, Any, Optional, Hashable
+from typing import List, Dict, TypedDict, Any, Optional, Hashable, NotRequired
 
 MessageType = Dict[str, str]
 ConversationType = List[MessageType]
@@ -31,9 +31,9 @@ class NamedWeightsUpdateRequest(TypedDict):
     names: List[str]
     dtypes: List[str]
     shapes: List[List[int]]
-    sizes: Optional[List[int]]
+    sizes: NotRequired[List[int]]
     extras: Optional[List[Dict[str, Any]]]
-    packed: Optional[bool]
+    packed: NotRequired[bool]
 
 
 class InferenceEngineInterface(ABC):
--- a/skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
+++ b/skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
@@ -143,6 +143,8 @@ class WorkerWrap:
             assert (
                 str_to_torch_dtype(dtypes[0]) == self.model_config.dtype
             ), f"mismatch dtype: src {dtypes[0]}, dst {self.model_config.dtype}"
+            assert len(sizes) == len(names), "sizes must be provided for packed weight update"
+            assert all(isinstance(size, int) for size in sizes), "sizes should be a list of integers"
 
             device = torch.cuda.current_device()
             props = torch.cuda.get_device_properties(device)


skyrl-train/skyrl_train/inference_engines/base.py
```diff
@@ -31,7 +31,9 @@ class NamedWeightsUpdateRequest(TypedDict):
     names: List[str]
     dtypes: List[str]
     shapes: List[List[int]]
+    sizes: List[int]
     extras: Optional[List[Dict[str, Any]]]
+    packed: bool
```
SumanthRH: The current PR will break compatibility for FSDP and DeepSpeed since they don't send these arguments. 

You should probably update the "sizes" to be a NotRequired argument. And you should update Fsdp and deepspeed worker files to send packed=False.

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
```diff
@@ -371,7 +398,9 @@ async def update_named_weights(self, request: NamedWeightsUpdateRequest):
                     request["names"],
                     request["dtypes"],
                     request["shapes"],
+                    request["sizes"],
```
SumanthRH: This should be an optional entry. Should use safer .get

erictang000: done

skyrl-train/skyrl_train/inference_engines/base.py
erictang000: good catch, updated and tested both

skyrl-train/skyrl_train/inference_engines/base.py
```diff
@@ -31,7 +31,9 @@ class NamedWeightsUpdateRequest(TypedDict):
     names: List[str]
     dtypes: List[str]
     shapes: List[List[int]]
+    sizes: Optional[List[int]]
```
SumanthRH: Nit: this is `NotRequired` . Optional means that hte field exists but is None

```python
from typing import NotRequired
```

skyrl-train/skyrl_train/inference_engines/base.py
```diff
@@ -31,7 +31,9 @@ class NamedWeightsUpdateRequest(TypedDict):
...
     extras: Optional[List[Dict[str, Any]]]
+    packed: Optional[bool]
```
SumanthRH: Same for this

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
```diff
@@ -127,29 +127,56 @@ def update_weights(self, names: List[str], dtypes: List[str], shapes: List[List[
...
-            # the key is to change device id to the current device id
-            # in case two processes have different CUDA_VISIBLE_DEVICES
             list_args[6] = device_id
-            weight = func(*list_args)
-            weight_list.append((name, weight))
+            packed_tensor = func(*list_args)
+
+            offset = 0
+            for name, shape, size in zip(names, shapes, sizes):
+                weight_list.append((name, packed_tensor[offset : offset + size].view(*shape)))
```
SumanthRH, line 162: dtype assert for this?

PR #491, opened 2025-10-16
Support configurable attention head sharding

tyler-griggs:
For dense models, we typically want to `tp` shard attention heads given that they make up a significant portion of total model parameters, For MoE models, however, attention head linear layers are a relatively small portion of parameters (due to the many expert layers), and it's not as useful to shard attention heads. 

With this motivation, this PR adds support for configuring whether attention heads are sharded.

### DIFF
--- a/skyrl-tx/tx/models/qwen3.py
+++ b/skyrl-tx/tx/models/qwen3.py
@@ -40,8 +40,7 @@ class Qwen3Attention(nnx.Module):
         if shard_attention_heads:
             assert self.num_heads % tp == 0, f"num_heads={self.num_heads} must be divisible by tp={tp}"
             assert self.num_kv_heads % tp == 0, f"num_kv_heads={self.num_kv_heads} must be divisible by tp={tp}"
-        qkv_shard_spec = jax.P(None, "tp") if shard_attention_heads else jax.P(None, None)
-        o_shard_spec = jax.P("tp", None) if shard_attention_heads else jax.P(None, None)
+        tp_shard = "tp" if shard_attention_heads else None
 
         self.head_dim = getattr(config, "head_dim", None) or config.hidden_size // self.num_heads
         max_lora_adapters = getattr(config, "max_lora_adapters", 0)
@@ -55,7 +54,7 @@ class Qwen3Attention(nnx.Module):
             dtype=dtype,
             param_dtype=dtype,
             use_bias=False,
-            kernel_init=nnx.with_partitioning(nnx.initializers.lecun_normal(), qkv_shard_spec),
+            kernel_init=nnx.with_partitioning(nnx.initializers.lecun_normal(), jax.P(None, tp_shard)),
             rngs=rngs,
         )
         self.k_proj = LoRALinear(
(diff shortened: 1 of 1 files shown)


skyrl-tx/tx/tinker/engine.py
```diff
@@ -387,7 +388,7 @@ def process_forward_backward_batch(
         )
         loss_mask = jnp.array(
             [all_token_weights[i] + [0] * (max_len - len(all_input_ids[i])) for i in range(len(all_token_weights))],
-            dtype=jnp.int32,
+            dtype=jnp.bool_,
```
tyler-griggs, line 391: I started getting an error from inside jax when calling `softmax_cross_entropy_with_integer_labels` that said:
```
    def check_where(name: str, where: ArrayLike | None) -> Array | None:
      if where is None:
        return where
      where = ensure_arraylike(name, where)
      if where.dtype != bool:
>       raise ValueError(
          f"jnp.{name}: where must be None or a boolean array; got {where.dtype=}."
        )
E       ValueError: jnp.max: where must be None or a boolean array; got where.dtype=dtype('int32').
```

So I switched this mask to a boolean array.

skyrl-tx/tx/tinker/engine.py
```diff
@@ -116,6 +116,7 @@ def __init__(
         # Configure LoRA settings
         self.model_config.max_lora_adapters = self.config.max_lora_adapters
         self.model_config.max_lora_rank = self.config.max_lora_rank
+        self.model_config.shard_attention_heads = self.config.shard_attention_heads
```
pcmoritz, line 119: We will need to find a better solution for this going forward but for now we can just keep adding it to the model_config :D

pcmoritz (approved):
Nice, this looks great :)

skyrl-tx/tx/models/qwen3.py
```diff
@@ -36,8 +36,13 @@ def __init__(self, config: Qwen3Config, *, dtype: jnp.dtype, rngs: nnx.Rngs) ->
         self.num_heads = config.num_attention_heads
         self.num_kv_heads = config.num_key_value_heads
         tp = get_abstract_mesh().shape.get("tp", 1)
-        assert self.num_heads % tp == 0, f"num_heads={self.num_heads} must be divisible by tp={tp}"
-        assert self.num_kv_heads % tp == 0, f"num_kv_heads={self.num_kv_heads} must be divisible by tp={tp}"
+        shard_attention_heads = getattr(config, "shard_attention_heads", True)
+        if shard_attention_heads:
+            assert self.num_heads % tp == 0, f"num_heads={self.num_heads} must be divisible by tp={tp}"
+            assert self.num_kv_heads % tp == 0, f"num_kv_heads={self.num_kv_heads} must be divisible by tp={tp}"
+        qkv_shard_spec = jax.P(None, "tp") if shard_attention_heads else jax.P(None, None)
```
pcmoritz: A different way would be `tp_shard = "tp" if shard_attention_heads else None` and then do `jax.P(None, tp_shard)` and `jax.P(tp_shard, None)` inline below, seems a little simple, I wonder if you had considered it (but not a big difference).

skyrl-tx/tx/tinker/engine.py
tyler-griggs: Yes agreed!

PR #503, opened 2025-10-17
[tx] resume training runs with sl_loop.py

Guido1Alessandro1Trevisan:
Resume checkpoint now works with sl_loop. Added the Checkpoint type to pave the way for storing checkpoint metadata in the database (I think someone is already working on it).


pcmoritz (approved):
Awesome, thanks for the contribution! I lifted the data validation into the API endpoint and introduced some common types to do the validation :) Also added a test.

PR #508, opened 2025-10-17
Remove truncation logic, fix corresponding tests

devpatelio:
*Early exit agent loop on exceeding max_input_length to avoid post-generation truncation*

This PR modifies the SkyRLGymGenerator to stop trajectory generation once the token count exceeds the maximum input length, instead of truncating trajectories after completion. This simplifies control flow and removes redundant truncation logic in skyrl_gym_generator.py. Additionally, the test was failing because the mock environment wasn't respecting the max_turns parameter and was returning incorrect rewards. The environment was hardcoded to do 2 turns regardless of max_turns setting, and the reward logic was placing rewards at the wrong token positions.

**Changes:**
- In 'skyrl_gym_generator', I added early-exit logic to stop trajectory generation once the token count exceeds max_input_length, preventing unnecessary continuation beyond limits.

- In 'test_agent_loop_truncation_drops_out_of_range_rewards', I updated TruncEnv.step() logic to correctly handle max_turns=1 case (was prev. letting 2 turns happen, causing 9 = 5 error). I also fixed reward placement to expect reward at last EOS token position as we removed the premature truncation step such that we instead add the EOS token manually (thus, reward expected at last token) --> expect reward=2.0 at index 4 (EOS token). I also changed stop_reason expectation from "length" to "stop" 

- In 'test_apply_overlong_filtering_non_batched', we expect 5 tokens but got 6 because we add the EOS token. If we don't want to add an EOS token for stop_reason = "length", then I updated the skyrl_gym_generator logic to account for this. Additionally, there was another test with stop reason "length" but actually expected an EOS token so I changed the stop reason to "stop" and that resolved test errors.

If the stop reason = length, we do not add the EOS token. If the stop reason = stop, we do add the EOS token. We update the loss mask and reward placement accordingly.

### DIFF
--- a/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
+++ b/skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
@@ -157,7 +157,6 @@ class SkyRLGymGenerator(GeneratorInterface):
         session_id = (
             f"{trajectory_id.instance_id}_{trajectory_id.repetition_id}" if trajectory_id is not None else uuid4().hex
         )
-
         done = False
 
         # Instantiate chat_history and chat_end_index, which are only used if `retokenize_chat_history==True`.
@@ -200,7 +199,6 @@ class SkyRLGymGenerator(GeneratorInterface):
                 engine_input = InferenceEngineInput(
                     prompt_token_ids=[input_ids], session_ids=[session_id], sampling_params=sampling_params
                 )
-
             engine_output = await self.inference_engine_client.generate(engine_input)
             output = engine_output["responses"][0]
             output_ids = engine_output["response_ids"][0]
@@ -284,7 +282,6 @@ class SkyRLGymGenerator(GeneratorInterface):
         assert len(loss_mask) == len(response_ids), "loss_mask and response_ids should have the same length"
 
         appended_eos_token = False
-
         if not self.use_conversation_multi_turn:
             if stop_reason != "length" and response_ids and response_ids[-1] != self.tokenizer.eos_token_id:
                 response_ids.append(self.tokenizer.eos_token_id)


skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -181,10 +182,14 @@ async def agent_loop(
         loss_mask = []  # this excludes the prompt
         rollout_logprobs = None
         # Accumulate per-step rewards. Format: (reward, response_end_token_idx)
-        per_step_rewards: List[Tuple[float, Optional[int]]] = []
+        per_step_rewards: List[Tuple[Optional[float], Optional[int]]] = []
```
tyler-griggs: Why is this changed to `Optional`?

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -181,10 +182,14 @@ async def agent_loop(
...
 
         while not done:
-            # 1. Generate output
```
tyler-griggs, line 187: Please keep this comment, moved down above `if retokenize_chat_history:`

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -281,25 +283,13 @@ async def agent_loop(
...
                 response_ids.append(self.tokenizer.eos_token_id)
                 loss_mask.append(1)
                 appended_eos_token = True
 
-        # need to truncate loss mask correctly for responses that go to max length
-        if self.max_turns > 1:
-            # max total resp length = max tokens (max length of final turn generation) + max_input_length (max input for any generation turn) - len(original prompt)
-            max_response_tokens = max_tokens + max_input_length - initial_prompt_length
-        else:
-            max_response_tokens = max_tokens
```
tyler-griggs, line 296: Is `max_tokens` now completely unused in `agent_loop`? If so, let's remove it from the method arguments

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator.py
```diff
@@ -759,13 +759,24 @@ def mock_apply_chat_template(messages, **kwargs):
     generator.base_conversation_token_ids = []  # to make sure observation_ids are encoded correctly
 
     # First test: response that doesn't end with eos token (should be filtered)
-    mock_llm.generate = AsyncMock(
-        return_value={
-            "responses": ["truncated response"],
-            "stop_reasons": ["length"],
-            "response_ids": [[10, 11, 12, 13, 14, 15, 16, 17, 18, 19]],  # 10 tokens, will be truncated
+    # TODO(Dev): Fix is fine for now, will revisit in case issues arise
```
tyler-griggs: What `fix` is this referring to? Is this comment needed? If so, please add detail so we know what it's referring to. If not, let's delete

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator.py
```diff
@@ -777,23 +788,35 @@ def mock_apply_chat_template(messages, **kwargs):
 
     # Verify truncated response has zeroed loss mask
     assert len(output_truncated["loss_masks"]) == 1
-    assert len(output_truncated["loss_masks"][0]) == 5  # Truncated to max_generate_length=5
+    # assert len(output_truncated["loss_masks"][0]) == 6  # Truncated to max_generate_length=5
```
tyler-griggs: Delete this commented out code?

skyrl-train/tests/cpu/generators/test_skyrl_gym_generator.py
```diff
@@ -1195,15 +1226,16 @@ class TruncEnv(BaseTextEnv):
...
             return prompt, {}
 
         def step(self, action):
             self.turns += 1
-            if self.turns == 1:
+            if self.turns < self.max_turns:
                 return BaseTextEnvStepOutput(observations=[], reward=1.0, done=False, metadata={})
-            else:
+            else:  ## reach max turns, return final reward
                 return BaseTextEnvStepOutput(observations=[], reward=2.0, done=True, metadata={})
```
tyler-griggs:
```suggestion
            else:  
                # On the final turn, return the final reward.
                return BaseTextEnvStepOutput(observations=[], reward=2.0, done=True, metadata={})
```

PR #511, opened 2025-10-17
[tx] Added lora embed

Guido1Alessandro1Trevisan:
First attempt at adding LoRA support for the embeddings layer.
I’ve verified the forward backward passes and optimizer step everything runs and the loss decreases as expected.

I’m wondering what better tests are there to confirm that LoRA is functioning correctly. Subtle token routing bugs could really ruin someone’s day.

### DIFF
--- a/skyrl-tx/tests/tinker/test_api.py
+++ b/skyrl-tx/tests/tinker/test_api.py
@@ -32,6 +32,9 @@ def api_server():
             "8000",
             "--base-model",
             BASE_MODEL,
+            # Set number of LoRA adapters lower to avoid OOMs in the CI
+            "--max-lora-adapters",
+            "4",
         ],
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,


skyrl-tx/tx/layers/lora.py
```diff
@@ -84,6 +84,74 @@ def apply_lora(
...
+            sharding_A=jax.sharding.PartitionSpec(None, sharding[0], None),
+            sharding_B=jax.sharding.PartitionSpec(None, None, sharding[1]),
+            dtype=param_dtype,
+            rngs=rngs,
+        )
+
+    def __call__(self, x: jax.Array, adapter_indices: jax.Array | None = None) -> jax.Array:
+        base_out = super().__call__(x)
+
+        if self.max_lora_adapters == 0 or adapter_indices is None:
```
pcmoritz: I think here we should re-use self.apply_lora (we might need to parametrize or modify that slightly)

Guido1Alessandro1Trevisan: Definitely. I was hesitant on the first attempt since there are quite a few small differences here and there between the embedding lora and the linear lora. a better apply_lora is something to brood on

Guido1Alessandro1Trevisan:
Bumping this for merge. parametrized apply_lora,  fixed the test and an issue with the lora expert layer.

pcmoritz (approved):
I simplified the code some more now and added a test, it should be ready to merge! Thanks a lot for the contribution :)

PR #515, opened 2025-10-18
[GPTOSS] Integrate Unsloth's flex attention implementation for attention sink

SumanthRH:
# What does this PR do?

Integrates unsloth's flex attention implementation for GPT-OSS. Validated for single turn training for GSM8K. 

This PR extends Unsloth's flex attention implementation by adding support for user specified attention masks.

## Overall Changes
- Added flex attention implementation files in `patches/gptoss` based on Unsloth's implementation
       Key Changes: I used their `old_flex_attention_with_sink` function because it is meant for use in training. The original function doesn't support user specified attention mask. I added support for an attention mask in this PR. 
- Introduce a custom attention impl `custom_flex` that wraps the `old_flex_attention_with_sink` function. This was not enough, and I also had to implement a custom attn mask implementation `custom_attention_mask`. This is because the default attention mask function used is `flash_attention_mask` in transformers: https://github.com/huggingface/transformers/blob/2e67a9b602a8aa5fcf9f82e25b4e331f3328976c/src/transformers/masking_utils.py#L581 . This will set the attention_mask to be `None` and assume that position_ids wil be used (but our custom flex attention function doesn't use position ids): https://github.com/huggingface/transformers/blob/2e67a9b602a8aa5fcf9f82e25b4e331f3328976c/src/transformers/masking_utils.py#L612-L615

## Micro Benchmarking

I created a script `bench_flex_attn.py` to benchmark the flex attention implementation against eager attention. Currently, the forward pass with flex attention is fast enough - it beats eager attention as you scale context length > 1K. However, the backward pass is inefficient and can be optimized. Further, flex attention can scale to larger context lengths like 16K for a single attention layer whereas eager will OOM.

### Benchmarking results

Context Length | Method | Fwd Time (ms) ± std | Bwd Time (ms) ± std
-- | -- | -- | --
512 | Eager | 1.994 ± 0.0466 | 2.400 ± 0.1201
512 | Flex Attn | 2.777 ± 0.3506 | 52.976 ± 0.2773
1024 | Eager | 5.241 ± 0.0331 | 6.644 ± 0.0492
1024 | Flex Attn | 3.274 ± 0.0729 | 205.246 ± 0.7510
2048 | Eager | 20.701 ± 0.0095 | 22.270 ± 0.0636
2048 | Flex Attn | 5.415 ± 0.5166 | 860.371 ± 2.2019
4096 | Eager | 59.065 ± 0.1233 | 84.267 ± 0.0776
4096 | Flex Attn | 11.944 ± 0.4277 | 3610.017 ± 3.3435
8192 | Eager | OOM | OOM
8192 | Flex Attn | 34.053 ± 0.4521 | 14770.799 ± 9.1954
16384 | Eager | OOM | OOM
16384 | Flex Attn | 103.907 ± 0.3850 | 56920.245 ± 17.4912

## Correctness

For testing correctness for the patch, I have a simple test script `test_gptoss.py` to compare logprobs for eager and flex attention. Logprobs match upto 2 decimal places. I've also noticed that the difference can increase with longer context lengths, I plan to investigate this in a future PR. 

## End-to-End validation

I've validated training for GSM8K in this PR. results are similar to eager. 

Screenshot 2025-10-21 at 8 10 17 PM
Screenshot 2025-10-21 at 8 10 23 PM
Screenshot 2025-10-21 at 8 10 37 PM

## What is not supported
- Sequence packing
- Sequence parallelism: To be added in a followup

### DIFF
--- a/skyrl-train/skyrl_train/model_wrapper.py
+++ b/skyrl-train/skyrl_train/model_wrapper.py
@@ -127,6 +127,7 @@ class HFModelWrapper(nn.Module):
                     from skyrl_train.patches.gptoss.patch_transformers import (
                         custom_attention,
                         custom_attention_mask,
+                        patch_GptOssAttention,
                     )
                     from transformers import AttentionInterface, AttentionMaskInterface
 
@@ -135,6 +136,9 @@ class HFModelWrapper(nn.Module):
                     # set attention implementation to be `custom_flex`
                     self.model.set_attn_implementation("custom_flex")
                     self.attn_implementation = "custom_flex"
+                    # NOTE: Even though we set a custom attn implementation, we
+                    # also patch the full attention function for GPT OSS
+                    patch_GptOssAttention()
 
             # LoRA
             if lora_rank > 0:


skyrl-train/examples/gptoss/bench_flex_attn.py
```diff
@@ -0,0 +1,163 @@
+from transformers.models.gpt_oss.modeling_gpt_oss import GptOssAttention, GptOssRotaryEmbedding
```
erictang000, line 1: are we keeping this test here in examples? seems ok to keep just checking

erictang000 (approved):
LGTM! just question about the example test scripts

skyrl-train/examples/gptoss/bench_flex_attn.py
SumanthRH: Yeah I don't see why not. Unsloth only had their patches but didn't have any benchmarking scripts, so I added this micro benchmark. I think this script is helpful for people to know what they are getting with the patches we have.

PR #516, opened 2025-10-18
[tx] independent optimizer per-adapter + per-request optim_step hyperparam overrides

cwwarren:
Following previous PR https://github.com/NovaSky-AI/SkyRL/pull/470 - a simple sparse approach to per-request hyperparams and per-adapter optimizer independence.

Each adapter now has an independent `nnx.Optimizer`. The `lr` field is renamed to `learning_rate` and beta1/beta2/eps are added to match the [official client](https://github.com/thinking-machines-lab/tinker/blob/main/src/tinker/types/optim_step_request.py). We default to the same values as the official client.

### DIFF
--- a/skyrl-tx/tx/tinker/api.py
+++ b/skyrl-tx/tx/tinker/api.py
@@ -180,10 +180,10 @@ class ForwardBackwardInput(BaseModel):
 
 
 class AdamParams(BaseModel):
-    learning_rate: float = Field(default=1e-4, ge=0.0, lt=1.0)
+    learning_rate: float = Field(default=1e-4, ge=0.0)
     beta1: float = Field(default=0.9, ge=0.0, lt=1.0)
     beta2: float = Field(default=0.95, ge=0.0, lt=1.0)
-    eps: float = Field(default=1e-12, gt=0.0, lt=1.0)
+    eps: float = Field(default=1e-12, gt=0.0)
 
     def to_adam_params(self) -> types.AdamParams:
         return types.AdamParams(learning_rate=self.learning_rate, beta1=self.beta1, beta2=self.beta2, eps=self.eps)


skyrl-tx/tx/tinker/api.py
```diff
@@ -180,7 +180,10 @@ class ForwardBackwardInput(BaseModel):
 
 
 class AdamParams(BaseModel):
-    lr: float = 1e-4
+    learning_rate: float | None = Field(default=None, ge=0.0, lt=1.0)
+    beta1: float | None = Field(default=None, ge=0.0, lt=1.0)
+    beta2: float | None = Field(default=None, ge=0.0, lt=1.0)
+    eps: float | None = Field(default=None, gt=0.0, lt=1.0)
```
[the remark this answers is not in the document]
cwwarren: Good catch Gemini on the discrepancy, although 1.0 values for any of these wouldn't be expected. I updated the PR description to be `<` for consistency but would love any thoughts on whether it's worth allowing `1.0` for the hyperparams. (This is also trivially changed at any time)

skyrl-tx/tx/tinker/engine.py
```diff
@@ -495,10 +514,9 @@ def expand_adapter_grads(lora_param, adapter_grad):
 
         full_lora_grads = jax.tree.map(expand_adapter_grads, self.lora_params, adapter_grads)
 
-        # Apply optimizer update -- going forward we need to figure out how to use different learning rates per adapter
-        adam_params = request_data.adam_params
-        assert adam_params.lr == LEARNING_RATE, f"Currently we only support a fixed learning rate {LEARNING_RATE}"
-        self.optimizer.update(self.lora_params, full_lora_grads)
+        self._set_opt_hyperparams(model_id, request_data.adam_params)
+        self.optimizers[model_id].update(self.lora_params, full_lora_grads)
+        self._set_opt_hyperparams(model_id, DEFAULT_ADAM_PARAMS)
```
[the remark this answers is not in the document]
cwwarren: If an exception was thrown in `.update(...)` it would leave stale overridden hyperparams however we explicitly set all hyperparams to the default or overridden value before each call to `.update(...)` so there's no impact from the stale values.

I'd be happy to make this change to avoid the stale values if desired, but I'd rather rely on always explicitly setting the hyperparams before an optim step which avoids all potential stale-hyperparam issues.

skyrl-tx/tx/tinker/api.py
```diff
@@ -353,7 +356,14 @@ async def optim_step(request: OptimStepRequest, session: AsyncSession = Depends(
...
         model_id=request.model_id,
-        request_data=types.OptimStepInput(adam_params=types.AdamParams(lr=request.adam_params.lr)),
+        request_data=types.OptimStepInput(
+            adam_params=types.AdamParams(
+                learning_rate=request.adam_params.learning_rate,
+                beta1=request.adam_params.beta1,
+                beta2=request.adam_params.beta2,
+                eps=request.adam_params.eps,
+            )
+        ),
```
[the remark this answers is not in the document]
cwwarren: I intentionally was explicit here to avoid any silent fallbacks to default if the models were mismatched in any way (e.g. `lr` vs `learning_rate`).

We could unify the types since they are identical, but there's also a good argument for keeping API types separate from internal types.

skyrl-tx/tests/tinker/test_engine.py
```diff
@@ -200,3 +203,59 @@ def test_micro_batch_grad_accumulation():
...
+
+    # Apply override step on the first adapter.
+    tiny_norm = apply_step(1, low_adapter, tiny_request)
+
+    # Verify the request hyperparameter overrides do not persist after the step.
+    hyperparams = engine.optimizers[low_adapter].opt_state.hyperparams
+    assert hyperparams["learning_rate"].value == pytest.approx(DEFAULT_ADAM_PARAMS.learning_rate, rel=2e-3)
+    assert hyperparams["b1"].value == pytest.approx(DEFAULT_ADAM_PARAMS.beta1, rel=2e-3)
+    assert hyperparams["b2"].value == pytest.approx(DEFAULT_ADAM_PARAMS.beta2, rel=2e-3)
+    assert hyperparams["eps"].value == pytest.approx(DEFAULT_ADAM_PARAMS.eps, rel=2e-3)
```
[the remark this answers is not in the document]
cwwarren: The current code is more clear imo.

skyrl-tx/tx/tinker/engine.py
```diff
@@ -216,6 +212,21 @@ def loss_for_lora(
...
+
+    def _set_opt_hyperparams(self, model_id: str, adam_params: types.AdamParams) -> None:
+        """Set optimizer hyperparameters, defaulting unspecified values."""
+        hp = self.optimizers[model_id].opt_state.hyperparams
+        hp["learning_rate"][...] = (
+            DEFAULT_ADAM_PARAMS.learning_rate if adam_params.learning_rate is None else adam_params.learning_rate
+        )
+        hp["b1"][...] = DEFAULT_ADAM_PARAMS.beta1 if adam_params.beta1 is None else adam_params.beta1
+        hp["b2"][...] = DEFAULT_ADAM_PARAMS.beta2 if adam_params.beta2 is None else adam_params.beta2
+        hp["eps"][...] = DEFAULT_ADAM_PARAMS.eps if adam_params.eps is None else adam_params.eps
```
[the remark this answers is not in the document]
cwwarren: This is much less clear than the current code.

skyrl-tx/tx/tinker/api.py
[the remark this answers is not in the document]
pcmoritz: I would say let's make the fields in the internal type `types.AdamParams` required (the one in `types.py`), and then I think we can do this simplification safely. In general I think it is best if defaults get filled in as early as possible (e.g. at the request layer), and make all the internal type fields and database columns required, it avoids multiple possible code paths and makes things easier to debug.

pcmoritz:
Thanks, this looks really great! I'll take a pass and will try to simplify it a little bit, in particular, always fill out the defaults at the request level, and don't reset the parameters to the defaults (since all of them get set on each request, either with the user provided values, or the defaults from the request).

cwwarren:
I wasn't sure if we'd want to preserve a marker for overridden/default in the engine but if there's no need for that your approach makes sense to simplify the defaults/None handling upfront into the original request object.

I'll quickly knock that out, no worries!

pcmoritz:
Yeah, let's just make sure the parameters are always set on each request, and then it doesn't even matter what they get initialized with initially (we can even omit that), so at the end the only numbers we provide are the defaults in the pydantic schema of the API request type :)

cwwarren:
One quick question however, if we make the new values (beta1/beta2/eps) required fields on types.AdamParams without a default it may break loading existing requests from the DB since they wouldn't have those fields set. Are there any concerns with that breakage?

pcmoritz:
Yeah, at the moment we don't guarantee any schema evolutions yet for the database. I hope we can get to a 0.1 release pretty soon (where "most" of the functionality that a user might reasonable need is implemented -- this PR is certainly one of the features users will expect), and from that point on we can provide schema migrations and be careful with the schema evolution. Right now, people should just delete their local sqlite db when the code gets incompatible (I think that's a reasonable stance for a sub 0.1 release).

cwwarren:
Very reasonable and thanks for the feedback! I pushed a new commit with those simplifications made, let me know if there's anything else.

skyrl-tx/tx/tinker/api.py
```diff
@@ -353,7 +356,7 @@ async def optim_step(request: OptimStepRequest, session: AsyncSession = Depends(
         session=session,
         request_type=types.RequestType.OPTIM_STEP,
         model_id=request.model_id,
-        request_data=types.OptimStepInput(adam_params=types.AdamParams(lr=request.adam_params.lr)),
+        request_data=types.OptimStepInput(adam_params=request.adam_params.model_dump()),
```
pcmoritz: You actually don't need the `.model_dump()` here

skyrl-tx/tx/tinker/engine.py
```diff
@@ -329,6 +328,11 @@ def process_create_model(self, model_id: str, request_data: types.CreateModelInp
         )
         self.accumulated_grads[model_id] = AccumulatedGradients(grad_sum=None, denominator=0)
 
+        with jax.set_mesh(self.mesh):
+            # These values are always overridden by the hyperparams in the optim_step request.
+            tx = optax.inject_hyperparams(optax.adamw)(learning_rate=1e-4, b1=0.9, b2=0.95, eps=1e-12)
```
pcmoritz: Why don't we just do `tx = optax.inject_hyperparams(optax.adamw)(learning_rate=0.0)`, this will make it even clearer that these are not values that are going to be used and remove "magic" constants (I know you also say that in the comment, but why not also make it clear in the code).

skyrl-tx/tests/tinker/test_engine.py
```diff
@@ -200,3 +204,52 @@ def test_micro_batch_grad_accumulation():
...
+        assert isinstance(result, types.OptimStepOutput)
+        params_after = nnx.to_arrays(nnx.pure(engine.lora_params))
+
+        delta = jax.tree.map(lambda a, b: a - b, params_after, params_before)
+        return jnp.sqrt(jax.tree.reduce(lambda a, x: a + (x.astype(jnp.float32) ** 2).sum(), delta, 0.0)).item()
+
+    tiny_request = types.OptimStepInput(
+        adam_params=types.AdamParams(learning_rate=1e-8, beta1=1e-8, beta2=1e-8, eps=1e-9)
+    )
+    default_request = types.OptimStepInput(adam_params=api.AdamParams().model_dump())
```
pcmoritz: Here you also don't need the `.model_dump()` I think

pcmoritz (approved):
Awesome thanks a lot for the contribution, this looks really great! I'll try it out a little more later tonight and then get it merged :)

skyrl-tx/tests/tinker/test_engine.py
cwwarren: I didn't think so either but when I ran the server to test 🤷‍♂️

```
  File "/home/colin/dev/SkyRL/skyrl-tx/.venv/lib/python3.12/site-packages/pydantic/main.py", line 250, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 1 validation error for OptimStepInput
adam_params
  Input should be a valid dictionary or instance of AdamParams [type=model_type, input_value=AdamParams(learning_rate=..., beta2=0.95, eps=1e-12), input_type=AdamParams]
    For further information visit https://errors.pydantic.dev/2.12/v/model_type
```

cwwarren: I just tested removing model_dump from both locations in the tests and api.py and they both raise that exception without it. 

I remember I used to be able to do something like this in Pydantic, I'm not sure if this was a v1→v2 change or if it's due to subtle differences in the fields (e.g. default values/validation).

skyrl-tx/tx/tinker/engine.py
cwwarren: I pushed a small commit to make these all explicitly 0.0.

skyrl-tx/tests/tinker/test_engine.py
pcmoritz: Ah yes you are right, it seems to be needed, sorry for the confusion :)

pcmoritz: I added a small method to do this now, so if one of the pydantic models changes in the future, the type checker can spot it before a mismatch happens at runtime :)

PR #523, opened 2025-10-20
Add implementation of CISPO loss

vutrung96:
Did a quick test run on gsm8k with Qwen-2.5-0.5B. We seem to converge quicker but the asymptote is roughly the same as PPO

Screenshot 2025-10-19 at 6 11 39 PM

The CISPO-trained model does seem more verbose which might suggest that "reflective tokens" are less suppressed and the model can reason for longer, but policy entropy is actually lower than PPO, so I'm not sure.

Screenshot 2025-10-19 at 6 19 25 PM
Screenshot 2025-10-19 at 6 19 33 PM

Full wandb report:
https://api.wandb.ai/links/bespoke-labs/de4bz2z5

### DIFF
--- a/skyrl-train/docs/configuration/config.rst
+++ b/skyrl-train/docs/configuration/config.rst
@@ -397,6 +397,11 @@ Algorithm Configuration
         kl_cov_frac: 0.2 # percentage of tokens to apply KL regularization to (20%)
         ppo_kl_coef: 1.0 # coefficient for KL regularization term
 
+      # cispo parameters (only used when policy_loss_type: "cispo")
+      cispo:
+        clip_low: 0.2 # lower bound for importance sampling ratio clipping
+        clip_high: 0.2 # upper bound for importance sampling ratio clipping
+
       # value loss parameters
       value_clip: 0.2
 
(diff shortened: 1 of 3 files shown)


vutrung96:
@erictang000 i used the pre-commit from the repo and it also re-formatted other parts of the code that are not changed... apologies if the diffs are hard to understand

erictang000:
@vutrung96 thanks, this looks right to me!

One main note, could we add separate clipping parameters for cispo, similar to what we're doing for clip_cov right now (add `algorithm.cispo.clip_low`/`algorithm.cispo.clip_high`)? Then we can support the general formulation for CISPO that they mention using the existing `eps_clip_high`/`eps_clip_low` ppo clipping params.

image

there's a pending PR in verl that has added this for reference: https://github.com/jzstudent/verl/blob/88a3f2296d487012a3a6f5686ea8d3efac5b4c08/verl/trainer/ppo/core_algos.py#L1216

vutrung96:
> @vutrung96 thanks, this looks right to me!
> 
> One main note, could we add separate clipping parameters for cispo, similar to what we're doing for clip_cov right now (add `algorithm.cispo.clip_low`/`algorithm.cispo.clip_high`)? Then we can support the general formulation for CISPO that they mention using the existing `eps_clip_high`/`eps_clip_low` ppo clipping params.
> 
> image
> there's a pending PR in verl that has added this for reference: https://github.com/jzstudent/verl/blob/88a3f2296d487012a3a6f5686ea8d3efac5b4c08/verl/trainer/ppo/core_algos.py#L1216

this makes sense @erictang000. I've added a separate set of CISPO params instead of reusing PPO params. PTAL!

erictang000 (approved):
looks great thanks for adding this! 🚀🚀

tyler-griggs:
Awesome!!

PR #525, opened 2025-10-20
[train] Increase default timeout for placement groups to 180s

SumanthRH:
# What does this PR do?

Increases the default timeout for placement groups to 3 minutes. I've noticed that in multi-node clusters the initial worker startup time can be easily ~ 1 min because it takes time to download all required packages with uv. Thus, we need to account for additional time at startup. 

This could be increased even more, but this should work okay for small clusters for now (the tradeoff is that you might wait too long even though pg is not satisfiable)


SumanthRH:
Test failures not related

PR #532, opened 2025-10-21
[skyrl-train] Add DAPO 7B recipe, and 32B training script

erictang000:
Adds tested DAPO 7B recipe and adds a README.md

### DIFF
--- a/skyrl-train/examples/algorithms/dapo/README.md
+++ b/skyrl-train/examples/algorithms/dapo/README.md
@@ -1,4 +1,11 @@
 # DAPO
+
+## Reproduction Runs
+| Recipe  | Model  |  AIME24 Pass@32 | AIME24 Avg Score | Hardware  | WandB |
+|---|---|---|---|---|---|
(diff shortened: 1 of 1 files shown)


skyrl-train/skyrl_train/trainer.py
```diff
@@ -189,7 +189,6 @@ def train(self):
                         generator_output, uids, keep_sampling = self.handle_dynamic_sampling(generator_output, uids)
                         if keep_sampling:  # continue sampling
                             # update progress bar for current batch (but not global step)
-                            pbar.update(1)
```
SumanthRH, line 192: Why is this change needed? I believe we explicitly changed the progress bar to "Training Batches Processed" because of dynamic sampling.... 

We also won't know the size of the progress bar if we did "Training steps" only ....

erictang000: hmm yeah i think i just wanted to see the actual training step while training. reverted

SumanthRH: Isn't that available as `global_step` ?

PR #534, opened 2025-10-21
 [tx] Adds `prompt_logprobs` parameter to sampling API

Guido1Alessandro1Trevisan:
Adds prompt_logprobs flag to generator to output the prompt_logprobs of the model


pcmoritz:
Btw @Guido1Alessandro1Trevisan if you want to remove the conflicts and make this mergeable, I think now the code has exactly the right hole that would be filled by this PR (basically initializing all_logprobs from the prefill results, and passing it down). We probably don't even need a flag for it since passing down the logprobs should be basically free. No rush, this is not time critical, but also shouldn't be much work :)

Guido1Alessandro1Trevisan:
> Btw @Guido1Alessandro1Trevisan if you want to remove the conflicts and make this mergeable, I think now the code has exactly the right hole that would be filled by this PR (basically initializing all_logprobs from the prefill results, and passing it down). We probably don't even need a flag for it since passing down the logprobs should be basically free. No rush, this is not time critical, but also shouldn't be much work :)

Sounds good i'll close this pr and open a new one to add prompt_logprobs

Issue #520, opened 2025-10-19
[tx] Support postgres as the database


pcmoritz:
People who actually want to run SkyRL tx in production will likely want to use postgres as the database, so we should support it.

atemaguer:
I'll take a stab at this and work on this #519 issue afterwards as well

pcmoritz:
Awesome, that would be great :)

pcmoritz:
This is now merged -- at some point it would be nice to have CI tests against postgres too, but that can come later, I'll close this for now :)

PR #545, opened 2025-10-21
[tx] Postgres Database Support

atemaguer:
This PR implements the #520 issue and introduces PostgreSQL database support. Now users can just define a `TINKER_DATABASE_URL` environment variable and the engine will persist data to this remote Postgres database when skyrl-tx is deployed as a service.

### DIFF
--- a/skyrl-tx/pyproject.toml
+++ b/skyrl-tx/pyproject.toml
@@ -40,8 +40,6 @@ tinker = [
     "sqlalchemy[asyncio]",
     "aiosqlite",
     "asyncpg",
-    "psycopg2-binary",
-    "python-dotenv",
 ]
 
 aws = [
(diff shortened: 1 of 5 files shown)


pcmoritz:
Thanks a lot for this PR @atemaguer, it looks great! Can you make a few small changes before we merge it?

1. Remove some of the spurious formatting changes (like in the readme etc) -- if you really want to make them, that's ok but it should be a separate PR in that case
2. Change the environment variable from `TINKER_` to `TX_` -- we shouldn't be claiming the `TINKER_` environment variable namespace, they might clash e.g. with the tinker SDK
3. Instead of parsing the urls manually, the suggestion that gemini made with `sqlalchemy.engine.url` seems nice, do you want to adopt it?
4. Let's remove the manual loading of the .env variable so there are not too many ways to configure this -- if people do want to use the .env file, it will be better if they use https://docs.astral.sh/uv/concepts/configuration-files/#env instead of us handling it in the code

atemaguer:
I'm on it.

PR #559, opened 2025-10-23
[tx] Add gradient checkpointing support

tyler-griggs:
This PR adds support for using jax's gradient `checkpoint_policies` to reduce memory pressure during the fwd_bwd execution.

The gradient checkpoint policy is configurable between jax's policies: checkpoint_dots_with_no_batch_dims, dots_with_no_batch_dims_saveable, nothing_saveable, or `None` for the default policy of full recomputation.

### DIFF
--- a/skyrl-tx/tests/tinker/test_engine.py
+++ b/skyrl-tx/tests/tinker/test_engine.py
@@ -259,19 +259,17 @@ def test_gradient_checkpointing():
     """
     Verify gradient checkpointing doesn't affect loss values.
     """
-    base_kwargs = dict(
-        base_model="Qwen/Qwen3-0.6B",
-        enforce_eager=False,
-        train_batch_size=2,
-        micro_batch_size=1,
-        max_lora_adapters=1,
-        max_lora_rank=4,
-    )
-
     losses = []
     for use_gradient_checkpointing in (False, True):
-        cfg = EngineConfig(**base_kwargs)
-        cfg.gradient_checkpointing = use_gradient_checkpointing
+        cfg = EngineConfig(
+            base_model="Qwen/Qwen3-0.6B",
+            enforce_eager=False,
+            train_batch_size=2,
+            micro_batch_size=1,
+            max_lora_adapters=1,
+            max_lora_rank=4,
+            gradient_checkpointing=use_gradient_checkpointing,
+        )
         engine = TinkerEngine(cfg)
 
         # Create batch


skyrl-tx/tx/tinker/config.py
```diff
@@ -26,6 +26,12 @@ class EngineConfig(BaseModel):
         default=True,
         description="Whether to shard attention linear layers (qkvo projections) across tensor parallel devices",
     )
+    gradient_checkpointing: bool = Field(default=False, description="Whether to use gradient checkpointing")
+    gradient_checkpoint_policy: str | None = Field(
```
pcmoritz: Currently it seems like something like `str | None` is not supported in the argparse stuff:
```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/ray/default/SkyRL/skyrl-tx/tx/tinker/api.py", line 720, in <module>
    add_model(parser, EngineConfig)
  File "/home/ray/default/SkyRL/skyrl-tx/tx/tinker/config.py", line 66, in add_model
    parser.add_argument(f"--{arg_name}", **kwargs)
  File "/home/ray/anaconda3/lib/python3.12/argparse.py", line 1487, in add_argument
    raise ValueError('%r is not callable' % (type_func,))
ValueError: str | None is not callable
```

I think it is possible to work around this right now with
```python
diff --git a/skyrl-tx/tx/tinker/config.py b/skyrl-tx/tx/tinker/config.py
index 2ae8f8b..4c641f1 100644
--- a/skyrl-tx/tx/tinker/config.py
+++ b/skyrl-tx/tx/tinker/config.py
@@ -27,8 +27,8 @@ class EngineConfig(BaseModel):
         description="Whether to shard attention linear layers (qkvo projections) across tensor parallel devices",
     )
     gradient_checkpointing: bool = Field(default=False, description="Whether to use gradient checkpointing")
-    gradient_checkpoint_policy: str | None = Field(
-        default=None,
+    gradient_checkpoint_policy: str = Field(
+        default="",
         description="Gradient checkpointing policy. If None, jax defaults to full recomputation and no checkpointing. "
         "Options: checkpoint_dots_with_no_batch_dims, dots_with_no_batch_dims_saveable, nothing_saveable",
     )
diff --git a/skyrl-tx/tx/tinker/engine.py b/skyrl-tx/tx/tinker/engine.py
index 2f809b7..8207bd9 100644
--- a/skyrl-tx/tx/tinker/engine.py
+++ b/skyrl-tx/tx/tinker/engine.py
@@ -178,8 +178,8 @@ class TinkerEngine:
         if self.config.gradient_checkpointing:
             policy = (
                 None
-                if self.config.gradient_checkpoint_policy is None
-                else getattr(ckpt_policies, self.config.gradient_checkpoint_policy)
+                if self.config.gradient_checkpoint_policy == ""
+                else self.config.gradient_checkpoint_policy
             )
             _model_forward = nnx.remat(_model_forward, policy=policy)
 
```
but I'll also think about a proper fix to support optional args in the command line parser :)

tyler-griggs: Ah good point. I suppose no longer a problem now that we've switched to a boolean flag, but good to think about

skyrl-tx/tests/tinker/test_engine.py
```diff
@@ -254,3 +253,53 @@ def apply_step(request_id: int, model_id: str, request: types.OptimStepInput) ->
...
+            target_ids,
+            loss_mask,
+            loss_fn_types,
+            sampling_logprobs,
+            advantages,
+        )
+        losses.append(float(loss_full))
+
+    # Check relative difference between losses is small
+    assert abs(losses[0] - losses[1]) / abs(losses[0]) < 5e-3
```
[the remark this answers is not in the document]
tyler-griggs: Not needed because we use a single policy.

skyrl-tx/tests/tinker/test_engine.py
```diff
@@ -254,3 +253,53 @@ def apply_step(request_id: int, model_id: str, request: types.OptimStepInput) ->
...
+        enforce_eager=False,
+        train_batch_size=2,
+        micro_batch_size=1,
+        max_lora_adapters=1,
+        max_lora_rank=4,
+    )
+
+    losses = []
+    for use_gradient_checkpointing in (False, True):
+        cfg = EngineConfig(**base_kwargs)
```
pcmoritz: I think it will be slightly cleaner / more typechecking to get rid of the dict and do
```python
cfg = EngineConfig(
    base_model="Qwen/Qwen3-0.6B",
    enforce_eager=False,
    train_batch_size=2,
    micro_batch_size=1,
    max_lora_adapters=1,
    max_lora_rank=4,
    gradient_checkpointing=use_gradient_checkpointing,
)
```

pcmoritz (approved):
Looks great, I pointed out one small nit :)

PR #539, opened 2025-10-21
[tx] Add generation profiling + Improve generation speed by jitting

Guido1Alessandro1Trevisan:
This PR improves generation speed substantially (note this is fp32 currently):

Before:
```
tests/models/test_qwen3_generate2.py Generation stats (50 tokens, 5 runs):
Mean time: 73036.76 ± 16852.31 ms
Min/Max: 64259.65 / 106729.98 ms
Tokens/sec: 7.52
```

After:
```
tests/models/test_qwen3_generate.py Generation stats (50 tokens, 5 runs):
Mean time: 6905.05 ± 534.00 ms
Min/Max: 6418.45 / 7929.51 ms
Tokens/sec: 79.51
```

PR #560, opened 2025-10-23
[tx] Batched sampler

Guido1Alessandro1Trevisan:
Implemented batched sampling. There’s still a lot of room for improvement in input shapes and the use of better JAX functions.

#534 is ready for review once it gets merged I will add the prompt log_probs to the batching as well.
#511 is also ready. fixes also a bug for the lora test when using moe.

Currently branched off of #539 but batching is still very slow but there is a lot of headroom. (will rebase when the pr is merged to main)

### DIFF
--- a/skyrl-tx/tx/tinker/engine.py
+++ b/skyrl-tx/tx/tinker/engine.py
@@ -589,12 +589,8 @@ class TinkerEngine:
 
         with jax.set_mesh(self.mesh):
             model = nnx.merge(self.graphdef, self.lora_params, self.non_lora_params)
-
             result = model.generate(
-                input_ids,
-                attention_mask,
-                sampling_params=all_sampling_params,
-                adapter_indices=all_adapter_indices,
+                input_ids, attention_mask, sampling_params=all_sampling_params, adapter_indices=all_adapter_indices
             )
 
         all_sequences = [


Guido1Alessandro1Trevisan:
Next Steps:
 - Checkpoint loading from disk when checkpoint_id is provided.
 if request_data.checkpoint_id:
    load_lora_checkpoint(self.model, adapter_index, checkpoint_path)

pcmoritz:
@Guido1Alessandro1Trevisan I unified the code a little bit with the other batching code. I think there are two things missing right now:

1. Making sure the right adapters are loaded before running the inference -- do you want to take a stab at this? I think we can do it by just extending `load_sampler_weights` to be able to load all adapters (if they are not already loaded) and return the full `adapter_indices` array. We can support inference from the base model by loading a LoRA adapter that is initialized with zeros.
2. We need to also batch the actual calls to model.generate, for which we need to pass all the sampling parameters as tensors. I can take a stab at that and also see if I can get https://github.com/NovaSky-AI/SkyRL/pull/541 merged as part of that :)

Guido1Alessandro1Trevisan:
@pcmoritz Okay sounds good i'll do that

Guido1Alessandro1Trevisan:
Increased batching efficiency requests are now grouped based on seed, max_tokens, and temperature. Also when the engine initializes the LoRA module with adapter index 0 is loaded even if no active LoRA weights are present.

Guido1Alessandro1Trevisan:
Weights are loaded from disk every time, even though they should be loaded from RAM if available.
Edit: added a simple check by updating the ModelMetadata (doesn't introduce incompatibility with tinker api).

skyrl-tx/tx/tinker/engine.py
```diff
@@ -497,6 +516,120 @@ def process_forward_backward_batch(
...
+        )
+        all_adapter_indices = jnp.array(all_adapter_indices, dtype=jnp.int32)
+
+        # Collect per-sample outputs
+        sequences_out = [None] * int(input_ids.shape[0])
+
+        with jax.set_mesh(self.mesh):
+            model = nnx.merge(self.graphdef, self.lora_params, self.non_lora_params)
+
+            # Build groups for batching depending on max_params, seed, and temperature:
```
pcmoritz: The nice thing with the new generate method https://github.com/NovaSky-AI/SkyRL/pull/567 is that we don't need to do this any more and we can just share the same generate method without having to group things. For memory reasons, we might still want to allow micro-batching like we do for the forward_backward method though, but we can implement that separately :)

skyrl-tx/tx/tinker/engine.py
```diff
@@ -110,6 +111,7 @@ def __init__(
 
             # Split model into LoRA and non-LoRA parameters
             self.graphdef, self.lora_params, self.non_lora_params = nnx.split(self.model, self.model.is_lora_param, ...)
+            update_adapter_config(self.model, adapter_index=0, lora_rank=1, lora_alpha=1.0)
```
pcmoritz, line 114: Yes, this is great and makes a lot of sense. There is a small variation of this that doesn't need to load the extra all-zeros lora adapter onto adapter_index 0 -- we can just use a non-existent adapter_index -1, and then in https://github.com/NovaSky-AI/SkyRL/blob/main/skyrl-tx/tx/layers/lora.py#L78 pass those activations through without putting them into the ragged_dot. It means we do not need to waste a lora adapter on all zeros, in OSS people typically operate close to the memory limit so it would be nice to be able to save that memory.

While thinking more about it, I noticed however that the LoRA expert code will also get slightly more complicated, so maybe something we can consider in the future. We might add a flag to the engine that turns off base model inference and then skips loading the additional adapter. E.g. for RL you typically don't need to sample from the base model anyways.

skyrl-tx/tx/tinker/engine.py
pcmoritz: (I updated this now)

Guido1Alessandro1Trevisan:
@pcmoritz Let me know when it's ready to merge. I'm excited to finally get the RL loop working properly. I’ll also work on some LoRA eviction logic to prevent memory overflow.

pcmoritz:
Sounds great, it is very close now I think :)

For the memory management, I think ideally we can tie it to the clients (each client is associated with on model_id I believe). If we can track when a client disconnects, we know when it is safe to remove the corresponding adapter and make it available to the next client :)

skyrl-tx/tx/tinker/engine.py
```diff
@@ -317,10 +347,30 @@ def find_batchable_forward_backward(self, session: Session) -> list[FutureDB]:
...
+        """
+        sample_query = (
+            select(FutureDB)
+            .where(FutureDB.request_type == types.RequestType.SAMPLE)
+            .where(FutureDB.status == RequestStatus.PENDING)
+            .order_by(FutureDB.request_id)
+        )
+        sample_ops = session.exec(sample_query).all()
+
+        return sample_ops
```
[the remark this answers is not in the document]
pcmoritz: This is ok, because for sampling, (if needed) we always load the right model weights before we evaluate the model.

pcmoritz (approved):
Nice, this is good to merge now I think :)

skyrl-tx/tx/tinker/engine.py
```diff
@@ -632,103 +740,46 @@ def process_save_weights_for_sampler(
...
+                if self.models[model_id].loaded_checkpoint_id == request_data.checkpoint_id:
+                    adapter_indices_list.append(adapter_index)
+                # Load model from disk
+                else:
+                    checkpoint_path = self.config.checkpoints_base / model_id / f"{request_data.checkpoint_id}.tar.gz"
+                    logger.info(f"Loading LoRA sampler checkpoint from {checkpoint_path}")
+                    load_lora_checkpoint(self.model, adapter_index, checkpoint_path)
+
+                    self.models[model_id].loaded_checkpoint_id = request_data.checkpoint_id
+                    logger.info(f"Loaded LoRA sampler weights for model {model_id} at adapter index {adapter_index}")
```
tyler-griggs, line 772: This may be a niche edge case, but I'm curious if we can hit a scenario where there are requests to sample from a single `model_id` at different `checkpoint_id`s? If this were to happen, we would batch these sample requests together but only one `checkpoint_id` would be loaded (namely, the one that comes last in the batch).

Probably not something we're worried about now, but wanted to call out that we're assuming only a single checkpoint ID for a given model can be loaded at a time.

tyler-griggs (approved):
Looks great!

skyrl-tx/tx/tinker/engine.py
pcmoritz: Yes you are absolutely right, this can happen and we need to either handle it correctly or at least give an error in that case (instead of wrong results). Let me think about how to best handle it. One possibility is to handle it in `find_batchable_sample` and only allow one checkpoint_id per model_id.

This can of course be made more efficient going forward -- if there is enough adapter slots, there is no reason why we couldn't load all checkpoints_ids at the same time. But for now I think that's a bit of a niche edge case, for now it will be enough to be correct in that case and leave some performance on the table.

Guido1Alessandro1Trevisan: I think we could have a single function to load checkpoints either with or without optimizers and gradients, depending on whether they’re needed for training or sampling.
The ModelMetadata should use an adapter dictionary instead of an adapter integer mapping checkpoint IDs to adapter indices.
Does that sound like a reasonable approach? Once this is merged i’d like to open a pr to implement it.

Ties well with the lora evicting logic

pcmoritz: Ok, so for this PR, I have now added some simple logic so only sampling requests are batched together that load only one checkpoing per model_id. I also added an assert in the `load_sampler_weights` method to make sure adapters are not overwritten (which should never happen with the added logic to find_batchable_sample). I think this should be ready to merge now.

Going forward, likely the right strategy will be to completely get rid of the model_id <-> adapter_index association, and just load checkpoints completely dynamically (unless it is already loaded, which can be tracked in a global adapter dictionary). Is this what you also mean Guido? Before we implement the new strategy, it will probably be best to do some benchmarking of the two approaches, and before that makes sense, we should probably fix other low hanging performance improvements that might impact such a benchmark :)

The reason why I think this is the right strategy is that (a) most MultiLoRA inference engines are built that way and (b) it removes any limitations that the presence of limited adapter slots might impose on number of supported clients. However, it might incur more LoRA params reloading, which means we have to be extra careful to make sure loading LoRA weights is fast :)

pcmoritz: This will also handle the adapter_index 0 (base model) situation more elegantly, because if the adapters are assigned dynamically, the base model will only use an adapter if somebody actually performs inference on the base model :)

Guido1Alessandro1Trevisan: Sounds good. I feel like there are still some things to fiddle with for the weight loading. When loading a checkpoint  if it’s for sampling, you only load the weights, if it’s for training you load both the weights and the optimizer. And if it’s for training again you’d first check whether the weights are already loaded and then load the optimizer on top of that.

pcmoritz:
Alright, time to merge this now :)

PR #528, opened 2025-10-20
[dependencies] separate vllm + megatron + bump vllm back to 0.11.0 + pin minimum uv version for extra-build-dependencies

erictang000:
## Separates vllm + megatron deps
After #481, there were some megatron flashinfer issues with --extra vllm. This PR separates out the version of vllm that megatron relies on from the general vllm version, allowing us to bump vllm to 0.11.0 for the rest of the training stack. 

## Update flash-attn installation
Updates flash-attn installation to use the `extra-build-dependencies` feature from uv, requiring us to use a uv version >= 0.8.10. This feature allows us to do the following, removing the need to deal with markers + extras to specify a url source for each set of extras.

```
[tool.uv.extra-build-dependencies]
flash-attn = [{requirement = "torch", match-runtime = true}]

[tool.uv.extra-build-variables]
flash-attn = { FLASH_ATTENTION_SKIP_CUDA_BUILD = "TRUE"}

[project.optional-dependencies]
vllm = [
    "vllm==0.11.0",
    "flash-attn==2.8.3",
...
]
mcore = [
     "flash-attn==2.7.4.post1"
...
]
```

PR #561, opened 2025-10-23
Add YaRN support for VLLM and HF

sergeypastukhov-ddog:


skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
```diff
@@ -143,6 +145,20 @@ def create_ray_wrapped_inference_engines(
...
+            if rope_scaling:
+                rope_engine_kwargs["rope_scaling"] = rope_scaling
+                if "max_model_len" not in engine_init_kwargs:
+                    rope_factor = rope_scaling.get("factor", None)
+                    rope_max_pos = rope_scaling.get("original_max_position_embeddings", None)
+                    assert rope_factor is not None, "Please provide rope scaling `factor` to compute model max length"
+                    assert (
+                        rope_max_pos is not None
+                    ), "Please provide rope `original_max_position_embeddings` to compute model max length"
+                    rope_engine_kwargs["max_model_len"] = int(rope_factor * rope_max_pos)
```
SumanthRH, line 158: QQ: Could you explain why setting `max_model_len` here was needed? 

Also, we should have an `else` block here that handles the case where `max_model_len` is provided by the user and `rope_scaling` is not None. Perhaps we raise an error if it's incompatible?

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -31,6 +31,11 @@ trainer:
           alpha: 16
           dropout: 0
           lora_sync_path: "/tmp/skyrl_lora_sync"
+      rope_scaling:
+        rope_type: yarn
+        factor: 4.0
+        original_max_position_embeddings: 32768
```
SumanthRH: Why are there two versions here? Could we just have one at the `trainer` level since it will be used for ref model and the policy model

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -166,6 +171,10 @@ trainer:
   logger: "wandb"
   dump_data_batch: false
   dump_eval_results: true
+  rope_scaling:
+    rope_type: yarn
+    factor: 4.0
+    original_max_position_embeddings: 32768
```
SumanthRH: `rope_theta` is not added here?

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -166,6 +171,10 @@ trainer:
   logger: "wandb"
   dump_data_batch: false
   dump_eval_results: true
+  rope_scaling:
```
SumanthRH: Could we add a comment that this supports all parameters in https://huggingface.co/docs/transformers/main/en/internal/rope_utils#transformers.RopeParameters? 

```suggestion
  rope_scaling: # supports all parameters in https://huggingface.co/docs/transformers/main/en/internal/rope_utils#transformers.RopeParameters
```

skyrl-train/skyrl_train/entrypoints/main_base.py
```diff
@@ -66,6 +66,11 @@ def create_ray_wrapped_inference_engines_from_config(cfg: DictConfig, colocate_p
         engine_kwargs["sleep_level"] = 1
         engine_kwargs["max_loras"] = 1
 
+    if (rope_scaling := cfg.trainer.policy.model.get("rope_scaling", None)) is not None:
+        engine_kwargs["rope_scaling"] = rope_scaling
+    if (rope_theta := cfg.trainer.policy.model.get("rope_theta", None)) is not None:
+        engine_kwargs["rope_theta"] = rope_theta
```
SumanthRH:
```suggestion
    if (rope_scaling := cfg.trainer.get("rope_scaling", None)) is not None:
        engine_kwargs["rope_scaling"] = rope_scaling
    if (rope_theta := cfg.trainer.get("rope_theta", None)) is not None:
        engine_kwargs["rope_theta"] = rope_theta
```

Let's just use the rope parameters at the trainer level?

skyrl-train/skyrl_train/config/ppo_base_config.yaml
SumanthRH: Looks like `rope_theta` is a part of `RopeParameters` already: https://huggingface.co/docs/transformers/main/en/internal/rope_utils#transformers.RopeParameters

wondering why vllm separates it out then 🤔

skyrl-train/skyrl_train/config/ppo_base_config.yaml
SumanthRH: Oh I see now that the `RopeParameters` class is only introduced in transformers 4.57.1, which we upgraded to only recently after #528

skyrl-train/skyrl_train/config/ppo_base_config.yaml
SumanthRH: Oh I see now that the RopeParameters class is only introduced in transformers 4.57.1, which we upgraded to only recently after https://github.com/NovaSky-AI/SkyRL/pull/528

SumanthRH: Ok I believe the upgrade on Huggingface's side is backwards compatible. So we can probably have these fields in the older format like above, i.e have `rope_scaling` and `rope_theta` separate, and we can move to the newer  `RopeParameters` after we move completely to transformers 4.57.1 and above (some paths like flashrl are still behind)

skyrl-train/skyrl_train/model_wrapper.py
```diff
@@ -61,6 +61,7 @@ def __init__(
         sequence_parallel_size=1,
         use_sample_packing: bool = False,
         use_torch_compile: bool = False,
+        rope_scaling: Dict[str, Any] = {},
```
SumanthRH, line 64: Could we add `rope_theta` as well here?

SumanthRH: Will also need to pass it through the FSDP and DeepSpeed Worker classes

skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
sergeypastukhov-ddog: `max_model_len` is a parameter to configure max model length in VLLM. By default it is taken from model config, and it'll be shorter than what we want using YaRN. See feature request example [here](https://github.com/vllm-project/vllm-ascend/issues/760):

`--rope-scaling '{"rope_type":"yarn","factor":4.0,"original_max_position_embeddings":32768}' --max-model-len 131072`

Regarding `else`, I thought about it, maybe in future there will be some other context increasing methods and then this check will be annoying. So if user wants to have some weird value in their config - let it be.

Also `max_model_len` supports k/m/g/K/M/G in human-readable format (like "128K") ([doc](https://docs.vllm.ai/en/v0.10.1/cli/serve.html?h=rope+scaling#-max-model-len)) and in order to correctly verify I would need to parse all of that somehow. Curious to hear your thoughts on this

skyrl-train/skyrl_train/config/ppo_base_config.yaml
sergeypastukhov-ddog: There may be a case when user wants to increase context length only for the inference, but not for the trainer. For example, this would allow to have more thinking tokens, without increasing memory footprint much (training on longer context needs more memory comparing to inference on longer context).

I could write structure like this if you prefer:
```
trainer:
  policy:
    model:
        rope_scaling: ${rope_scaling}
  rope_scaling: ${rope_scaling}

rope_scaling:
   ...
```

skyrl-train/skyrl_train/model_wrapper.py
sergeypastukhov-ddog: Ah missed it, one moment

sergeypastukhov-ddog: Done!

skyrl-train/skyrl_train/entrypoints/main_base.py
sergeypastukhov-ddog: Discussing this in another comments thread

skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
```diff
@@ -91,6 +84,8 @@ def create_ray_wrapped_inference_engines(
     max_lora_rank=64,
     max_loras=1,
     engine_init_kwargs: Dict[str, Any] = {},
+    rope_scaling: Dict[str, Any] = {},
+    rope_theta: int | None = None,
```
tyler-griggs: I believe `rope_theta` is a `float`. Could you update this throughout the code?

skyrl-train/skyrl_train/config/ppo_base_config.yaml
SumanthRH: I would probably not have this structure, I think having both of these in the `trainer` configuration is confusing - since `trainer.policy.model.rope_scaling` is only used to set the parameter for the inference engine (generator). 

Maybe we can make this a part of the `generator` config? 

```yaml
trainer:
    # rope parameters
    rope_scaling: ...
    rope_theta: ... 

generator: 
    # rope parameters, can be optionally different from the trainer , useful in some cases like with thinking models.
    rope_scaling: ${trainer.rope_scaling}
    rope_theta: ${trainer.rope_theta}
```

skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
sergeypastukhov-ddog: Done!

skyrl-train/skyrl_train/config/ppo_base_config.yaml
sergeypastukhov-ddog: Done! I have added some print statements for testing please ignore, I'll remove them after I confirm changes are ok on our side

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -167,6 +167,13 @@ trainer:
   dump_data_batch: false
   dump_eval_results: true
 
+  # YaRN:
+  rope_scaling: null
```
SumanthRH, line 171:
```suggestion
  rope_scaling: null
  rope_theta: null
```

sergeypastukhov-ddog: Done!

Issue #535, opened 2025-10-21
[tx] Make rl_loop.py example work


pcmoritz:
We are almost there with making the `rl_loop.py` example from the tinker cookbook work (all the features are there).

I think the remaining items are:

1. Make batching work for the sampling (otherwise it is very slow)
2. Find the right configurations for the database sessions, e.g. something like

```python
from sqlalchemy.pool import StaticPool
from sqlalchemy import text
```
```python
        self.db_engine = create_engine(
            f"sqlite:///{db_path}",
            echo=False,
            poolclass=StaticPool,  # Single persistent connection
            connect_args={
                "timeout": 30.0,  # Wait up to 30s for locks
                "check_same_thread": False,  # Allow connection sharing
            }
        )

        # Enable WAL mode for better concurrency
        with self.db_engine.connect() as conn:
            conn.execute(text("PRAGMA journal_mode=WAL"))
            conn.commit()
```

VThejas:
Hey @pcmoritz! I would like to pick this up, could you share some details on how I can proceed?

My current thought process:
- copy and run the rl_loop.py script and see what's missing
- I can investigate the batching of sampling calls which you mention above

Could you share more details on what the issue with the database configuration is?

pcmoritz:
Hey @VThejas thanks for your interest, I believe this issue is now already resolved. I created a new issue that might be a good issue to get started: https://github.com/NovaSky-AI/SkyRL/issues/646 -- I wonder if you would be interested in that. Otherwise, implementing any other missing functionality right now would be very much appreciated. I would encourage you to just try stuff out that you are interested in and if you see anything missing that you need, implement it :)

I'll close this issue for now, we can discuss more in https://github.com/NovaSky-AI/SkyRL/issues/402 if you want to generally talk about what is currently missing :)

PR #564, opened 2025-10-24
[tx] PostgreSQL Database support

atemaguer:
This PR implements the https://github.com/NovaSky-AI/SkyRL/issues/520 issue and introduces PostgreSQL database support. Now users can just define a TK_DATABASE_URL environment variable, and the engine will persist data to this remote Postgres database when skyrl-tx is deployed as a service.

### DIFF
--- a/skyrl-tx/tx/tinker/db_models.py
+++ b/skyrl-tx/tx/tinker/db_models.py
@@ -43,8 +43,7 @@ def get_async_database_url(db_url: str | None = None) -> str:
     Raises:
         ValueError: If the database scheme is not supported.
     """
-    url_str = get_database_url(db_url)
-    parsed_url = sqlalchemy_url.make_url(url_str)
+    parsed_url = sqlalchemy_url.make_url(get_database_url(db_url))
 
     match parsed_url.get_backend_name():
         case "sqlite":


atemaguer:
@pcmoritz, Any chance you know why the CPU tests are failing?

pcmoritz:
Hmm, not sure, but sometimes it just gets cancelled (doesn't seem to happen often), let's just re-run it :)

pcmoritz:
I can review and get this PR merged now, I wonder did you have a chance to try it out with Postgres / are there any other problems besides being able to configure the URL (e.g. around the queries). After we finish fixing https://github.com/NovaSky-AI/SkyRL/issues/535, I might go back to getting some docs set up, then we can also add some docs on how to run things with postgres :)

atemaguer:
Yes. I did test it with a Postgres database, and things worked fine. I've also added support for alembic in another branch, which I'll push once postgres support is merged.

pcmoritz:
Awesome, thank you so much, this is great!

skyrl-tx/tx/tinker/config.py
```diff
@@ -51,7 +57,15 @@ def add_model(parser: argparse.ArgumentParser, model: type[BaseModel]) -> None:
         else:
             # Add type if available
             if field.annotation is not None:
-                kwargs["type"] = field.annotation
+                # Handle Union types (e.g., str | None)
```
pcmoritz: This handling looks pretty brittle (especially going forward if we want to handle more complex types), I think it will be better to add something like `json_schema_extra={"argparse_type": str},` to the pydantic field, and then use the type from that (and fall back to field.annotation if that's not defined).

pcmoritz:
The PR looks great, I made a small suggestion to improve it (about the typing), if you have time to update it go for it, otherwise I'm happy to apply the fix after dinner and get it merged :)

pcmoritz:
I'm back now, working on the few small changes and getting it merged :)

pcmoritz (approved):
Very nice, thanks a lot for contributing this, let's merge it now :)

PR #569, opened 2025-10-24
[skyrl-train] Enforce eager by default

SumanthRH:
# What does this PR do?

Set `enforce_eager` to true by default. 

We've seen issues with cuda graphs affecting convergence on some training runs. This is likely drift due to train <> rollout logprobs mismatch, but we need to investigate this better. 

Until then, it is best to turn off cuda graphs by default. Training runs are stable by default, more performant if the user chooses so.


SumanthRH:
The only example scripts where we explicitly set `enforce_eager=false` are FlashRL scripts, which IMO is fine to stay as is due to TIS

PR #570, opened 2025-10-25
[skyrl-train][Examples] Support truncated importance sampling for `StepWiseGenerator`

SumanthRH:
# What does this PR do?

Adds support for truncated importance sampling (TIS ) with the step wise example. This means that TIS can now be used with multi-turn training

The implementation is mostly straightforward. One thing to note is that we append an EOS token in the agent loop when stop tokens are passed. This is a new token that was not generated by the model and thus logprobs are not available - thus this token is masked out.

### DIFF
--- a/format.sh
+++ b/format.sh
@@ -1,7 +1,7 @@
 set -e
 
 if command -v uv >/dev/null 2>&1; then
-    uv pip install --system -q pre-commit
+    uv pip install -q pre-commit
 else 
     pip install -q pre-commit
 fi


skyrl-train/skyrl_train/utils/utils.py
```diff
@@ -287,11 +287,6 @@ def validate_cfg(cfg: DictConfig):
 
         if cfg.generator.backend == "sglang":
             raise NotImplementedError("`trainer.algorithm.use_tis` doesn't support Sglang backend, please use vLLM")
-
```
SumanthRH, line 290: This validation is too restrictive so I have removed it. 

Now, we can use TIS with multi-turn generation in the step wise example

skyrl-train/examples/step_wise/step_wise_generator.py
```diff
@@ -61,6 +61,12 @@ def __init__(
         if not self.use_conversation_multi_turn:
             raise ValueError("`StepWiseGenerator` doesn't support `use_conversation_multi_turn=False`")
 
+    def _validate_cfg(self, generator_cfg: DictConfig):
+        if len(generator_cfg.chat_template_kwargs) and generator_cfg.batched:
+            raise ValueError(
+                "`chat_template_kwargs` is not compatible with `batched=True` since the chat templating is handled by the inference engine"
+            )
```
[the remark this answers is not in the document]
SumanthRH: It ok

skyrl-train/examples/step_wise/step_wise_generator.py
```diff
@@ -333,12 +343,16 @@ def _get_next_input_ids_with_multiturn_chat_template(
...
         Returns:
             input_ids: List[int]
             loss_mask: List[int]
         """
         input_ids += output_ids
-        loss_mask = [1] * len(output_ids)
+        # if `added_eos` is `True`, then  the EOS token was not generated and only added in the
+        # agent loop. For consistency with other entities like logprobs , we ignore it in the loss
+        # mask
+        loss_mask = [1] * len(output_ids) if not added_eos else [1] * (len(output_ids) - 1) + [0]
```
CharlieFRuan, line 359: Good point! Could you make this change to the base `skyrl_gym_generator.py` instead? So that we have the same behavior and only coded in one place.

skyrl-train/examples/step_wise/step_wise_generator.py
```diff
@@ -137,19 +140,24 @@ async def agent_loop(
             output = engine_output["responses"][0]
             output_ids = engine_output["response_ids"][0]
             stop_reason = engine_output["stop_reasons"][0]
+            response_logprobs = engine_output.get("response_logprobs", None)
+            if response_logprobs is not None:
+                response_logprobs = response_logprobs[0]
```
CharlieFRuan, line 145: Can we add an assert here `assert len(output_ids) == len(response_logprobs)`? Similarly for `generate_batched` 

I see we do `response_logprobs += [0] * (len(loss_mask) - len(response_logprobs))` later on here, and `generate_batched` does `sample_logprobs = logprobs[i][: len(response_ids)]`. I am worried that this hides potential mismatch of logprobs length and output ids length.

SumanthRH: Done! actually I added the assert at the very end to catch issues. Should be good

skyrl-train/examples/step_wise/step_wise_generator.py
```diff
@@ -162,11 +170,10 @@ async def agent_loop(
...
             # Follow multi-turn chat history format.
             input_ids, loss_mask = self._get_next_input_ids_with_multiturn_chat_template(
-                input_ids,
-                output_ids,
-                new_obs,
-                done,
+                input_ids, output_ids, new_obs, done, added_eos
             )
+            if response_logprobs is not None:
+                response_logprobs += [0] * (len(loss_mask) - len(response_logprobs))
```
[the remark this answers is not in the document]
SumanthRH: bruh that is loss mask  in teh other generator @gemini-code-assist

skyrl-train/examples/step_wise/step_wise_generator.py
CharlieFRuan: But if you add it to the very end, it is after `response_logprobs += [0] * (len(loss_mask) - len(response_logprobs))` right? Which hides potential mismatch before this. I was hoping to add a `len(output_ids) == len(response_logprobs)` right after parsing them. There can be possible mismatch due to re-tokenization

CharlieFRuan: But I guess it is trivial because we do TI/TO when we use logprobs. Though I'm worried this assumption could change in later PRs

SumanthRH: I think we can revisit this later - I don't think the assertion you mentioed should be here - it should be a unit test for the inference engine abstractions

PR #575, opened 2025-10-26
refactor(tx): introduce proper types for Qwen3Model and Qwen3ForCausalLM …

### LINKED ISSUE #521: [tx] Introduce proper types for the return values of Qwen3Model.__call__ and Qwen3ForCausalLM.__call__
Currently they return dicts and the keys of that dict are not well specified. We should instead return a proper type like huggingface does. Maybe a dataclass? It will need to be made compatible with Jax, see https://docs.jax.dev/en/latest/_autosummary/jax.tree_util.register_dataclass.html

gitlost-murali:
…return values

Replace dict return values with JAX-compatible dataclasses to improve type safety and code clarity.

- Add Qwen3ModelOutput dataclass with fields: last_hidden_state, kv_cache, hidden_states
- Add Qwen3CausalLMOutput dataclass with fields: logits, last_hidden_state, kv_cache, hidden_states
- Both dataclasses registered with jax.tree_util.register_dataclass for JAX compatibility
- Update Qwen3Model.__call__ to return Qwen3ModelOutput instead of dict
- Update Qwen3ForCausalLM.__call__ to return Qwen3CausalLMOutput instead of dict
- Update all call sites to use attribute access instead of dict key access
- Export new types in tx.models.__all__

Fixes #521

### DIFF
--- a/skyrl-tx/tx/models/outputs.py
+++ b/skyrl-tx/tx/models/outputs.py
@@ -1,7 +1,4 @@
-"""Model-specific output dataclasses.
-
-This module contains model-specific output types for various model architectures.
-"""
+"""Model output dataclasses."""
 
 from __future__ import annotations
 from dataclasses import dataclass


pcmoritz (approved):
❤️  Thanks a lot for the contribution @gitlost-murali ! I made some small fixes and renamed the output classes to have generic names (very likely once we support more models than qwen3, we can reuse them, and it is also more consistent with the fact that they live outside of the qwen3.py file) :)

PR #577, opened 2025-10-27
[tx] added prompt logprobs

Guido1Alessandro1Trevisan:
To not break jiting even if only one request has prompt_logprobs=True all logprobs are calculated.

### DIFF
--- a/skyrl-tx/tx/tinker/engine.py
+++ b/skyrl-tx/tx/tinker/engine.py
@@ -618,7 +618,7 @@ class TinkerEngine:
             return results
 
         # Computes prompt_logprobs for the whole batch if any request asked for them
-        needs_prompt_logprobs = any(request_data.prompt_logprobs for _, (_, request_data) in valid_requests.items())
+        needs_prompt_logprobs = any(request_data.prompt_logprobs for (_, request_data) in valid_requests.values())
 
         all_prompts = []
         all_sampling_params = []
--- a/skyrl-tx/tx/utils/generator.py
+++ b/skyrl-tx/tx/utils/generator.py
@@ -278,7 +278,7 @@ class GeneratorMixin:
             stop_reasons=["stop" if stop_pos_host[i, 0] >= 0 else "length" for i in range(batch_size)],
             logprobs=[all_logprobs_host[i][: end_positions_host[i]].tolist() for i in range(batch_size)],
             prompt_logprobs=(
-                [prompt_logprobs_host[i, : int(prompt_lengths_host[i]) - 1].tolist() for i in range(batch_size)]
+                [prompt_logprobs_host[i, : prompt_lengths_host[i] - 1].tolist() for i in range(batch_size)]
                 if prompt_logprobs
                 else None
             ),


skyrl-tx/tx/utils/generator.py
```diff
@@ -217,4 +231,9 @@ def scan_fn(carry, _):
             generated_ids=[generated_ids[i, prompt_length : end_positions[i]].tolist() for i in range(batch_size)],
             stop_reasons=["stop" if stop_pos[i, 0] >= 0 else "length" for i in range(batch_size)],
             logprobs=[all_logprobs[i, prompt_length : end_positions[i]].tolist() for i in range(batch_size)],
+            prompt_logprobs=(
+                [prompt_logprobs_array[i, : int(prompt_lengths[i]) - 1].tolist() for i in range(batch_size)]
+                if prompt_logprobs
+                else None
+            ),
```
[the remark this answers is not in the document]
Guido1Alessandro1Trevisan: @pcmoritz Just out of curiosity do you think this is an actual issue? Is host syncing really that bad here?

skyrl-tx/tx/tinker/engine.py
```diff
@@ -623,7 +626,7 @@ def process_sample_batch(
                 all_sampling_params.append(request_data.sampling_params)
                 all_adapter_indices.append(adapter_indices_batch[i])
 
-            request_batch_slices.append((request_id, model_id, request_start, len(all_prompts)))
+            request_batch_slices.append((future.request_id, model_id, request_start, len(all_prompts), request_data))
```
pcmoritz: This should be `request_id` instead of `future.request_id`

skyrl-tx/tx/tinker/types.py
```diff
@@ -172,6 +172,7 @@ class SampleInput(BaseModel):
     sampling_params: SamplingParams
     num_samples: int
     checkpoint_id: str
+    prompt_logprobs: bool = False
```
pcmoritz: It would be best to omit the default here -- we always initialize it and it is an internal type, so there is really no reason for the default, it can only lead to bugs going forward :)

skyrl-tx/tx/utils/generator.py
```diff
@@ -247,4 +258,9 @@ def generate(
             generated_ids=[generated_ids_host[i][: end_positions_host[i]].tolist() for i in range(batch_size)],
             stop_reasons=["stop" if stop_pos_host[i, 0] >= 0 else "length" for i in range(batch_size)],
             logprobs=[all_logprobs_host[i][: end_positions_host[i]].tolist() for i in range(batch_size)],
+            prompt_logprobs=(
+                [prompt_logprobs_array[i, : int(prompt_lengths[i]) - 1].tolist() for i in range(batch_size)]
```
pcmoritz: I think currently there is some stuff missing, you need to call `compute_prompt_logprobs` to fill the prompt part of `all_logprobs`, and then remove the slicing in the `device_get` and return them :)

tyler-griggs: I forgot to respond to this, but I indeed has some non-pushed code. Thanks!

tyler-griggs:
I made some small clean-up modifications, and added a couple tests. Should be ready to review!

skyrl-tx/tx/utils/generator.py
```diff
@@ -117,6 +118,15 @@ def next_token_and_logprobs(s: DecodeState) -> tuple[jax.Array, jax.Array, jax.A
     return next_rngs, next_token, all_logprobs, stop_pos
 
 
+def compute_prompt_logprobs(prefill_logits: jax.Array, input_ids: jax.Array) -> jax.Array:
+    """Compute log probabilities of prompt tokens from prefill logits"""
+    logits_for_prompt = prefill_logits[:, :-1, :]
+    log_probs = jax.nn.log_softmax(logits_for_prompt, axis=-1)
```
pcmoritz, line 125: For future reference, I think we can optimize this going forward (from Gemini) -- for now it is probably best to keep it simple as it is:

Memory Optimization in compute_prompt_logprobs
In skyrl-tx/tx/utils/generator.py, the compute_prompt_logprobs function uses jax.nn.log_softmax.

Issue: log_softmax materializes a tensor of shape (Batch, Seq, Vocab). For a vocabulary of 128k and a sequence of 1024, this single operation allocates ~500MB of temporary memory per batch item, which can cause OOMs on smaller GPUs.

Simplification (Optimization): Since you only need the logprobs for specific target tokens, you can calculate logits[target] - logsumexp(logits). This avoids allocating the full Vocab dimension for the probability tensor.

```Python
def compute_prompt_logprobs(prefill_logits: jax.Array, input_ids: jax.Array) -> jax.Array:
    """Compute log probabilities of prompt tokens efficiently."""
    logits_for_prompt = prefill_logits[:, :-1, :]
    prompt_tokens = input_ids[:, 1:]
    
    # 1. Gather logits for the specific target tokens (Shape: Batch, Seq)
    target_logits = jnp.take_along_axis(
        logits_for_prompt, 
        prompt_tokens[..., None], 
        axis=-1
    ).squeeze(-1)
    
    # 2. Compute normalizer (Shape: Batch, Seq)
    # This avoids creating the massive (Batch, Seq, Vocab) log_softmax tensor
    normalizers = jax.nn.logsumexp(logits_for_prompt, axis=-1)
    
    return target_logits - normalizers
```

skyrl-tx/tx/tinker/engine.py
```diff
@@ -617,6 +617,9 @@ def process_sample_batch(
         if not valid_requests:
             return results
 
+        # Computes prompt_logprobs for the whole batch if any request asked for them
+        needs_prompt_logprobs = any(request_data.prompt_logprobs for (_, request_data) in valid_requests.values())
```
pcmoritz, line 621: Probably this should be done per micro-batch going forward to avoid unnecessary computations, and I was also a little confused why it was
```
                if needs_prompt_logprobs and result.prompt_logprobs:
                    all_prompt_logprobs.extend(result.prompt_logprobs[:batch_size])
```
below and not (unconditionally)
```
all_prompt_logprobs.extend(result.prompt_logprobs[:batch_size] if result.prompt_logprobs else [None] * batch_size)
```
but it makes sense since `needs_prompt_logprobs` applies for the whole batch :)

pcmoritz: For now totally fine to keep it simple as it is :)

skyrl-tx/tx/utils/generator.py
tyler-griggs: Good point, I left a todo in code for this.

skyrl-tx/tx/tinker/engine.py
tyler-griggs: Ah good call, I will add per-micro-batch in a follow-up PR.

PR #584, opened 2025-10-27
[tx] Make token weights optional

pcmoritz:
If the token weights are not provided by the user, they will be filled in to all be 1.0. Also fixes the predicted tokens and token weights in the test (doesn't impact the test, but it is good to fix it).

This is needed for https://github.com/NovaSky-AI/SkyRL/issues/535

### DIFF
--- a/skyrl-tx/tests/utils/test_generator.py
+++ b/skyrl-tx/tests/utils/test_generator.py
@@ -1,0 +1,89 @@
+from types import SimpleNamespace
+
+import jax.numpy as jnp
+from tx.tinker.types import SamplingParams
+from tx.utils.generator import GenerateOutput, GeneratorMixin, KVCache
+
+
+class DummyModel(GeneratorMixin):
+    def __init__(self, vocab_size: int = 16):
+        self.vocab_size = vocab_size
+
+    def __call__(self, input_ids, attention_mask=None, positions=None, kv_cache=None, adapter_indices=None):
+        """Simple dummy model for testing generator behavior."""
+        batch_size, seq_len = input_ids.shape
+        base = jnp.arange(self.vocab_size, dtype=jnp.float32)
+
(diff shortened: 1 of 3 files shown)


skyrl-tx/tests/tinker/test_api.py
```diff
@@ -64,40 +83,12 @@ def test_training_workflow(service_client):
...
-
-        # Target tokens are shifted by 1
-        target_tokens = all_tokens[1:] + [tokenizer.eos_token_id]
-
-        # Create Datum
-        datum = types.Datum(
-            model_input=types.ModelInput.from_ints(all_tokens[:-1]),
-            loss_fn_inputs={
-                "weights": weights[:-1],
-                "target_tokens": target_tokens[:-1],
```
pcmoritz, line 96: This was actually wrong before, since the eos token is removed by the slicing

skyrl-tx/tests/tinker/test_api.py
```diff
@@ -64,40 +83,12 @@ def test_training_workflow(service_client):
...
-            weights = [0.0] * len(prompt_tokens) + [1.0] * len(completion_tokens)
-
-        # Target tokens are shifted by 1
-        target_tokens = all_tokens[1:] + [tokenizer.eos_token_id]
-
-        # Create Datum
-        datum = types.Datum(
-            model_input=types.ModelInput.from_ints(all_tokens[:-1]),
-            loss_fn_inputs={
-                "weights": weights[:-1],
```
pcmoritz, line 95: This was wrong before, consider e.g. `a=1` where `1` is the target, then `weights = (0, 0, 1)`, but we actually want `weights = (0, 1, 1)` weight 0 for the matching of `a` to `=`, weight 1 for the matching of `=` to `1` and weight 1 for the matching of `1` to `<eos>`.

PR #585, opened 2025-10-27
[skyrl-train] Add example for on-policy distillation

erictang000:
# Overview
Adds an example of extending the `RayPPOTrainer` to do on policy distillation by providing a custom `apply_reward_kl_penalty` function, a pass through advantage estimator, and using the `importance_sampling` loss function as detailed in the [thinky blog](https://tinker-docs.thinkingmachines.ai/losses#policy-gradient-importance_sampling).

### distilling rl trained qwen3-4b-base (dapo recipe) back into qwen3-4b-base
image

### DIFF
--- a/skyrl-train/examples/on_policy_distillation/README.md
+++ b/skyrl-train/examples/on_policy_distillation/README.md
@@ -6,19 +6,19 @@ This folder contains scripts for running On-Policy Distillation on SkyRL.
   <img src="https://github.com/user-attachments/assets/255c46bd-f443-4f36-b8fc-9546c3a01cf1" width="45%" />
 </p>
 
-On-Policy Distillation is a technique recently highlighted by [Thinking Machines](https://thinkingmachines.ai/blog/on-policy-distillation), which combines the benefits of on-policy RL style training, with the dense reward signal of distillation. The main idea is to collect on-policy samples from a student model, use a teacher to grade each token from the student samples, update the student policy accordingly, and repeat. On-Policy distillation has previously been shown to be a more compute efficient approach than RL for efficiently post-training models ([Agarwal et al](https://arxiv.org/abs/2306.13649), [Gu et al](https://arxiv.org/abs/2306.08543), [Qwen3 team](https://arxiv.org/abs/2505.09388).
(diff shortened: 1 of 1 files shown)


skyrl-train/retry.sh
SumanthRH, line 1: Remove?

skyrl-train/examples/on_policy_distillation/run_on_policy_distill_math.sh
```diff
@@ -0,0 +1,78 @@
...
+  trainer.eval_interval=5 \
+  trainer.update_epochs_per_batch=1 \
+  trainer.train_batch_size=1024 \
+  trainer.policy_mini_batch_size=256 \
+  trainer.micro_forward_batch_size_per_gpu=2 \
+  trainer.micro_train_batch_size_per_gpu=1 \
+  trainer.ckpt_interval=10 \
+  trainer.max_prompt_length=2048 \
+  generator.sampling_params.max_generate_length=8192 \
+  trainer.policy.optimizer_config.lr=1.0e-6 \
```
SumanthRH: Thinky loss function uses a plain`.sum()` and doesn't normalize by sequence length. We normalize by max sequence length. Should the LR change? https://tinker-docs.thinkingmachines.ai/losses#policy-gradient-importance_sampling .

skyrl-train/retry.sh
erictang000: yes will do

skyrl-train/examples/on_policy_distillation/run_on_policy_distill_math.sh
erictang000: hmm yeah you're right i guess it makes sense to not have to normalize by sequence length for per token rewards as well. So the code should be:

```python
torch.mean(torch.sum(loss * loss_mask, dim=-1))
```

erictang000: couldn't find details on learning rate in their post, I assume they kept it constant relative to the pure RL runs

erictang000: jk: 5e-5 was used: https://github.com/thinking-machines-lab/tinker-cookbook/tree/main/tinker_cookbook/recipes/distillation#on-policy-distillation

SumanthRH: Resolved offline. Seems like they implicitly normalize by max tokens somewhere because a plain sum doesn't make much sense with this LR

skyrl-train/examples/on_policy_distillation/run_on_policy_distill_math.sh
```diff
@@ -0,0 +1,103 @@
+set -x
+
+# Running on policy distillation for Math on GSM8K
+# Uses Qwen-3-1.7B-Base as the student model and an RL trained Qwen-3-4B as the teacher model
```
tyler-griggs: I believe this example is not using 1.7B, right?

PR #591, opened 2025-10-29
[tx] Got the RL Loop working!

Guido1Alessandro1Trevisan:
I got the rl_loop.py loop working from the tinker cookbook!

Main fix:
The checkpoint_type field should also be a primary key, since the loop generates checkpoints with the same ID for both the sampler and the training runs.

In the API, weights are optional and are automatically padded if they are not provided before being passed to the model.


Guido1Alessandro1Trevisan:
noticed that the weights issue was already dealt with! What would be best a fallback logic in case the checkpoint already or three primary keys? I went with the latter

pcmoritz (approved):
Awesome, thank a lot for fixing this! I updated the checkpoint type for archive to be "sampler", since you can only create checkpoint archives for samplers (this is already covered by the test_api.py test fortunately).

PR #594, opened 2025-10-29
[tx] Support micro batching sample requests 

tyler-griggs:
To support arbitrarily large batches in the sample requests, we add support for specifying a max batch size supported by the engine based on memory limits. 

This is currently implemented as sequential "micro" batches, but will be updated to continuous batching.

### DIFF
--- a/skyrl-tx/tx/tinker/config.py
+++ b/skyrl-tx/tx/tinker/config.py
@@ -24,11 +24,11 @@ class EngineConfig(BaseModel):
     tensor_parallel_size: int = Field(default=1, description="Tensor parallelism degree to use for the model")
     train_micro_batch_size: int = Field(
         default=0,
-        description="Micro-batch size in sequences for gradient accumulation; 0 means disabled (use full batch)",
+        description="Micro-batch size (measured in number of sequences) for gradient accumulation; 0 means disabled (use full batch)",
     )
     sample_max_num_sequences: int = Field(
         default=0,
-        description="Maximum batch size in sequences for sampling/generation; 0 means disabled (use full batch)",
+        description="Maximum batch size (measured in number of sequences) for sampling/generation; 0 means disabled (use full batch)",
     )
     enforce_eager: bool = Field(default=False, description="Disable JAX JIT compilation")
     shard_attention_heads: bool = Field(


skyrl-tx/tx/tinker/config.py
```diff
@@ -26,6 +26,10 @@ class EngineConfig(BaseModel):
         default=0,
         description="Micro-batch size for gradient accumulation; 0 means disabled (use full batch)",
     )
+    sample_micro_batch_size: int = Field(
```
pcmoritz: This is good, would it also make sense to rename the current `micro_batch_size` to `train_micro_batch_size` so it is very clear which micro batch size is which?

skyrl-tx/tx/tinker/engine.py
```diff
@@ -579,25 +579,41 @@ def process_sample_batch(
...
+
+                result = model.generate(
+                    input_ids,
+                    attention_mask,
+                    sampling_params=sampling_params,
+                    adapter_indices=adapter_indices,
+                )
+                all_sequences.extend(
+                    types.GeneratedSequence(stop_reason=sr, tokens=toks, logprobs=lps)
+                    for sr, toks, lps in zip(result.stop_reasons, result.generated_ids, result.logprobs)
```
pcmoritz: Let's be slightly less cryptic here and keep `stop_reason, tokens, logprob`?

skyrl-tx/tests/tinker/test_engine.py
```diff
@@ -301,3 +301,65 @@ def test_gradient_checkpointing():
...
+        return types.SampleInput(
+            base_model=BASE_MODEL,  # Sample from base model (no LoRA)
+            prompt=types.ModelInput(chunks=[types.ModelInputChunk(tokens=tokens)]),
+            sampling_params=sampling_params,
+            num_samples=1,
+            checkpoint_id="",  # Empty for base model sampling
+        )
+
+    # Build a batch of 5 sample requests
+    reqs = [(FutureStub(rid), "", make_sample_input(tokens)) for rid, tokens in enumerate(prompts, start=1)]
```
pcmoritz: how about we use `request_id` instead of rid, since everywhere we call it `request_id` :)

skyrl-tx/tests/tinker/test_engine.py
```diff
@@ -301,3 +301,65 @@ def test_gradient_checkpointing():
...
+
+    # Process sample requests.
+    results = engine.process_sample_batch(reqs)
+
+    # Verify results
+    assert len(results) == len(prompts), f"Expected {len(prompts)} results, got {len(results)}"
+    for rid in range(1, len(prompts) + 1):
```
pcmoritz: Same as above

skyrl-tx/tests/tinker/test_engine.py
pcmoritz: Also, curious why you have it start at 1. It also makes the loop below slightly awkward, maybe just let it start at 0?

skyrl-tx/tests/tinker/test_engine.py
```diff
@@ -301,3 +301,65 @@ def test_gradient_checkpointing():
...
+        result = results[rid]
+
+        assert len(result.sequences) == 1, f"Request {rid}: expected 1 sequence, got {len(result.sequences)}"
+        seq = result.sequences[0]
+        tokens = seq.tokens if isinstance(seq.tokens, list) else seq.tokens.tolist()
```
pcmoritz: I think seq.tokens is always a list, right? So we can get rid of this

skyrl-tx/tx/tinker/engine.py
tyler-griggs: Yes, I need to keep claude on a shorter leash with variable naming :D

skyrl-tx/tests/tinker/test_engine.py
tyler-griggs: Ah I hit a weird error that I thought was related to request_id==0 being invalid, but this was not the root cause and forgot to revert this change. Updated!

skyrl-tx/tx/tinker/config.py
tyler-griggs: Naming updated after offline chat!

skyrl-tx/tx/tinker/config.py
```diff
@@ -22,9 +22,13 @@ class EngineConfig(BaseModel):
     max_lora_adapters: int = Field(default=32, description="Maximum number of LoRA adapters")
     max_lora_rank: int = Field(default=32, description="Maximum LoRA rank")
     tensor_parallel_size: int = Field(default=1, description="Tensor parallelism degree to use for the model")
-    micro_batch_size: int = Field(
+    train_micro_batch_size: int = Field(
         default=0,
-        description="Micro-batch size for gradient accumulation; 0 means disabled (use full batch)",
+        description="Micro-batch size in sequences for gradient accumulation; 0 means disabled (use full batch)",
```
pcmoritz: to make it clear, maybe `(measured in number of sequences)` instead of `in sequences`? Same comment for the other doc string below :)

Issue #598, opened 2025-10-30
[tx] Connecting to external Postgres Database bug


atemaguer:
The current external database support implementation contains a bug where the database URL password is being masked, leading to authentication problems when connecting to the database.

pcmoritz:
This should now be fixed through https://github.com/NovaSky-AI/SkyRL/pull/600 :)

Issue #439, opened 2025-10-09
[tx] Standardize the configuration options


pcmoritz:
We are starting to have configuration options like the checkpoints base directory (and also things like max lora rank etc). It would be great to have a standardized way to configure these so in the future people will be adding them in the same way. Some options are:

- Using environment variables
- Using a configuration file (e.g. toml which is part of the python standard library now)
- Store the configs in a table in the database

It is worth thinking about this and seeing how other projects of similar nature handle this.

pcmoritz:
A little bit of progress on this has been made in https://github.com/NovaSky-AI/SkyRL/pull/464, but there is still lots of room for improvement here :)

Guido1Alessandro1Trevisan:
I would use environment variables. vllm uses them and they’re convenient when working with containers like docker. Let me know though what you think and what configs should be implemented.

PR #600, opened 2025-10-30
[tx] Postgres Database Connection Bug Fix

atemaguer:
This PR fixes this issue: #598

### DIFF
--- a/skyrl-tx/tx/tinker/db_models.py
+++ b/skyrl-tx/tx/tinker/db_models.py
@@ -25,14 +25,16 @@ def get_async_database_url(db_url: str) -> str:
 
     match parsed_url.get_backend_name():
         case "sqlite":
-            return parsed_url.set(drivername="sqlite+aiosqlite").render_as_string(hide_password=False)
+            async_url = parsed_url.set(drivername="sqlite+aiosqlite")
         case "postgresql":
-            return parsed_url.set(drivername="postgresql+asyncpg").render_as_string(hide_password=False)
+            async_url = parsed_url.set(drivername="postgresql+asyncpg")
         case _ if "+" in parsed_url.drivername:
             # Already has an async driver specified, keep it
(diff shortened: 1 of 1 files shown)


atemaguer:
@pcmoritz, It seems I hadn't extensively tested the implementation that introduced Postgres support. This PR fixes a bug in that implementation. Now  external database support with Postgres should work correctly

skyrl-tx/tx/tinker/api.py
```diff
@@ -37,13 +37,17 @@ async def lifespan(app: FastAPI):
...
+        echo=False,
+    )
 
     async with app.state.db_engine.begin() as conn:
         await conn.run_sync(SQLModel.metadata.create_all)
 
     # Build subprocess command with engine config parameters
-    cmd = ["uv", "run", "--extra", "tinker", "-m", "tx.tinker.engine"]
+    # Use --env-file to ensure the engine subprocess loads .env file
+    cmd = ["uv", "run", "--env-file", ".env", "--extra", "tinker", "-m", "tx.tinker.engine"]
```
pcmoritz: I've been thinking about this a little bit more, it would be good not to hard code an explicit .env. Instead I'm thinking about this:

Introduce a new `json_schema_extra` option in EngineConfig, named "env_var", and if it is set, the order of initialization will be `explicitly specified arg` > `take value from env variable` > `use default`, this way the db config will just be passed from the api pod to the engine via the config mechanism, and the user can specify the .env file when they start the api server. It will also unify the config handling and make it consistent for all the options.

This also makes some progress on https://github.com/NovaSky-AI/SkyRL/issues/439 and will allow us to easily support more environment variables going forward (and have a standardized handling for them)

Let me know what you think! I can make a quick PR for this and then port your PR over to the new scheme :)

pcmoritz (approved):
Thanks a lot for fixing this!

Issue #602, opened 2025-10-30
[tx] API returns 422 errors on /api/v1/asample endpoint due to mismatched request schema


atemaguer:
The /api/v1/asample endpoint is rejecting all requests from the Tinker SDK with 422 Unprocessable Entity errors due to two schema mismatches:

Missing type discriminator: The API expected ModelInputChunk as {"tokens": [...]} but the SDK sends {"tokens": [...], "type": "encoded_text"}
Stop sequences type mismatch: The API only accepted integer token IDs in stop, but the SDK sends string stop sequences like ["\n\nUser:"]

To reproduce this issue, run the `tinker_cookbook.recipes.rl_basic`example.

atemaguer:
@pcmoritz, @Guido1Alessandro1Trevisan. Let me know if you can reproduce this issue and if so, I can work on a fix and push it

Guido1Alessandro1Trevisan:
I’ll be going on a pilgrimage to learn JAX properly so I’ll be inactive for a while.

pcmoritz:
@atemaguer Are you running the example with any special arguments? If I run it with `uv run rl_basic.py  base_url=http://localhost:8000 lora_rank=1`, it seems to be working for me, I tried it on commit `406e6fe79227d71c86b94fca134d64c509688c45` of the cookbook :) I would love to repro this!

Maybe you used a different dataset than the default?

atemaguer:
So, yeah. I've just been able to reproduce this issue again at the commit hash you provided above as well as using the same command. The backend logs look like this:

> INFO:     127.0.0.1:58284 - "POST /api/v1/asample HTTP/1.1" 422 Unprocessable Entity
INFO:     127.0.0.1:58285 - "POST /api/v1/asample HTTP/1.1" 422 Unprocessable Entity
INFO:     127.0.0.1:58286 - "POST /api/v1/asample HTTP/1.1" 422 Unprocessable Entity
INFO:     127.0.0.1:58287 - "POST /api/v1/asample HTTP/1.1" 422 Unprocessable Entity
INFO:     127.0.0.1:58288 - "POST /api/v1/asample HTTP/1.1" 422 Unprocessable Entity
INFO:     127.0.0.1:58005 - "POST /api/v1/asample HTTP/1.1" 422 Unprocessable Entity

While the client logs look like this:

[...]

PR #613, opened 2025-11-01
[tx] Fix asample 422 validation errors

atemaguer:
This PR fixes an issue where the client was getting back 422 validation errors after hitting the sampling endpoint. This happened in cases where the client defined a stop token as part of the sampling parameters. The related issue is #602

### DIFF
--- a/skyrl-tx/tx/tinker/api.py
+++ b/skyrl-tx/tx/tinker/api.py
@@ -1,7 +1,6 @@
 import fastapi
 from fastapi import FastAPI, HTTPException, Depends, Request
-from fastapi.responses import StreamingResponse, RedirectResponse, JSONResponse
-from fastapi.exceptions import RequestValidationError
+from fastapi.responses import StreamingResponse, RedirectResponse
 from pydantic import BaseModel, Field, model_validator
 from typing import Literal, Any, AsyncGenerator, Sequence, Annotated, Union
 from datetime import datetime, timedelta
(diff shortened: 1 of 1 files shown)


pcmoritz:
Thanks a lot for creating the PR! I think the stop string sequences actually need to be implemented differently, we cannot just convert the stop strings to tokens (since the concrete tokenization depends on what comes before in the sequence). The right way to do it is to detokenize the sequence when we are decoding it (in the engine), and then keep comparing if the stop strings are part of the decoded sequence. E.g. in vllm this is how it is done: https://github.com/vllm-project/vllm/blob/c2ed069b32e2805c05a858c6157f4c6393b145a8/vllm/v1/engine/detokenizer.py#L65

In order to concretely implement this in our engine, we might need to run the scan jax loop for the decoding for a number of tokens (e.g. 32 or 64 tokens), and then give control back to python and check for the stop tokens. Or maybe we can just run the incremental detokenization inside the scan, it is worth trying since that would be simpler.

How about we just merge the other part of this PR and make a separate PR to implement the stop string sequence?

atemaguer:
Awesome. This is great feedback. Let me work on an update.

atemaguer:
Also, I'm happy with whichever. I can just push an update to this PR, or have this one merged and then create a new branch that incorporates the updates. Let me know whichever path you decide

pcmoritz:
Why don't you remove the stop token handling from this PR and then we merge it, and you implement the stop token handling in a separate PR. Also can you be a little more specific in which situation the encoded_text / image_asset_pointer type was passed in (see my comment in https://github.com/NovaSky-AI/SkyRL/issues/602#issuecomment-3476675115), I'd love to repro it and add a test (if adding a simple test for this to `test_sample` is easy for you to do that would also be nice).

atemaguer:
Check this out comment for an update on this: [#602 (comment)](https://github.com/NovaSky-AI/SkyRL/issues/602#issuecomment-3476839584)

PR #633, opened 2025-11-07
[skyrl-train] Upgrade Ray to 2.51.1

### LINKED ISSUE #632: Upgrade Ray version to 2.50.1

SumanthRH:
# What does this PR do?

Upgrades Ray to 2.51. Closes #632 

Note that this is a hard upgrade. This is because SkyRL is changing so fast that the docker image only packages the dependencies and not a versioned package of the repo. Further, the typical workflow is that users add to the SkyRL repo to add their own example, etc. 

SkyRL can still be run with ray < 2.5.1, and that section of the docs is retained as is. 

We will improve this dependency story in the future.

TODO:
- [x] Publish new docker image
- [x] Update docs

### DIFF
--- a/skyrl-train/docs/getting-started/installation.rst
+++ b/skyrl-train/docs/getting-started/installation.rst
@@ -43,9 +43,9 @@ That is it! You should now be able to run our :doc:`quick start example <quickst
 
 .. warning::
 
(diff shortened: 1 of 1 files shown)


SumanthRH:
Running GPU CI for confirmation: https://github.com/NovaSky-AI/SkyRL/actions/runs/19445629619

SumanthRH:
Test passed successfully (CI workflow is randomly marking it as failed because there were concurrent runs)

PR #634, opened 2025-11-07
[tx] Applyt jax.jit to prefill during generation and bin sequence lengths

pcmoritz:
This PR introduces a number of optimizations:

- Jitting prefill for decoding
- Bin sequence lengths during decoding
- Pad batch size during decoding
- Refactor scan, so the scan function doesn't change for different calls to .generate (before, it closed over local variables and the closure is a new variable in each call to .generate), so jitted functions get cached across function invocations

These optimizations bring each RL step of rl_loop down from 400s to about 150s.

### DIFF
--- a/skyrl-tx/tests/utils/test_generator.py
+++ b/skyrl-tx/tests/utils/test_generator.py
@@ -1,11 +1,11 @@
-from types import SimpleNamespace
-
+from flax import nnx
 import jax.numpy as jnp
+from tx.models.outputs import CausalLMOutput
 from tx.tinker.types import SamplingParams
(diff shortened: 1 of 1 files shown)


.github/workflows/cpu_skyrl_tx.yaml
```diff
@@ -47,7 +47,7 @@ jobs:
     #     uv run --extra tinker --extra dev ty check
     - name: Run pytest
       run: |
-        uv run --extra tinker --extra dev pytest -s tests
+        uv run --extra tinker --extra dev pytest --forked -s tests
```
pcmoritz, line 50: Added --forked so each test gets run in a separate process, this reduces the memory usage

Issue #629, opened 2025-11-05
[tx] Improve model configuration for LoRA parameters


pcmoritz:
Currently we have statements like
```
        max_lora_adapters = getattr(config, "max_lora_adapters", 0)
        max_lora_rank = getattr(config, "max_lora_rank", 8)
```
everywhere in the code base. We should get rid of them and handle this better. One way to do it would be to have our own subclass of `Qwen3Config` and just add these parameters as required attributes.

Maybe there is a better way too :)

pcmoritz:
This is now fixed by https://github.com/NovaSky-AI/SkyRL/pull/636

PR #636, opened 2025-11-07
[tx] Refactor LoRA config to eliminate getattr duplication

hezyin:
## Summary

This PR addresses #629 by replacing repetitive `getattr()` calls with a proper configuration class that extends `Qwen3Config`.

## Changes

- Add `tx/models/configs.py` with `Qwen3Config` class
- Update all model classes to use `Qwen3Config` type hints
- Replace `getattr()` calls with direct attribute access
- Add tests for new config class
- Update existing tests to use new config

## Benefits

- Type-safe access to LoRA parameters
- Single source of truth for default values
- Eliminates code duplication

### DIFF
--- a/skyrl-tx/tests/utils/test_models.py
+++ b/skyrl-tx/tests/utils/test_models.py
@@ -19,10 +19,7 @@ from tx.utils.storage import download_and_unpack
 
 
 def create_test_model(rank: int, alpha: int, adapter_index: int):
-    """Create a small Qwen3 model for testing with LoRA enabled.
-
-    Returns both the tx Qwen3Config and the base HuggingFace config for testing.
-    """
+    """Create a small Qwen3 model for testing with LoRA enabled."""
     base_config = AutoConfig.from_pretrained("Qwen/Qwen3-0.6B")
     # Make it smaller for testing
     base_config.num_hidden_layers = 1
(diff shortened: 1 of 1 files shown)


pcmoritz:
Awesome, thanks a lot for making the PR, it looks great!

Can you rename Qwen3ConfigWithLoRA to just Qwen3Config? The reasons are, first of all, the added configs are not necessarily related to LoRA (like shard_attention_heads), and second, wherever possible we try to treat LoRA natively as part of the models (you might have seen this -- there is no "with lora" / "without lora" model, there is only one). I don't think there will be clashes with the huggingface Qwen3Config , in fact going forward we may never actually import the huggingface Qwen3Config, and just import PretrainedConfig from huggingface. Meanwhile if we do and need both, we can import it as `from transformers import Qwen3Config as HfQwen3Config` :)

Also there are a few small linting things from the CI, can you fix them as well?

There are a few smaller refactors I would like to try out, I'll just apply them to the PR directly after you made the above changes (I don't know if they will be good yet) :)

hezyin:
@pcmoritz Thanks, Philipp. All makes sense, I’ll make those changes

hezyin:
@pcmoritz PR updated with defaults removed.

PR #640, opened 2025-11-07
Add a configuration for max tokens per microbatch

justinvyu:
## Summary

Introduces token-based binpacking for microbatches, enforcing a `max_tokens_per_microbatch` limit on the total number of real tokens in a microbatch.

Replaces the even sample-based chunking `BatchIterator` with a `BalancedBatchIterator` that respects the microbatch token limits.

## Problem
Previously, microbatches created by chunking sequences evenly so that each microbatch was a certain batch size (`micro_train_batch_size_per_gpu`), which could lead to:
* Sequence packed microbatches can exceed GPU memory limits when sequences vary in length, so you need to conservatively set the micro batch size.
* Inefficient GPU utilization due to uneven token distribution across microbatches

## Solution
Introduce `balanced_binpacking` and `BalancedBatchIterator` utilities that:
* Enforce max_tokens_per_microbatch — no microbatch exceeds the token limit
* Roughly balance the microbatches so that we avoid straggler microbatches.
* Ensure that every worker still receives the same number of microbatches to satisfy Deepspeed/FSDP requirements (every DP worker must do the same number of forward passes).

## API changes

* Adds a `BalancedBatchIterator` that follows the `BatchIterator` interface, which is just an iterator over `Experience`s.
* Unifies `Worker.forward` to also use the `BatchIterator` interface, rather than call `TrainingBatch.chunk` directly.
* Replaces the `BatchIterator` with `BalancedBatchIterator` when the `max_tokens_per_microbatch` is set, for the `forward`, `Critic.ppo_train` and `Policy.ppo_train`.

## Example

Example ppo_train call from the unit test:

```
minibatch_sequences=tensor([[88, 96,  3, 23, 44, 65, 55,  0, 66, 49],
        [25, 85, 10, 86, 31, 55, 31, 51,  4, 56],
        [28, 38, 76,  8,  0,  0,  0,  0,  0,  0],
        [21, 97, 46, 75, 73,  0,  0,  0,  0,  0]])
minibatch_attention_mask=tensor([[1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
        [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
        [1, 1, 1, 1, 1, 0, 0, 0, 0, 0],
        [1, 1, 1, 1, 1, 0, 0, 0, 0, 0]])
Policy Train epoch [1/1]:   0%|          | 0/2 [00:00<?, ?it/s]
(DeepSpeedPolicyWorkerBase pid=182679) micro_batches_per_mini_batch=2
(DeepSpeedPolicyWorkerBase pid=182679) experience.sequences=tensor([[88, 96,  3, 23, 44, 65, 55,  0, 66, 49],
(DeepSpeedPolicyWorkerBase pid=182679)         [28, 38, 76,  8,  0,  0,  0,  0,  0,  0]])
Policy Train epoch [1/1]:  50%|█████     | 1/2 [00:01<00:01,  1.28s/it, pg=-0.203, glen=4, policy_lr=1e-6, ent=3.08]
(DeepSpeedPolicyWorkerBase pid=182679) experience.sequences=tensor([[25, 85, 10, 86, 31, 55, 31, 51,  4, 56],
(DeepSpeedPolicyWorkerBase pid=182679)         [21, 97, 46, 75, 73,  0,  0,  0,  0,  0]])
```

## Follow-up

This only does balanced microbatching at the individual worker level. We can further balance the minibatches sent to each worker at a global level to ensure every DP worker has roughly the same amount of work.

There's also different algorithms that we could plug into the microbatch chunking step.

### DIFF
--- a/skyrl-train/skyrl_train/workers/worker.py
+++ b/skyrl-train/skyrl_train/workers/worker.py
@@ -32,7 +32,7 @@ from loguru import logger
 from skyrl_train.distributed.ulysses import set_ulysses_sequence_parallel_group, apply_monkey_patch
 from skyrl_train.distributed.utils import init_custom_process_group
 from skyrl_train.utils.ppo_utils import PolicyLossRegistry, ppo_critic_loss, compute_approx_kl
-from skyrl_train.workers.worker_utils import BatchIterator, BalancedBatchIterator, reduce_metrics
+from skyrl_train.workers.worker_utils import BatchIterator, TokenBasedBatchIterator, reduce_metrics
 from skyrl_train.dataset.replay_buffer import Experience
 from skyrl_train.training_batch import TrainingInputBatch, TrainingOutputBatch
 from skyrl_train.inference_engines.inference_engine_client import InferenceEngineClient
@@ -329,7 +329,7 @@ class Worker(DistributedTorchRayActor):
         # Check if token-based chunking is enabled
         if self.cfg.trainer.max_tokens_per_microbatch > 0:
             # Use token-based chunking
-            micro_batch_iterator = BalancedBatchIterator(
+            micro_batch_iterator = TokenBasedBatchIterator(
                 data=data,
                 max_tokens_per_microbatch=self.cfg.trainer.max_tokens_per_microbatch,
             )
@@ -655,7 +655,7 @@ class PolicyWorkerBase(Worker):
         global_step = train_data.metadata["global_step"]
 
         if self.cfg.trainer.max_tokens_per_microbatch is not None:
-            dataloader = BalancedBatchIterator(
+            dataloader = TokenBasedBatchIterator(
                 train_data, max_tokens_per_microbatch=self.cfg.trainer.max_tokens_per_microbatch
             )
             accumulation_steps = dataloader.num_microbatches
@@ -958,7 +958,7 @@ class CriticWorkerBase(Worker):
 
         # TODO: Move this to the base class since it's common to both policy and critic workers.
         if self.cfg.trainer.max_tokens_per_microbatch is not None:
-            dataloader = BalancedBatchIterator(
+            dataloader = TokenBasedBatchIterator(
                 train_data, max_tokens_per_microbatch=self.cfg.trainer.max_tokens_per_microbatch
             )
             accumulation_steps = dataloader.num_microbatches
@@ -968,7 +968,7 @@ class CriticWorkerBase(Worker):
             )
 
             # TODO: Make `num_microbatches` a property of the dataloader instead of computing it here.
-            # Ex: see the BalancedBatchIterator class.
+            # Ex: see the TokenBasedBatchIterator class.
             micro_batches_per_mini_batch = (
                 self.policy_mini_batch_size_per_gpu // self.cfg.trainer.micro_train_batch_size_per_gpu
             )
--- a/skyrl-train/skyrl_train/workers/worker_utils.py
+++ b/skyrl-train/skyrl_train/workers/worker_utils.py
@@ -146,7 +146,7 @@ def balanced_binpacking(token_counts: List[int], max_tokens_per_microbatch: int)
     return microbatch_indices
 
 
-class BalancedBatchIterator(BaseBatchIterator):
+class TokenBasedBatchIterator(BaseBatchIterator):
     """An iterator that chunks microbatches based on real token count.
 
     Packs samples into microbatches, ensuring each microbatch doesn't exceed


skyrl-train/skyrl_train/workers/worker.py
```diff
@@ -322,11 +322,21 @@ def forward(
...
-        # TODO (sumanthrh): this can be in the policy/critic impl if the micro batch size can be specific to policy, critic, etc.
-        micro_batches = data.chunk(self.cfg.trainer.micro_forward_batch_size_per_gpu)
+        # Check if token-based chunking is enabled
+        if self.cfg.trainer.max_tokens_per_microbatch > 0:
+            # Use token-based chunking
+            micro_batch_iterator = BalancedBatchIterator(
+                data=data,
+                max_tokens_per_microbatch=self.cfg.trainer.max_tokens_per_microbatch,
+            )
+            micro_batches = list(micro_batch_iterator)
```
justinvyu: todo: reorder after forward

skyrl-train/skyrl_train/workers/worker_utils.py
```diff
@@ -69,3 +75,171 @@ def batch_to_experience(batch: TrainingInputBatch):
...
+        for microbatch_idx, original_indices in enumerate(self._microbatches):
+            for sample_idx, original_idx in enumerate(original_indices):
+                original_idx_to_microbatch_idx[original_idx] = (microbatch_idx, sample_idx)
+
+        # Reorder the microbatches to match the original data order
+        reordered_microbatches = []
+        for original_idx in range(len(self._token_counts)):
+            microbatch_idx, sample_idx = original_idx_to_microbatch_idx[original_idx]
+            reordered_microbatches.append(non_padding_microbatches[microbatch_idx][sample_idx])
+        return TensorBatch.cat(reordered_microbatches)
```
justinvyu: TODO: I think this is inefficient

skyrl-train/tests/gpu/gpu_ci/test_ppo_train.py
```diff
@@ -77,6 +77,74 @@ def test_ppo_train_basic_execution(ray_init_fixture, cfg):
...
+            "response_mask": torch.ones((batch_size, num_actions), dtype=int, device="cpu"),
+        }
+    )
+    data.metadata = {"response_length": num_actions}
+    return data
+
+
+
+@pytest.mark.parametrize("worker_type", ["policy", "critic"])
+def test_max_tokens_per_microbatch(ray_init_fixture, cfg, worker_type):
```
justinvyu, line 118: TODO: test a multi-worker case where you have padding microbatches

skyrl-train/skyrl_train/workers/worker_utils.py
```diff
@@ -37,15 +55,18 @@ def __len__(self):
...
         except StopIteration:
             self._iter = iter(self._chunks)
             raise StopIteration
 
+    def reorder_microbatches(self, microbatches: List[TensorBatch]) -> TensorBatch:
+        """Reorder microbatches to match the order of the original data."""
+        # TODO: Define a common interface for batch iterators.
+        return TensorBatch.cat(microbatches)
+
     @staticmethod
```
SumanthRH, line 70: This `batch_to_experience` should be moved to `BaseBatchIterator` it looks like?

skyrl-train/skyrl_train/workers/worker.py
```diff
@@ -953,7 +986,8 @@ def ppo_train(self, train_data: TrainingInputBatch) -> TrainingOutputBatch:
                 desc=f"Critic Train epoch [{epoch + 1}/{self.cfg.trainer.update_epochs_per_batch}]",
                 disable=not self.strategy.is_rank_0(),
             )
-            for local_step, experience in enumerate(pbar):
+            for local_step, batch in enumerate(pbar):
+                experience = BatchIterator.batch_to_experience(batch)
```
SumanthRH, line 990:
```suggestion
                experience = BaseBatchIterator.batch_to_experience(batch)
```

skyrl-train/skyrl_train/workers/worker.py
```diff
@@ -932,19 +955,29 @@ def save_hf_model(self, export_dir: str, tokenizer):
...
-        )
+
+        # TODO: Move this to the base class since it's common to both policy and critic workers.
+        if self.cfg.trainer.max_tokens_per_microbatch is not None:
+            dataloader = TokenBasedBatchIterator(
+                train_data, max_tokens_per_microbatch=self.cfg.trainer.max_tokens_per_microbatch
+            )
+            accumulation_steps = dataloader.num_microbatches
+        else:
+            dataloader = BatchIterator(
```
SumanthRH, line 966:
```suggestion
            dataloader = SampleBasedBatchIterator(
```

skyrl-train/skyrl_train/workers/worker.py
```diff
@@ -660,7 +680,9 @@ def ppo_train(self, train_data: TrainingInputBatch) -> TrainingOutputBatch:
                 desc=f"Policy Train epoch [{epoch + 1}/{self.cfg.trainer.update_epochs_per_batch}]",
                 disable=not self.strategy.is_rank_0(),
             )
-            for local_step, experience in enumerate(pbar):
+            for local_step, batch in enumerate(pbar):
+                experience = BatchIterator.batch_to_experience(batch)
```
SumanthRH, line 684:
```suggestion
                experience = BaseBatchIterator.batch_to_experience(batch)
```

skyrl-train/skyrl_train/workers/worker.py
```diff
@@ -32,7 +32,7 @@
 from skyrl_train.distributed.ulysses import set_ulysses_sequence_parallel_group, apply_monkey_patch
 from skyrl_train.distributed.utils import init_custom_process_group
 from skyrl_train.utils.ppo_utils import PolicyLossRegistry, ppo_critic_loss, compute_approx_kl
-from skyrl_train.workers.worker_utils import BatchIterator, reduce_metrics
+from skyrl_train.workers.worker_utils import BatchIterator, TokenBasedBatchIterator, reduce_metrics
```
SumanthRH, line 35:
```suggestion
from skyrl_train.workers.worker_utils import BaseBatchIterator, SampleBasedBatchIterator, TokenBasedBatchIterator, reduce_metrics
```

skyrl-train/skyrl_train/workers/worker.py
```diff
@@ -640,15 +653,22 @@ def _normalize_mini_batch_size(self):
...
+        else:
+            dataloader = BatchIterator(
+                train_data, sample_batch_size=self.cfg.trainer.micro_train_batch_size_per_gpu, drop_last=False
+            )
+
+            micro_batches_per_mini_batch = (
+                self.policy_mini_batch_size_per_gpu // self.cfg.trainer.micro_train_batch_size_per_gpu
+            )
+            # The number of steps (over micro batches) to accumulate gradients before taking an optimizer step.
+            accumulation_steps = micro_batches_per_mini_batch
```
SumanthRH, line 671: QQ: Should the dynamic batch size related packing happen at the batch level or the mini-batch level? 

The most efficient one would be at the batch level, but then we won't have a good way to get mini batches afterwards. It seems like we should just do this at the mini batch level first. (you'd need to introduce an explicit loop over mini batches)

SumanthRH: This also allows us to naturally define the accumulation steps

justinvyu: > Should the dynamic batch size related packing happen at the batch level or the mini-batch level?

Yeah it should also be done at the global batch level, but I'm leaving that for a follow-up optimization, see the end of the PR description. The global batch level batch balancing is useful to assign each worker a similar number of tokens.

This PR's main goal is the max tokens per microbatch parameter. The balancing of microbatches is kind of a convenient feature add-on that is a slight optimization.

Here's why the balancing at the microbatch level is useful:
* Worker 0 mini-batch contains 4 sequences of length: [10, 10, 5, 5]
* Worker 0 has a max microbatch token length of 15.
* If you just do a single front-to-back pass, you'll end up with 3 microbatches: [10], [10, 5], [5]
* However, the better way to chunk this would be [10, 5], [10, 5], which is what the balanced microbatching method does here.
* Basically, we want to minimize number of microbatches while satisfying max_tokens_per_microbatch.

justinvyu: > This also allows us to naturally define the accumulation steps

Could you elaborate on this a bit more?

SumanthRH: Oh sorry here's what I meant;

We have "mini batches" per worker that is set based on `policy_mini_batch_size`.  This is separate from the per-worker shard you're referring to.


 `train_batch_size / policy_mini_batch_size` dictates the number of optimization steps with the current training batch. 
 
 So for example, with `train_batch_size=16, num_workers=8, policy_mini_batch_size=8` , each worker will receive a shard of 2 samples, and this wil be split further into `train_batch_size/ policy_mini_batch_size=2` mini batches. Each mini batch will correspond to one optimization step. 
 
 For each mini batch per worker, we further split it into micro batches. 
 
 My point was that the dynamic batch size related binpacking should happen at the worker-level mini batches not the worker-level overall batch.

justinvyu:
Closing this PR since it's a bit outdated. Will redo this now that https://github.com/NovaSky-AI/SkyRL/pull/817 has landed.

PR #649, opened 2025-11-10
[WIP] [tx] Torch definition for qwen3 and LoRA

tyler-griggs:
Implements a torch definition of Qwen3 and multi-LoRA. 

The intent was to closely match the interface and implementation of the jax-based code, with exceptions for differences in the canonical torch library (e.g., the interface to attention is a little different with tensor dimensions in a different order). 

This PR focuses on dense Qwen3 with LoRA applied to the linear layers. This PR does not add support for Qwen3 MoE or LoRA in the embedding or expert layers. 

There are also performance improvements not included. For example, `apply_lora` loops over adapter indices and performs individual mm's instead of a raggeddot/gemm. These will be resolved in following PRs.

### DIFF
--- a/skyrl-tx/tx/torch/layers/util.py
+++ b/skyrl-tx/tx/torch/layers/util.py
@@ -51,8 +51,7 @@ def prepare_routing(
     sorted_adapter_indices = None if adapter_indices is None else adapter_indices[sort_idx]
 
     # Compute group sizes (minlength guarantees output length)
-    sorted_indices = indices[sort_idx]
-    group_sizes = torch.bincount(sorted_indices, minlength=num_groups)
+    group_sizes = torch.bincount(indices, minlength=num_groups)
 
     # Inverse permutation to restore original order
     unsort_indices = torch.argsort(sort_idx)


skyrl-tx/tx/torch/layers/lora.py
```diff
@@ -0,0 +1,199 @@
...
+        self.max_lora_rank = int(max_lora_rank)
+
+        if self.max_lora_adapters == 0:
+            self.lora_scaling = None
+            self.lora_ranks = None
+            self.lora_A = None
+            self.lora_B = None
+            return
+
+        if device is None:
```
pcmoritz: Should we remove this and always make device required (and elsewhere too)?

skyrl-tx/tx/torch/layers/lora.py
```diff
@@ -0,0 +1,199 @@
...
+        self.register_buffer(
+            "lora_ranks",
+            torch.full((self.max_lora_adapters,), self.max_lora_rank, dtype=torch.int32, device=device),
+            persistent=True,
+        )
+
+        A = torch.empty(*shape_A, dtype=dtype, device=device)
+        B = torch.zeros(*shape_B, dtype=dtype, device=device)
+        nn.init.kaiming_uniform_(A, a=math.sqrt(5))  # He-uniform A
+        self.lora_A = nn.Parameter(A, requires_grad=True)
```
pcmoritz: I wonder if we can / if it makes sense to have something a little more like our `Param` class s we can do `Param(*shape, init=..., sharding=...)` going forward (we can also experiment with this as a follow up / later). The imperative `nn.init.kaiming_uniform_` initialization is slightly ugly :D

skyrl-tx/tx/torch/layers/lora.py
```diff
@@ -0,0 +1,199 @@
...
+
+        Returns:
+          base_output + lora_output with per-adapter routing and scaling
+        """
+        if self.max_lora_adapters == 0 or adapter_indices is None:
+            return base_output
+
+        if x.dim() != 3:
+            raise ValueError("x must be [B, T, in_features].")
+        B, T, in_features = x.shape
```
pcmoritz, line 96: We will need to make sure this can support https://github.com/NovaSky-AI/SkyRL/pull/511

skyrl-tx/tx/torch/layers/lora.py
```diff
@@ -0,0 +1,199 @@
...
+      base_out = F.linear(x, weight, bias)
+      return self.apply_lora(x, base_out, adapter_indices)
+    """
+
+    def __init__(
+        self,
+        in_features: int,
+        out_features: int,
+        *,
+        max_lora_adapters: int = 0,
```
pcmoritz: It would probably be best if we get rid of all the defaults here going forwards (I know the current code has them, but I don't think it is good, it can only lead to errors if somebody forgets to pass the parameter and the default was not good)

skyrl-tx/tx/torch/layers/util.py
```diff
@@ -0,0 +1,37 @@
...
+        unsort_indices: Indices to restore original order after ragged operations
+        sorted_adapter_indices: Adapter indices sorted with tokens (or None if not provided)
+    """
+    # Sort by group index
+    sort_idx = torch.argsort(indices)
+    sorted_tokens = tokens[sort_idx]
+    sorted_adapter_indices = None if adapter_indices is None else adapter_indices[sort_idx]
+    
+    # Compute group sizes (minlength guarantees output length)
+    sorted_indices = indices[sort_idx]
```
pcmoritz: Hmm, this is not needed, right? E.g. bincount is permutation invariant so we can just pass indices in there and otherwise we don't need sorted_indices.

skyrl-tx/tx/torch/layers/util.py
```diff
@@ -0,0 +1,37 @@
...
+        adapter_indices: Optional tensor of shape (num_tokens,) to be sorted together with tokens
+
+    Returns:
+        sorted_tokens: Tokens sorted by group index
+        group_sizes: Number of tokens in each group
+        unsort_indices: Indices to restore original order after ragged operations
+        sorted_adapter_indices: Adapter indices sorted with tokens (or None if not provided)
+    """
+    # Sort by group index
+    sort_idx = torch.argsort(indices)
```
pcmoritz, line 49: Let's call this `sort_indices` so it is consistent with `unsort_indices` below and make clear it is multiple indices? (or alternatively maybe sort_perm and unsort_perm for "permutation" if you prefer)?

skyrl-tx/tx/torch/models/qwen3.py
```diff
@@ -0,0 +1,319 @@
...
+            T_new = k.size(2)
+            with torch.no_grad():
+                k_cache[:, :, cache_pos : cache_pos + T_new, :].copy_(k)
+                v_cache[:, :, cache_pos : cache_pos + T_new, :].copy_(v)
+
+            k, v = k_cache, v_cache
+
+        updated_cache = (k, v)
+
+        # Attention (causal only during prefill, GQA handled via repeat)
```
pcmoritz, line 131: You should be able to just set `enable_gqa=True` in `scaled_dot_product_attention`, right?

skyrl-tx/tx/torch/models/qwen3.py
```diff
@@ -0,0 +1,319 @@
...
+        hidden_states = residual + hidden_states
+
+        return hidden_states, updated_cache
+
+
+class Qwen3Model(nn.Module):
+    def __init__(self, config: Qwen3Config, *, dtype: torch.dtype):
+        super().__init__()
+        self.config = config
+        max_lora_adapters = getattr(config, "max_lora_adapters", 0)
```
pcmoritz, line 247: Since https://github.com/NovaSky-AI/SkyRL/pull/636, you don't need these any more, you can just do config.{max_lora_adapters, max_lora_rank), and it is probably easiest to just inline it in the `Qwen3DecoderLayer` constructor below

skyrl-tx/tests/torch/models/test_qwen3.py
```diff
@@ -0,0 +1,347 @@
...
+
+def load_model_from_hf_checkpoint(
+    checkpoint_dir: str, config: Qwen3Config, device, dtype: torch.dtype = torch.float32
+) -> Qwen3ForCausalLM:
+    """Load our model from a HuggingFace checkpoint directory."""
+    model = Qwen3ForCausalLM(config, dtype=dtype)
+
+    # Load all safetensors files
+    state_dict = {}
+    from pathlib import Path
```
pcmoritz, line 63: Let's move this to the top?

skyrl-tx/tx/torch/layers/lora.py
tyler-griggs: Gave this a shot!

skyrl-tx/tx/torch/layers/lora.py
tyler-griggs: Yes, it won't right now, but will be updated to support embedding layer adapters

skyrl-tx/tx/torch/layers/lora.py
tyler-griggs: Good point, this is a good chance to clean it up. I set all optional arguments to default to None, but otherwise removed the defaults

skyrl-tx/tx/torch/layers/util.py
tyler-griggs: Oh good catch

PR #657, opened 2025-11-11
[tx] Add LLama3 support

atemaguer:
This PR adds LLama3.2 models support to Tx. LLlam3 and Qwen3 mostly share the same architecture except for the QK-Norm layers present in Qwen3 but absent in LLama3. So both models share certain layers.

### DIFF
--- a/skyrl-tx/tx/models/llama3.py
+++ b/skyrl-tx/tx/models/llama3.py
@@ -251,6 +251,7 @@ class Llama3Model(nnx.Module):
         if output_hidden_states:
             all_hidden_states.append(hidden_states)
 
+        # Increment cache_position if cache exists, or use sequence length for new cache
         new_cache_position = kv_cache.cache_position + 1 if kv_cache is not None else input_ids.shape[1]
 
         return ModelOutput(
--- a/skyrl-tx/tx/models/qwen3.py
+++ b/skyrl-tx/tx/models/qwen3.py
@@ -358,6 +358,7 @@ class Qwen3Model(nnx.Module):
         if output_hidden_states:
             all_hidden_states.append(hidden_states)
 
+        # Increment cache_position if cache exists, or use sequence length for new cache
         new_cache_position = kv_cache.cache_position + 1 if kv_cache is not None else input_ids.shape[1]
 
         return ModelOutput(


atemaguer:
@pcmoritz , any thoughts about this PR?

atemaguer:
@tyler-griggs, got thoughts about this PR?

skyrl-tx/tx/tinker/api.py
```diff
@@ -270,7 +270,7 @@ class SaveWeightsForSamplerRequest(BaseModel):
 class SamplingParams(BaseModel):
     max_tokens: int | None = None
     seed: int | None = None
-    stop: Sequence[int] | None = None
+    stop: Sequence[int] | Sequence[str] | None = None
```
tyler-griggs: I'm curious why these updates to `stop` were needed -- could you please explain?

skyrl-tx/tx/models/qwen3.py
tyler-griggs, line 1: Based on the observation you made that Qwen3 and Llama3 are very similar, could we simplify this even further such that Qwen3 inherits from Llama3 and adds fairly light-weight additional logic? E.g., the transformers library takes this pattern: https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3/modular_qwen3.py

We could likely further reduce the duplicate code between the two models, which will make it easier to keep them consistent going forward. But let me know if this actually doesn't make sense.

skyrl-tx/tx/tinker/api.py
atemaguer: I think it's because the Tinker client API update supports providing stop sequences as well. Also, some of the examples were failing, especially when using Llama 3 as the base model.

skyrl-tx/tx/models/qwen3.py
atemaguer: On it

atemaguer:
@tyler-griggs , let me know what you think of these new updates

skyrl-tx/tx/models/llama3.py
```diff
@@ -0,0 +1,318 @@
...
+
+
+class Llama3Model(nnx.Module):
+    """LLaMA 3 base model (without language modeling head)."""
+
+    def __init__(self, config: LlamaConfig, *, dtype: jnp.dtype, rngs: nnx.Rngs) -> None:
+        self.config = config
+
+        # Token embeddings
+        self.embed_tokens = nnx.Embed(
```
tyler-griggs: I assume we'd like to have LoRA in the embedding layer for Llama3 (as we do in Qwen3). Is the plan to do this in a follow up PR?

skyrl-tx/tx/models/qwen3.py
```diff
@@ -105,13 +34,22 @@ def __call__(
...
         v = self.v_proj(x, adapter_indices=adapter_indices).reshape(B, T, self.num_kv_heads, self.head_dim)
 
+        # Apply QK-Norm (Qwen3-specific)
+        q = self.q_norm(q)
+        k = self.k_norm(k)
+
+        # Get rope_theta from config
+        from tx.layers.common import apply_rope
+
+        rope_theta = getattr(self.config, "rope_theta", 500000.0)
```
tyler-griggs: Ideally we would not set default parameters in the code like this, but rather inherit the defaults from the base config itself. Is there a reason to use `getattr` here or can we avoid it?

skyrl-tx/tx/tinker/api.py
tyler-griggs: It makes sense that it should be updated if the Tinker client API should support stop sequences, but I don't think our generation logic actually supports stop sequences right now. It seems like we should push this set of changes into a later PR when we actually add support for stop strings sequences rather than just stop tokens.

atemaguer:
@tyler-griggs , let me know what you think about these new updates

atemaguer:
Btw @pcmoritz, it appears like this PR will cause conflicting changes with the FSDP support changes. Any chance you could review and merge this PR and then add the FSDP changes after?

skyrl-tx/tx/tinker/engine.py
```diff
@@ -2,7 +2,6 @@
 
 import argparse
```
pcmoritz, line 3: I'm a little suprised about the diff in this file, maybe you need to merge main into the PR? These changes should already be on the current master :)

atemaguer:
@pcmoritz, I merged the changes from main. let me know if there's anything else that needs changing or improving

pcmoritz:
Thanks a lot for all the work, I'll get this PR merged next :)

I'm planning do a little bit of restructuring, it is probably easiest to just edit it directly :)

PR #658, opened 2025-11-12
Added support for saving shard as HuggingFace safetensors for Megatron backend (WIP)

zhenga1:
Added new runnable script `main_load` which can be referenced with uv pip. 

Added support for new functions in `worker.py`. Provided example implementation of working conversion using a new script `skyrl-train/examples/gsm8k/convert_megatron_to_hf.sh`

To run the script, copy the original `.sh` file used for training and change the entrypoint from `skyrl_train.entrypoints.main_base` to `skyrl_train.entrypoints.main_load`. 

Please comment and give feedback.


zhenga1:
Fixed the other suggestions, have not changed the entrypoint `main_base.py`, since there may be many training scripts that reference `main_base.py` already. In other words, the `main_load.py` and `main_base.py` still both exists.

PR #641, opened 2025-11-07
[tx] Add fine-grained LoRA control (train_attn, train_mlp, train_unembed)

hezyin:
## Summary

Adds three boolean configuration options for fine-grained control over which layers use LoRA:
- `train_attn`: Enable LoRA on attention layers (q, k, v, o projections)
- `train_mlp`: Enable LoRA on MLP layers (gate, up, down projections + MoE experts)
- `train_unembed`: Enable LoRA on the unembedding/LM head layer

All three can be configured independently via API or CLI arguments.

## Implementation Details

**Mechanism**: When a flag is `False`, we pass `max_lora_adapters=0` to layer constructors. This prevents LoRA parameter allocation entirely—the layer behaves as if LoRA doesn't exist.

**Key changes**:
1. **Qwen3Attention** (qwen3.py:~48): Conditionally enables LoRA on q/k/v/o projections based on `train_attn`
2. **Qwen3MLP** (qwen3.py:~145): Conditionally enables LoRA on gate/up/down projections based on `train_mlp`
3. **Qwen3Experts** (qwen3.py:~192): Conditionally enables LoRA on MoE expert layers based on `train_mlp`
4. **Qwen3ForCausalLM** (qwen3.py:~403): Converted `lm_head` from `nnx.Linear` to `LoRALinear` with conditional enabling based on `train_unembed`

## Changes

- **tx/tinker/types.py**: Added `train_attn`, `train_mlp`, `train_unembed` to `LoraConfig`
- **tx/models/configs.py**: Added train flags to `Qwen3Config` with defaults (True/True/False)
- **tx/tinker/config.py**: Added CLI argument support for train flags
- **tx/tinker/engine.py**: Passes train flags to `Qwen3Config` constructor
- **tx/models/qwen3.py**: Implemented conditional LoRA in 4 model classes
- **tests/models/test_qwen3_config.py**: Added tests for default and custom flag values

## Notes

- **Defaults preserve current behavior**: `train_attn=True`, `train_mlp=True`, `train_unembed=False`
- **Breaking change for lm_head**: Converted from `nnx.Linear` to `LoRALinear`. Even with `train_unembed=False` (default), the layer type changes, but behavior remains identical since `max_lora_adapters=0` means no LoRA parameters exist.
- All tests pass (27/27)

## Dependencies

- Depends on #636 (Qwen3Config refactor)

Fixes #630.

PR #664, opened 2025-11-13
[tx] Add per-layer LoRA control

### LINKED ISSUE #630: [tx] Add config options for LoRA configs
Implement the `train_unembed`, `train_mlp`, `train_attn` parameters in https://github.com/thinking-machines-lab/tinker/blob/main/src/tinker/types/lora_config.py#L18

hezyin:
## Summary

Adds fine-grained per-layer LoRA control and extends LoRA support to the lm_head (unembedding) layer. This PR implements three independent boolean flags that control which model layers use LoRA:

- `train_attn`: Enable LoRA on attention layers (q, k, v, o projections) - default: `True`
- `train_mlp`: Enable LoRA on MLP layers (gate, up, down projections + MoE experts) - default: `True`  
- `train_unembed`: Enable LoRA on the lm_head/unembedding layer - default: `False`

### 1. Per-Layer LoRA Control (train_attn, train_mlp, train_unembed)
- **Mechanism**: When a flag is `False`, the effective rank is set to 0 for that layer type
- **Implementation**: Modified `update_adapter_config()` in `layers/lora.py` to check layer type and set rank=0 for disabled layers
- **Affected layers**:
  - `train_attn=False`: Disables q_proj, k_proj, v_proj, o_proj
  - `train_mlp=False`: Disables gate_proj, up_proj, down_proj, and MoE expert layers
  - `train_unembed=False`: Disables lm_head

### 2. lm_head LoRA Support
- **Changed**: Converted `lm_head` from `nnx.Linear` to `LoRALinear` in `qwen3.py:397-407`
- **Behavior**: With `train_unembed=False` (default), lm_head has rank=0, so it behaves identically to the old `nnx.Linear` implementation
- **Note**: Only applies to models with `tie_word_embeddings=False` (models with separate lm_head, like the tiny test model)

### 3. MoE Expert Rank Normalization
- Fixed MoE expert LoRA layers to properly handle rank scaling when `train_mlp=False`
- Ensures expert layers behave consistently with regular MLP layers

### 4. Checkpoint Save/Load with Zero-Sized Arrays
**Problem**: When a layer has rank=0 (disabled LoRA), it creates zero-sized arrays like `lora_A[adapter_idx, ..., :, :0]`. Flax's `save_checkpoint()` rejects these with error:
```
Cannot save arrays with zero size: ParamInfo: [name=lora_weights.lm_head.lora_A]
```

**Solution**: 
- Filter out zero-sized arrays before saving to Flax checkpoints (in `process_save_weights()`)
- Modified `insert_adapter_state()` to handle missing keys during restore by keeping original values
- This allows rank=0 parameters to be correctly preserved through save/load cycles

### 5. Optimizer State Path Handling
- Fixed `get_rank_path()` in `utils/models.py` to correctly handle optimizer state paths
- Discovered the actual structure: `('opt_state', 'inner_state', idx, moment_name, ...)`
- Previous assumption was: `('opt_state', idx, moment_name, ...)` (missing 'inner_state')

## Files Modified

- **tx/tinker/types.py**: Added `train_attn`, `train_mlp`, `train_unembed` to `LoraConfig`
- **tx/models/qwen3.py**: Added lm_head LoRA support (converted to `LoRALinear`)
- **tx/layers/lora.py**: Added layer-specific rank control in `update_adapter_config()`
- **tx/tinker/engine.py**: Added zero-sized array filtering in `process_save_weights()`
- **tx/utils/models.py**: 
  - Fixed optimizer state path handling in `get_rank_path()`
  - Fixed checkpoint restore in `insert_adapter_state()` to handle filtered arrays
- **tests/**: All existing tests updated and passing

## Test Results

All 24 tests passing, including:
- ✅ Checkpoint save/load with disabled LoRA layers (`test_training_workflow`)
- ✅ Sampling with LoRA adapters (`test_sample[lora_model]`)
- ✅ Forward/backward with different layer configurations

## Breaking Changes

None. Default behavior is preserved:
- `train_attn=True` (same as before)
- `train_mlp=True` (same as before)
- `train_unembed=False` (lm_head had no LoRA before, so disabled by default)

Fixes #630
Supersedes #641

### DIFF
--- a/skyrl-tx/tx/layers/lora.py
+++ b/skyrl-tx/tx/layers/lora.py
@@ -286,9 +286,9 @@ def update_adapter_config(model: nnx.Module, adapter_index: int, lora_config: Lo
         # Determine if this layer should be trained based on layer type
         if not lora_config.train_attn and "self_attn" in path_str:
             effective_rank = 0
-        elif not lora_config.train_mlp and ("mlp" in path_str or "experts" in path_str):
+        if not lora_config.train_mlp and ("mlp" in path_str or "experts" in path_str):
             effective_rank = 0
-        elif not lora_config.train_unembed and ("embed_tokens" in path_str or "lm_head" in path_str):
+        if not lora_config.train_unembed and ("embed_tokens" in path_str or "lm_head" in path_str):
             effective_rank = 0
 
         if path[-2].key == "lora_ranks":


skyrl-tx/tx/tinker/engine.py
```diff
@@ -766,11 +763,26 @@ def process_save_weights(self, model_id: str, request_data: types.SaveWeightsInp
                 adapter_index, nnx.state(self.optimizers[model_id]), self.non_lora_params
             )
 
+            # Filter out zero-sized arrays (disabled LoRA layers with rank=0)
```
pcmoritz: I'm pretty sure this is not actually needed for the training checkpoints -- saving the rank 0 arrays and loading them later is totally fine (I just tested it after the fix https://github.com/NovaSky-AI/SkyRL/pull/666). I'm going to remove this code :)

pcmoritz: Hehe ok I hit the orbax codepath that complains about rank 0 arrays now. Very unfortunate that orbax doesn't allow this, I'll try to bring this code back in the form of a wrapper function around orbax that handles this correctly without needing to modify the rest of the code :)

pcmoritz: I also opened an upstream issue about this: https://github.com/google/orbax/issues/2587

skyrl-tx/tx/layers/lora.py
```diff
@@ -268,20 +269,39 @@ def update_adapter_config(model: nnx.Module, adapter_index: int, lora_rank: int,
...
+        elif not lora_config.train_unembed and "lm_head" in path_str:
+            effective_rank = 0
+
         if path[-2].key == "lora_ranks":
-            return value.at[adapter_index].set(lora_rank)
+            return value.at[adapter_index].set(effective_rank)
         if path[-2].key == "lora_scaling":
-            return value.at[adapter_index].set(scaling)
+            # Set scaling to 0.0 if rank is 0
+            return value.at[adapter_index].set(lora_config.alpha / effective_rank if effective_rank > 0 else 0.0)
```
pcmoritz, line 298: I think this should be `lora_config.alpha / effective_rank` and not `lora_config.alpha / lora_config.rank` because we are trying to make the learning rate independent of rank and we are summing effective_rank many numbers. It would be good to check this experimentally.

skyrl-tx/tx/utils/models.py
```diff
@@ -214,7 +226,7 @@ def insert_state(path: tuple, p: jax.Array, new: jax.Array):
         elif path[-1].key == "lora_B":
             return p.at[adapter_index, ..., :rank, :].set(new)
 
-    updated = jax.tree.map_with_path(insert_state, nnx.to_pure_dict(lora_params), new_params)
```
pcmoritz, line 217: This change was needed, since new_params now contains less keys than lora_params so we cannot just map over both, instead we have to map over the one with lesser keys and then get the values from lora_params via the dictionary.

tyler-griggs (approved):

tyler-griggs (approved):

yuhao-zh:
What is the specific parameter meaning of ```train_unembed```, whether it include ```lmhead and embedding``` or only ```lmhead```? Is this a bug ?

pcmoritz:
I took it to mean if train_unembed is set, both lmhead and embedding is trained, and if it is not set, neither of them are trained. Treating them separately didn't seem natural, since they might be tied, and also there is not a separate flag for lmhead/embedding. The parameter was initially introduced so it is easy to train models that can be served on a serving system that doesn't support custom embedding / lmhead.

However, it is possible that the above interpretation is incorrect -- it is worth checking how exactly the parameter is implemented in the tinker service, and mirror it here too. If you end up investigating this more, keep us updated :)

The official documentation I'm aware of is here https://tinker-docs.thinkingmachines.ai/tinker/api-reference/types/loraconfig/?h=train_unembed

yuhao-zh:
> I took it to mean if train_unembed is set, both lmhead and embedding is trained, and if it is not set, neither of them are trained. Treating them separately didn't seem natural, since they might be tied, and also there is not a separate flag for lmhead/embedding. The parameter was initially introduced so it is easy to train models that can be served on a serving system that doesn't support custom embedding / lmhead.我的理解是，如果设置了 `train_unembed`，则 lmhead 和 embedding 都会被训练；如果未设置，则两者都不被训练。将它们分开处理似乎不太自然，因为它们可能相互关联，而且也没有单独的标志位来区分 lmhead 和 embedding。最初引入这个参数是为了方便训练那些可以在不支持自定义 embedding/lmhead 的服务系统上运行的模型。
> 
> However, it is possible that the above interpretation is incorrect -- it is worth checking how exactly the parameter is implemented in the tinker service, and mirror it here too. If you end up investigating this more, keep us updated :)然而，上述解释可能并不正确——值得检查一下 Tinker 服务中该参数的具体实现方式，并将其反映在这里。如果您对此进行了更深入的研究，请随时告知我们 :)
> 
> The official documentation I'm aware of is here https://tinker-docs.thinkingmachines.ai/tinker/api-reference/types/loraconfig/?h=train_unembed我所知的官方文档在这里： [https://tinker-docs.thinkingmachines.ai/tinker/api-reference/types/loraconfig/?h](https://tinker-docs.thinkingmachines.ai/tinker/api-reference/types/loraconfig/?h=train_unembed) =train_unembed

yeah, I also think this setting is more reasonable. However, during training, I found that the training effect was worse when unembed was enabled than when it was not. Specifically, it converged faster and had lower precision. I'm looking into the reason and I also checked the official documentation of tinker. train_unembed refers to the unembedding layer, that is, lmhead. I will conduct a more in-depth study before leaving you a message. Thank you very much for your reply.

PR #669, opened 2025-11-15
[tx] Clean up rank handling for extract and insert adapter state

pcmoritz:
Implement the suggestion from https://github.com/google/flax/issues/5030 and also prepare for jit compiling the loading and saving of the adapter. This means we will always save the full lora `rank` tensors, even for expert layers. But the performance improvements from jitting should outweight the efficiency loss, and going forward we can optimize further and get rid of the additional overhead by slicing the tensors in `save_safetensors` (and restoring in `load_safetensors`).


pcmoritz:
TODO: Before merging this, we have to bring back the removal of LoRA weights that are not trained from sampling checkpoints

PR #674, opened 2025-11-16
[tx] fsdp support for qwen3

taroigarashi2001:
FSDP support for Qwen3 inside `skyrl-tx`. This is my first PR into SkyRL and deep dive into LLM training code in general, so I apologize in advance for any obvious mistakes or issues.

Benchmarks on 2 x H100 with 2048 batch size and 10 runs are shown below. The results to me seem reasonable, TP should increase inference throughput by reducing the workload per node. FSDP should slow things down slightly due to the communication overhead in exchange for less memory utilization.

```
RUNS=10 BATCH_SIZE=2048 CUDA_VISIBLE_DEVICES=0,1 uv run --extra dev --extra tinker --extra gpu pytest -v -s tests/models/test_qwen3_generate.py

tests/models/test_qwen3_generate.py::test_qwen3_generate_speed[0]
Testing with mesh configuration: basic (dp=1, tp=1, fsdp=False, batch size=2048, runs=10)
You are using a model of type qwen3 to instantiate a model of type . This is not supported for all configurations of models and can yield errors.
Generation stats (50 tokens, 10 runs):
Mean time: 1163.93 ± 3.59 ms
Min/Max: 1160.15 / 1173.37 ms
New tokens/sec: 87977.77
PASSED

tests/models/test_qwen3_generate.py::test_qwen3_generate_speed[1]
Testing with mesh configuration: tp (dp=1, tp=2, fsdp=False, batch size=2048, runs=10)
You are using a model of type qwen3 to instantiate a model of type . This is not supported for all configurations of models and can yield errors.
Generation stats (50 tokens, 10 runs):
Mean time: 819.94 ± 7.46 ms
Min/Max: 815.67 / 842.08 ms
New tokens/sec: 124887.53
PASSED

tests/models/test_qwen3_generate.py::test_qwen3_generate_speed[2]
Testing with mesh configuration: fsdp (dp=2, tp=1, fsdp=True, batch size=2048, runs=10)
You are using a model of type qwen3 to instantiate a model of type . This is not supported for all configurations of models and can yield errors.
Generation stats (50 tokens, 10 runs):
Mean time: 1358.68 ± 102.92 ms
Min/Max: 1321.38 / 1667.26 ms
New tokens/sec: 75367.24
PASSED
```

Thank you to @atemaguer for walking me through many of the metal concepts of FSDP and TP.

### DIFF
--- a/skyrl-tx/tests/models/test_qwen3_generate.py
+++ b/skyrl-tx/tests/models/test_qwen3_generate.py
@@ -16,31 +16,8 @@ from tx.tinker import types
 from tx.utils.models import load_safetensors
 
 
-@pytest.fixture
-def mesh_configs():
-    """Generate mesh configurations to test."""
-    num_devices = len(jax.devices())
-    configs = [
-        {"name": "basic", "fsdp": 1, "tp": 1},
-    ]
-
-    if num_devices > 1:
-        configs.append({"name": "tp", "fsdp": 1, "tp": num_devices})
-        configs.append({"name": "fsdp", "fsdp": num_devices, "tp": 1})
-
-    return configs
-
-
-@pytest.mark.parametrize("mesh_config_idx", [0, 1, 2])
-def test_qwen3_generate(mesh_config_idx, mesh_configs):
+def test_qwen3_generate():
     """Test batched text generation with KV caching matches HuggingFace."""
-    if mesh_config_idx >= len(mesh_configs):
-        pytest.skip(f"Mesh config index {mesh_config_idx} not available (only {len(mesh_configs)} configs)")
-    mesh_config = mesh_configs[mesh_config_idx]
-    print(
-        f"\nTesting with mesh configuration: {mesh_config['name']} (fsdp={mesh_config['fsdp']}, tp={mesh_config['tp']})"
-    )
-
     model_name = "Qwen/Qwen3-0.6B"
     tokenizer = AutoTokenizer.from_pretrained(model_name, padding_side="left")
     hf_model = AutoModelForCausalLM.from_pretrained(model_name, attn_implementation="eager", use_safetensors=True)
@@ -65,7 +42,7 @@ def test_qwen3_generate(mesh_config_idx, mesh_configs):
         base_config = PretrainedConfig.from_pretrained(model_name)
         config = Qwen3Config(base_config, max_lora_adapters=32, max_lora_rank=32, shard_attention_heads=True)
 
-        mesh = jax.make_mesh((mesh_config["fsdp"], mesh_config["tp"]), ("fsdp", "tp"))
+        mesh = jax.make_mesh((1, 1), ("fsdp", "tp"))
         with jax.set_mesh(mesh):
             model = Qwen3ForCausalLM(config, dtype=jnp.float32, rngs=nnx.Rngs(0))
         load_safetensors(tmp, config, model)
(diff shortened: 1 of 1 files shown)


skyrl-tx/tx/models/qwen3.py
```diff
@@ -182,14 +186,16 @@ class Qwen3Experts(nnx.Module):
 
     def __init__(self, config: Qwen3Config, *, dtype: jnp.dtype, rngs: nnx.Rngs) -> None:
         self.config = config
+        fsdp_shard = "dp" if config.fsdp else None
```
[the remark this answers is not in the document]
taroigarashi2001: Resolved in next commit.

skyrl-tx/tx/models/qwen3.py
```diff
@@ -43,6 +43,8 @@ def __init__(self, config: Qwen3Config, *, dtype: jnp.dtype, rngs: nnx.Rngs) ->
             assert self.num_heads % tp == 0, f"num_heads={self.num_heads} must be divisible by tp={tp}"
             assert self.num_kv_heads % tp == 0, f"num_kv_heads={self.num_kv_heads} must be divisible by tp={tp}"
         tp_shard = "tp" if shard_attention_heads else None
+        fsdp_shard = "dp" if config.fsdp else None
+        column_parallel_shard = tuple(s for s in [fsdp_shard, tp_shard] if s is not None) or None
```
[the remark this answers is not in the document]
taroigarashi2001: Ignoring since tp_shard convention was already in the codebase. Happy to change if it's unruly.

taroigarashi2001:
@gemini-code-assist review

taroigarashi2001:
One thing that I am confused about is how to implement FSDP for `LoRAExpert`. The current code implements TP/FSDP via:

```py
self.gate_proj = LoRAExpert(
    config.num_experts,
    config.hidden_size,
    config.moe_intermediate_size,
    max_lora_adapters=config.max_lora_adapters,
    max_lora_rank=config.max_lora_rank,
    dtype=dtype,
    kernel_init=nnx.with_partitioning(nnx.initializers.lecun_normal(), jax.P(None, fsdp_shard, "tp")),
    rngs=rngs,
)
```

However, I am reading [online](https://huggingface.co/blog/moe#parallelism:~:text=Expert%20parallelism%3A%20experts%20are%20placed%20on%20different%20workers.%20If%20combined%20with%20data%20parallelism%2C%20each%20core%20has%20a%20different%20expert%20and%20the%20data%20is%20partitioned%20across%20all%20cores) that the right way to implement parallelism for MoeExperts is by placing each expert on it's own GPU/TPU. Is this an inference only optimization? I could see this being difficult for the backwards pass to handle if too many tokens get routed to a single expert.

skyrl-tx/tests/models/test_qwen3_generate.py
```diff
@@ -84,20 +105,36 @@ def test_qwen3_generate():
                 hf_logprobs = torch.nn.functional.log_softmax(hf_logits, dim=-1)
                 expected_logprob = float(hf_logprobs[token_id])
 
-                assert np.isclose(our_logprob, expected_logprob, rtol=1e-3, atol=1e-3), (
+                assert np.isclose(our_logprob, expected_logprob, rtol=5e-3, atol=5e-3), (
```
taroigarashi2001: Same comment as above.

taroigarashi2001: FYI, adjusted this so this test passes on GPU. Discussed with @pcmoritz in Slack that it's expected that the kernels are a bit more accurate than CPU (only tested on CPU historically).

taroigarashi2001:
Let's make sure to squash this PR (or happy to rebase) to avoid spamming main with a bunch of commits.

skyrl-tx/tx/models/qwen3.py
```diff
@@ -182,14 +186,17 @@ class Qwen3Experts(nnx.Module):
 
     def __init__(self, config: Qwen3Config, *, dtype: jnp.dtype, rngs: nnx.Rngs) -> None:
         self.config = config
+        fsdp_shard = "dp" if config.fsdp else None
```
pcmoritz: Let's just call the axis "fsdp" (that's also what is e.g. used in https://docs.jax.dev/en/latest/the-training-cookbook.html)

Also, I'd say let's always annotate the axis as "fsdp" (and not None), and in the case where the tensors are not sharded over the fsdp axis, let's just make the "fdsp" dimension of the mesh equal to 1 (similar to how we handle "tp")

taroigarashi2001: Will fix.

taroigarashi2001: Looking at this more closely, in the current code (on main) "tp" gets set to None if `config.shard_attention_heads` is set to False, which is why I implemented it like this. 

I'll adjust the code for "fsdp" but let me know if you'd like to also change it for "tp".

taroigarashi2001:
@pcmoritz This should be ready for another review + running CI.

I wanted to echo my comment above as well (in case you also want me to fix this, I could also be misunderstanding why it's set to None in that case):

> Looking at this more closely, in the current code (on main) "tp" gets set to None if config.shard_attention_heads is set to False, which is why I implemented it like this.
>
> I'll adjust the code for "fsdp" but let me know if you'd like to also change it for "tp".

pcmoritz (approved):

PR #694, opened 2025-11-21
[skyrl-train][step-wise] 1/N - Support step-wise training with `step_wise_training` flag

SumanthRH:
# What does this PR do?

Supports step wise training natively with `step_wise_training`

Currently step-wise training introduces a new generator output format, and some custom book-keeping in the agent loop. For the first integration, we add this functionality as a separate generator. 

I plan to have a follow-up PR where we simplify this and have the logic in the base generator.

TODO:
- [x] Migrate `step_wise` example to use the same flag
- [x] Update docs
- [x] Clean up trainer

E2E Run: 

Step wise training for SkyRL-SQL:

image

Reference run: 

image

### DIFF
--- a/skyrl-train/examples/step_wise/run_skyrl_sql_step_wise.sh
+++ b/skyrl-train/examples/step_wise/run_skyrl_sql_step_wise.sh
@@ -19,7 +19,7 @@ MAX_GENERATE_LENGTH=3000
 TRAIN_BATCH_SIZE=256
 MAX_TURNS=6
 
-uv run --isolated --extra vllm --with ray==2.48.0 -m skyrl_train.entrypoints.main_base \
+uv run --isolated --extra vllm -m skyrl_train.entrypoints.main_base \
   trainer.algorithm.advantage_estimator="grpo" \
   data.train_data="['$DATA_DIR/train.parquet']" \
   data.val_data="['$DATA_DIR/validation.parquet']" \


skyrl-train/skyrl_train/generators/base.py
```diff
@@ -35,6 +38,9 @@ class GeneratorOutput(TypedDict):
     stop_reasons: Optional[List[str]]
     rollout_metrics: Optional[Dict[str, Any]]
     rollout_logprobs: Optional[List[List[float]]]
+    trajectory_ids: Optional[List[TrajectoryID]]
+    # Applicable only for step-wise training
+    is_last_step: Optional[List[bool]]
```
SumanthRH, line 43: Both the fields are optional right now since it's not a hard requirement for all generators to send this over, and only required for step wise training

skyrl-train/skyrl_train/trainer.py
```diff
@@ -685,6 +774,47 @@ def dump_data(self, data: TrainingInputBatch, file_name: str):
...
+        new_tensors = {}
+        training_input.metadata["pad_size"] = pad_size
+        if pad_size == 0:
+            return training_input
+        for key, tensor in training_input.items():
+            if tensor is not None:
+                additional_dims = tuple(tensor.shape[1:]) if len(tensor.shape) > 1 else ()
+
+                if key == "is_last_step":
+                    padding_tensor = torch.ones(pad_size, *additional_dims, dtype=tensor.dtype, device=tensor.device)
```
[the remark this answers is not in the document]
SumanthRH: This is incorrect

skyrl-train/skyrl_train/generators/step_wise_generator.py
```diff
@@ -188,6 +175,7 @@ async def agent_loop(
                 prompt_ids=copy.deepcopy(input_ids[:current_prompt_length]),
```
CharlieFRuan, line 175: This commet is for the line `response_ids = copy.deepcopy(input_ids[current_prompt_length:])`.

This `input_ids` is after we added the observation tokens and next turn's generation prompt right? Shouldn't the response IDs just be `output_ids`?

skyrl-train/skyrl_train/trainer.py
```diff
@@ -538,16 +555,27 @@ def convert_to_training_input(self, generator_output: GeneratorOutput, uids: Lis
...
+            "trajectory_ids": [trajectory_id.to_string() for trajectory_id in generator_output["trajectory_ids"]],
         }
         # padded response length
         training_input.metadata["response_length"] = response_masks_tensor.shape[1]
         training_input.metadata["avg_response_length"] = sum(
             len(sample_response_ids) for sample_response_ids in response_ids
         ) / len(response_ids)
+
+        logger.info(f"Number of sequences before padding: {len(training_input['sequences'])}")
+        training_input = self.pad_batch(training_input)
```
CharlieFRuan, line 576: Just wanted to make sure the PR doesn't break existing flow. Would this be a no-op if we're not doing step-wise training?

skyrl-train/skyrl_train/trainer.py
```diff
@@ -632,27 +679,69 @@ def compute_advantages_and_returns(self, data: TrainingInputBatch) -> TrainingIn
...
-        avg_rewards: float = return_sums.mean().item()
+        # remove padding while calculating metrics
+        pad_size = data.metadata.get("pad_size", 0)
+        num_samples = len(token_level_rewards)
+
+        return_sums = token_level_rewards.sum(dim=-1)[: num_samples - pad_size]
+        if self.cfg.trainer.step_wise_training:
+            avg_rewards: float = return_sums[data["is_last_step"][: num_samples - pad_size]].mean().item()
+        else:
+            avg_rewards: float = return_sums.mean().item()
```
CharlieFRuan, line 737: would the changes here be no-op if we're not doing step-wise training?

skyrl-train/skyrl_train/generators/step_wise_generator.py
SumanthRH: Hmm aren't both equivalent ways for step wise training? i.e you could treat (assistant response + obs, reward) as a step vs just count (assistant resp, reward) as a step?

skyrl-train/skyrl_train/trainer.py
SumanthRH: That's correct! The `pad_batch` logic is actually very similar to the initial `_remove_tail_data` logic  - if the batch is already divisible by the DP dimensions then there's no need for padding. The `pad_batch` statement is also inserted in `convert_to_training_input` after generation has fully finished. Without step wise training, in existing flow, there are two branches: with and without dynamic sampling. In both cases, the batch size should be `train_batch_size*num_prompts` (with tail data trimmed) - and padding should be zero

skyrl-train/skyrl_train/trainer.py
SumanthRH: Yes! `pad_size` is 0

skyrl-train/skyrl_train/generators/step_wise_generator.py
CharlieFRuan: True, was just wondering if this adds additional computation. Indeed I don't think the current way affects correctness

skyrl-train/skyrl_train/trainer.py
CharlieFRuan: got it, thanks for the explanation!

skyrl-train/skyrl_train/trainer.py
CharlieFRuan: Go it!

skyrl-train/skyrl_train/trainer.py
```diff
@@ -538,16 +555,27 @@ def convert_to_training_input(self, generator_output: GeneratorOutput, uids: Lis
...
+                "is_last_step": (
+                    torch.tensor(generator_output["is_last_step"], dtype=torch.bool)
+                    if generator_output.get("is_last_step", None) is not None
+                    else None
+                ),
             },
         )
         training_input.metadata = {
             "uids": uids,
+            "trajectory_ids": [trajectory_id.to_string() for trajectory_id in generator_output["trajectory_ids"]],
```
CharlieFRuan, line 567: Just realized this is breaking. We now require `trajectory_ids` to be a required field.

PR #696, opened 2025-11-21
Remove tokens after last response in agent_loop

Swynfel:
Currently, the agent_loop can return trajectories that are "too long", namely the length of `prompt_ids` + `response_ids` can be higher than `max_input_length + max_tokens`. This is because, even if the loop will be interrupted due to a "length too long" introduced in #508 , `response_ids` doesn't _just_ contain the last response generated by the model, but also the content of `new_obs`, that is unbound. Thus the env's last step can append an arbitrarily long sequence

As these last observations are not actually needed for the gradient, I fixed the issue by always removing them.

Alternatively, the loop could be reworked, for example by keeping new observation tokens in a separate object. Feel free to suggest a different approach

The PR was tested with our code, but not with the tests. It is clearly non-exhaustive as it only passed the "if" branch.


erictang000:
hey @Swynfel can you run the linter: https://skyrl.readthedocs.io/en/latest/getting-started/development.html#linting-and-formatting

and check that cpu tests are passing, otherwise this makes sense and we can merge soon. Thanks!

Swynfel:
Thanks for the documentation, the code should be properly formatted now. I also ran all cpu tests (except flash-attention ones), and indeed one test was now incorrect, so I updated it. (I also rebased on head)

PR #697, opened 2025-11-21
[skyrl-train][megatron] Remove use of PYTHONPATH for getting around transformer-engine installation

erictang000:
# Overview
Previously, we were setting the `PYTHONPATH` manually in order to point uv to the cached installation of transformer-engine in the docker image. This was hacky and could cause issues with dependencies. Now with the upgraded uv version allowing us to install transformer-engine with matched runtime with torch, and adding ninja to the base dependencies for faster build, this hack is no longer needed. Leaving the code path for backwards compatibility, but removing it from scripts and docs.

### DIFF
--- a/skyrl-train/docs/examples/megatron.rst
+++ b/skyrl-train/docs/examples/megatron.rst
@@ -90,9 +90,8 @@ Setting up the Docker image
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 To get started, you can follow the instructions for installing via Docker in the :doc:`../getting-started/installation` page, but using the ``novaskyai/skyrl-train-ray-2.51.1-py3.12-cu12.8-megatron`` image instead of the default image.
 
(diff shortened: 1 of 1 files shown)


skyrl-train/docs/examples/megatron.rst
```diff
@@ -92,19 +92,9 @@ To get started, you can follow the instructions for installing via Docker in the
 
 This ensures that the necessary dependencies needed for Megatron (i.e. ``TransformerEngine``) are installed and don't need to be built on each node for each run, which can be time consuming.
 
-Environment Variables
-~~~~~~~~~~~~~~~~~~~~~
-After following the installation instructions, set the following environment variables for the ``TransformerEngine`` dependency to be correctly picked up by the uv + ray integration (since we currently exclude it from the pyproject.toml file to avoid building it on each node):
-
-.. code-block:: bash
-
-    export SKYRL_PYTHONPATH_EXPORT=1
```
SumanthRH, line 101: Could you explicitly mention here a note that setting pythonpath is no longer needed ?

PR #716, opened 2025-11-28
Add empty_cuda_cache flag in the megatron config

Swynfel:
This PR adds a `empty_cuda_cache` flag in the megatron config of the policy to indicate that torch's cuda cache should be manually cleared between the forward/backward pass and the optimizer step

We found out that reserved but unallocated memory was high at this point, and frequent OoMs happened in the next step, likely indicating the cache wasn't automatically cleared even when needed in the optimizer (our hypothesis is because the NCCL distribution handles allocation at a low level directly and isn't compatible with torch's reservation). Manually clearing the cache at this point generally made tests and trainings more stable, and even allowed training that previous consistently threw OoM errors


Swynfel:
Just rebased on head commit!

Issue #646, opened 2025-11-09
[tx] Implement list_training_runs of the RestClient


pcmoritz:
Currently it doesn't work, e.g. start the engine and then
```python
export TINKER_API_KEY=dummy

uv run --with tinker python
Python 3.12.11 | packaged by conda-forge | (main, Jun  4 2025, 14:45:31) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import tinker
>>> service_client = tinker.ServiceClient(base_url="http://localhost:8000")
>>> rest_client = service_client.create_rest_client()
>>> rest_client.list_training_runs().result()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ray/anaconda3/lib/python3.12/concurrent/futures/_base.py", line 456, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/home/ray/anaconda3/lib/python3.12/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/home/ray/.cache/uv/archive-v0/p0jPctnnKYkJjQe2NXfiE/lib/python3.12/site-packages/tinker/lib/public_interfaces/rest_client.py", line 165, in _list_training_runs_async
    return await self.holder.execute_with_retries(_send_request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ray/.cache/uv/archive-v0/p0jPctnnKYkJjQe2NXfiE/lib/python3.12/site-packages/tinker/lib/internal_client_holder.py", line 229, in execute_with_retries
    raise e
  File "/home/ray/.cache/uv/archive-v0/p0jPctnnKYkJjQe2NXfiE/lib/python3.12/site-packages/tinker/lib/internal_client_holder.py", line 190, in execute_with_retries
    return await func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ray/.cache/uv/archive-v0/p0jPctnnKYkJjQe2NXfiE/lib/python3.12/site-packages/tinker/lib/public_interfaces/rest_client.py", line 159, in _send_request
    return await client.get(
           ^^^^^^^^^^^^^^^^^
[...]

PR #720, opened 2025-11-30
[tx] Implement `/api/v1/training_runs`

VThejas:
Based on issue #646, I've implemented the training runs API (exact same signature as the tinker version). 

~I have a question about unit tests, I don't see that the API methods have unit tests. Are there any conventions for the same? Thanks!~

Update: I've added a unit test for the functionality!


VThejas:
Hey @pcmoritz just a gentle ping requesting your review on this! Thank you very much!

PR #728, opened 2025-12-01
[skyrl-train] Move `target_modules` and `exclude_modules` into the lora config

erictang000:
Previously `target_modules` and `exclude_modules` were specified at the `trainer` config level. Moves them into the `trainer.policy.model.lora` and `trainer.critic.model.lora` configs, and marks the trainer level flags for deprecation.


tyler-griggs (approved):
thanks!

PR #730, opened 2025-12-02
Use truncated responses ids for overlong filtering instead of strings

yarinshaked:
apply_overlong_filtering method expects response_ids such it can compare the last element to a token_id. In the current code it passes strings instead of ids such that the apply_overlong_filtering method masks all of the tokens.


erictang000:
thanks for noticing this and putting this PR up @yarinshaked - I had also noticed this and fixed it in: https://github.com/NovaSky-AI/SkyRL/pull/764, so closing this for now

PR #736, opened 2025-12-03
[tx] weights_info to fix CI

agolajko:
Some tests are failing due to the api not implementing `/api/v1/weights_info`

Seems like it's been the case with previously merged PRs as well eg: https://github.com/NovaSky-AI/SkyRL/pull/701

This adds the endpoint

cc @pcmoritz @tyler-griggs


pcmoritz (approved):
Thanks a lot for fixing this! :)

PR #760, opened 2025-12-09
[tx] Optimize save_safetensors function by batching host transfers

pcmoritz:


pcmoritz:
The overhead of this is very small, so probably not worth the extra complexity :)

Issue #763, opened 2025-12-09
[skyrl-train][bug] SkyRLGymGenerator batched generation logs higher max_num_tokens than expected


erictang000:
It seems like sampling parameters are having unexpected behavior when used with generator.batched=True. 

After #652 moved 

```python
responses = truncated_responses
```

to be after 

```python
 rollout_metrics = get_rollout_metrics(responses, rewards, env_metrics, env_classes)
```

we are seeing the max number of tokens be higher than `generator.max_generate_length`. We can see this on the nightly gsm8k CI, which should have max_num_tokens=1024 (https://github.com/NovaSky-AI/SkyRL/blob/e3e8dab7308f5bc4b78d2f441fc817bb8c4fdd8b/skyrl-train/examples/gsm8k/run_gsm8k.sh#L43)

<img width="1639" height="630" alt="Image" src="https://github.com/user-attachments/assets/33a35a81-cb40-42c6-8b7a-505065a59dec" />

This has not caused correctness issues, since we truncate overlong responses, however the metrics are off.

erictang000:
closed by #764

PR #764, opened 2025-12-10
[skyrl-train][bugfix] fix bug causing metrics + overlong filtering to use strings instead of token ids for batched generation

erictang000:
Fixes bug described in #763. Adds check to GPU CI to check rollout metrics to catch this in the future

Pink metrics are with the fix, green is before.

image

### DIFF
--- a/skyrl-train/skyrl_train/generators/utils.py
+++ b/skyrl-train/skyrl_train/generators/utils.py
@@ -248,9 +248,6 @@ def get_rollout_metrics(
     Returns:
         Dictionary of aggregated metrics
     """
-    assert (
-        isinstance(responses, list) and isinstance(responses[0], list) and isinstance(responses[0][0], int)
-    ), "responses must be a list of list of token ids"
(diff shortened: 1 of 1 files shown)


skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -376,20 +376,18 @@ async def generate_batched(
...
+            env_step_output: BaseTextEnvStepOutput = await self._run_in_executor_if_available(env.step, output)
             reward = env_step_output["reward"]
             rewards.append(reward)
 
-            if len(response_ids) > max_tokens:
-                response_ids = response_ids[:max_tokens]
-            loss_masks.append([1] * len(response_ids))
-            truncated_responses.append(response_ids)
+            if len(response) > max_tokens:
+                response = response[:max_tokens]
```
CharlieFRuan, line 386: @gemini-code-assist does this in-place update the `responses` which will be used later?

PR #777, opened 2025-12-12
[tx] (Not for merging) Test cudnn backend for attention implementation

pcmoritz:
Since https://docs.jax.dev/en/latest/_autosummary/jax.nn.dot_product_attention.html supports the cudnn flash attention implementation, I was curious how the performance compares and whether we leave performance on the table by not using it. I ran it with
```
uv run --extra aws --extra gpu --extra tinker -m tx.tinker.api --base-model Qwen/Qwen3-4B --max-lora-adapters 3 --max-lora-rank 1 --tensor-parallel-size 4 --train-micro-batch-size 8 --sample-max-num-sequences 256 > out.log
```
and
```
uv run --with Pillow --with wandb --with tinker rl_loop.py base_url=http://localhost:8000/ model_name="Qwen/Qwen3-4B" lora_rank=1 max_length=1024 max_tokens=512 save_every=100
```

The timing with cudnn is slightly worse than the XLA timing, so there is no point in merging this PR, but I wanted to open it for documentation purposes.

Screenshot 2025-12-12 at 2 01 27 PM

### DIFF
--- a/skyrl-tx/tx/models/qwen3.py
+++ b/skyrl-tx/tx/models/qwen3.py
@@ -122,12 +122,19 @@ class Qwen3Attention(nnx.Module):
         updated_cache = (k, v)
 
         # Attention (causal only during prefill, GQA handled natively by dot_product_attention)
+        # Expand mask to [B, 1, Q_len, KV_len] for cuDNN compatibility (no broadcasting)
+        Q_len = q.shape[1]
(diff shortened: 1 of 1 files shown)


pcmoritz:
This is replaced by https://github.com/NovaSky-AI/SkyRL/pull/879

PR #778, opened 2025-12-15
[tx] Implement string stop sequences

pcmoritz:
In order to be able to run the rl_loop.py with https://github.com/NovaSky-AI/SkyRL/pull/657, we need to implement string stop sequences.

### DIFF
--- a/skyrl-tx/tx/utils/generator.py
+++ b/skyrl-tx/tx/utils/generator.py
@@ -103,7 +103,9 @@ def find_string_stop_position(
     stream = DecodeStream(skip_special_tokens=False)
     text = ""
     for i, token in enumerate(tokens):
-        text += stream.step(tokenizer._tokenizer, token)
+        chunk = stream.step(tokenizer._tokenizer, token)
+        if chunk is not None:
+            text += chunk
         for stop_string in stop_strings:
             if stop_string in text:
                 return i + 1


skyrl-tx/tx/utils/generator.py
```diff
@@ -77,6 +78,41 @@ def compute_positions(attention_mask: jax.Array) -> jax.Array:
...
+    # Incremental decode using DecodeStream
+    stream = DecodeStream(skip_special_tokens=False)
+    text = ""
+    for i, token in enumerate(tokens):
+        chunk = stream.step(tokenizer._tokenizer, token)
+        if chunk is not None:
+            text += chunk
+        for stop_string in stop_strings:
+            if stop_string in text:
+                return i + 1
```
tyler-griggs, line 111: I *think* it's possible that this can return extra characters, but I'm not sure what the desired behavior is. Namely, we are checking the stop condition by string/characters, but returning the index into the tokens. E.g., if the stop string is "stop" but the final token decodes to "stop!", then this code will return "stop!" instead of "stop". But I haven't referenced other implementations and whether they include the last full token or not, so maybe this is expected behavior

pcmoritz: Thanks for pointing this out -- I tried it now and Tinker will return the last full token, even if it generated a larger string (which makes a lot of sense I think). This is the test case:
```python
#!/usr/bin/env python3
"""
Minimal test: stop on '..' when '...' is a single token.
Shows that Tinker returns the full token even if stop string is a substring.
"""

from tinker import ServiceClient, types
from transformers import AutoTokenizer

MODEL = "Qwen/Qwen3-8B"
tokenizer = AutoTokenizer.from_pretrained(MODEL, trust_remote_code=True)
client = ServiceClient().create_sampling_client(base_model=MODEL)

# Show tokenization
print("Tokenization:")
print(f"  '..'  -> {tokenizer.encode('..', add_special_tokens=False)}")
print(f"  '...' -> {tokenizer.encode('...', add_special_tokens=False)}")

# Sample with stop=".."
prompt = "End with ellipsis: thinking"
result = client.sample(
    prompt=types.ModelInput.from_ints(tokenizer.encode(prompt, add_special_tokens=False)),
    sampling_params=types.SamplingParams(max_tokens=30, temperature=0.0, stop=[".."]),
    num_samples=1,
).result()

seq = result.sequences[0]
decoded = tokenizer.decode(seq.tokens)

print(f"\nStop string: '..'")
print(f"Output tokens: {seq.tokens}")
print(f"Decoded: {repr(decoded)}")
```
The output is
```
Tokenization:
  '..'  -> [496]
  '...' -> [1112]

Stop string: '..'
Output tokens: [311, 6133, 979, 498, 2299, 10507, 4849, 4325, 11, 892, 646, 387, 264, 1841, 315, 458, 61845, 19984, 1707, 476, 458, 1112]
Decoded: " to yourself when you're idealizing someone, which can be a sign of an insecure attachment style or an..."
```

PR #783, opened 2025-12-16
[tx] Factor out LoRA routing

pcmoritz:
While this doesn't have impact on the performance (since XLA is already smart enough to do this optimization), it drastically reduces the compile time of the model if we use FSDP, so this is a preparation for merging FSDP support.


pcmoritz:
This turns out to not be needed any more

Issue #566, opened 2025-10-24
[skyrl-train] Flexible weight syncing support


SumanthRH:
# Summary

The current weight syncing logic in SkyRL-Train is straightforward: policy worker classes implement a `broadcast_to_inference_engines` method, while the inference engine worker classes implement an `update_weights`  method for receiving weights from the trainer (Ex: vLLM workers implement `update_weights_cuda_ipc` for CUDA IPC transport, and `update_weights` for non-colocated case).  However this is inflexible and we want to allow for different weight syncing strategies:

1. Support different communication backends for weight syncing: For example, allowing users to use the Ray Direct Transport APIs for weight syncing https://docs.ray.io/en/latest/ray-core/direct-transport.html for weight transfer in lieu of regular torch distributed. 
2. Support different weight syncing implementations: There are cases where we’d want to have multiple weight syncing implementations and allow the user to choose one depending on their setup. For example, an optimized weight sync implementation for RDMA enabled clusters, or one targeting quantized rollouts. This might also be training framework specific (ex: Megatron only)

PR #785, opened 2025-12-16
[train] Weight Sync Interface 3/3: transfer strategies and final integration

raulchen:
This is the final PR of refactoring the weight sync interface. 
Linked issue: https://github.com/NovaSky-AI/SkyRL/issues/566
Design doc: https://docs.google.com/document/d/17UeCgTfEeG0jXcES4Z7pc2u8_zXQ6oUN_B1eZTcKeD0/edit?tab=t.6ickfkx55idl

Key changes

  - Introduce strategy pattern for weight transfer between training workers and inference engines. Replacing ad-hoc type checking with a clean,  extensible architecture.
  - Add WeightTransferStrategy, WeightTransferSender, and WeightTransferReceiver abstractions.
  - Implement BroadcastTransferStrategy and CudaIpcTransferStrategy as concrete strategies.
  - Consolidate weight sync logic from workers into reusable strategy classes. 
     - FSDP/DeepSpeed worker now supports packing tensors for the cuda ipc path as Megatron already does. 
  - Replace use_cuda_ipc flag with automatic strategy selection based on config.
  - Add typed dataclass-based request objects (BroadcastWeightUpdateRequest, CudaIpcWeightUpdateRequest).

  Architecture

```
  WeightTransferStrategy (factory)
  ├── create_init_info(cfg) -> WeightSyncInitInfo
  ├── create_sender(init_info, client) -> WeightTransferSender
  └── create_receiver(init_info) -> WeightTransferReceiver

  Implementations:
  ├── BroadcastTransferStrategy (NCCL broadcast for distributed setups)
  └── CudaIpcTransferStrategy (CUDA IPC for colocated setups)

  Sender side (training worker):
  strategy_cls = get_transfer_strategy_cls(cfg)
  init_info = strategy_cls.create_init_info(cfg)
  sender = strategy_cls.create_sender(init_info, inference_client)
  await sender.send_chunks(chunks)

  Receiver side (inference engine):
  strategy_cls = init_info.strategy_type()
  receiver = strategy_cls.create_receiver(init_info)
  for name, tensor in receiver.receive_weights(request):
      model.load_weights([(name, tensor)])
```
  
  Test Plan

  - Existing CPU weight_sync tests
  - Added new CPU tests for transfer strategies. 
  - GPU integration test: tests/gpu/gpu_ci/test_transfer_strategies_e2e.py
  - Manual E2E tests: 
    - test_policy_local_engines_e2e.py
    - test_megatron_worker.py
    - run_gsm8k.sh
    - remote inference engine test

### DIFF
--- a/skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
+++ b/skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
@@ -88,6 +88,7 @@ class WorkerWrap:
         strategy_cls = init_info.strategy_type()
 
         if hasattr(self, "_weight_receiver") and self._weight_receiver is not None:
+            # TODO(haochen): we should get rid of this flag and override existing receiver.
             if init_info.override_existing_receiver:
                 self._weight_receiver.teardown()
                 self._weight_receiver = None
--- a/skyrl-train/skyrl_train/inference_engines/vllm/vllm_server.py
+++ b/skyrl-train/skyrl_train/inference_engines/vllm/vllm_server.py
@@ -95,6 +95,11 @@ class VllmServer:
             import pickle
             from skyrl_train.weight_sync import BroadcastWeightUpdateRequest
 
+            # Convert the HTTP request to a BroadcastWeightUpdateRequest
+            # TODO(haochen): only the broadcast strategy is currently supported
+            # for the remote inference engine path.
+            # To support other strategies, we'll need to add a "strategy=xxx"
+            # parameter in the HTTP request.
             data = await request.json()
             weight_request = BroadcastWeightUpdateRequest(**data)
 


skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
```diff
@@ -72,62 +71,49 @@ def test_rpc(self, *args, **kwargs):
...
-                destroy_process_group(self._model_update_group)
-                self._model_update_group = None
+        # Unpickle init_info to restore the original object type
+        assert isinstance(init_info, bytes), f"Expected bytes, got {type(init_info).__name__}"
+        init_info = pickle.loads(init_info)
+
+        strategy_cls = init_info.strategy_type()
+
+        if hasattr(self, "_weight_receiver") and self._weight_receiver is not None:
+            if init_info.override_existing_receiver:
```
raulchen, line 92: @SumanthRH is there any reason we don't want to override the existing group?

skyrl-train/skyrl_train/inference_engines/vllm/vllm_server.py
```diff
@@ -101,24 +92,26 @@ async def _reset_prefix_cache(request: Request):
 
         @app.post("/update_weights")
         async def _update_weights(request: Request):
+            import pickle
+            from skyrl_train.weight_sync import BroadcastWeightUpdateRequest
```
SumanthRH, line 96: The current approach requires the server to know about the specific weight sync strategy used. This seems to be because we are using pickle to get around collective_rpc's limitation with serializing dataclasses. 

Instead of the current pickling based approach, what about:

1. Trainer communicates with engine and server with dicts. So something like `inference_engine_client.update_weights(init_info=asdict(UpdateWeightRequest(....))`
2. Server and Engine don't know about the specific strategy dataclasses and simply forwards the request to the worker 
3. The current way is that `init_info` has the strategy class information. We can invert the relationship - the strategy class has attributes `init_info_cls: Type[WeightSyncInitInfo]` and `update_info_cls: Type[WeightUpdateRequest]` that store the dataclass for the appropriate request. 

```python
class WeightTransferStrategy:
     # to be set by subclasses
     init_info_cls: Type[WeightSyncInitInfo]
     update_info_cls: Type[WeightUpdateRequest]
```

3. At the worker, we do the following:
a. Initialize weight receiver with a special `init_weight_receiver` method that is called by the engine at init time. Engine will pass the strategy name that can be used with `get_transfer_strategy_cls(name)`. 
b. The worker methods `init_weight_update_communicator` and `update_weights` will read the request and instantiate the appropriate dataclass with `self.weight_sync_strategy_cls.init_info_cls(**request)`

This way, 
1. the strategy class implementations like `WeighTransferReceiver` can be typed with dataclasses
2. we still work well with vLLM's serialization
3. server and engine implementation is weight sync strategy agnostic

I have something similar here: https://github.com/vllm-project/vllm/compare/v0.11.2...SumanthRH:vllm:weight_transfer_prototype

Wdyt?

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
SumanthRH: Can't think of any! This was mostly left as an escape hatch, we can get rid of this parameter (and always override) in another PR

raulchen: SG

skyrl-train/skyrl_train/inference_engines/vllm/vllm_server.py
raulchen: oh, actually, only the remote inference engine path explicitly deals with the broadcast strategy. 
The regular vLLM engine path transparently pickles and passes the init_info to the WorkerWrap.
The reason is because the remote engine path only supports the Broadcast strategy now, which I think is because cuda ipc doesn't make sense as the engine is remote. 
If we want to support other strategies in the future, what we need is to add a "strategy=xxx" param in the HTTP request. 
The pickle serialization should already be compatible. 
I can add a TODO here.

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
raulchen: added a todo

PR #792, opened 2025-12-18
[skyrl-train][mcore] upgrade megatron to 0.15.0 and vllm for mcore to 0.11.0

erictang000:
Bumping dependencies for megatron-core + vllm

Ran into some issues with megatron==0.15.0 not importing correctly on cpu machines - raised an issue here: https://github.com/NVIDIA/Megatron-LM/issues/2715. We can try pinning to 0.15.0.rc6 for now, while these dependency issues are fixed.

Verified all megatron tests are passing.

### DIFF
--- a/skyrl-train/tests/gpu/gpu_ci/test_trainer_full_checkpointing.py
+++ b/skyrl-train/tests/gpu/gpu_ci/test_trainer_full_checkpointing.py
@@ -9,7 +9,7 @@ For FSDP and DeepSpeed, run:
 uv run --isolated --extra dev --extra deepspeed --extra vllm pytest tests/gpu/gpu_ci/test_trainer_full_checkpointing.py -m "not megatron"
 
 For Megatron, run:
-uv run --isolated --extra dev --extra mcore --extra vllm pytest tests/gpu/gpu_ci/test_trainer_full_checkpointing.py -m "megatron"
+uv run --isolated --extra dev --extra mcore pytest tests/gpu/gpu_ci/test_trainer_full_checkpointing.py -m "megatron"
 """
 
 import ray


skyrl-train/pyproject.toml
```diff
@@ -71,12 +73,19 @@ conflicts = [
 no-build-isolation-package = [
     "transformer-engine-torch",
     "transformer-engine",
+    "nv-grouped-gemm",
+]
+override-dependencies = [
+    "nvidia-resiliency-ext; sys_platform == 'never'",
+    "mamba-ssm; sys_platform == 'never'",
+    "causal-conv1d; sys_platform == 'never'",
+    "transformer-engine[pytorch]==2.9.0",
```
erictang000, line 82: megatron-bridge attempts to pin a specific github revision for transformer-engine, so we need to override it to not build from source (which is very expensive)

skyrl-train/pyproject.toml
```diff
@@ -71,12 +73,19 @@ conflicts = [
...
+    "megatron-core==0.15.0"
```
erictang000, line 83: similarly, megatron-bridge attempts to pin a dev branch of megatron-core that's not stable.

skyrl-train/pyproject.toml
```diff
@@ -71,12 +73,19 @@ conflicts = [
 no-build-isolation-package = [
     "transformer-engine-torch",
     "transformer-engine",
+    "nv-grouped-gemm",
+]
+override-dependencies = [
+    "nvidia-resiliency-ext; sys_platform == 'never'",
```
erictang000, line 79: currently disabling these sub-dependencies of megatron-bridge since we are not utilizing fault tolerance or hybrid model features in SkyRL, and these add significant installation build overhead and dependency chains

PR #798, opened 2025-12-22
[Tinker-ify 1/N] Refactor training worker to implement `forward_backward` and `optim_step`

tyler-griggs:
### Overview
Split training_step into forward_backward and optim_step

### Summary
Refactors PolicyWorkerBase and CriticWorkerBase to decompose the monolithic training_step method into two composable methods:
`forward_backward(experience, accumulation_steps)` - handles forward pass, loss computation, and backward pass
`optim_step(local_step, accumulation_steps)` - handles gradient accumulation check and optimizer step

Note:
* Updated ppo_train() to use the new methods directly instead of calling training_step()
* Updated training_step() to compose the new methods (tests depend on this, but will be removed
* No functional changes

### DIFF
--- a/skyrl-train/skyrl_train/workers/worker.py
+++ b/skyrl-train/skyrl_train/workers/worker.py
@@ -724,15 +724,13 @@ class PolicyWorkerBase(Worker):
 
         return status
 
-    def optim_step(self, local_step: int, accumulation_steps: int) -> Optional[float]:
+    def optim_step(self) -> float:
         """
-        Perform optimizer step after accumulating gradients.
+        Perform optimizer step and return the gradient norm.
         """
-        grad_norm = None
-        if (local_step + 1) % accumulation_steps == 0:
-            grad_norm = self.strategy.optimizer_step(self.optimizer, self.model, self.scheduler, name="actor")
-            if grad_norm is not None:
-                grad_norm = grad_norm.detach().cpu().item()
+        grad_norm = self.strategy.optimizer_step(self.optimizer, self.model, self.scheduler, name="actor")
+        if grad_norm is not None:
+            grad_norm = grad_norm.detach().cpu().item()
         return grad_norm
 
     def ppo_train(self, train_data: TrainingInputBatch) -> TrainingOutputBatch:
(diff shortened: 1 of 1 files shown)


skyrl-train/skyrl_train/workers/worker.py
```diff
@@ -728,103 +831,22 @@ def training_step(self, experience: Experience, global_step, local_step, accumul
...
-            "policy_entropy": entropy.item(),
-        }
-        if self.cfg.trainer.algorithm.use_kl_loss:
-            status["policy_kl"] = kl_loss.item()
-
+        status["policy_lr"] = self.scheduler.get_last_lr()[0]
         if grad_norm is not None:
             status["raw_grad_norm"] = grad_norm
 
         for k, v in experience.info.items():
```
erictang000: i think this whole for loop block is actually legacy code and we should remove it entirely!

skyrl-train/skyrl_train/workers/worker.py
```diff
@@ -642,6 +642,99 @@ def _normalize_mini_batch_size(self):
...
+    def optim_step(self, local_step: int, accumulation_steps: int) -> Optional[float]:
+        """
+        Perform optimizer step after accumulating gradients.
+        """
+        grad_norm = None
+        if (local_step + 1) % accumulation_steps == 0:
+            grad_norm = self.strategy.optimizer_step(self.optimizer, self.model, self.scheduler, name="actor")
+            if grad_norm is not None:
+                grad_norm = grad_norm.detach().cpu().item()
+        return grad_norm
```
SumanthRH: This should probably be just `optim_step` without any args:

```
def optim_step() -> float:
      """Perform optimizer step and returns the gradient norm"""
```

And then the training loop in `ppo_train` will call it only after accumulation_steps:

```
                status = self.forward_backward(experience, accumulation_steps)
                if (local_step + 1) % accumulation_steps == 0: 
                       grad_norm = self.optim_step()
```

skyrl-train/skyrl_train/workers/worker.py
tyler-griggs: Oh can you say more about this? Do you mean we are not actually using these fields returned in the status so we can just remove the updates the for loop makes to the `experience` object?

tyler-griggs: Oh nevermind, I see it's just not used/populated, okay will remove

erictang000: yeah i think this was like the very very old way we got the kl, i think we don't update experience.info anywhere in the code anymore. If you check it's always just an empty dictionary

skyrl-train/skyrl_train/workers/worker.py
tyler-griggs: Yes

PR #799, opened 2025-12-22
[AsyncRL][N/N] Support TIS to fully async and multi-turn SkyRLGymGenerator

CharlieFRuan:
This PR supports TIS for fully async RL and thus conclude the support for fully async RL in SkyRL.

Besides, with this PR, we can use fully async for any task supported in `SkyRLGymGenerator`, and hence we can remove the `SkyRLGymHTTPGenerator`.

We also support TIS for mutli-turn generation in `SkyRLGymGenerator` (async or not) -- this was already supported, but just needed some config tuning.

Separately, we fix a bug in computing `expected_consumed_in_epoch` when we are at the boundary of an epoch.

### Tests

We validate multi-turn TIS with Search r1, with the curve comparison below. We indeed see some small improvement with TIS.
image

Also tested fully async DAPO w/ TIS, will add the script and results in a follow up PR.

### DIFF
--- a/skyrl-train/skyrl_train/generators/utils.py
+++ b/skyrl-train/skyrl_train/generators/utils.py
@@ -44,23 +44,6 @@ CUSTOM_CHAT_TEMPLATES = {
         "{% endif %}"
         "{% endfor %}"
     ),
-    # Qwen2.5 chat template but with `generation` and `endgeneration` tags, and simplified
-    "qwen2_5_with_generation_tag_simplified": (
-        "{% for message in messages %}"
-        "{% if (message.role == 'user') or (message.role == 'system' and not loop.first) %}"
-        "{{ '<|im_start|>' + message.role + '\n' + message.content + '<|im_end|>' + '\n' }}"
-        "{% elif message.role == 'assistant' %}"
-        "{{ '<|im_start|>' + message.role + '\n'}}"
-        "{% generation %}"
-        "{{ message.content + '<|im_end|>'}}"
-        "{% endgeneration %}"
-        "{{ '\n' }}"
-        "{% endif %}"
-        "{% endfor %}"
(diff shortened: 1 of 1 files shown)


skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -896,7 +895,7 @@ def _update_agent_loop_state_with_multiturn_chat_template(
             agent_loop_state.response_end_idx = len(agent_loop_state.input_ids) + len(turn_output.output_ids) - 1
             agent_loop_state.input_ids += turn_ids
             agent_loop_state.loss_mask += loss_mask_for_turn
-            if agent_loop_state.rollout_logprobs is not None:
+            if agent_loop_state.rollout_logprobs is not None and rollout_logprobs_for_turn is not None:
```
CharlieFRuan, line 898: Adding the second condition because when we eval, `agent_loop_state.rollout_logprobs` is not `None` (since we use the `generator_config.sampling_params`) but `rollout_logprobs_for_turn` is `None` since eval does not return logprobs

skyrl-train/skyrl_train/generators/skyrl_gym_generator.py
```diff
@@ -435,6 +432,8 @@ async def agent_loop(
             assert response_ids is not None and loss_mask is not None
             if stop_reason != "length" and response_ids and response_ids[-1] != self.tokenizer.eos_token_id:
                 response_ids.append(self.tokenizer.eos_token_id)
+                # TODO(Charlie): this should be 0? Otherwise logprobs will be extremely off. But if it is loss
```
CharlieFRuan, line 435: Come back to this in a future PR

PR #810, opened 2025-12-29
[tx] Implement multi-node support

pcmoritz:
This makes it possible to run SkyRL tx on multiple nodes. Rank 0 will run the full engine (including the database operations), and the other ranks will only execute the jax operations like forward_backward and sampling.

This can e.g. be run on a single node like
```
CUDA_VISIBLE_DEVICES=0,1,2,3 uv run --extra aws --extra gpu --extra tinker -m tx.tinker.api --base-model Qwen/Qwen3-8B --backend-config '{"max_lora_adapters": 3, "max_lora_rank": 1, "tensor_parallel_size": 4, "fully_sharded_data_parallel_size": 2, "train_micro_batch_size": 8, "sample_max_num_sequences": 256, "coordinator_address": "localhost:7777", "num_processes": 2}' > out1.log
```

and

```
CUDA_VISIBLE_DEVICES=4,5,6,7 uv run --extra aws --extra gpu --extra tinker -m tx.tinker.backends.jax --coordinator-address "localhost:7777" --num-processes 2 --process-id 1 > out2.log
```

### DIFF
--- a/skyrl-tx/tx/tinker/backends/jax.py
+++ b/skyrl-tx/tx/tinker/backends/jax.py
@@ -53,7 +53,6 @@ from tx.utils.models import (
     round_up_seq_len,
     resolve_model_path,
 )
-from tx.utils.storage import pack_and_upload, download_and_unpack
 from tx.utils.log import logger
 
 


skyrl-tx/tests/tinker/test_engine.py
```diff
@@ -14,6 +14,13 @@
 BASE_MODEL = "trl-internal-testing/tiny-Qwen3ForCausalLM"
 
 
+def make_engine(config: EngineConfig) -> TinkerEngine:
+    """Create TinkerEngine with JaxBackendImpl directly for test access to internals."""
+    engine = TinkerEngine(config)
+    engine.backend = engine.backend._backend
```
pcmoritz: I created an issue to fix this in https://github.com/NovaSky-AI/SkyRL/issues/819

skyrl-tx/tx/tinker/backends/backend.py
```diff
@@ -142,17 +142,6 @@ def save_sampler_checkpoint(self, output_path, model_id: str) -> None:
         """
         pass
 
-    @abstractmethod
-    def load_sampler_checkpoint(self, model_id: str, checkpoint_id: str, checkpoint_path) -> None:
```
pcmoritz, line 146: This is only called internally in JaxBackend in sample, so not part of the public interface (i.e. the functions that the engine needs to do its job)

Issue #812, opened 2025-12-30
SkyRL Refactor to the Tinker API


tyler-griggs:
## SkyRL Refactor to the Tinker API

Migrating `skyrl-train` and `skyrl-tx` -> `skyrl`: https://github.com/NovaSky-AI/SkyRL/issues/1145

tl;dr The SkyRL team is kicking off a refactor to make **SkyRL fully Tinker-compatible**.  We gladly welcome contributions to this effort! If you are interested, please comment in this issue or message us in the [slack workspace](https://join.slack.com/t/skyrl/shared_invite/zt-3f6ncn5b8-QawzK3uks6ka3KWoLwsi5Q).

### Overview

Tinker is a simple training API [introduced](https://thinkingmachines.ai/tinker/) by Thinking Machines Lab that nicely separates algorithm logic from infrastructure logic. We've spent quite a lot of time testing out the Tinker API through our **SkyRL tx** [project](https://github.com/NovaSky-AI/SkyRL/tree/main/skyrl-tx), and we’ve found the Tinker API to be compelling for a few key reasons:

- **API at the right leveI**: Tinker’s API surface is small and simple (`forward_backward`, `optim_step`, `sample`, checkpointing) and we’ve found it very effective at pushing all infrastructure logic below the API (worker onloading/offloading, micro-batching, gradient management) while leaving users with full algorithm and dataflow control above the API surface.
- **Flexibility**: Tinker’s low-level API of `forward_backward` and `optim_step` supports any weight-updating algorithms, allowing for flexible adaption to how researchers and practitioners want to perform post-training. In contrast, SkyRL’s training worker APIs are currently specific to algorithms like GRPO and PPO (such as explicit worker roles of Policy, Reference, etc.). Moving to the Tinker API allows the underlying worker implementations to also easily support SFT, DPO, or other weight-updating algorithms. 
- **Standardization & network effects**: We believe Tinker can become a widely-adopted standard for post-training, akin to OpenAI API’s adoption for inference, and are excited about the network effects this can unlock in open source. Today, it is difficult to adapt code written for one post-training framework to run effectively on another. In a future where more post-training stack developers implement Tinker backends and more researchers build atop the Tinker API, training scripts and libraries can be shared and adopted more easily across the community.

### Our goal
- **Full Tinker API compatibility**: Training scripts that target the Tinker API and can run on Thinking Machines’ hosted service will _run without any code changes on your own hardware using SkyRL's `skyrl-train` library._
[...]

PR #813, opened 2025-12-30
[tx] Add weight_decay to AdamParams for Tinker API compatibility

johnkitaoka:
Adds `weight_decay` parameter to `AdamParams` to match Tinker SDK expectations. Without this change, `weight_decay` values passed from Tinker SDK scripts are silently ignored by Pydantic's default `extra="ignore"` behavior. The optimizer runs with `weight_decay=0` regardless of user intent. 

Contributes to #812

### DIFF
--- a/skyrl-tx/tests/tinker/test_engine.py
+++ b/skyrl-tx/tests/tinker/test_engine.py
@@ -240,7 +240,7 @@ def test_process_optim_step_hyperparams_behavior():
         return float(optax.global_norm(delta))
 
     tiny_request = types.OptimStepInput(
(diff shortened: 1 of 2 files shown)


johnkitaoka:
Updated with the change in `skyrl-tx/tx/tinker/api.py` !

pcmoritz (approved):
Thanks a lot for the contribution! I pushed a small fix that removes the default from the type in types.py, it is better to not have defaults for these (where it makes sense) and only have the defaults at the API layer, so no default can accidentally sneak in.

PR #827, opened 2026-01-02
Remove deepspeed

tyler-griggs:
We are officially deprecating the DeepSpeed training backend in SkyRL. 

**If you are currently using DeepSpeed and need guidance migrating to FSDP or Megatron, please reach out via issues or in slack, we’re happy to help.**

SkyRL currently supports three training backends: DeepSpeed, FSDP, and Megatron. The dev effort to ensure feature parity, maintain rigorous testing, and properly dogfood all three backends has grown quite high, and the majority of our users are not currently using DeepSpeed. SkyRL v0.3.1 is the final release that includes DeepSpeed support, after which we will narrow our focus on making FSDP and Megatron pleasant and effective to use.


skyrl-train/skyrl_train/distributed/ulysses/utils.py
```diff
@@ -1,5 +1,3 @@
-# DeepSpeed Ulysses Sequence Parallelism borrowed from VERL
```
erictang000, line 1: we should probably keep this attribution

tyler-griggs: Claude code got a bit aggressive.. reverting :D

Issue #841, opened 2026-01-05
[tx] Properly clean up adapters that are not used any more


pcmoritz:
Currently, new model adapter indices are allocated in https://github.com/NovaSky-AI/SkyRL/blob/c07e8088c3a38339ba8ef6b732b3d11d273fbf9d/skyrl-tx/tx/tinker/backends/jax.py#L420 but never freed. If a client disconnects and we don't get health checks from it any more, we can free up the adapter index so it can be used by another client.

One possible way to implement this is to introduce a new method `delete_model` to the AbstractBackend and handle it properly in the JaxBackend.

A good end-to-end check on whether this works is to start an engine with a small max_lora_adapters, do a training run, and then after the training run finishes run it again and make sure it can be run again.

This was also pointed out by @OhadRubin in https://github.com/NovaSky-AI/SkyRL/pull/787#issuecomment-3693944111

Issue #819, opened 2025-12-31
[tx] Refactor test_engine.py to be test_backend.py


pcmoritz:
While working on https://github.com/NovaSky-AI/SkyRL/pull/810, I noticed that test_engine is really more of a test_backend test after https://github.com/NovaSky-AI/SkyRL/pull/787 got merged. We should refactor it to be a test of JaxBackend directly (or maybe AbstractBackend if it makes sense, but I think for most tests probably currently it doesn't).

pcmoritz:
Fixed by https://github.com/NovaSky-AI/SkyRL/issues/819

PR #844, opened 2026-01-06
[tx] implement model unloading

raulchen:
Closes https://github.com/NovaSky-AI/SkyRL/issues/841

This PR implements model unloading (manual via `unload_model` or automatic on session timeout). 

# Changes

New API endpoint:
- POST /api/v1/unload_model - unloads a model and frees its LoRA adapter slot

Automatic garbage collection:
- Engine periodically checks for sessions with stale heartbeats (configurable via --session-cleanup-interval-sec and --sessio
n-timeout-sec)
- Models from expired sessions are automatically unloaded, freeing adapter slots for reuse

Backend changes:
- AbstractBackend.delete_model() - new abstract method
- JaxBackend.delete_model() - clears LoRA weights and frees adapter index
- clear_adapter_config() - zeros out lora_A, lora_B, scaling, and rank for an adapter
- Adapter index allocation now reuses freed slots instead of always incrementing

# Test plan

- Integration tests in `test_api.py`
- unit tests in "test_engine.py" and "test_jax_backend.py"

### DIFF
--- a/skyrl-tx/tests/tinker/test_jax_backend.py
+++ b/skyrl-tx/tests/tinker/test_jax_backend.py
@@ -50,23 +50,18 @@ def test_adapter_slot_reuse():
     """Test that deleted adapter slots are reused."""
     backend = create_backend()
 
-    # Create 3 models
-    idx1 = create_model(backend, "model_1")
-    idx2 = create_model(backend, "model_2")
-    idx3 = create_model(backend, "model_3")
-    assert idx1 == 1
-    assert idx2 == 2
-    assert idx3 == 3
+    # Create 3 models and check adapter indices
+    assert create_model(backend, "model_1") == 1
+    assert create_model(backend, "model_2") == 2
+    assert create_model(backend, "model_3") == 3
 
     # Delete first model, new model should reuse index 1
     backend.delete_model("model_1")
-    idx4 = create_model(backend, "model_4")
-    assert idx4 == 1
+    assert create_model(backend, "model_4") == 1
 
     # Delete middle model, new model should fill gap at index 1
     backend.delete_model("model_2")
-    idx5 = create_model(backend, "model_5")
-    assert idx5 == 2
+    assert create_model(backend, "model_5") == 2
 
 
 def test_max_adapters_limit():
@@ -94,9 +89,8 @@ def test_max_adapters_after_delete():
     # Delete one model
     backend.delete_model("model_0")
 
-    # Now we should be able to create a new model
-    idx = create_model(backend, "model_new")
-    assert idx == 1  # Should reuse the freed slot
+    # Now we should be able to create a new model which should reuse the freed slot
+    assert create_model(backend, "model_new") == 1
 
 
 def test_clear_adapter_config():
--- a/skyrl-tx/tests/tinker/test_api.py
+++ b/skyrl-tx/tests/tinker/test_api.py
@@ -352,7 +352,7 @@ def test_unload_model(api_server):
     # Create a training client (which creates a model) and verify it works
     _, training_client = create_service_and_training_client(base_url=f"http://0.0.0.0:{TEST_SERVER_PORT}/")
 
-    async def _unload():
+    async def unload_model():
         async with tinker._client.AsyncTinker(
             api_key="dummy", base_url=f"http://0.0.0.0:{TEST_SERVER_PORT}/"
         ) as client:
@@ -365,8 +365,7 @@ def test_unload_model(api_server):
                     return result
                 await asyncio.sleep(0.1)
 
-    result = asyncio.run(_unload())
-    assert isinstance(result, types.UnloadModelResponse)
+    assert isinstance(asyncio.run(unload_model()), types.UnloadModelResponse)
 
     # Verify model no longer works after unload
     with pytest.raises(Exception):
--- a/skyrl-tx/tests/tinker/test_engine.py
+++ b/skyrl-tx/tests/tinker/test_engine.py
@@ -579,8 +579,6 @@ def test_cleanup_stale_sessions():
         )
         session.commit()
 
-    # Run cleanup
-    unloaded_count = engine.cleanup_stale_sessions()
-
-    assert unloaded_count == 1
+    # Run cleanup and assert one model was unloaded
+    assert engine.cleanup_stale_sessions() == 1
     assert not engine.backend.has_model(model_id)


pcmoritz:
Thanks a lot for implementing this! Some suggestions before merging this:

- Remove the manual unload API call -- I think it is too early to think about diverging from the upstream API / SDK (in the best case it is an extra API call that people don't really use and in the worst case there will be similar functionality upstream at some point with a [slightly] different API and then we need to reconcile it -- better to not add it until we think it is really needed)
- Make it possible to turn off this functionality by setting `session_cleanup_interval_sec` and/or `session_timeout_sec` to `-1` respectively (I'd suggest doing this since it is a very simple change, technically it can already be achieved by setting the timeouts to very large values, but it is probably better to have a documented way to do this)
- Add and index for ModelDB.status, since it will make sure the queries in cleanup_stale_sessions can be done in constant time

raulchen:
The manual `unload_model` API [already exists](https://github.com/thinking-machines-lab/tinker/blob/2d8e9d5e00f746f39148a5d0cb760dff3f2eed43/src/tinker/resources/models.py#L153) in Tinker

Will address the other 2 comments.

pcmoritz:
Got it that makes sense, can you try to modify the test code to use the SDK calls instead of plain http?

skyrl-tx/tests/tinker/test_jax_backend.py
```diff
@@ -0,0 +1,116 @@
+"""Unit tests for JaxBackend."""
```
raulchen, line 1: Current test_engine.py mixes tests for both TinkerEngine and JaxBackend. 
I'd be nice to separate them as the implementations have been separated. 
I've done so for the new tests in this PR. 
Will refactor the remaining tests in a follow-up PR.

skyrl-tx/tx/layers/lora.py
```diff
@@ -70,7 +70,7 @@ def apply_lora(
 
         (batch_size, seq_len, *dims) = x.shape
         assert len(self.lora_A.shape) == 3
-        assert len(dims) == 0 if isinstance(self, nnx.Embed) else tuple(dims) == self.lora_A.value.shape[1:-1]
+        assert len(dims) == 0 if isinstance(self, nnx.Embed) else tuple(dims) == self.lora_A[...].shape[1:-1]
```
raulchen, line 73: this is a drive-by fix because ".value" is deprecated.

skyrl-tx/tx/tinker/backends/jax.py
```diff
@@ -416,11 +416,12 @@ def create_model(self, model_id: str, lora_config: types.LoraConfig) -> None:
...
-        # Allocate adapter index for this model_id
-        adapter_index = max((m.adapter_index for m in self.models.values()), default=0) + 1
-
-        if adapter_index >= self.config.max_lora_adapters:
+        # Allocate adapter index for this model_id (find first available slot)
+        used_indices = {m.adapter_index for m in self.models.values()}
+        available_indices = set(range(1, self.config.max_lora_adapters + 1)) - used_indices
+        if not available_indices:
             raise ValueError(f"Maximum number of LoRA adapters ({self.config.max_lora_adapters}) reached")
+        adapter_index = min(available_indices)
```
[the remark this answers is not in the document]
raulchen: nice catch. 
so there are actually "max_lora_adapters - 1" available slots for the user. 
This might be confusing and unintuitive to users. should we  change the behavior? @pcmoritz

skyrl-tx/tx/tinker/engine.py
```diff
@@ -344,6 +347,79 @@ def process_create_model(self, model_id: str, request_data: types.CreateModelInp
...
         )
 
+    def process_unload_model(self, model_id: str, request_data: types.UnloadModelInput) -> types.UnloadModelOutput:
+        """Unload a model and free all resources."""
+        if self.backend.has_model(model_id):
+            self.backend.delete_model(model_id)
+        else:
+            logger.warning(f"Ignoring unload request for model {model_id} that is not loaded.")
+
+        self.backend.delete_model(model_id)
```
pcmoritz: This was already called above, I think it is a mistake to call it again, right?

pcmoritz: (actually this is probably the reason why the test is currently failing)

skyrl-tx/tx/tinker/types.py
```diff
@@ -80,6 +81,15 @@ class CreateModelOutput(BaseModel):
     lora_config: LoraConfig
 
 
+class UnloadModelInput(BaseModel):
+    pass
+
+
+class UnloadModelOutput(BaseModel):
+    model_id: str
+    status: str = "unloaded"
```
pcmoritz: Let's remove the default for this, I tried to keep defaults out of types in `types.py` (haven't always succeeded, but if we can remove more, that's a win I think). Defaults can lead to bugs where we forget to fill the value and it will be the wrong value because of that :)

skyrl-tx/tx/layers/lora.py
pcmoritz: very nice, thanks for fixing this!

skyrl-tx/tests/tinker/test_jax_backend.py
pcmoritz: Thanks for getting this started, there is also an issue about it https://github.com/NovaSky-AI/SkyRL/issues/819 :)

skyrl-tx/tx/tinker/engine.py
raulchen: it's an accidental bug introduced in the last commit. fixed now

OhadRubin:
I should probably open an issue for this but:
@raulchen @pcmoritz clear_adapter_config results in no gradient flow on subsequent non initial runs since both A and B are 0.
Should init the cleared loras weights with with the original default initilizers.
Or Al least only zero out one of the matrices (I think A is initted with zeros usually)

pcmoritz:
@OhadRubin Yes you are right! I briefly thought about this when reviewing the PR and thought it was reinitializing when the new adapter is allocated but that's not true. The best solution is probable to fix that. Let me know if you would like to make a PR for that, otherwise I'm happy to do it too :)

raulchen:
Thanks for reporting this issue. I can also put up a fix later today in case you don't have bandwidth.

pcmoritz:
@raulchen Why don't you make a PR for it (in `update_adapter_config`, get the initializer of `lora_A` and reinitialize it, and we should do that before we clear the out-of-rank entries) and we can discuss more in the PR :)

I propose we also rename `update_adapter_config` to `reinitialize_adapter` so it is clear what this function does. And we should also initialize lora_B to zero so everything is clear and ready and we don't depend on the previous values.

I think it is best to do it when creating the model instead of doing it when deleting the model -- when the model is deleted, the responsibility should be to ensure no information from this user leaks, when the model is initialized, the responsibility should be to make sure it is ready for the next training.

raulchen:
hey @OhadRubin @pcmoritz , here is the fix https://github.com/NovaSky-AI/SkyRL/pull/852

PR #847, opened 2026-01-06
[skyrl-train] Add `max_tokens_per_microbatch` configuration

justinvyu:
## Summary

Introduces token-based chunking  for microbatches, enforcing a `max_tokens_per_microbatch` limit on the total number of real tokens in a microbatch.

Replaces the even sample-based chunking `BatchIterator` with a `BalancedBatchIterator` that respects the microbatch token limits.

Screenshot 2026-01-06 at 2 55 33 PM

After this PR, the global training batch sharding and forming minibatches is still even sample-based chunking. The new functionality is adding token-based chunking at the bottom microbatching layer, which enforces a total token maximum per microbatch.

## Problem
Previously, microbatches created by chunking sequences evenly so that each microbatch was a certain batch size (`micro_train_batch_size_per_gpu`), which could lead to:
* Sequence-packed microbatches can exceed GPU memory limits when sequences vary in length, so you need to conservatively set the micro batch size.
* Inefficient GPU utilization due to uneven token distribution across microbatches. Token imbalance across DP workers leads to stragglers.

## Solution
Introduce `balanced_binpacking` and `TokenBasedBatchIterator` utilities that:
* Enforce max_tokens_per_microbatch — no microbatch exceeds the token limit
* Roughly balance the microbatches so that we avoid straggler microbatches.
* Ensure that every worker still receives the same number of microbatches by adding padding batches to satisfy FSDP requirements (every DP worker must do the same number of forward/backward passes).

## API changes

* External: added a `config.trainer.max_tokens_per_microbatch` configuration. Defaults to the old behavior (sample based microbatching).
* Adds `TokenBasedBatchIterator` and renames `BatchIterator -> SampleBasedBatchIterator`.
* Unifies `Worker.forward` to also use the `BatchIterator` interface, rather than call `TrainingBatch.chunk` directly.
* Use `TokenBasedBatchIterator` when `max_tokens_per_microbatch` is set for the `forward`, `Critic.ppo_train` and `Policy.ppo_train` methods.

## Example

Example ppo_train call from the unit test:

```
minibatch_sequences=tensor([[88, 96,  3, 23, 44, 65, 55,  0, 66, 49],
        [25, 85, 10, 86, 31, 55, 31, 51,  4, 56],
        [28, 38, 76,  8,  0,  0,  0,  0,  0,  0],
        [21, 97, 46, 75, 73,  0,  0,  0,  0,  0]])
minibatch_attention_mask=tensor([[1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
        [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
        [1, 1, 1, 1, 1, 0, 0, 0, 0, 0],
        [1, 1, 1, 1, 1, 0, 0, 0, 0, 0]])
Policy Train epoch [1/1]:   0%|          | 0/2 [00:00<?, ?it/s]
(DeepSpeedPolicyWorkerBase pid=182679) micro_batches_per_mini_batch=2
(DeepSpeedPolicyWorkerBase pid=182679) experience.sequences=tensor([[88, 96,  3, 23, 44, 65, 55,  0, 66, 49],
(DeepSpeedPolicyWorkerBase pid=182679)         [28, 38, 76,  8,  0,  0,  0,  0,  0,  0]])
Policy Train epoch [1/1]:  50%|█████     | 1/2 [00:01<00:01,  1.28s/it, pg=-0.203, glen=4, policy_lr=1e-6, ent=3.08]
(DeepSpeedPolicyWorkerBase pid=182679) experience.sequences=tensor([[25, 85, 10, 86, 31, 55, 31, 51,  4, 56],
(DeepSpeedPolicyWorkerBase pid=182679)         [21, 97, 46, 75, 73,  0,  0,  0,  0,  0]])
```

## Follow-up

This only does balanced microbatching at the individual worker level. We can further balance at the minibatch level sent to each worker at a global level to ensure every DP worker has roughly the same amount of work.

There's also different algorithms that we could plug into the microbatch chunking step.

We should also update the megatron codepath as a followup.

### DIFF
--- a/skyrl-train/skyrl_train/config/ppo_base_config.yaml
+++ b/skyrl-train/skyrl_train/config/ppo_base_config.yaml
@@ -184,6 +184,7 @@ trainer:
   critic_mini_batch_size: 256
   micro_train_batch_size_per_gpu: 1
   micro_forward_batch_size_per_gpu: 1
+  max_tokens_per_microbatch: -1 # TODO: Maybe split this between forward and train; -1 means no token-based chunking
   update_ref_every_epoch: false
   use_sample_packing: true
   eval_batch_size: 1024
--- a/skyrl-train/skyrl_train/workers/worker.py
+++ b/skyrl-train/skyrl_train/workers/worker.py
@@ -32,7 +32,12 @@ from transformers import PreTrainedModel
 from loguru import logger
 from skyrl_train.distributed.ulysses import set_ulysses_sequence_parallel_group, apply_monkey_patch
 from skyrl_train.utils.ppo_utils import PolicyLossRegistry, ppo_critic_loss, compute_approx_kl
-from skyrl_train.workers.worker_utils import BaseBatchIterator, SampleBasedBatchIterator, reduce_metrics
+from skyrl_train.workers.worker_utils import (
+    BaseBatchIterator,
+    SampleBasedBatchIterator,
+    TokenBasedBatchIterator,
+    reduce_metrics,
+)
 from skyrl_train.dataset.replay_buffer import Experience
 from skyrl_train.training_batch import TrainingInputBatch, TrainingOutputBatch
 from skyrl_train.inference_engines.inference_engine_client import InferenceEngineClient
@@ -787,11 +792,18 @@ class PolicyWorkerBase(Worker):
                 disable=not self.strategy.is_rank_0(),
             )
             for minibatch in minibatch_pbar:
-                microbatch_iterator = SampleBasedBatchIterator(
-                    minibatch, sample_batch_size=self.cfg.trainer.micro_train_batch_size_per_gpu, drop_last=False
-                )
-                num_microbatches = len(microbatch_iterator)
-                microbatch_weight = 1.0 / num_microbatches
+                if self.cfg.trainer.max_tokens_per_microbatch > 0:
+                    microbatch_iterator = TokenBasedBatchIterator(
+                        minibatch, max_tokens_per_microbatch=self.cfg.trainer.max_tokens_per_microbatch
+                    )
+                    num_microbatches = len(microbatch_iterator)
+                    microbatch_weight = 1.0 / num_microbatches  # TODO: return the correct weight from the iterator
+                else:
+                    microbatch_iterator = SampleBasedBatchIterator(
+                        minibatch, sample_batch_size=self.cfg.trainer.micro_train_batch_size_per_gpu, drop_last=False
+                    )
+                    num_microbatches = len(microbatch_iterator)
+                    microbatch_weight = 1.0 / num_microbatches
 
                 for microbatch_idx, microbatch in enumerate(microbatch_iterator):
                     microbatch_experience = BaseBatchIterator.batch_to_experience(microbatch)
--- a/skyrl-train/skyrl_train/workers/worker.py
+++ b/skyrl-train/skyrl_train/workers/worker.py
@@ -811,10 +811,15 @@ class PolicyWorkerBase(Worker):
                 )
                 num_microbatches = len(microbatch_iterator)
 
+                temp = []
                 for microbatch_idx, microbatch in enumerate(microbatch_iterator):
                     microbatch_experience = BaseBatchIterator.batch_to_experience(microbatch)
                     microbatch_weight = len(microbatch) / len(minibatch)
                     status = self.forward_backward(microbatch_experience, microbatch_weight=microbatch_weight)
+
+                    print("!!! loss", self.mesh_rank, status["final_loss"])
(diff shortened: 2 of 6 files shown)


skyrl-train/skyrl_train/workers/worker_utils.py
```diff
@@ -67,3 +58,246 @@ def batch_to_experience(batch: TrainingInputBatch):
...
+
+        This iterator evenly splits the input into batches of the same size,
+        so there's no need to reorder."""
+        return TensorBatch.cat(batches)
+
+
+def balanced_binpacking(token_counts: List[int], max_tokens_per_microbatch: int) -> List[List[int]]:
+    """Chunk a list of token counts into microbatches so that each
+    microbatch's total token count does not exceed `max_tokens_per_microbatch`,
+    and the microbatches roughly balanced.
```
justinvyu, line 102: Verl uses a Kamarkar-Karp bin balancing algorithm instead, which minimizes differences across K bins, and estimates K with `total_token_count / max_tokens_per_microbatch`. But it doesn't guarantee that every bin's token count is actually below the max token configuration. The upside is that they define the bincount upfront and don't need to inject dummy padding batches.

I'm doing balanced binpacking with capacity constraint can guarantee the max token count constraint, but requires adding dummy padding batches since each worker can come up with a different number of bins.

skyrl-train/tests/gpu/gpu_ci/test_ppo_train.py
```diff
@@ -218,3 +218,75 @@ def test_gradient_accumulation_scenarios(
...
+        )
+
+        train_data = make_dummy_batch([10, 10, 6, 5, 10, 10, 5, 5], num_actions=4)
+        # Expect:
+        # - dp=0: 3 microbatches with [10],    [10],    [11]
+        # - dp=1: 3 microbatches with [10, 5], [10, 5], [<dummy padding batch>]
+        train_data.metadata["global_step"] = 0
+
+        results = ray.get(actor_group.async_run_ray_method("mesh", "forward", train_data))
+        results = ray.get(actor_group.async_run_ray_method("mesh", "ppo_train", train_data))
```
justinvyu, line 285: I tested this with some print statements and it works as expected, but any ideas how to assert these in this test?

erictang000: hmm yeah good question - is there a way you could mock the actual model forward and make some assertion on model length that way?

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -184,6 +184,7 @@ trainer:
   critic_mini_batch_size: 256
   micro_train_batch_size_per_gpu: 1
   micro_forward_batch_size_per_gpu: 1
+  max_tokens_per_microbatch: -1 # TODO: Maybe split this between forward and train; -1 means no token-based chunking
```
erictang000, line 187: do we need some config validation to make sure that this is larger than the max response length? Maybe in utils/utils.py?

skyrl-train/skyrl_train/workers/worker.py
```diff
@@ -784,16 +804,23 @@ def record_status(status: Dict[str, float]):
...
+                    max_tokens_per_microbatch=self.cfg.trainer.max_tokens_per_microbatch,
                 )
                 num_microbatches = len(microbatch_iterator)
-                microbatch_weight = 1.0 / num_microbatches
 
+                temp = []
                 for microbatch_idx, microbatch in enumerate(microbatch_iterator):
-                    microbatch_experience = BatchIterator.batch_to_experience(microbatch)
+                    microbatch_experience = BaseBatchIterator.batch_to_experience(microbatch)
+                    microbatch_weight = len(microbatch) / len(minibatch)
```
justinvyu, line 817: This is a bit confusing for cases where there's padding batches. This is because the padding batch cannot be empty and must have seqlen >= 2.

Example: with 2 real microbatches of 15, and then 1 padding, the weights would be: 15/30, 15/30, 2/30.
* This counts on the dummy microbatch having a fully 0 loss mask, which zeros out this weight.

skyrl-train/tests/gpu/gpu_ci/test_ppo_train.py
justinvyu: lemme try

skyrl-train/skyrl_train/workers/worker.py
erictang000: why does seqlen have to be >= 2?

erictang000:
closed by #1477

PR #849, opened 2026-01-07
[skyrl-train] Refactor TIS to use more comprehensive off policy correction config

erictang000:
# Overview
- Marks `trainer.algorithm.use_tis` and `trainer.algorithm.tis_imp_ratio_cap` for deprecation
- Introduces new `trainer.algorithm.off_policy_correction` config (see new config below)
- Updates loss functions to return a `LossMetrics` TypedDict containing loss metrics (previously returned just loss, clip_ratio)
- Updates workers to all reduce mean/max/min appropriately, and to propagate loss metrics back up to the trainer.

### Off Policy Correction Config
```yaml
# To be deprecated in favor of off_policy_correction.tis_ratio_type = "token"
# and "token_tis_ratio_clip_high"
tis_imp_ratio_cap: -1.0
use_tis: false

off_policy_correction:
      # type of importance sampling ratio to use for ppo loss correction
      # here importance sampling ratio refers to exp(logprobs_{policy_old} - logprobs_{rollout_policy})
      tis_ratio_type: null # null, "token", "sequence"

      # used if tis_ratio_type = "token", 1.5-5.0 is recommended for "token" tis_ratio_type
      token_tis_ratio_clip_high: 2.0
      # used if tis_ratio_type = "sequence", 2.0-10.0 is recommended for "sequence" tis_ratio_type
      sequence_tis_ratio_clip_high: 5.0

      # method of masking out sequences with cumulative importance sampling ratios outside the cap
      # "product" masks out sequences with product of importance ratios outside the cap
      # "geometric" masks out sequences with geometric mean of importance ratios outside the cap
      sequence_mask_metric: null # null, "product", "geometric"

      # used if sequence_mask_metric = "geometric"
      # values around 0.99-1.01 are recommended for "geometric" sequence_mask_metric - MoE models may need larger allowed ranges due to higher mismatch
      geo_mask_high: 1.01
      geo_mask_low: 0.99

      # used if sequence_mask_metric = "product"
      # values around 0.5-2.0 are recommended for "sequence" sequence_mask_metric
      product_mask_high: 2.0
      product_mask_low: 0.5

      # separate from sequence_mask_metric and tis_ratio_type 
      # if any off_policy_correction is enabled, masks out sequences with any token having importance ratio
      # far outside an acceptable range (low and high thresholds)
      outlier_token_is_threshold_low: null
      outlier_token_is_threshold_high: null
```

### DIFF
--- a/docs/content/docs/algorithms/off_policy_correction.mdx
+++ b/docs/content/docs/algorithms/off_policy_correction.mdx
@@ -8,6 +8,25 @@ SkyRL provides built-in utilities for correcting off-policy drift from trainer/i
 2. **Algorithmic corrections** — importance sampling and sequence masking techniques
 3. **Configuration in SkyRL** — how to enable these corrections in your training runs
 
+# TLDR
+
+We recommend adding the following configs in order to your training runs to help address off-policy drift:
+
+```yaml
+# we recommend trying basic TIS correction first
+trainer.algorithm.off_policy_correction.tis_ratio_type="token"
+trainer.algorithm.off_policy_correction.token_tis_ratio_clip_high=2.0
+
+# for long context + MoE models, try geometric sequence masking - tune geo_mask_high/geo_mask_low as needed
+trainer.algorithm.off_policy_correction.sequence_mask_metric="geometric"
+trainer.algorithm.off_policy_correction.geo_mask_high=1.01
+trainer.algorithm.off_policy_correction.geo_mask_low=0.99
+
+# for longer context + MoE, you can also try outlier based sequence masking, which stacks on top of geometric sequence masking
+trainer.algorithm.off_policy_correction.outlier_token_is_threshold_low=1e-4
+trainer.algorithm.off_policy_correction.outlier_token_is_threshold_high=100
+```
+
 # Setup
 For common RL objectives (i.e. PPO/GRPO variants), we typically seek to optimize a token-wise objective of the form:
 
--- a/skyrl-train/skyrl_train/config/config.py
+++ b/skyrl-train/skyrl_train/config/config.py
@@ -255,6 +255,7 @@ class CISPOConfig(BaseConfig):
     cispo_eps_clip_high: float = 5.0
 
 
+# see https://docs.skyrl.ai/docs/algorithms/off_policy_correction for more details
 @dataclass
 class OffPolicyCorrectionConfig(BaseConfig):
     tis_ratio_type: Optional[str] = None
--- a/skyrl-train/skyrl_train/config/ppo_base_config.yaml
+++ b/skyrl-train/skyrl_train/config/ppo_base_config.yaml
@@ -127,6 +127,7 @@ trainer:
     # references
     # - https://github.com/szrlee/verl/blob/yingru/rollout_correction/docs/advance/rollout_corr_math.md
     # - https://fengyao.notion.site/off-policy-rl
+    # see https://docs.skyrl.ai/docs/algorithms/off_policy_correction for more details
     off_policy_correction:
       # type of importance sampling ratio to use for ppo loss correction
       # here importance sampling ratio refers to exp(logprobs_{policy_old} - logprobs_{rollout_policy})
--- a/skyrl-train/skyrl_train/utils/ppo_utils.py
+++ b/skyrl-train/skyrl_train/utils/ppo_utils.py
@@ -895,7 +895,7 @@ def cross_entropy_loss(
     config: Union[AlgorithmConfig, DictConfig],
     loss_mask: Optional[torch.Tensor] = None,
     rollout_logprobs: Optional[torch.Tensor] = None,
-) -> Tuple[torch.Tensor, float]:
+) -> Tuple[torch.Tensor, dict[str, float]]:
     """
     Cross-entropy loss for supervised fine-tuning (SFT).
 
@@ -926,7 +926,7 @@ def cross_entropy_loss(
         loss = elementwise_loss.sum()
 
     # No clipping in cross-entropy loss
-    return loss, 0.0
+    return loss, {"clip_ratio": 0.0}
 
 
 def reduce_loss(


skyrl-train/docs/configuration/config.rst
```diff
@@ -489,6 +524,35 @@ Algorithm Configuration
   - ``tau_pos``: Temperature for gating function for tokens with positive advantages.
   - ``tau_neg``: Temperature for gating function for tokens with negative (or zero) advantages.
 
+Off Policy Correction Configuration
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
CharlieFRuan, line 525: Let's cite the blogpost here as well

skyrl-train/docs/configuration/config.rst
```diff
@@ -389,6 +389,45 @@ Algorithm Configuration
       # dual clip parameters
       clip_ratio_c: 3.0
 
+      # To be deprecated in favor of off_policy_correction.tis_ratio_type = "token"
```
CharlieFRuan, line 390: I think `off_policy_correction` deserves a documentation page under the Algorithms section. We can learn from veRL's and give some canonical pre-set example, and some intuitions on when to use which. Especially these configs are kind of hierarchical (`token_tis_ratio_clip_high` only applicable when `tis_ratio_type` is `token`). We can do it in a followup PR.

Perhaps we can tell users what is the basic way of doing TIS (token-level), which only involves two configs. Then if they're advanced enough they can refer to the blogs and further tune the configs.

From my understanding of these two blogs:
- https://yingru.notion.site/When-Speed-Kills-Stability-Demystifying-RL-Collapse-from-the-Training-Inference-Mismatch-271211a558b7808d8b12d403fd15edda#27b211a558b78099ba48fa8849ab54c8
- https://richardli.xyz/post/rl-collapse-part3/

The best config seems to be token, and geometric (figures 16-18)? Especially for long-horizon tool call? If that's the impression you got from the blogs, we should make a comment so users can pick that "preset"

skyrl-train/skyrl_train/utils/ppo_utils.py
```diff
@@ -552,6 +556,252 @@ def _safe_exp_delta(delta: torch.Tensor, clip: float = 20.0, out_dtype=None) ->
     return y.to(out_dtype or delta.dtype)
 
 
+def compute_tis_ratio(
```
CharlieFRuan: These are great! Can we put them into a separate file? Our `ppo_utils.py` is 1.4k LOCs now.

In long term we could break `ppo_utils.py` down, but for now let's create a file of `off_policy_correction_utils.py` (or some other name you see fit) with all these methods you added. We can keep the rest in where they currently are and come back later if we'd want to further clean up.

skyrl-train/skyrl_train/utils/ppo_utils.py
```diff
@@ -196,6 +196,10 @@ def ppo_critic_loss(
     return 0.5 * loss, clipfrac
 
 
+class LossMetrics(TypedDict, total=False):
```
CharlieFRuan: this single-field TypedDict is a bit confusing. I know we can extend this with a lot of fields depending on the set up. Is there a better solution? Do we plan to add more fields to this? If not, should we remove this class for now?

skyrl-train/skyrl_train/utils/ppo_utils.py
```diff
@@ -552,6 +556,252 @@ def _safe_exp_delta(delta: torch.Tensor, clip: float = 20.0, out_dtype=None) ->
...
+        )
+        loss_mask = loss_mask * sequence_mask
+        metrics.update(sequence_mask_metrics)
+
+    return tis_ratio, metrics, loss_mask
+
+
 @register_policy_loss(PolicyLossType.REGULAR)
 @register_policy_loss(PolicyLossType.DUAL_CLIP)
 def ppo_policy_loss(
```
CharlieFRuan, line 548: The return is typed as `Tuple[torch.Tensor, float]`, which isn't correct right, due to it currently returning `loss_metrics`. Depending on what we do with `LossMetrics` as noted in the other comment on `LossMetrics`, we could make it `dict[str, float]`

skyrl-train/skyrl_train/utils/ppo_utils.py
```diff
@@ -581,17 +831,20 @@ def ppo_policy_loss(
...
-    if config.use_tis:
-        from loguru import logger as logger_
+    loss_metrics = LossMetrics(clip_ratio=clip_ratio)
 
-        logger_.debug(f"Using TIS with dtype: {rollout_logprobs.dtype}")
-        # Apply truncated importance sampling -> https://fengyao.notion.site/off-policy-rl
-        tis_imp_ratio = _safe_exp_delta(old_log_probs - rollout_logprobs, clip=20.0, out_dtype=log_probs.dtype)
-        tis_imp_ratio = torch.clamp(tis_imp_ratio, max=config.tis_imp_ratio_cap)
-        loss = loss * tis_imp_ratio
+    # apply off policy correction
```
CharlieFRuan, line 577: these seem redundant, used in sapo, gspo, cispo, and ppo. Can we write a functional helper to extract these out?

skyrl-train/skyrl_train/utils/ppo_utils.py
```diff
@@ -581,17 +831,20 @@ def ppo_policy_loss(
...
+        if tis_ratio is not None:
+            loss = loss * tis_ratio
+        loss_metrics.update(off_policy_correction_metrics)
 
     loss = reduce_loss(loss, loss_mask, loss_reduction, config.max_seq_len)
-    return loss, clip_ratio
+    return loss, loss_metrics
 
 
 @register_policy_loss(PolicyLossType.SAPO)
```
CharlieFRuan, line 587: Might be a dumb question: why is off-policy correction only used in sapo, gspo, cispo, and ppo, not the other loss functions like `compute_policy_loss_clip_cov`, `compute_policy_loss_kl_cov`?

skyrl-train/docs/configuration/config.rst
CharlieFRuan: Depends on how you add the separate correction doc page (see other comment). But it'd be easier for the user if we can do the following. Basically help the uesrs understand each config (3 groups of them) one-by-one by pointing them to other resources.

### 1. Group these three together

- ``algorithm.off_policy_correction.tis_ratio_type``
- ``algorithm.off_policy_correction.token_tis_ratio_clip_high``
- ``algorithm.off_policy_correction.sequence_tis_ratio_clip_high``

and tell them:
- how to do the basic TIS proposed in https://fengyao.notion.site/Your-Efficient-RL-Framework-Secretly-Brings-You-Off-Policy-RL-Training-237721e3f6c48094ad67dad3ac091c56
  - i.e. token level, and a default clip value
- what is sequence level, the difference, and the motivation of doing that; perhaps simply refer to section 4.2 of https://yingru.notion.site/When-Speed-Kills-Stability-Demystifying-RL-Collapse-from-the-Training-Inference-Mismatch-271211a558b7808d8b12d403fd15edda#27b211a558b78099ba48fa8849ab54c8

### 2. Then group these together
```
sequence_mask_metric: null # null, "product", "geometric"
geo_mask_high: 1.01
geo_mask_low: 0.99
product_mask_high: 2.0
product_mask_low: 0.5
```

### 3. Then group the outlier threshold together

```
outlier_token_is_threshold_low: 1e-4
outlier_token_is_threshold_high: 100
```

### other remarks

Then pointing to our implementation would also be helpful. Namely the `rollout_corrections.py` or whatever name you decided in the end

skyrl-train/docs/configuration/config.rst
```diff
@@ -389,6 +389,45 @@ Algorithm Configuration
...
+
+        # used if sequence_mask_metric = "product"
+        # values around 0.5-2.0 are recommended for "product" sequence_mask_metric
+        product_mask_high: 2.0
+        product_mask_low: 0.5
+
+        # separate from sequence_mask_metric and tis_ratio_type 
+        # if any off_policy_correction is enabled, masks out sequences with any token having importance ratio
+        # far outside an acceptable range (low and high thresholds)
+        outlier_token_is_threshold_low: 1e-4
```
CharlieFRuan: can we set the default value to `null` for `outerlier_token_is_xxx`? I know the current values are really high, but I feel like `null` and no outerlier masking is a better default. Seems to align with veRL's approach

https://verl.readthedocs.io/en/latest/examples/config.html

```
   rollout_correction:
     rollout_is: token # IS weights: token/sequence/null
     rollout_is_threshold: 2.0 # Upper threshold for IS weights
     rollout_rs: null # Rejection sampling: token/sequence/geometric/null
     rollout_rs_threshold: null # RS upper threshold
     rollout_rs_threshold_lower: null # RS lower threshold
     rollout_token_veto_threshold: null # Per-token veto (null to disable)
```

Not sure if you'd need to make changes on the implementation side to treat `None` as not masking

skyrl-train/tests/cpu/algorithms/test_losses.py
```diff
@@ -620,4 +638,575 @@ def gate_function(x, tau):
     torch.testing.assert_close(actual_loss, expected_loss, rtol=1e-5, atol=1e-8)
 
     # SAPO should always report clip_ratio = 0.0
-    assert actual_clip_ratio == 0.0
+    assert loss_metrics["clip_ratio"] == 0.0
+
+
+# ============================================================================
```
CharlieFRuan: Can we move everything below to a separate test file

skyrl-train/skyrl_train/utils/ppo_utils.py
erictang000: hmm you're right, let me just change the convention to return a python dictionary of metrics

erictang000:
image
image
image

erictang000:
image

skyrl-train/docs/configuration/config.rst
erictang000: done! pasted screenshots below

skyrl-train/docs/configuration/config.rst
erictang000: done

skyrl-train/skyrl_train/utils/ppo_utils.py
erictang000: done!

skyrl-train/skyrl_train/utils/ppo_utils.py
erictang000: hmm it was because the covariance calculation could include masked out samples, so just adding the sequence masking before reduce_loss didn't seem sufficient.

erictang000: i think we could add it but i would vote to just skip for now since these are not commonly used anyway

skyrl-train/examples/megatron/run_megatron_dapo_qwen3_30b_a3b_lora.sh
```diff
@@ -41,7 +41,7 @@ MINI_BATCH_SIZE=32
 N_SAMPLES_PER_PROMPT=16
 EVAL_N_SAMPLES_PER_PROMPT=32
 ENFORCE_EAGER=true # cuda graphs can cause some instability
-LR=3e-5
+LR=1e-5 # 10x compared to full finetuning
```
CharlieFRuan, line 44: just wanted to make sure the changes to LR, LORA_RANK, LORA_ALPHA are all intentional right?

skyrl-train/skyrl_train/config/config.py
```diff
@@ -255,6 +255,20 @@ class CISPOConfig(BaseConfig):
     cispo_eps_clip_high: float = 5.0
 
 
+@dataclass
+class OffPolicyCorrectionConfig(BaseConfig):
```
CharlieFRuan, line 260: let's add a link to the doc you wrote here perhaps?

skyrl-train/skyrl_train/config/ppo_base_config.yaml
```diff
@@ -118,9 +118,47 @@ trainer:
...
     # dual clip parameters
     clip_ratio_c: 3.0
-    # Truncated Importance Sampling as proposed in https://fengyao.notion.site/off-policy-rl 
+
+    # To be deprecated in favor of off_policy_correction.tis_ratio_type = "token"
+    # and "token_tis_ratio_clip_high"
     tis_imp_ratio_cap: -1.0
     use_tis: false
+
+    # references
```
CharlieFRuan, line 127: Shall we point to our doc link here as well

skyrl-train/skyrl_train/utils/ppo_utils.py
```diff
@@ -878,7 +884,7 @@ def compute_policy_loss_kl_cov(
     )
 
     # NOTE (sumanthrh): Since the pg clip ratio is not applicable for KL-COV so we just use 0.0
-    return pg_loss, 0.0
+    return pg_loss, {"clip_ratio": 0.0}
 
 
 @register_policy_loss(PolicyLossType.CROSS_ENTROPY)
```
CharlieFRuan, line 890: Should this return `return loss, {"clip_ratio": 0.0}` and change return type to `Tuple[torch.Tensor, dict[str, float]]`? Though we only use it in SFT, but it might make sense to keep it consistent

docs/content/docs/algorithms/off_policy_correction.mdx
CharlieFRuan, line 1: This doc is just great! Thank you so much for the effort!

Some really nits:
- For a user that just wants to pick a correction config, they might not want to finish reading everything. Could we give some quick pointers (like TLDR, here are the configs you can start from) at the top? Like, just use 
```
algorithm.off_policy_correction.tis_ratio_type=xxx
algorithm.off_policy_correction.token_tis_ratio_clip_high=xxx
```
if you want to do the most popular (or basic?) TIS.

Use xxx if you want to follow this blog, etc. 

And could we add a reference section at the top or the bottom as well please

skyrl-train/examples/megatron/run_megatron_dapo_qwen3_30b_a3b_lora.sh
erictang000: yep

docs/content/docs/algorithms/off_policy_correction.mdx
erictang000: > And could we add a reference section at the top or the bottom as well please

already there! (just didn't screenshot)

erictang000: > For a user that just wants to pick a correction config, they might not want to finish reading everything. Could we give some quick pointers (like TLDR, here are the configs you can start from) at the top? Like, just use


good point, added!
image

PR #671, opened 2025-11-15
[skyrl-train] Support older vllm versions till 0.9.2

SumanthRH:
# What does this PR do?

Adds support for older vllm versions.

With these changes, I am able to use older vllm versions as follows:

```bash
uv run --isolated --extra vllm --extra dev --with vllm==0.9.2 --with transformers==4.53.0 --with torch==2.7.0 --with "flash-attn@https://github.com/Dao-AILab/flash-attention/releases/download/v2.8.3/flash_attn-2.8.3+cu12torch2.7cxx11abiTRUE-cp312-cp312-linux_x86_64.whl" -- pytest -s -vvv tests/gpu/gpu_ci/test_engine_generation.py::test_token_based_generation -m "vllm"
```

Will add docs for this in a future PR

PR #851, opened 2026-01-07
No need to check vllm_version >= 0.10.0, pyproject.toml does it

AYou0207:
Because skyrl-train requires `vllm==0.11.0`, we do not need check `version.parse(vllm.__version__) >= version.parse("0.10.0")`.

And, `ErrorInfo` is used by default in the codebase, e.g., [class ErrorInfo(BaseModel):](https://github.com/NovaSky-AI/SkyRL/blob/2a7a5721b8a625801aa3d1b7988928974646bf7b/skyrl-train/skyrl_train/inference_engines/inference_engine_client_http_endpoint.py#L42).

---
The better way is to use `ErrorInfo & ErrorResponse` defined in `skyrl_train` not in `vllm`. 
👀

### DIFF
--- a/skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
+++ b/skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
@@ -385,7 +385,7 @@ class InferenceEngineClient(InferenceEngineInterface):
         # NOTE(Charlie): do not reuse headers here as the single request may become various new requests
         headers = {"Content-Type": "application/json"}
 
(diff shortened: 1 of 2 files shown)


erictang000:
hey @AYou0207, thanks for taking a look at this - because of the FlashRL integration, which uses an old vLLM wheel built on 0.9.2, we still need this guard. See PR: https://github.com/NovaSky-AI/SkyRL/pull/671

Closing this for now, but once we upstream the changes from the FlashRL integration, we should be able to update this.

PR #852, opened 2026-01-07
[tx] Fix lora_A not reinitialized after clearing lora adapter and support seeds in LoraConfig

raulchen:
Summary

  - Fix gradient flow bug when reusing adapter slots after deletion
  - Rename update_adapter_config → init_lora_adapter and clear_adapter_config → clear_lora_adapter

  Problem

  When an adapter slot is cleared and reused, both lora_A and lora_B are zero, resulting in zero gradients during training.

  Solution

  init_lora_adapter now fully initializes the adapter (lora_A with he_uniform, lora_B with zeros) instead of assuming lora_A ha
s valid random values.

  Test plan

  - Added test_adapter_reuse_initializes_lora_adapter to verify proper initialization on slot reuse
  - Run existing LoRA training tests

### DIFF
--- a/skyrl-tx/benchmarks/benchmark_engine.py
+++ b/skyrl-tx/benchmarks/benchmark_engine.py
@@ -47,7 +47,7 @@ def build_engine(config: EngineConfig, num_adapters: int) -> TinkerEngine:
         engine.process_single_request(
             types.RequestType.CREATE_MODEL,
             model_id,
-            {"lora_config": {"rank": config.backend_config["max_lora_rank"], "alpha": 32}},
+            {"lora_config": {"rank": config.backend_config["max_lora_rank"], "alpha": 32, "seed": i}},
         )
         # Mark as loaded so sampling uses in-memory weights
         engine.backend.models[model_id].loaded_checkpoint_id = model_id


skyrl-tx/tx/layers/lora.py
```diff
@@ -294,24 +295,32 @@ def update_lora_config(path, value):
...
+        if key_name == "lora_scaling":
             # Set scaling to 0.0 if rank is 0
             return value.at[adapter_index].set(lora_config.alpha / effective_rank if effective_rank > 0 else 0.0)
-        if path[-2].key == "lora_A":
-            # Zero out columns beyond the rank for this adapter; lora_B is already zero
-            return value.at[adapter_index, ..., effective_rank:].set(0.0)
+        if key_name == "lora_A":
+            # Reinitialize with he_uniform, then zero columns beyond rank
+            shape = value[adapter_index].shape
+            new_A = nnx.initializers.he_uniform()(model.rngs.params(), shape, value.dtype)
```
pcmoritz: Hmm, I think we should try to access the initializer from the model/module definition here and call it. The reasons are:

- Different models could have different initializers
- Somebody might want to change the initializer (e.g. if they are doing research) and it will be confusing if they have to change it in two places. The model definition should be the canonical place where it is defined.

pcmoritz: Even better if we can call the initializer for both lora_A and lora_B, e.g. if somebody changes lora_B to not be initialized with zeros in the model definition. This way we could possibly also consolidate the two if branches.

pcmoritz: It looks like this is not easily possible. Instead we can just initialize them to zero in https://github.com/NovaSky-AI/SkyRL/blob/main/skyrl-tx/tx/layers/lora.py#L46 and leave a comment that the actual initialization happens in reinit_lora_adapter, and leave reinit_lora_adapter as you have it at the moment.

skyrl-tx/tests/tinker/test_api.py
```diff
@@ -381,7 +382,13 @@ def test_stale_session_cleanup(api_server_fast_cleanup):
...
     _, log_path = api_server_fast_cleanup
     base_url = f"http://0.0.0.0:{TEST_SERVER_PORT_FAST_CLEANUP}/"
-    service_client, training_client = create_service_and_training_client(base_url=base_url)
+    # Skip verification because we've reduced cleanup timeout in this test.
+    # When running on a slow machine, the model may have already be cleaned
+    # up at verification time.
+    service_client, training_client = create_service_and_training_client(
+        base_url=base_url,
+        skip_verify=True,
+    )
```
raulchen, line 391: also de-flakey this test that was also introduced in the last PR

skyrl-tx/tx/layers/lora.py
raulchen: done

skyrl-tx/tx/layers/lora.py
```diff
@@ -294,24 +298,32 @@ def update_lora_config(path, value):
...
-            return value.at[adapter_index, ..., effective_rank:].set(0.0)
+        if key_name == "lora_A":
+            # Reinitialize with he_uniform, then zero columns beyond rank
+            shape = value[adapter_index].shape
+            new_A = nnx.initializers.he_uniform()(rngs.params(), shape, value.dtype)
+            new_A = new_A.at[..., effective_rank:].set(0.0)
+            return value.at[adapter_index].set(new_A)
+        if key_name == "lora_B":
+            # Explicitly zero lora_B
+            return value.at[adapter_index].set(0.0)
```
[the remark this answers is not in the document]
pcmoritz: Not clear that this is better

pcmoritz (approved):

PR #817, opened 2025-12-31
[skyrl-train] Refactor training loop structure to explicitly batch at two levels (minibatch -> microbatch)

justinvyu:
## Summary

This PR refactors the training loop structure in both `PolicyWorkerBase` and `CriticWorkerBase` to use a consistent two-level batching strategy (minibatch → microbatch) with a `microbatch_weight` used for gradient accumulation.

Refactored ppo_train to use a two-level loop: iterate over minibatches, then subdivide each minibatch into microbatches
* Changed forward_backward signature from accumulation_steps: int to microbatch_weight: float
* Loss is now scaled by microbatch_weight (i.e., 1.0 / num_microbatches) instead of dividing by accumulation_steps
* Optimizer step is now called once per minibatch (after all microbatches are processed), rather than conditionally based on step count

## Motivation

The motivation for this PR was for batch balancing which I tried adding in this PR: https://github.com/NovaSky-AI/SkyRL/pull/640

The problem was that minibatch boundaries are not explicitly defined right now. We have 2 configurations `policy_mini_batch_size` and `micro_train_batch_size_per_gpu`, and the minibatch is implicitly constructed by doing gradient accumulation over `(policy_mini_batch_size // micro_train_batch_size_per_gpu)` identical microbatches. https://github.com/NovaSky-AI/SkyRL/pull/640 breaks the "same sized microbatches" assumption, which is why we need to partition explicitly at the minibatch level first, before chunking into possibly uneven microbatches. That way, it's simpler to tell when to stop accumulating the gradient.

Also, the introduction of a more general `microbatch_weight` is also motivated by dynamic microbatch sizes introduced by https://github.com/NovaSky-AI/SkyRL/pull/640. Each microbatch should contribute `N_i / sum(N_j) * loss_i` to the accumulated gradient. In the default case, this is just `1/accumulation_steps`.

## Testing

* ✅ [FSDP metrics are equivalent to master](https://github.com/NovaSky-AI/SkyRL/pull/817#issuecomment-3712474195)
* ✅ [Megatron codepath still works](https://github.com/NovaSky-AI/SkyRL/pull/817#issuecomment-3712589412)

PR #859, opened 2026-01-09
[Tinker] Refactor trainer and worker (to move algo to trainer and infra to worker)

tyler-griggs:
## PR Overview

This PR has a pretty significant architectural refactor. The primary changes are:

1. Move algorithm logic into the trainer and infra logic into the workers.  

    -  Remove `ppo_train` (from FSDP workers for now, @erictang000 to remove from Megatron)
    - Move micro-batching logic from trainer to worker (worker doesn't know mini batches, trainer doesn't know micro batches
    - Worker onloading/offloading moved out of trainer
2. Introduce WorkerDispatch: an intermediate layer between the training loop and workers. Later, this is the entity that will sit beneath the Tinker API server. It is the entity that handles a "pool" of workers. Currently it dispatches requests to the pool of workers (by mesh or by pass_through) and handles onloading / offloading workers. 
  i. I'm very open to renaming suggestions.

---

### Architecture
Here's a diagram of what the breakdown of responsibilities now looks like. Note that some of this will still need to change. E.g., the trainer will not call `dispatch` in the future, it will call into the Tinker API server. 

```
┌─────────────────────────────────────────────────────────────┐
│                    TRAINER (Algorithm)                       │
│  - PPO algorithm implementation                              │
│  - Knows only mini batches                                   │
│  - Calls dispatch.forward_backward() + dispatch.optim_step() │
└─────────────────────────────────────────────────────────────┘
                              │
                              ▼
┌─────────────────────────────────────────────────────────────┐
│                 WORKER DISPATCH (Coordination)               │
│  - Manages all actor groups (policy, critic, ref)           │
│  - Handles GPU state (offload/backload) automatically       │
│  - Routes calls to appropriate workers                      │
│  - Handles DP sharding via MeshDispatch                     │
└─────────────────────────────────────────────────────────────┘
                              │
                              ▼
┌─────────────────────────────────────────────────────────────┐
│                    WORKERS (Execution)                       │
│  - Execute forward/backward passes                          │
│  - Handle micro-batching internally                         │
│  - Scale gradients at optim_step                            │
│  - Model-specific implementations (FSDP, Megatron)          │
└─────────────────────────────────────────────────────────────┘
```

---

### Tests

**Deleted Tests**
- `test_ppo_train.py` - tested removed `ppo_train` method

**Updated Tests**
- `test_training_step.py` - uses `WorkerDispatch` for policy tests
- `test_worker_offload.py` - updated to work with new interfaces
- `test_save_load_checkpoint.py` - updated imports
- `test_trainer.py` - rewrote `test_normalize_mini_batch_size`

---

### Gradient scaling

Since the worker no longer knows mini batches, we scale gradients instead of loss:

**Old (scale loss during backward)**
```
for i in 1..N:
    grad += (1/N) * ∂loss_i/∂param
optimizer.step(grad)
```

**New (scale gradients at optim_step)**
```
for i in 1..N:
    grad += ∂loss_i/∂param
grad *= 1/N
optimizer.step(grad)
```

Both produce: `grad = (1/N) * Σ ∂loss_i/∂param`

---

## What's next for training refactor?

- [ ] Remove weight sync logic from trainer. This should not be explicitly triggered by the trainer
- [ ] Remove other “infra” calls from trainer, such as empty_cache
- [ ] Create separate entry points for a) launching workers and b) launching training. Currently trainer.py  sets up the workers (build_models) and launches training
- [ ] Update Megatron (@erictang000) to bring to same state instead of branching on which backend is used


skyrl-train/skyrl_train/trainer.py
```diff
@@ -354,11 +355,7 @@ def _remove_tail_data(self, entries: List[Any]) -> List[Any]:
         training we care that the total number of samples is nicely splittable
         across the (combined) data-parallel size of all enabled models.
         """
-        lcm_dp_size = self.policy_model.actor_infos[0].rank.dp_size
-        if self.critic_model is not None:
-            lcm_dp_size = math.lcm(lcm_dp_size, self.critic_model.actor_infos[0].rank.dp_size)
-        if self.ref_model is not None:
-            lcm_dp_size = math.lcm(lcm_dp_size, self.ref_model.actor_infos[0].rank.dp_size)
+        lcm_dp_size = self.dispatch.get_lcm_dp_size()
```
tyler-griggs, line 367: This also needs to be moved out of trainer. The trainer shouldn't know/care about dp size.

skyrl-train/skyrl_train/workers/worker.py
```diff
@@ -937,25 +942,63 @@ def forward_backward(self, experience: Experience, microbatch_weight: float) ->
...
+
+        Returns:
+            The gradient norm (before scaling, after clipping)
         """
+        # Scale accumulated gradients by 1/N to get correct average
+        if self._micro_batches_accumulated > 0:
+            scale = 1.0 / self._micro_batches_accumulated
+            for param in self.model.parameters():
+                if param.grad is not None:
+                    param.grad.mul_(scale)
```
erictang000, line 998: I get that this is cleaner to not worry about loss scaling during forward (which requires pre-knowing the details of how we break down the microbatch or how often we call optim_step).

But are you sure that this is exactly numerically equivalent + compatible with autocasting/mixed precision? Scaling the loss prior to backward seems cleaner in that sense, can we verify that this is numerically identical (and that there aren't overflow edge cases here) to scaling the loss in pytorch?

skyrl-train/skyrl_train/workers/worker_dispatch.py
```diff
@@ -0,0 +1,314 @@
...
+        if not self.colocate_all:
+            return
+        # Ensure policy model is on GPU (will offload others in colocation group)
+        self._ensure_on_gpu("policy", need_optimizer=False, need_model=True)
+        # Offload optimizer if it's on GPU
+        if self._gpu_state["policy"].optimizer_on_gpu:
+            self._offload("policy", offload_optimizer=True, offload_model=False)
+
+    def finish_weight_sync(self) -> None:
+        """Finish weight sync: offload model."""
```
erictang000, line 270: Mostly just asking for understanding, is the intent for this worker_dispatch class to eventually handle the inference engine as well? At that point we could get rid of this and offload the model greedily?

skyrl-train/skyrl_train/trainer.py
```diff
@@ -893,81 +889,27 @@ def fwd_logprobs_values_reward(
...
-            # handle colocate policy and ref model
-            if self.cfg.trainer.placement.colocate_policy_ref or self.colocate_all:
-                all_rank_base_log_probs: List[TrainingOutputBatch] = ray.get(base_action_log_probs_refs)
-                base_log_probs = collect_results(self.ref_model.actor_infos, all_rank_base_log_probs, key="output")
-                self.ref_model.offload_to_cpu()
-                ray.get(self.ref_model.async_run_ray_method("pass_through", "empty_cache"))
-        else:
-            base_log_probs = None
+        # Critic forward (dispatch handles offload/backload automatically)
+        if self.has_critic:
```
erictang000, line 923: one reason that this logic was so messy before was that you could technically overlap the forward pass for critic/ref/policy, which is something you lose here. This probably wasn't obvious or used very much in most of our configs since everything was colocated, but could matter more at scale.

Once we make the api async (and handle some requests queue on the server side) we could add that functionality back - do you think that's worth adding a TODO?

erictang000 (commented):
looks pretty good to me! will work on the megatron worker update off of this branch

Also calling out some work that's actively ongoing for adding dynamic batch sizing that's touching relevant codepaths:
https://github.com/NovaSky-AI/SkyRL/pull/817
https://github.com/NovaSky-AI/SkyRL/pull/847

maybe the best thing to do would be to rebase #847 on top of the tinkerify branch? (and move down the micro batching logic updates from there into into forward_backward)

skyrl-train/skyrl_train/workers/worker_dispatch.py
erictang000: ok i am reviewing the next PR and i see it now

skyrl-train/skyrl_train/workers/worker.py
erictang000: coming back to this now that i'm implementing this for megatron - I think it overall seems cleaner and more flexible to handle scaling loss for gradient accumulation in `forward_backward`, where we could just set the default to weight each micro batch evenly and scale the loss for each microbatch by (1/num_microbatches). We could even theoretically provide an optional parameter for `forward_backward` that specifies a loss scaling term (default 1), so that users could flexibly scale the loss for different mini-batches differently (i.e. if you have 2 mini batches of different sizes, you might want to make sure that the 2nd mini batch loss is scaled to size). 

```python
def _forward_backward_micro(experience, microbatch_weight):
     ...
     loss = loss * microbatch_weight
     loss.backward()
     ...
def forward_backward(data, loss_weight=1.0):
     micro_batch_iterator = BatchIterator(data, micro_batch_size, drop_last=False)
     for micro_batch in micro_batch_iterator:
            metrics = self._forward_backward_micro(micro_batch, microbatch_weight=loss_weight / len(micro_batch_iterator))
      ...
```

This would still be compatible with the upstream tinker API which doesn't let you specify a loss weight.

wdyt? @tyler-griggs

skyrl-train/skyrl_train/trainer.py
tyler-griggs: Yeah I was thinking the same thing. I think it's fine to lose overlapping for a bit, especially given that critic usage is presumably quite low. But it will be possible to add it back once we update the async request handling.

PR #879, opened 2026-01-15
[tx] Add cuDNN flash attention

raulchen:
## Summary
  - Add `dot_product_attention()` in `tx/layers/attention.py` that
  uses cuDNN on GPU for both causal (prefill) and non-causal
  (decode) attention, with XLA fallback for CPU/TPU
  - Update Llama3 and Qwen3 models to use the shared attention
  function
  - Add GPU CI workflow via Anyscale

  ## Test plan
  - [ ] GPU tests verify cuDNN output matches XLA mask-based
  attention
  - [ ] Tests cover: causal with padding, no padding, mixed lengths,
   GQA, decode (non-causal)
  - [ ] CPU CI excludes GPU tests (`--ignore=tests/gpu`)
  
Benchmark results https://github.com/NovaSky-AI/SkyRL/issues/891#issuecomment-3775792201

### DIFF
--- a/skyrl-train/skyrl_train/inference_engines/base.py
+++ b/skyrl-train/skyrl_train/inference_engines/base.py
@@ -1,9 +1,9 @@
 from abc import ABC, abstractmethod
-from typing import List, Dict, TypedDict, Any, Optional, Hashable, TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Dict, Hashable, List, Optional, TypedDict
 
 if TYPE_CHECKING:
-    from skyrl_train.weight_sync.transfer_strategy import WeightSyncInitInfo
     from skyrl_train.weight_sync import WeightUpdateRequest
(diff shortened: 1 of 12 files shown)


skyrl-tx/ci/anyscale_gpu_ci.yaml
```diff
@@ -0,0 +1,10 @@
+name: skyrl-tx-gpu-ci
+entrypoint: bash ci/gpu_ci_run.sh
+image_uri: novaskyai/skyrl-train-ray-2.51.1-py3.12-cu12.8
+cloud: sky-anyscale-aws-us-east-1
+ray_version: "2.51.1"
+compute_config: l4_ci
+working_dir: .
+env_vars:
```
pcmoritz: I'm pretty sure this is not needed any more on modern Ray versions :)

raulchen: yeah, copied from other files. removed all

PR #882, opened 2026-01-15
Minor bug fix: Make SkyRL's vllm engine compatible with higher versions of VLLM

zhenga1:
OpenAIServingModels from VLLM has different signatures depending on version


zhenga1:
Closed because relevant PR has been merged

PR #899, opened 2026-01-20
[skyrl-train][inference] Inference Server Refactor (1/N)

kouroshHakha:
## Why
We're building HTTP-based inference serving for RL workloads. This enables:
- **Decoupled training ↔ inference** via standard HTTP (no Ray object refs)
- **Flexible backends** (vLLM now, SGLang later via protocol)
- **Dynamic weight sync** between trainer and inference servers

## What
```
skyrl_train/inference_servers/
├── router.py             # HTTP proxy: load balancing + control plane fan-out
├── server_group.py       # Manages N vLLM actors on a placement group
├── vllm_server_actor.py  # Ray actor wrapping vLLM OpenAI server
├── protocols.py          # ServerActorProtocol interface
├── vllm_worker.py        # Worker extension for NCCL weight sync
└── common.py             # Utilities (get_open_port, get_node_ip)
```

## Key Design Decisions
| Decision | Rationale |
|----------|-----------|
| `ServerActorProtocol` | Swap vLLM for SGLang without changing `ServerGroup` |
| Router as HTTP proxy | Compatible with any HTTP client; no vLLM-specific deps |
| Control plane fan-out | `/pause`, `/resume`, `/init_weight_transfer` hit all servers |
| Session-aware hashing | Sticky routing for multi-turn conversations |

## Review Guide
1. **Start here**: `protocols.py` (30 lines) - the interface contract
2. **Core logic**: `router.py` - proxy routing + fan-out behavior  
3. **Glue code**: `server_group.py` + `vllm_server_actor.py` - Ray actor lifecycle
4. **Tests**: `test_weight_sync.py` - trainer→inference weight sync flow

## Testing
```bash
# GPU CI (requires 4 GPUs)
uv run pytest tests/gpu/gpu_ci/test_inference_server_group.py -v
uv run pytest tests/gpu/gpu_ci/test_weight_sync.py -v
```

## Next
`RemoteInferenceClient` - typed client for trainer to call inference servers

### DIFF
--- a/skyrl-train/tests/gpu/gpu_ci/inference_servers/test_inference_server_group.py
+++ b/skyrl-train/tests/gpu/gpu_ci/inference_servers/test_inference_server_group.py
@@ -15,32 +15,24 @@ import time
 
 import httpx
 import pytest
-import torch
+import argparse
 
-# Skip entire module if vllm is not installed (e.g., when running sglang tests)
-vllm = pytest.importorskip("vllm")
-from vllm.entrypoints.openai.cli_args import make_arg_parser  # noqa: E402
-from vllm.utils.argparse_utils import FlexibleArgumentParser  # noqa: E402
-
-from skyrl_train.inference_servers.common import get_open_port  # noqa: E402
-from skyrl_train.inference_servers.router import InferenceRouter  # noqa: E402
-from skyrl_train.inference_servers.server_group import ServerGroup  # noqa: E402
+from skyrl_train.inference_servers.common import get_open_port
+from skyrl_train.inference_servers.router import InferenceRouter
+from skyrl_train.inference_servers.server_group import ServerGroup
 
 MODEL = "Qwen/Qwen2.5-0.5B-Instruct"
-
-
-# Skip entire module if not enough GPUs
-_gpu_count = torch.cuda.device_count() if torch.cuda.is_available() else 0
-if _gpu_count < 4:
-    pytest.skip(f"Need 4 GPUs for full test suite, found {_gpu_count}", allow_module_level=True)
 
 
 def make_vllm_cli_args(
     model: str,
     tp_size: int = 2,
     load_format: str = "auto",
-) -> FlexibleArgumentParser:
+) -> argparse.Namespace:
     """Create CLI args for vLLM server using official parser."""
+    from vllm.entrypoints.openai.cli_args import make_arg_parser
+    from vllm.utils.argparse_utils import FlexibleArgumentParser
+
     parser = FlexibleArgumentParser(description="vLLM server")
     parser = make_arg_parser(parser)
     return parser.parse_args(
@@ -109,6 +101,7 @@ def server_group_and_router(ray_init_fixture):
     group.shutdown()
 
 
+@pytest.mark.vllm
 class TestServerGroupAndRouter:
     """Tests for ServerGroup + InferenceRouter with 2 TP=2 servers."""
 
--- a/skyrl-train/skyrl_train/inference_servers/server_group.py
+++ b/skyrl-train/skyrl_train/inference_servers/server_group.py
@@ -13,7 +13,6 @@ from ray.util.scheduling_strategies import PlacementGroupSchedulingStrategy
 from skyrl_train.inference_servers.common import ServerInfo
 from skyrl_train.inference_servers.protocols import ServerActorProtocol
 from skyrl_train.inference_servers.server_pool import ServerActorPool
-from skyrl_train.inference_servers.vllm_server_actor import VLLMServerActor
 
 logger = logging.getLogger(__name__)
 
@@ -42,7 +41,7 @@ class ServerGroup:
         enable_dp: bool = False,
         enable_pd: bool = False,
         nixl_side_channel_base: int = 5600,
-        server_actor_cls: Type[ServerActorProtocol] = VLLMServerActor,
+        server_actor_cls: Optional[Type[ServerActorProtocol]] = None,
     ):
         """
         Initialize the server group.
@@ -61,10 +60,12 @@ class ServerGroup:
             nixl_side_channel_base: Base port for NIXL side channels. Each
                 server will be assigned a port of nixl_side_channel_base +
                 server_idx.
-            server_actor_cls: Server actor class implementing ServerActorProtocol.
-                Defaults to VLLMServerActor.
+            server_actor_cls: Server actor class implementing
+                ServerActorProtocol. Defaults to VLLMServerActor.
         """
-        self._server_actor_cls = server_actor_cls
+        from skyrl_train.inference_servers.vllm_server_actor import VLLMServerActor
+
+        self._server_actor_cls = server_actor_cls or VLLMServerActor
         self._cli_args = cli_args
         self._num_servers = num_servers
         self._start_port = start_port
@@ -77,10 +78,10 @@ class ServerGroup:
         self._internal_pg: Optional[PlacementGroup] = None
 
         # Query the actor class for GPU requirements
-        self._num_gpus_per_server = server_actor_cls.compute_num_gpus_per_server(cli_args)
+        self._num_gpus_per_server = self._server_actor_cls.compute_num_gpus_per_server(cli_args)
 
         logger.info(
-            f"ServerGroup: actor_cls={server_actor_cls.__name__}, "
+            f"ServerGroup: actor_cls={self._server_actor_cls.__name__}, "
             f"num_servers={num_servers}, "
             f"gpus_per_server={self._num_gpus_per_server}, "
             f"enable_dp={enable_dp}, enable_pd={enable_pd}, "
--- a/skyrl-train/tests/gpu/gpu_ci/conftest.py
+++ b/skyrl-train/tests/gpu/gpu_ci/conftest.py
@@ -35,7 +35,10 @@ def ray_init_fixture():
     env_vars["NVTE_FUSED_ATTN"] = "0"
 
     if SKYRL_PYTHONPATH_EXPORT:
-        env_vars["PYTHONPATH"] = os.environ.get("PYTHONPATH")
+        pythonpath = os.environ.get("PYTHONPATH")
+        if pythonpath is None:
+            raise RuntimeError("SKYRL_PYTHONPATH_EXPORT is set but PYTHONPATH is not defined in environment")
+        env_vars["PYTHONPATH"] = pythonpath
 
     logger.info(f"Initializing Ray with environment variables: {env_vars}")
     ray.init(runtime_env={"env_vars": env_vars})
--- a/skyrl-train/tests/gpu/gpu_ci/inference_servers/test_weight_sync.py
+++ b/skyrl-train/tests/gpu/gpu_ci/inference_servers/test_weight_sync.py
@@ -18,24 +18,17 @@ import httpx
 import pytest
 import ray
 import torch
+import argparse
+
 from ray.util.placement_group import placement_group
 from ray.util.scheduling_strategies import PlacementGroupSchedulingStrategy
 from transformers import AutoModelForCausalLM
 
-# Skip entire module if vllm is not installed (e.g., when running sglang tests)
-vllm = pytest.importorskip("vllm")
-from vllm.entrypoints.openai.cli_args import make_arg_parser  # noqa: E402
-from vllm.utils.argparse_utils import FlexibleArgumentParser  # noqa: E402
-
-from skyrl_train.inference_servers.common import get_node_ip, get_open_port  # noqa: E402
-from skyrl_train.inference_servers.router import InferenceRouter  # noqa: E402
-from skyrl_train.inference_servers.server_group import ServerGroup  # noqa: E402
+from skyrl_train.inference_servers.common import get_node_ip, get_open_port
+from skyrl_train.inference_servers.router import InferenceRouter
+from skyrl_train.inference_servers.server_group import ServerGroup
 
 MODEL = "Qwen/Qwen2.5-0.5B-Instruct"
-
-
-# Check GPU availability for test requirements
-_gpu_count = torch.cuda.device_count() if torch.cuda.is_available() else 0
 
 
 def make_vllm_cli_args(
@@ -43,8 +36,11 @@ def make_vllm_cli_args(
     tp_size: int = 2,
     load_format: str = "auto",
     gpu_memory_utilization: float = 0.5,
-) -> FlexibleArgumentParser:
+) -> argparse.Namespace:
     """Create CLI args for vLLM server using official parser."""
+    from vllm.entrypoints.openai.cli_args import make_arg_parser
+    from vllm.utils.argparse_utils import FlexibleArgumentParser
+
     parser = FlexibleArgumentParser(description="vLLM server")
     parser = make_arg_parser(parser)
     return parser.parse_args(
@@ -157,8 +153,6 @@ def weight_update_env(ray_init_fixture):
     - 4 GPUs total: trainer on GPUs 0-1, server on GPUs 2-3
     - Uses NCCL broadcast for weight sync
     """
-    if _gpu_count < 4:
-        pytest.skip(f"Need 4 GPUs for non-colocated test, found {_gpu_count}")
 
     # Create server with dummy weights (TP=2)
     cli_args = make_vllm_cli_args(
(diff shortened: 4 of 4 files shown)


skyrl-train/tests/cpu/inference_servers/test_common.py
```diff
@@ -0,0 +1,37 @@
...
+        port = get_open_port()
+        assert isinstance(port, int)
+        assert 1 <= port <= 65535
+        self._verify_port_is_free(port)
+
+    def _verify_port_is_free(self, port: int):
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+            s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+            s.bind(("", port))
+            s.listen(1)
```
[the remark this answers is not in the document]
kouroshHakha: The race window is microseconds — Between releasing and re-binding, the window is extremely narrow. For this to fail, another process would need to bind to that exact ephemeral port in that tiny window.

Also this reflects the real case failure if it ever happens, so the test is correct.

skyrl-train/skyrl_train/env_vars.py
```diff
@@ -0,0 +1,26 @@
...
+"""
+Offset for the data parallel port of the vLLM server.
+"""
+SKYRL_WAIT_UNTIL_INFERENCE_SERVER_HEALTHY_TIMEOUT_S = int(
+    os.environ.get("SKYRL_WAIT_UNTIL_INFERENCE_SERVER_HEALTHY_TIMEOUT_S", 600)
+)
+"""
+Timeout for waiting until the inference server is healthy.
+"""
+SKYRL_INCLUDE_PYTHONPATH_IN_RUNTIME_ENV = str(
```
erictang000: does this overlap with these?: https://github.com/NovaSky-AI/SkyRL/blob/6a9ab82c006e6c85a313b60120a8282a5ca4ed5a/skyrl-train/skyrl_train/utils/constants.py#L26

kouroshHakha: oh I didn't know about this. yes. it overlaps. let me consolidate.

kouroshHakha: ok need to merge this first: https://github.com/NovaSky-AI/SkyRL/pull/903

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
CharlieFRuan, line 1: can we run the full GPU CI for this PR to ensure it is not breaking anything? Though it doesn't seem likely to break things, but just wanted to make sure. Also a good thing to check that the new GPU CI tests you added work well with our GPU CI environment. 

You can do this in `Actions` after pushing this branch to NovaSky-AI's fork

skyrl-train/skyrl_train/inference_servers/router.py
```diff
@@ -0,0 +1,296 @@
...
+]
+
+
+class InferenceRouter:
+    """
+    HTTP proxy router for multiple vLLM servers.
+
+    Routing behavior:
+    - Data plane (generation requests): Routes to ONE server.
+      - If X-Session-ID header present: consistent hash to same backend
```
CharlieFRuan, line 47: The current `skyrl-train/skyrl_train/inference_engines/inference_engine_client_http_endpoint.py` allows the users to, in their agent harness, instead using a say OpenAI endpoint, to use the HTTP endpoint that SkyRL spins up.

With this refactoring, the user will use `http://0.0.0.0/8080/v1/chat/completions` for the router `router = InferenceRouter(server_urls, host="0.0.0.0", port=8080)`.

My question is, for the sticky routing, how will the user add the header? Would it be simpler if the session ID is in the body? 

Currently for terminal bench, it will have a `session_id` field in the request:
- https://github.com/NovaSky-AI/SkyRL/blob/b9a6307ca4bed5f5c055032650ad5ae68da694ed/skyrl-train/examples/terminal_bench/generator/terminal_bench_generator.py#L105
  - TBench side will forward this kwarg in the request body
- Then on SkyRL, upon receiving this request, pops the `session_id` here: https://github.com/NovaSky-AI/SkyRL/blob/b9a6307ca4bed5f5c055032650ad5ae68da694ed/skyrl-train/skyrl_train/inference_engines/inference_engine_client.py#L367-L377

skyrl-train/skyrl_train/inference_servers/router.py
```diff
@@ -0,0 +1,296 @@
...
+
+from skyrl_train.inference_servers.common import get_node_ip
+from skyrl_train.env_vars import SKYRL_WAIT_UNTIL_INFERENCE_SERVER_HEALTHY_TIMEOUT_S
+
+logger = logging.getLogger(__name__)
+
+
+# Control plane routes are broadcast to ALL servers.
+# Everything else (data plane) is load-balanced to ONE server.
+CONTROL_PLANE_ROUTES = [
```
CharlieFRuan, line 25: From my understanding, these `CONTROL_PLANE_ROUTES` route to **each** `VLLMServerActor`, and all the rest of the methods route to **only one**.

Would it be too implicit if a user uses methods beyond these control plane methods and the typical `/chat/completion` / `/completion`, such as `/is_paused` you used in the unit test, and yet the semantic is actually only one engine tells you whether it is paused?

I may not have the best knowledge here, but was just curious if this is a conventional semantic.

An explicit way is to list out the data plane routes as well, write about the semantic for each (e.g. `/is_paused` only tells you the status of one engine), and assert that a request is either in `CONTROL_PLANE_ROUTES` or `DATA_PLANE_ROUTES`

skyrl-train/skyrl_train/inference_servers/router.py
```diff
@@ -0,0 +1,296 @@
...
+        """
+        Determine server for a request.
+
+        If X-Session-ID header is present, use consistent hashing.
+        Otherwise, use round-robin.
+        """
+        session_id = request.headers.get("X-Session-ID")
+        if session_id:
+            return self._get_server_for_session(session_id)
+        return self._get_server_round_robin()
```
CharlieFRuan, line 107: So this method will be called every time when a data plane request hits the router. Our goal is to do round_robin for `/chat/completions` and `/completions` requests, but there are other data plane request that might mess with the load balancing (e.g. tokenize, detokenize, etc.).

Do you think this would be an issue? If not, please ignore the comment

skyrl-train/tests/gpu/gpu_ci/inference_servers/test_inference_server_group.py
CharlieFRuan, line 1: minor: shall we also create a folder of `inference_servers` for GPU CI, just like the CPU ones?

skyrl-train/tests/gpu/gpu_ci/test_weight_sync.py
```diff
@@ -0,0 +1,308 @@
...
+            torch.distributed.destroy_process_group(self.pg)
+            self.pg = None
+
+
+@pytest.fixture(scope="class")
+def weight_update_env(ray_init_fixture):
+    """
+    Create environment for weight update testing:
+    - Trainer with real weights on GPU 0
+    - 1 vLLM server with TP=2 and DUMMY weights (uses GPU 1,2)
```
CharlieFRuan: Could we also add a colocated weight sync test? If you think we should do it in a future PR that is also fine!

skyrl-train/skyrl_train/inference_servers/router.py
kouroshHakha: I am not sure if I understood the question / concern. So you can keep control plane ops in inference router (explicit or implicit) but some routers won't. For example, the vllm-router as it stands right now, does not handle the control plane ops. So from the client side, one can use each engine's api server to perform the control plane operations. At the same time router can do this for you as well, in which case you can just pass in the [proxy_url] as the external urls. 


Whether a method is gonna be a control plane or data plane is a decision we have to make at the router layer depending on what that method should do, so I think by nature it will be implicit. The explicit way you suggested above, does not generalize to APIs that will get added later, we cannot keep and maintain a list on skyrl where the ground truth depends on vllm's implementation unless we use it. 

Actually is_paused should have been a control plane op. Can add it if / when we need to use it.

skyrl-train/skyrl_train/inference_servers/router.py
kouroshHakha: Not for now. I think it is fair assumption for now to just make routing decision uniform across all data plane ops.

skyrl-train/skyrl_train/inference_servers/router.py
kouroshHakha: so we need to make it compatible with industry standards. if you look at sgl router, vllm-router, etc. the dominant way to pass in the session id is through passing a request header that has `X-Session-ID`. I don't know what the benefit of this is over keeping it in the body, but I think this is more of a standard for decoupling middle-ware / routing ops from the request body.

kouroshHakha: session_id is not part of chat or completion api standard btw. So I'd say we need to change the agent harness / generator to the header instead.

skyrl-train/skyrl_train/inference_engines/vllm/vllm_engine.py
kouroshHakha: https://github.com/NovaSky-AI/SkyRL/actions/runs/21322295774/job/61373903070

skyrl-train/skyrl_train/inference_servers/router.py
CharlieFRuan: > The explicit way you suggested above, does not generalize to APIs that will get added later, we cannot keep and maintain a list on skyrl where the ground truth depends on vllm's implementation unless we use it.

I see your point. But what are some data plane methods beyond `/chat/completion`, `/completion`, `/tokenize`, `/detokenize`? As long as we keep it clear that those data plane methods only get mapped to **a single engine**, I think it's fine.

For instance, currently `f"{router_url}/health"` returns a random engine's health, which doesn't sound quite right...

> Actually is_paused should have been a control plane op. Can add it if / when we need to use it.

Got it!

skyrl-train/skyrl_train/inference_servers/router.py
CharlieFRuan: > if you look at sgl router, vllm-router, etc. the dominant way to pass in the session id is through passing a request header that has X-Session-ID

Thanks for the info, didn't know before! Let's keep it as `X-Session-ID` for now! We just need to remember properly telling users about how to do sticky routing somewhere. Made an issue here: https://github.com/NovaSky-AI/SkyRL/issues/939

skyrl-train/skyrl_train/inference_servers/router.py
```diff
@@ -0,0 +1,296 @@
...
+        logger.info(f"Router started at {router_url}")
+        logger.info("  GET /servers - list servers")
+        logger.info("  GET /get_server_info - get parallelism info")
+        return router_url
+
+    def _wait_until_healthy(
+        self, router_url: str, timeout: float = SKYRL_WAIT_UNTIL_INFERENCE_SERVER_HEALTHY_TIMEOUT_S
+    ) -> None:
+        """Poll health endpoint until server is ready."""
+        health_url = f"{router_url}/health"
```
CharlieFRuan, line 280: This is the one I referred to in this comment: https://github.com/NovaSky-AI/SkyRL/pull/899/files#r2724313775

We are checking a random engine's health because this `/health` is implicitly a data plane request

skyrl-train/skyrl_train/inference_servers/router.py
kouroshHakha: > But what are some data plane methods beyond /chat/completion, /completion, /tokenize, /detokenize?

vllm has generate, embedding, etc. 

> For instance, currently f"{router_url}/health" returns a random engine's health, which doesn't sound quite right

I think we need to be explicit about control plane ops. I missed /health as another exmaple. Will add it.

skyrl-train/tests/gpu/gpu_ci/test_weight_sync.py
kouroshHakha: Good point. I'll add it in this PR.

kouroshHakha: Actually I just realized this is incomplete for using colocated and cuda_ipc. 

We need to defer that to when we merge the vllm native apis. Or if that drags longer I'll make a separate pr to support cuda_ipc fully before full flip. basically on 4+/N

CharlieFRuan: sounds good, thank you!

skyrl-train/skyrl_train/inference_servers/router.py
CharlieFRuan: > vllm has generate, embedding, etc.

Ah good point.

> I think we need to be explicit about control plane ops. I missed /health as another exmaple. Will add it.

Got it!

skyrl-train/skyrl_train/inference_servers/router.py
```diff
@@ -0,0 +1,298 @@
...
+CONTROL_PLANE_ROUTES = [
+    # BUILT-IN ROUTES
+    "/pause",
+    "/is_paused",
+    "/resume",
+    "/sleep",
+    "/wake_up",
+    "/reset_prefix_cache",
+    "/collective_rpc",
+    "/health",
```
CharlieFRuan: What is the semantic of `_proxy_to_all()` for `/health` and `/is_paused`? If one of the engines returns false, does the router return false altogether? Could we add a unit test for`/health` when one of the engine is not healthy?

I feel like the semantic of `/health` and `/is_paused` (GET methods) are different from the other control plane APIs (POST) for `_proxy_to_all()`.

skyrl-train/tests/gpu/gpu_ci/inference_servers/test_inference_server_group.py
```diff
@@ -0,0 +1,211 @@
...
+
+from skyrl_train.inference_servers.common import get_open_port  # noqa: E402
+from skyrl_train.inference_servers.router import InferenceRouter  # noqa: E402
+from skyrl_train.inference_servers.server_group import ServerGroup  # noqa: E402
+
+MODEL = "Qwen/Qwen2.5-0.5B-Instruct"
+
+
+# Skip entire module if not enough GPUs
+_gpu_count = torch.cuda.device_count() if torch.cuda.is_available() else 0
```
CharlieFRuan: could we remove this skip? Our CI always has 4 GPUs. I'm worried that we silently skip them due to unexpected reasons

skyrl-train/tests/gpu/gpu_ci/inference_servers/test_inference_server_group.py
```diff
@@ -16,12 +16,15 @@
...
 import pytest
 import torch
-from vllm.entrypoints.openai.cli_args import make_arg_parser
-from vllm.utils.argparse_utils import FlexibleArgumentParser
 
-from skyrl_train.inference_servers.common import get_open_port
-from skyrl_train.inference_servers.router import InferenceRouter
-from skyrl_train.inference_servers.server_group import ServerGroup
+# Skip entire module if vllm is not installed (e.g., when running sglang tests)
+vllm = pytest.importorskip("vllm")
```
CharlieFRuan: I'm worried that this will silently skip. We already run CI with `uv run --extra vllm pytest -s tests/gpu/gpu_ci -m "not (sglang or integrations or megatron)"`

Could we do this instead `pytestmark = pytest.mark.vllm  # Mark entire module as vllm`?

skyrl-train/skyrl_train/inference_servers/router.py
kouroshHakha: good q.

/is_paused --> routers returns a mapping from server_urls to is_paused state of each engine (same as any other control plane ops)

/health --> I think we should actually make it not a control plane endpoint. Basically if at least one engine exists that is healthy it should return true, By making /health go through the routing, if it returns true you are guaranteed that at least one engine is healhty. The router should smart enough to not route to the engines that are unhealthy (that's part of the responsibility of the router that we can add to our simple one if we want later)

I removed /health from the control plane endpoints because of above.

CharlieFRuan: sounds good, thanks!

> routers returns a mapping from server_urls to is_paused state of each engine (same as any other control plane ops)

Got it. Would be good to document this behavior somewhere! We can do that in another PR

CharlieFRuan: On /health, I made an issue here: https://github.com/NovaSky-AI/SkyRL/issues/958

PR #901, opened 2026-01-20
Unify Megatron and FSDP training interfaces with forward_backward + optim_step

tyler-griggs:
## Summary
- Add `forward_backward()` and `optim_step()` methods to `MegatronPolicyWorkerBase` to match FSDP worker interface
- Update trainer to use unified interface for both Megatron and FSDP strategies (removes strategy branching)
- Mark `ppo_train()` as deprecated (kept for backward compatibility)
- Update `test_megatron_worker.py` to use the new interface
- Add `get_lr` and `set_lr` to the megatron worker to be in line with behavior from #978 
- Add SFT behavior form #961, allowing the megatron backend to be used with the TX SkyRL-Train integration

This brings Megatron up to parity with FSDP following the refactoring in PR #859.

## Test plan
- [x] Run `test_megatron_worker.py` to verify forward_backward + optim_step works correctly
- [x] Verify metrics match between Megatron and FSDP implementations

Co-Authored-By: Eric Tang <erictang000@gmail.com>

### DIFF
--- a/skyrl-train/tests/gpu/gpu_ci/test_worker_dispatch_offload.py
+++ b/skyrl-train/tests/gpu/gpu_ci/test_worker_dispatch_offload.py
@@ -19,6 +19,7 @@ from skyrl_train.utils.utils import validate_cfg
 from skyrl_train.utils import get_ray_pg_ready_with_timeout
 from skyrl_train.entrypoints.main_base import config_dir
 from skyrl_train.workers.worker_dispatch import WorkerDispatch, GPUState
+from tests.gpu.utils import import_worker
 from skyrl_train.workers.fsdp.fsdp_worker import PolicyWorker, RefWorker, CriticWorker
 from skyrl_train.workers.worker import PPORayActorGroup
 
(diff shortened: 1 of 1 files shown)


skyrl-train/skyrl_train/workers/megatron/megatron_worker.py
```diff
@@ -519,136 +502,110 @@ def init_model(self, model_path, num_training_steps: int = 1e9):
...
+    def optim_step(self) -> Optional[float]:
+        """
+        Perform optimizer step.
+
+        Note: Unlike FSDP workers, Megatron doesn't need manual gradient scaling here
+        because Megatron Core's forward_backward_func handles loss scaling internally.
+
+        Returns:
+            The gradient norm (before scaling, after clipping), or None if unavailable.
+        """
```
erictang000, line 610: need the gradient scaling for grad accumulation here i think? looked into it briefly and it seemed doable

erictang000: @tyler-griggs given offline discussion of moving to do loss sums, maybe we don't need to do any scaling here anymore actually? We will need to figure out how the megatron internal gradient accumulation works and whether we can enforce that it also only does sums.

erictang000:
training still working, new tests pass

image

erictang000:
sft trainer script runs with the megatron backend: 
image

PR #903, opened 2026-01-20
[refactor] consolidate env vars into single module

kouroshHakha:
Moves all environment variable definitions from skyrl_train/utils/constants.py into skyrl_train/env_vars.py for better discoverability. This creates a single source of truth for configuration at the package root level.

### DIFF
--- a/skyrl-train/docs/api/constants.rst
+++ b/skyrl-train/docs/api/constants.rst
@@ -1,5 +1,0 @@
-Environment Variables
-=====================
-
-.. automodule:: skyrl_train.env_vars
-    :members:


erictang000 (commented):
Can you make the change here too: https://github.com/NovaSky-AI/SkyRL/blob/6a9ab82c006e6c85a313b60120a8282a5ca4ed5a/skyrl-train/skyrl_train/utils/utils.py#L20

kouroshHakha:
good catch 👍

erictang000:
sorry one more place: https://github.com/NovaSky-AI/SkyRL/blob/6a9ab82c006e6c85a313b60120a8282a5ca4ed5a/skyrl-train/docs/api/constants.rst#L3

then should be good to merge

PR #904, opened 2026-01-20
[skyrl-train][refactor] Inference Server Refactor -- RemoteInferenceClient 2/N

kouroshHakha:
**Summary:** Adds `RemoteInferenceClient`, a lightweight, fully serializable HTTP client that wraps inference server APIs. This client replaces the old `InferenceEngineInterface` for HTTP-based inference and can work with any HTTP-compatible inference backend (vLLM, sglang-router, Ray Serve LLM, etc.).

**Architecture:**
- **Router (InferenceRouter)**: Data plane only - routes requests to ONE backend (session-aware or round-robin)
- **Client (RemoteInferenceClient)**: Fully responsible for control plane fan-out to all backends

This separation allows using external routers (vllm-router, sglang-router) that only handle data plane.

**Key Features:**
- **Serializable**: Can be pickled and passed between Ray actors/processes
- **Two URL types**: `proxy_url` for data plane (router), `server_urls` for control plane (fan-out to backends)
- **Data plane**: `generate()`, `chat_completion()`, `completion()`, `tokenize()`, `detokenize()`
- **Control plane (fan-out)**: `pause()`, `resume()`, `sleep()`, `wake_up()`, `reset_prefix_cache()`
- **Weight sync (fan-out)**: `init_weight_transfer()`, `update_weights()`, `finalize_weight_update()`
- **PauseMode enum**: Forward-compatible with vLLM RFC #32103 pause modes
- **Built-in retry on abort**: Handles `stop_reason="abort"` during weight sync

**Comparison vs `InferenceEngineInterface` + `InferenceEngineClient`:**
- **Serializable** - Just URLs, no Ray actors/tokenizers/thread events
- **No local tokenizer** - Uses `/tokenize` endpoint instead
- **Server-side routing** - Router handles session routing via `X-Session-ID` header
- **Simplified parallelism** - Single `get_world_size()` vs separate `tp_size()`, `pp_size()`, `dp_size()`
- **No ABC hierarchy** - Simple dataclass with async methods
- **Backend-agnostic** - Works with any HTTP server (vLLM, sglang, Ray Serve LLM)

**Files Added:**
- `skyrl_train/inference_servers/remote_inference_client.py` - The client implementation
- `tests/cpu/inference_servers/test_remote_inference_client.py` - Unit tests

### DIFF
--- a/skyrl-train/skyrl_train/inference_engines/base.py
+++ b/skyrl-train/skyrl_train/inference_engines/base.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from typing import List, Dict, TypedDict, Any, Optional, Hashable, Protocol, runtime_checkable, TYPE_CHECKING
+from typing import List, Dict, TypedDict, Any, Optional, Hashable, TYPE_CHECKING
 
 if TYPE_CHECKING:
     from skyrl_train.weight_sync import WeightUpdateRequest
@@ -29,36 +29,6 @@ class InferenceEngineOutput(TypedDict):
     response_ids: List[List[int]]
     stop_reasons: List[str]
     response_logprobs: Optional[List[List[float]]]
-
-
-@runtime_checkable
-class InferenceClientProtocol(Protocol):
-    """
-    Structural protocol for inference clients.
-
-    Both InferenceEngineInterface and RemoteInferenceClient satisfy this protocol,
-    enabling code to work with either implementation via duck typing.
-
-    This is the minimal common interface for:
-    - Data plane: generate()
-    - Lifecycle: sleep(), wake_up(), teardown()
-    """
-
-    async def generate(self, input_batch: InferenceEngineInput) -> InferenceEngineOutput:
-        """Generate completions for a batch of inputs."""
-        ...
-
-    async def sleep(self, *args: Any, **kwargs: Any) -> Any:
-        """Put the engine into sleep/low-power mode."""
-        ...
-
-    async def wake_up(self, *args: Any, **kwargs: Any) -> Any:
-        """Wake the engine from sleep mode."""
-        ...
-
-    async def teardown(self) -> None:
-        """Clean up resources."""
-        ...
 
 
 class InferenceEngineInterface(ABC):


skyrl-train/skyrl_train/inference_servers/router.py
```diff
@@ -119,6 +119,11 @@ def _build_app(self) -> FastAPI:
             openapi_url=None,
         )
 
+        @app.get("/health")
+        async def health():
+            """Router health check (doesn't proxy to backends)."""
```
CharlieFRuan, line 113: Is this actually correct or should we add a TODO here? Maybe link to this issue https://github.com/NovaSky-AI/SkyRL/issues/958

skyrl-train/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -0,0 +1,661 @@
...
+server HTTP API. It replaces the old InferenceEngineInterface for HTTP-based
+inference servers.
+
+Key features:
+- Serializable: Can be pickled and passed between processes
+- Two URL types:
+  - proxy_url: Single URL for data plane operations (routed requests)
+  - server_urls: List of backend URLs for control plane operations (fan-out)
+- Lazy world_size fetching from /get_server_info
+- Pure HTTP: Uses /tokenize endpoint if token IDs are needed
```
CharlieFRuan: Typo on "Pure HTTP"?

skyrl-train/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -0,0 +1,661 @@
...
+            model_name=model_name,
+        )
+
+    @classmethod
+    def from_router(
+        cls,
+        router_url: str,
+        model_name: str = "default",
+    ) -> "RemoteInferenceClient":
+        """
```
CharlieFRuan: why do we need to support both? I see the redundancy of how `remote_inference_client.py` supports fan-out for control calls, and how `router.py` also support that. If we want to support arbitrary router, we could just make the assumption that all routers only support data-plane request right, to avoid duplicated logic. What would be your thought on this?

skyrl-train/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -0,0 +1,661 @@
...
+                async with session.request(method, url, json=json) as resp:
+                    body = await resp.json() if resp.content_length else None
+                    return server_url, {"status": resp.status, "body": body}
+            except Exception as e:
+                return server_url, {"status": 500, "error": str(e)}
+
+        results = await asyncio.gather(*[call_server(url) for url in self.server_urls])
+        return {url: resp for url, resp in results}
+
+    async def pause(self, mode: Union[PauseMode, str] = PauseMode.ABORT) -> Dict[str, Any]:
```
CharlieFRuan, line 446: In my implementation in `inference_engine_client.py`, I had a `self.generation_paused_event` to see when the inference is paused / unpaused.

We don't need that here because vllm will block internally when we `/pause` without `/resume`, is that correct?

skyrl-train/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -0,0 +1,661 @@
...
+        # Tokenize full accumulated text for response_ids
+        final_token_ids = (await self.tokenize([accum_text], add_special_tokens=False))[0] if accum_text else []
+
+        return {
+            "response": accum_text,
+            "stop_reason": stop_reason,
+            "response_ids": final_token_ids,
+        }
+
+    async def chat_completion(
```
CharlieFRuan, line 286: Currently `chat_completion()` doesn't support in-flight weight update right? would that be added at some point? It would be needed before we fully ship the refactor

skyrl-train/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -0,0 +1,661 @@
...
+        elif "max_completion_tokens" in sampling_params:
+            max_key = "max_completion_tokens"
+        original_max_tokens = sampling_params.get(max_key) if max_key else None
+
+        # Accumulate across retries
+        accum_text = ""
+        accum_token_ids: List[int] = []
+        stop_reason = "abort"
+
+        while stop_reason == "abort":
```
CharlieFRuan, line 241: we likely will need a more thorough test here.

I suppose as a part of rolling out process we will reach testing parity with the existing rollout stack, would that be the goal?

skyrl-train/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -0,0 +1,661 @@
...
+        max_key = None
+        if "max_tokens" in sampling_params:
+            max_key = "max_tokens"
+        elif "max_completion_tokens" in sampling_params:
+            max_key = "max_completion_tokens"
+        original_max_tokens = sampling_params.get(max_key) if max_key else None
+
+        # Accumulate across retries
+        accum_text = ""
+        accum_token_ids: List[int] = []
```
CharlieFRuan, line 238: logprobs would also be needed

skyrl-train/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -0,0 +1,661 @@
...
+            New requests are blocked. No retry needed.
+            Maps to: wait_for_inflight_request=True
+    """
+
+    ABORT = "abort"
+    FINISH = "finish"
+
+
+@dataclass
+class RemoteInferenceClient:
```
CharlieFRuan, line 108: If a user has their own agent that want to perform RL with, substituting their OpenAI URL with SkyRL's URL -- I suppose that would just be the router's URL right? Meaning that they wouldn't go through `RemoteInferenceClient` So likely `RemoteInferenceClient.chat_completion()` and `RemoteInferenceClient.completion()` will not be needed -- assuming vllm supports in-flight-weight update without our retry loop.

So I guess our implementation and the role of RemoteInferenceClient really depends on when that RFC will be merged in.

In the world where vllm supports in-flight-weight update internally with the RFC, is my understanding of the role of `RemoteInferenceClient` correct:
- Supports a pythonic `generate()`, which is a `/completion` behind the scene
- A central place for all the control fan-out ops

skyrl-train/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -0,0 +1,661 @@
...
+        based on session_id. All requests are made in parallel.
+
+        Args:
+            input_batch: Contains prompt_token_ids, sampling_params, and optional session_ids.
+
+        Returns:
+            InferenceEngineOutput with responses, response_ids, and stop_reasons.
+        """
+
+        prompt_token_ids = input_batch.get("prompt_token_ids")
```
CharlieFRuan, line 174: If we want to support a `generate()` endpoint to users, it's possible that the input is string rather than token IDs

skyrl-train/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -0,0 +1,661 @@
...
+import logging
+from dataclasses import dataclass, field
+from enum import Enum
+from typing import Any, Dict, List, Optional, Union, TYPE_CHECKING
+from dataclasses import asdict
+
+
+import aiohttp
+
+from skyrl_train.inference_engines.base import InferenceEngineInput, InferenceEngineOutput
```
CharlieFRuan, line 78: Whether to continue using these two APIs can be debated

CharlieFRuan (commented):
Left several comments, thank you!

skyrl-train/skyrl_train/inference_servers/remote_inference_client.py
kouroshHakha: discussed and agree with redundancy. Will modify the router to be only data plane.

skyrl-train/skyrl_train/inference_servers/remote_inference_client.py
kouroshHakha: correct.

skyrl-train/skyrl_train/inference_servers/remote_inference_client.py
kouroshHakha: yes. as we incrementally rollout and those features get upstreamed in vllm, we can keep skyrl lighter weight.

skyrl-train/skyrl_train/inference_servers/remote_inference_client.py
kouroshHakha: correct. We only flip the flag by default when all the previous code can safely be deleted.

skyrl-train/skyrl_train/inference_servers/remote_inference_client.py
kouroshHakha: in a world where vllm supports the in-flight-weight update internally all data ops from remote_inference_client can be removed and /chat or /completion apis should work out of box without any client code side.

skyrl-train/skyrl_train/inference_servers/router.py
kouroshHakha: I think the health check of router vs. health of replicas need to be decoupled. Added a todo for now, to think about this a bit later during parity testing.

skyrl-train/skyrl_train/inference_servers/remote_inference_client.py
kouroshHakha: I'll revisit the content of generate during incremental parity matching case by case (I think the vllm prs are a 1/2 weeks out from getting merged upstream)

skyrl-train/skyrl_train/inference_servers/remote_inference_client.py
kouroshHakha: agree. one data plane ops are removed this won't matter.

skyrl-train/skyrl_train/inference_servers/remote_inference_client.py
kouroshHakha: see https://github.com/NovaSky-AI/SkyRL/pull/904#discussion_r2743621364

Issue #891, opened 2026-01-16
[TX] support long sequence


raulchen:
I'm currently testing / optimizing skyrl-tx for long-sequence use cases. E.g., the tinker [search tool recipe](https://github.com/thinking-machines-lab/tinker-cookbook/tree/main/tinker_cookbook/recipes/search_tool)

Using this issue to track the related fixes and optimizations.

raulchen:
# TX Memory Benchmark Results

**Model:** Qwen/Qwen3-4B-Instruct-2507 | **TP:** 8 | **GPUs:** 8x A100-40GB

---

## Optimization Stack (Cumulative)

| Stage | Optimizations |
|-------|--------------|
| 0 | baseline (main) |
| 1 | + flash-attention https://github.com/NovaSky-AI/SkyRL/pull/879 |
| 2 | + skip-full-logits https://github.com/NovaSky-AI/SkyRL/pull/878 |
| 3 | + per-layer-checkpointing https://github.com/NovaSky-AI/SkyRL/pull/906 |
| 4 | + chunked-lm-head https://github.com/NovaSky-AI/SkyRL/pull/902 |

### What Each Optimization Does

**1. Flash Attention**
- Uses cuDNN flash attention instead of naive attention
- Memory: O(T) instead of O(T²). At seq=4k, avoids materializing 4096×4096 attention matrix per head
- Universal win with no tradeoffs

**2. Skip Full Logits**
- During prefill, only compute logits for last token instead of all tokens
- Memory: Avoids [batch, seq, vocab] tensor. At BS=16, seq=4k, vocab=152k = 20GB in fp16
- Only helps sampling (prefill). No effect on training

**3. Per-Layer Checkpointing**
- Use `jax.lax.scan` with checkpointing so XLA keeps only ONE layer's activations, recomputing during backward
- Memory: 36 layers → 1 layer activation storage (~36x reduction)
- Key training optimization. Trades compute for memory

**4. Chunked LM Head**
- Compute cross-entropy loss in chunks instead of materializing full logits tensor
- Memory: At BS=16, seq=4k, vocab=152k: 10GB → 300MB (with chunk_size=512)

---

## Sample Mode Results

Format: peak memory (MiB) / time (s)

### Seq Len = 4096

| Stage | BS=16 | BS=32 | BS=64 | BS=128 | BS=256 |
|-------|-------|-------|-------|--------|--------|
[...]

PR #906, opened 2026-01-21
[tx] add per-layer gradient checkpointing with scan for memory-efficient training

raulchen:
## Summary
  - Add per-layer gradient checkpointing using `jax.lax.scan` with `jax.checkpoint`
  - Reduces peak memory by ~num_layers factor (only one layer's activations held during backward)
  - Enable via `gradient_checkpointing=True` in model config

  ## Implementation
  Uses `jax.lax.scan` so XLA compiles ONE loop body and reuses buffers during backward recomputation. With a Python loop, XLA unrolls N separate checkpoint regions and can't optimize buffer reuse across them.

  Tradeoff: requires stacking all layer weights once per forward pass. This is acceptable because checkpointing already trades
compute for memory.

  ## Test plan
  - [x] Forward outputs match with/without checkpointing
  - [x] Hidden states length and values match
  - [x] `is_training=False` uses standard path with KV cache
  - [x] Gradients match between checkpointed and non-checkpointed paths

Benchmark results https://github.com/NovaSky-AI/SkyRL/issues/891#issuecomment-3775792201


skyrl-tx/tx/models/llama3.py
```diff
@@ -217,28 +218,34 @@ def __call__(
...
-        all_hidden_states: list[jax.Array] = []
-        updated_keys, updated_values = [], []
-
-        for layer_idx, layer in enumerate(self.layers):
-            if output_hidden_states:
-                all_hidden_states.append(hidden_states)
 
-            hidden_states, (k, v) = layer(
+        if is_training and self.config.gradient_checkpointing:
+            hidden_states, all_hidden_states = forward_layers_checkpointed(
```
pcmoritz: Can we unify these two methods into a single method `forward_layers` that will do the right thing depending on the parameters that are passed in? The reasoning is that model authors should not be concerned with the difference between checkpointing and non-checkpointing, they should just be able to write the model, and also we would want to avoid duplicating the code into all the model definitions.

Internally you can of course still have `_forward_layers_checkpointed` and `_ forward_layers` and implement the "public" forward_layers function in terms of those.

skyrl-tx/tx/models/qwen3.py
```diff
@@ -405,6 +412,7 @@ def __call__(
         output_hidden_states: bool | None = None,
         adapter_indices: jax.Array | None = None,
         kv_cache: KVCache | None = None,
+        is_training: bool = False,
```
pcmoritz: An alternative way to do this would be to use https://flax.readthedocs.io/en/v0.8.3/api_reference/flax.experimental.nnx/module.html#flax.experimental.nnx.Module.train (and https://flax.readthedocs.io/en/v0.8.3/api_reference/flax.experimental.nnx/module.html#flax.experimental.nnx.Module.eval) to switch the module to train/eval mode. This would have some benefits: (a) it is the standard way to do something like this (similar to how pytorch does it, there you see this kind of pattern a lot), (b) if we are going to add dropout in the future, handling training / eval like this will be much easier and (c) we don't need to pass the flags around if we want to access similar attributes. It is less explicit though, and we would need to provide our own attribute to set. Curious about your thoughts :)

skyrl-tx/tests/models/test_models_common.py
```diff
@@ -19,6 +19,16 @@
 MODEL_IDS = ["llama3", "qwen3"]
 
 
+def create_model(model_name, config_cls, model_cls, mesh_axes):
```
pcmoritz: If we want to introduce a separate function for creating a model without loading the checkpoint, we should reuse the code in `load_model` too.

skyrl-tx/tests/models/test_models_common.py
```diff
@@ -37,6 +47,69 @@ def load_model(tmp_dir, model_name, config_cls, model_cls, mesh_axes, *, loss_ch
...
+        attention_mask = jnp.ones((batch_size, seq_len), dtype=jnp.int32)
+
+        # Run without checkpointing
+        config.gradient_checkpointing = False
+        model.train()
+        out_no_ckpt = model(input_ids, attention_mask=attention_mask)
+        logits_no_ckpt = model.compute_logits(out_no_ckpt.last_hidden_state)
+
+        # Run with checkpointing
+        config.gradient_checkpointing = True
```
pcmoritz: I don't think this does what you might think it does -- after jax jit compiled the model, just changing the config won't be effective. The safest way to do this will be to create a new model with the new config.

If you need to create a new model, you need to make sure they have the same weights, maybe just use `load_model` with a checkpoint and remove create_model?

skyrl-tx/tests/models/test_models_common.py
```diff
@@ -37,6 +47,69 @@ def load_model(tmp_dir, model_name, config_cls, model_cls, mesh_axes, *, loss_ch
...
+
+        batch_size, seq_len = 2, 8
+        input_ids = jax.random.randint(jax.random.key(0), (batch_size, seq_len), 0, config.vocab_size)
+        attention_mask = jnp.ones((batch_size, seq_len), dtype=jnp.int32)
+
+        config.gradient_checkpointing = False
+        model.train()
+        out_no_ckpt = model(input_ids, attention_mask=attention_mask, output_hidden_states=True)
+
+        config.gradient_checkpointing = True
```
pcmoritz: Same comment here as above

skyrl-tx/tests/models/test_models_common.py
pcmoritz: Actually I might be wrong about this, but it would be good to check that the gradient_checkpointing branch is indeed taken.

pcmoritz (approved):

raulchen:
closing in favor of https://github.com/NovaSky-AI/SkyRL/pull/996

PR #912, opened 2026-01-21
Fix the probability diff operation

ashish230897:
This PR adds a small fix to the `prob_diff` variable used to compare the difference between trainer and generator log probabilities.


erictang000:
hmm @ashish230897, maybe what would be helpful here as well is just removing the `exp()` operation entirely and just recording the raw absolute logprob diffs, since this is what would be the most useful for tracking rollout/trainer logprob mismatches. wdyt? 

then we could follow the gemini suggestion in renaming these to just logprob diffs rather than recording the probability drift

ashish230897:
yes makes sense. Other works too generally plot the abs of the logprob diffs and not its exp, I will add this change.

ashish230897:
Added this change.

erictang000 (approved):
thanks!

PR #918, opened 2026-01-22
[tx] Use right padding of sequences during sampling prefill

pcmoritz:
This is in preparation for getting https://github.com/NovaSky-AI/SkyRL/pull/879 merged (the cudnn attention backend doesn't support left padding).


pcmoritz:
replaced with https://github.com/NovaSky-AI/SkyRL/pull/927

PR #898, opened 2026-01-19
[Tinker] Move weight sync logic into dispatch (out of trainer.py)

tyler-griggs:
## Summary

This PR moves the weight sync logic from trainer.py into WorkerDispatch, providing a cleaner API for syncing weights to inference engines.

### Changes:
- Add `save_weights_for_sampler()` method to WorkerDispatch that handles:
  - Preparing GPU state for weight sync
  - Waking up inference engine for weights
  - Broadcasting weights to inference engines  
  - Finishing weight sync and offloading model
  - Waking up inference engine for KV cache
- Add `inference_engine_client` parameter to WorkerDispatch constructor
- Add `set_inference_engine_client()` method for setting client after construction
- Simplify trainer.py weight sync calls to single `dispatch.save_weights_for_sampler()` call

### Benefits:
- Cleaner separation of concerns - trainer doesn't need to know about colocation details
- Single method call replaces 5-6 lines of boilerplate in trainer
- Encapsulates the weight sync protocol in one place

## Test plan
- [x] CPU tests pass (6/6)
- [x] GPU tests for weight sync

Generated with help from [Claude Code](https://claude.com/claude-code)

PR #922, opened 2026-01-22
[Tinker] Add save_weights_for_sampler() to WorkerDispatch

tyler-griggs:
## Summary
Adds `save_weights_for_sampler()` to WorkerDispatch as the single entry point for syncing policy weights to inference engines before sampling. Now, calls to `save_weights_for_sampler()` are the trigger for weight sync, rather than explicit weight synchronization logic in the trainer. This aligns with the Tinker API pattern where users explicitly call save_weights_for_sampler() after training and before sampling.

###  Changes:
  - WorkerDispatch: Added `save_weights_for_sampler()` (to reflect the Tinker library) that handles the full weight sync flow:
    - Prepares GPU state (offloads optimizer, keeps model on GPU)
    - Wakes inference engine for weight transfer (colocate_all only)
    - Broadcasts weights to inference engines
    - Offloads model after sync (colocate_all only)
    - Wakes inference engine for KV cache (colocate_all only)
  - Trainer: Replaced two explicit weight sync blocks with dispatch.save_weights_for_sampler():
    - After checkpoint load (before first generation)
    - After each training step (before next generation)

## Testing
Added `test_save_weights_for_sampler` tests, all passing.

### DIFF
--- a/skyrl-train/skyrl_train/trainer.py
+++ b/skyrl-train/skyrl_train/trainer.py
@@ -573,7 +573,7 @@ class RayPPOTrainer:
         Note: For new code, prefer using dispatch.save_weights_for_sampler() which
         handles the full weight sync protocol including offload/backload.
         This method is kept for backward compatibility with subclasses.
-        TODO(tgriggs): Remove this method when migration is compelte.
+        TODO(tgriggs): Remove this method when migration is complete.
         """
         return self.policy_model.async_run_ray_method(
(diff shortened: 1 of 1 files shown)


erictang000:
was this intended to be added since there's already https://github.com/NovaSky-AI/SkyRL/pull/898?

tyler-griggs:
Sorry I just got confused :D

PR #909, opened 2026-01-21
[wip] Fix loss reduction

justinvyu:

PR #925, opened 2026-01-23
[skyrl-train] Fix loss reduction by moving normalization to the advantage computation

justinvyu:
## Summary

The previous implementation for ppo policy loss reduction had a "mean of means" bias — when computing token-mean loss across micro-batches and workers with varying token counts, the naive averaging gave incorrect results where:
* microbatches with fewer tokens are weighted more heavily (since we take a mean across microbatches within a minibatch)
    * `Micro-batch 1: 100 tokens, average loss = 0.5, micro-batch 2: 900 tokens, average loss = 0.3`
    * -> `Naive mean: (0.5 + 0.3) / 2 = 0.4, Correct token-mean: (100×0.5 + 900×0.3) / 1000 = 0.32`
* worker minibatches with fewer tokens were weighted more heavily (since DDP all-reduce takes a mean across minibatches)
    * Same example as above, but the average is across workers instead.

After this PR, `ppo_policy_loss` used within `forward_backward` now just sums the per-token loss for all sequences and relies on the advantages passed in by the user to handle the loss normalization.

This aligns with [Tinker semantics](https://tinker-docs.thinkingmachines.ai/losses#proximal-policy-optimization-ppo):

> Notice that for all objectives we sum the token-level losses over the sequence length unlike some other loss implementations. If you would like to explore different aggregation schemes, you can include that in the advantage tensor computation.

Example for `loss_reduction="token_mean"`:
* Move the `1/num_minibatch_tokens` normalization into the advantage: `loss = sum( -advantage_i * ratio_i for i in range(num_minibatch_tokens) ) / num_minibatch_tokens`
* -> `sum( -(advantage_i / num_minibatch_tokens) * ratio_i for i in range(num_minibatch_tokens) )`

### DDP all-reduce

DDP/FSDP defaults to a mean all-reduce for gradients across workers. This PR counteracts this by multiplying by the DP world size.

## Additional details

This was the first attempt: https://github.com/NovaSky-AI/SkyRL/pull/909

This method was to track total tokens and then do one big normalization at the `optim_step` in order to get an average per-token loss. But, we decided to align with Tinker's way of just summing up the loss at the end, and pushing any loss normalization to the user's advantage calculation.

The benefit is that users have full control of customizing their loss reduction strategy, rather than having it happen in our opaque `forward_backward`, `optim_step` implementation which would require some configuration argument that diverges from tinker's API. For example, we would need to add a config somewhere to determine how to average/sum the loss:

```python
client.forward_backward(...)
client.optim_step(..., loss_reduction="token_mean")  # no longer tinker compatible
```

## Follow-up work

The `ppo_critic_loss` has the same problem but is not as important as the policy loss.

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/925" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/skyrl-train/tests/gpu/gpu_ci/test_megatron_worker.py
+++ b/skyrl-train/tests/gpu/gpu_ci/test_megatron_worker.py
@@ -76,13 +76,14 @@ def get_test_training_batch(batch_size=4) -> TrainingInputBatch:
 
     sequences = [tokenizer.encode(sentence) for sentence in sentences]
     attention_masks = [[1] * len(seq) for seq in sequences]
-    num_actions = 10
+    num_actions = 15
     # max seq len 1 longer than the longest sequence so we always have some padding
     max_seq_length = max([len(seq) for seq in sequences]) + 7
 
     pad_token_id = tokenizer.pad_token_id
     pad_before = [4, 0, 1, 6] * num_repeats
     pad_after = [max_seq_length - len(seq) - pad_before[i] for i, seq in enumerate(sequences)]
+    loss_masks = torch.stack([torch.cat([torch.ones(num_actions - pad_after[i]), torch.zeros(pad_after[i])]) for i in range(batch_size)])
 
     for i, (pad_before, pad_after) in enumerate(zip(pad_before, pad_after)):
         sequences[i] = [pad_token_id] * pad_before + sequences[i] + [pad_token_id] * pad_after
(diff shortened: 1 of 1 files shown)


skyrl-train/skyrl_train/workers/worker.py
```diff
@@ -778,17 +778,15 @@ def _forward_backward_micro(self, experience: Experience) -> Dict[str, float]:
...
-        # Scale accumulated gradients by 1/N to get correct average
-        if self._micro_batches_accumulated > 0:
-            scale = 1.0 / self._micro_batches_accumulated
-            for param in self.model.parameters():
-                if param.grad is not None:
-                    param.grad.mul_(scale)
+        # Scale gradients by data parallelism size to undo the DDP all-reduce mean.
+        for param in self.model.parameters():
+            if param.grad is not None:
+                param.grad.mul_(self.strategy.world_size)
```
justinvyu: we could do this at the advantage computation level, but i thought it was a bit weird to have ddp all-reduce implementation details there so i separated it to be here.

erictang000: yeah i agree that this is the right separation

justinvyu:
Closing in favor of https://github.com/NovaSky-AI/SkyRL/pull/1296

PR #927, opened 2026-01-23
[tx] Use right padding of sequences during sampling prefill, redux

pcmoritz:
This is in preparation for merging https://github.com/NovaSky-AI/SkyRL/pull/879.
Updated version of https://github.com/NovaSky-AI/SkyRL/pull/918, rebased on top of https://github.com/NovaSky-AI/SkyRL/pull/919

The reason we do this is, while it is slightly less natural for sampling prefill to be left aligned (right padded), it makes things more uniform, since that's the same alignment used during training, and the jax cudnn flash attention doesn't currently support left padding as far as we are aware. It is a small change and actually a little simpler, so what is not to like about it. Plus the work on the PR uncovered a bug in the prompt logprobs.

### DIFF
--- a/skyrl-tx/tests/models/test_qwen3_generate.py
+++ b/skyrl-tx/tests/models/test_qwen3_generate.py
@@ -59,7 +59,6 @@ def test_qwen3_generate():
             types.SamplingParams(max_tokens=2, temperature=0.0, seed=42, stop_tokens=[6149]),
         ]
 
-        # Batch with right-padding for our implementation
         batch = tokenizer(inputs, return_tensors="pt", padding=True)
         result = model.generate(
             batch.input_ids.numpy(),


skyrl-tx/tests/models/test_qwen3_generate.py
```diff
@@ -97,12 +105,27 @@ def test_qwen3_generate():
                     f"Ours: {our_logprob}, HF: {expected_logprob}, diff: {abs(our_logprob - expected_logprob)}"
                 )
 
+        # Compare prompt_logprobs against HuggingFace forward pass
```
pcmoritz, line 108: This is a regression test, it is actually failing before this PR, because when we extracted the prompt logprobs, we assumed right padding already (which was a bug)

PR #946, opened 2026-01-25
Add Pydantic Configuration System with CLI Override Support

tyler-griggs:
## 🎯 Overview

This PR introduces a comprehensive **Pydantic-based configuration system** for SkyRL training while maintaining **full backward compatibility** with existing YAML/Hydra workflows.

### 1. Type-Safe Pydantic Models
- 30+ Pydantic model classes with full type annotations
- 1:1 mapping to `ppo_base_config.yaml` structure
- IDE autocomplete and type checking support
- Runtime validation of configuration values

### 2. Unified Entry Point
- New `run_training(cfg: Union[DictConfig, SkyRLConfig])` function
- Accepts both YAML-based DictConfig and Python-based Pydantic configs
- Automatic type detection and conversion

### 3. CLI Override Support ✨
- NEW: CLI overrides work with Python configs!
- Syntax: `--trainer.epochs=30 --trainer.policy.model.path=Qwen/Qwen2.5-7B`
- Supports all Python types (int, float, bool, strings, lists)

### 4. Compositional Config Building
- RLlib-style nested configuration construction
- More readable than mutation-based style

### 5. Full Backward Compatibility
- All existing `.sh` scripts work unchanged
- YAML + Hydra CLI overrides still supported

## 📦 What's New

- `skyrl_train/config/configs.py` - Pydantic models (800+ lines)
- `skyrl_train/config/cli_overrides.py` - CLI override utilities  
- `examples/gsm8k/run_gsm8k.py` - Python-based config example
- `examples/gsm8k/test_run_gsm8k.py` - Test config
- `PYDANTIC_CONFIG_PROJECT.md` - Full documentation

## 🧪 Testing

✅ Successfully ran full training with Pydantic config on 4 GPUs
✅ Verified backward compatibility with YAML configs
✅ Tested CLI override parsing and application

See `PYDANTIC_CONFIG_PROJECT.md` for complete documentation!

### DIFF
--- a/skyrl-train/examples/gsm8k/run_gsm8k.py
+++ b/skyrl-train/examples/gsm8k/run_gsm8k.py
@@ -32,7 +32,7 @@ Usage:
 import os
 from pathlib import Path
 
-from skyrl_train.config.configs import create_default_config, SkyRLConfig
+from skyrl_train.config.configs import SkyRLConfig
 from skyrl_train.entrypoints.main_base import run_training
 
 
--- a/skyrl-train/skyrl_train/config/cli.py
+++ b/skyrl-train/skyrl_train/config/cli.py
@@ -75,114 +75,29 @@ def apply_overrides(config: SkyRLConfig, overrides: List[str]) -> SkyRLConfig:
     return config
 
 
-def create_override_parser(description: str = "SkyRL Training") -> argparse.ArgumentParser:
+def collect_overrides(args: Optional[List[str]] = None) -> List[str]:
     """
-    Create an ArgumentParser that captures remaining args as overrides.
+    Collect config overrides from command-line arguments.
+
+    Overrides are arguments in --key.path=value format.
 
     Args:
-        description: Description for the argument parser
+        args: Optional argument list (uses sys.argv[1:] if None)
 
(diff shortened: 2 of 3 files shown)


erictang000:
thoughts on making these just simple python dataclasses rather than opting for full blown pydantic configs?

just feeling that pydantic sometimes comes with heavy weight and annoying validation which can slow down projects (for example, it's sometimes pretty annoying for testing to construct valid pydantic configs that have just a subset of arguments). We do a lot of config validation in skyrl-train already, and adding new fields to a dataclass is easier. We could use pydantic selectively for some specific configs that need extra validation, which seems like a common pattern.

For example, vLLM and huggingface [EngineArgs](https://github.com/vllm-project/vllm/blob/8caffd92dfb011733b3c71c76756a687f64eb3ae/vllm/engine/arg_utils.py#L350) and [TrainingArgs](https://github.com/huggingface/transformers/blob/dfe30827b8ebdd974eb7ce69c7d5d8cf8e6cf852/src/transformers/training_args.py#L199)  use python dataclasses for the top level config, but have some specific configs that use pydantic for validation.

PR #982, opened 2026-01-28
[skyrl-train] temporarily fix ray serialization slowdown during training

erictang000:
After #859 pulled the training loop out of the workers, additional serialization overhead was added per mini batch, causing slowdowns of the `policy_train` method.

This PR does the following to fix that slow down:

1.  Add `forward_backward_from_staged`, `dispatch_from_staged`, and `stage_data` to stage the training data once in the ray object store and index into the object in the object store on each worker to avoid overhead 
2. Update `TensorBatch` serialization to use numpy by default, falling back to previous torch.load/save implementation for types unsupported in numpy.
3. Update default `dispatch` method to also pass `ObjectRef`s to workers, reducing serialization being repeated across workers within a DP group.

```python

        # Stage full batch in object store ONCE to avoid repeated serialization
        data_ref = self.dispatch.stage_data(data)

        # Training loop over epochs and mini-batches
        for _epoch in range(self.cfg.trainer.update_epochs_per_batch):
            num_mini_batches = len(data) // mini_batch_size
            for local_step in range(num_mini_batches):
                start_idx = local_step * mini_batch_size
                end_idx = (local_step + 1) * mini_batch_size

                # Workers fetch from object store and slice locally
                status = self.dispatch.forward_backward_from_staged(model, data_ref, start_idx, end_idx)
```

### Result
For the first batch of the qwen3-1.7B DAPO script, we can recover the previous training time (after #859, this blew up from ~240 -> 1600)
image

With fixes, both training and fwd pass are now faster than prior implementation!

image

image

### DIFF
--- a/skyrl-train/tests/cpu/distributed/test_dispatch.py
+++ b/skyrl-train/tests/cpu/distributed/test_dispatch.py
@@ -1,5 +1,5 @@
 """
-uv  run --isolated --extra dev pytest tests/cpu/distributed/test_dispatch.py
+uv run --isolated --extra dev pytest tests/cpu/distributed/test_dispatch.py
 """
 
 from skyrl_train.training_batch import TrainingInputBatch


skyrl-train/skyrl_train/distributed/dispatch.py
```diff
@@ -156,6 +160,52 @@ def sync_collect(cls, actor_infos: List[ActorInfo], object_refs: List[ObjectRef]
...
 
+    @classmethod
+    def dispatch_from_staged(
+        cls,
+        actor_infos: List[ActorInfo],
+        method: str,
+        data_ref: ObjectRef,
+        start_idx: int,
+        end_idx: int,
+    ) -> List[ObjectRef]:
```
SumanthRH, line 166: This should have a `kwargs` argument for pass-through kwargs similar to `MeshDispatch.dispatch` method

skyrl-train/examples/algorithms/dapo/run_dapo_qwen3_1.7b_aime.sh
```diff
@@ -94,7 +94,7 @@ uv run --isolated --extra vllm -m examples.algorithms.dapo.main_dapo \
   generator.gpu_memory_utilization=0.8 \
   trainer.logger="$LOGGER" \
   trainer.project_name="dapo_aime" \
-  trainer.run_name="dapo_qwen3_1.7b_base" \
+  trainer.run_name="dapo_qwen3_1.7b_base_6a9ab8" \
```
SumanthRH: revert?

skyrl-train/skyrl_train/training_batch.py
```diff
@@ -133,15 +132,25 @@ def device(self) -> torch.device:
...
         batch_dict = {}
         for key, value in self.items():
-            buffer = io.BytesIO()
-            torch.save(value, buffer)
-            batch_dict[key] = buffer.getvalue()
+            if value is None:
+                batch_dict[key] = None
+            else:
+                # Fast serialization: direct memory copy via numpy
+                arr = value.numpy()
```
SumanthRH: This will break for dtypes not supported by numpy

```ipython
In [5]: torch.tensor([1, 2, 3], dtype=torch.bfloat16).numpy()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[5], line 1
----> 1 torch.tensor([1, 2, 3], dtype=torch.bfloat16).numpy()

TypeError: Got unsupported ScalarType BFloat16
```

skyrl-train/skyrl_train/distributed/dispatch.py
```diff
@@ -156,6 +160,52 @@ def sync_collect(cls, actor_infos: List[ActorInfo], object_refs: List[ObjectRef]
         assert all(obj is None for obj in all_objects), "Got a mix of `None` and non-`None` objects"
         return
 
+    @classmethod
+    def dispatch_from_staged(
```
SumanthRH, line 164: Let's also add a test for this new dispatch method

erictang000: done!

skyrl-train/skyrl_train/distributed/dispatch.py
erictang000: done

skyrl-train/skyrl_train/training_batch.py
erictang000: added fallback to torch.load/save!

skyrl-train/skyrl_train/training_batch.py
```diff
@@ -151,10 +173,24 @@ def __getstate__(self):
...
+                # Fallback path: torch.load for unsupported dtypes
+                buffer = io.BytesIO(value["data"])
+                self[key] = torch.load(buffer, weights_only=True)
+            else:
+                # Fast path: reconstruct from numpy bytes
+                # Also handles legacy format without "format" key
+                arr = np.frombuffer(value["data"], dtype=np.dtype(value["dtype"]))
+                arr = arr.reshape(value["shape"])
+                # Convert to tensor (makes a copy, which is needed since frombuffer is read-only)
+                self[key] = torch.from_numpy(arr.copy())
```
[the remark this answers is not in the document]
erictang000: i think this is ok to break this one time

erictang000:
With fixes, both training and fwd pass are now faster than prior implementation!

image

image

PR #994, opened 2026-01-30
[tx] Fix pyproject.toml on macOS

pcmoritz:
This is a slightly more minimal version of https://github.com/NovaSky-AI/SkyRL/pull/991


pcmoritz:
I think https://github.com/NovaSky-AI/SkyRL/pull/991 is actually better because it doesn't need the override

PR #996, opened 2026-01-30
[tx] Per-layer gradient checkpointing with stacked decoder layers

raulchen:
This PR implements **per-layer gradient checkpointing** to reduce training memory, enabled by a **stacked layer weights**  architecture that allows using `jax.lax.scan` with `jax.checkpoint` over transformer layers.

  ## Key Changes

  ### 1. Per-Layer Gradient Checkpointing

  Checkpoint each transformer layer individually using `jax.lax.scan` with `jax.checkpoint`. Memory grows O(1) with layer count   instead of O(n) - only one layer's activations held in memory during backward pass.

  ### 2. Stacked Layer Weights

  Store all layer parameters as stacked arrays `(num_layers, ...)` instead of separate layer modules. This enables scan over  uniform data structures.

  **StackedDecoderLayers**: Manages stacked parameters with scan-based forward pass
  **MultiStackedDecoderLayers**: Supports heterogeneous architectures (e.g., DeepSeek with dense + MoE layer groups)

  ### 3. Checkpoint Loading

  `unstack_state()` transforms stacked format to per-layer view using `ArrayRef` write-through variables for transparent
  checkpoint loading.

  ## Architecture Support

  - **Homogeneous models** (Llama3, Qwen3): Single `StackedDecoderLayers`
  - **Heterogeneous models** (DeepSeek): Multiple groups via `MultiStackedDecoderLayers`

# Benchmark 

```
  ┌────────────────────────┬───────────────┬─────────────┬──────────┬───────────────┐
  │          Mode          │    Branch     │ Peak Memory │ JIT Time │ Post-JIT Time │
  ├────────────────────────┼───────────────┼─────────────┼──────────┼───────────────┤
  │ Sample (bs=16, seq=4k) │ main          │ 6,737 MiB   │ 331.5s   │ 21.28s        │
  ├────────────────────────┼───────────────┼─────────────┼──────────┼───────────────┤
  │                        │ stack-weights │ 6,727 MiB   │ 406.2s   │ 24.18s        │
  ├────────────────────────┼───────────────┼─────────────┼──────────┼───────────────┤
  │ Train (bs=1, seq=4k)   │ main          │ 22,451 MiB  │ 331.1s   │ 0.85s         │
  ├────────────────────────┼───────────────┼─────────────┼──────────┼───────────────┤
  │                        │ stack-weights │ 6,699 MiB   │ 227.2s   │ 0.51s         │
  └────────────────────────┴───────────────┴─────────────┴──────────┴───────────────┘
```

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/996" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


skyrl-tx/tx/layers/lora.py
```diff
@@ -1,13 +1,37 @@
...
+from jax.core import Tracer
 
-from tx.utils.models import filter_lora
+from tx.utils.models import filter_lora, is_stacked_lora_path
 from tx.layers.util import Param, prepare_routing, ragged_dot
 from tx.models.types import ModelForCausalLM
 from tx.tinker.types import LoraConfig
 
 
+def _get_sharding_spec(arr: jax.Array):
```
pcmoritz: It will be good to get rid of this function (so we also don't depend on the `jax.core.Tracer` class, which looks like more of an internal class).

The best way to handle this I think is to introduce a required `sharding` argument in the constructors of the classes `LoRA{Layer}` and then call `nnx.with_partitioning(...)` inside the classes. This way the sharding is always required and we can even get rid of the assert.

I'm happy to make a separate PR with the refactor (do let me know if you see a problems with this).

skyrl-tx/tx/models/deepseekv3.py
```diff
@@ -444,10 +445,29 @@ def __call__(
         return hidden_states, updated_cache
 
 
+class DeepseekV3DenseDecoderLayer(DeepseekV3DecoderLayer):
```
pcmoritz: Can we simplify this a little and get rid of these classes? What I have in mind is to just have the class DeepseekV3DecoderLayer from above and do
```python
class DeepseekV3DecoderLayer(nnx.Module):
    def __init__(self, mlp_cls: Callable[[ModelConfig, ...], nnx.Module], config: DeepseekV3Config, *, dtype: jnp.dtype, rngs: nnx.Rngs) -> None:
        # ...
        self.mlp = mlp_cls(config, dtype=dtype, rngs=rngs)
```
and then down below where the stacked layers are created, do
```python
def create_dense_layer(rngs: nnx.Rngs) -> DeepseekV3DecoderLayer:
    return DeepseekV3DecoderLayer(DeepseekV3MLP, config, dtype=dtype, rngs=rngs)
```

skyrl-tx/tx/models/deepseekv3.py
```diff
@@ -483,29 +518,77 @@ def __call__(
...
             )
-            updated_keys.append(k)
-            updated_values.append(v)
+            all_hidden_states.extend(moe_hidden_states)
 
         hidden_states = self.norm(hidden_states)
         if output_hidden_states:
             all_hidden_states.append(hidden_states)
 
+        # Merge KV caches from dense and MoE layers
```
pcmoritz: This merits a `KVCache.concatenate()` function I think, it could even be implemented generically like
```
@staticmethod
def concatenate(*kv_caches: KVCache) -> KVCache:
    return KVCache(
        keys=jnp.concatenate(kv_cache.key for kv_cache in kv_caches),
        values=jnp.concatenate(kv_cache.value for kv_cache in kv_caches),
        cache_position=kv_caches[0].cache_position,
    )
```
The reasoning is that we try to keep this kind of complexity out of the modeling files as much as possible.

skyrl-tx/tx/models/deepseekv3.py
```diff
@@ -483,29 +518,77 @@ def __call__(
 
         hidden_states = self.embed_tokens(input_ids, adapter_indices=adapter_indices)
         all_hidden_states: list[jax.Array] = []
-        updated_keys, updated_values = [], []
 
-        for layer_idx, layer in enumerate(self.layers):
-            if output_hidden_states:
-                all_hidden_states.append(hidden_states)
+        # Split KV cache for dense and MoE layers
```
pcmoritz: Similar to the below, it would be nice to write this as something like
```python
dense_kv_cache, moe_kv_cache = kv_cache.split(self.num_dense_layers, self.num_moe_layers)
```
If you wanted, you could probably implement this genericlly by lifting https://docs.jax.dev/en/latest/_autosummary/jax.numpy.split.html, something like
```python
def split(kv_cache: KVCache, indices_or_sections):
    keys_list = jnp.split(kv_cache.keys, indices_or_sections, axis=0)
    values_list =  jnp.split(kv_cache.keys, indices_or_sections, axis=0)
    return [KVCache(keys=keys, values=values, cache_position=kv_cache.cache_position) for (keys, values) in zip(keys_list, values_list))
```
but not sure if that's actually better. Otherwise just supporting the case of two splits is totally fine.

skyrl-tx/tx/layers/lora.py
```diff
@@ -1,13 +1,37 @@
...
+        aval = jax.typeof(arr)
+        if hasattr(aval, "sharding") and aval.sharding is not None:
+            return aval.sharding.spec
+        return None
+    if arr.sharding is not None:
+        return arr.sharding.spec
+    return None
+
+
+def _adapter_index(is_stacked: bool, adapter_index: int):
```
pcmoritz: I think we should change this function to
```python
def get_adapter_idx(path: str, adapter_index: int):
    is_stacked = _is_stacked(path)
    return (slice(None), adapter_index) if is_stacked else (adapter_index,)
```
and then use it everywhere in the PR. I believe all the call sited of  `_adapter_index` are already of this form, and there are lots more call sites where we can get rid of a pattern like
```python
if is_stacked:
     # Process stacked weights
else:
     # Process unstacked weights
```
This can always be done like
```python
idx = get_adapter_idx(path, adapter_index)
# Process weights with weights[idx,...]
```

skyrl-tx/tx/layers/lora.py
raulchen: yeah, this makes sense.

skyrl-tx/tx/utils/models.py
```diff
@@ -247,18 +362,32 @@ def get_optimizer(optimizer_name: OptimizerName, optimizer_args: dict) -> optax.
             raise ValueError("The 'learning_rate' key must be provided in optimizer_args.")
 
 
+def _lora_slice(is_stacked: bool, adapter_index: int, rank: int, is_lora_A: bool) -> tuple:
```
pcmoritz: Do we actually need this function? Can't we just do things like
```python
idx = get_adapter_idx(path, adapter_index)
p.at[idx, ..., :, :rank]
p.at[idx, ..., :rank, :]
```

skyrl-tx/tx/utils/models.py
```diff
@@ -76,19 +78,115 @@ def get_model_class(config: PretrainedConfig) -> Callable[..., nnx.Module]:
...
+    """
+    path_strs = [p.key if hasattr(p, "key") else str(p) for p in path]
+    # Check for split stacked layer names (DeepSeekV3)
+    if "dense_layers" in path_strs or "moe_layers" in path_strs:
+        return True
+    # Check for regular stacked layers (Qwen3/Llama3)
+    if "layers" not in path_strs:
+        return False
+    layers_idx = path_strs.index("layers")
+    if layers_idx + 1 < len(path_strs) and path_strs[layers_idx + 1].isdigit():
```
pcmoritz: Do you know why this case is needed? If it isn't, we only need one method here and can always use the logic of `is_stacked_lora_path`.

My rationale for this question is, if we always stack the layers that are stackable, a case like `('model', 'layers', '0', 'self_attn', ...)` should never happen, right?

skyrl-tx/tx/layers/lora.py
raulchen: good idea. code could be simpler

skyrl-tx/tx/utils/models.py
raulchen: this was added in the mid-ground state when both stacked and non-stacked are supported. 
Should be non longer needed.

skyrl-tx/tx/layers/stacked.py
```diff
@@ -0,0 +1,463 @@
...
+    """A Variable providing a view into an indexed slice of a parent Variable."""
+
+    def __init__(self, parent: nnx.Variable, idx: int):
+        super().__init__(parent[idx])
+        self.set_metadata("_parent", parent)
+        self.set_metadata("_idx", idx)
+
+    def __getitem__(self, key):
+        parent, idx = self.get_metadata("_parent"), self.get_metadata("_idx")
+        return parent[idx] if key is Ellipsis else parent[idx][key]
```
pcmoritz: I don't think we need a special case for the `Ellipsis` here and can just do
```
return parent[idx][key]
```
right? Since for `key = Ellipsis` that's equivalent to `return parent[idx]`.

skyrl-tx/tx/layers/stacked.py
```diff
@@ -0,0 +1,463 @@
...
+
+    def __setitem__(self, key, value):
+        """Write through to parent when value is set via indexing."""
+        parent, idx = self.get_metadata("_parent"), self.get_metadata("_idx")
+        if key is Ellipsis:
```
pcmoritz: Similar to above, I don't think we need to two different cases here

skyrl-tx/tx/layers/stacked.py
```diff
@@ -0,0 +1,463 @@
...
+            offset += group.num_layers
+
+        raise IndexError(f"Layer index {index} not found")
+
+    def __iter__(self):
+        """Iterate over all layers across all groups."""
+        for group in self.layer_groups:
+            yield from group
+
+    def get_stacked_layers_list(self) -> list[StackedDecoderLayers]:
```
pcmoritz: I belive this is not used any more and can be removed

skyrl-tx/tx/models/deepseekv3.py
```diff
@@ -472,14 +482,24 @@ def __init__(self, config: DeepseekV3Config, *, dtype: jnp.dtype, rngs: nnx.Rngs
...
+        def create_moe_layer(rngs: nnx.Rngs) -> DeepseekV3DecoderLayer:
+            return DeepseekV3DecoderLayer(config, mlp_cls=DeepseekV3MoE, dtype=dtype, rngs=rngs)
+
+        dense_layers = StackedDecoderLayers(create_dense_layer, self.num_dense_layers, rngs)
+        moe_layers = StackedDecoderLayers(create_moe_layer, self.num_moe_layers, rngs)
+        self.layers = MultiStackedDecoderLayers(dense_layers, moe_layers)
+
         self.norm = RMSNorm(config.hidden_size, eps=config.rms_norm_eps, dtype=dtype, rngs=rngs)
 
+    def get_stacked_layers_list(self):
```
pcmoritz: I believe this is not used any more and can be removed

skyrl-tx/tx/utils/models.py
```diff
@@ -252,13 +285,14 @@ def extract_adapter_state(adapter_index: int, lora_params: nnx.GraphState, rank:
...
             return p
-        assert p.ndim in {3, 4}, f"LoRA parameters must have 3 or 4 dimensions, got shape {p.shape}"
-        if path[-2].key == "lora_A":
-            return p[adapter_index, ..., :, :rank]
-        if path[-2].key == "lora_B":
-            return p[adapter_index, ..., :rank, :]
+        assert p.ndim in {3, 4, 5}, f"LoRA parameters must have 3-5 dimensions, got shape {p.shape}"
+        idx = get_adapter_idx(path, adapter_index)
+        if key == "lora_A":
+            return p[idx + (..., slice(None, rank))]
```
pcmoritz: It might be a little more readable to write this as `p[*idx, ..., :, :rank]` and similar for the others and also below

skyrl-tx/tx/utils/models.py
```diff
@@ -99,32 +128,32 @@ def load_safetensors(
     prefix: str = "",
     filter_fn: Callable[[tuple], bool] | None = None,
 ) -> None:
+    """Load safetensors weights into a model with stacked layers."""
+    from tx.layers.stacked import unstack_state
```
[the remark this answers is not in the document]
pcmoritz: This is actually a good comment, we can also do it in a separate PR

skyrl-tx/tx/models/deepseekv3.py
```diff
@@ -460,6 +467,9 @@ class DeepseekV3Model(nnx.Module):
 
     def __init__(self, config: DeepseekV3Config, *, dtype: jnp.dtype, rngs: nnx.Rngs) -> None:
         self.config = config
+        self.num_dense_layers = config.first_k_dense_replace
+        self.num_moe_layers = config.num_hidden_layers - config.first_k_dense_replace
+        self.num_layers = config.num_hidden_layers
```
pcmoritz: `self.num_layers` isn't actually used, right? Same for the other models

skyrl-tx/tx/layers/util.py
```diff
@@ -96,12 +96,21 @@ def shard_map_ep(module: nnx.Module, func, *args):
...
     """
     graphdef, state = nnx.split(module)
-    # Extract only 'ep' dims from PartitionSpecs, replacing others with None
-    state_specs = jax.tree.map(
-        lambda s: PartitionSpec(*(p if p == "ep" else None for p in s)) if isinstance(s, PartitionSpec) else s,
-        nnx.get_partition_spec(state),
-        is_leaf=lambda x: isinstance(x, PartitionSpec),
-    )
+
+    def make_ep_spec(spec, value):
```
pcmoritz, line 100: I find this function pretty confusing, clearer is: Rename `is_stacked_lora_path` in the other file to `is_stacked_path`, and then replace `jax.tree.map` with `jax.tree.map_with_path`, and then write this function like
```python
def make_ep_spec(path, s):
        if not isinstance(s, PartitionSpec):
            return s
        # Strip leading stacking dimension if path is stacked
        dims = s[1:] if is_stacked_path(path) else s
        # Extract only 'ep' dims from PartitionSpecs, replacing others with None
        return PartitionSpec(*(p if p == "ep" else None for p in dims))

state_specs = jax.tree_util.tree_map_with_path(
        make_ep_spec, nnx.get_partition_spec(state), is_leaf=lambda x: isinstance(x, PartitionSpec)
    )
```

skyrl-tx/tests/models/test_qwen3.py
```diff
@@ -105,6 +105,28 @@ def load_lora_weights(
     jax_module.lora_ranks[...] = jax_module.lora_ranks[...].at[adapter_idx].set(rank)
 
 
+def load_stacked_lora_weights(
```
pcmoritz: This isn't needed any more with the automatic view of stacked layers as unstacked

pcmoritz:
You can revert most of the changes to the test (except for the new tests and factoring out the common methods) -- see https://github.com/NovaSky-AI/SkyRL/pull/1018 where the tests just pass with the existing code by using the stacked abstraction

skyrl-tx/tx/layers/util.py
raulchen: I tried the is_stacked_path + tree_map_with_path approach but it doesn't work for shard_map_ep. The issue is that shard_map_ep receives a sub-module (e.g. self.experts) that was extracted from a stacked layer during scan. The paths in that sub-module's state don't contain _stacked — that's the parent module's attribute name. So is_stacked_path returns False and  the leading stacking dim isn't stripped, causing a spec/tensor rank mismatch.

skyrl-tx/tx/layers/stacked.py
raulchen: this actually only supports the ... key. See the latest comment.

skyrl-tx/tx/utils/models.py
```diff
@@ -214,21 +284,19 @@ def save_lora_checkpoint(
         adapter_index: Index of the adapter to save
         output_path: Path to save the checkpoint tar.gz file
     """
-    _, lora_params, _ = nnx.split(model, model.is_lora_param, ...)
-
-    adapter_lora_params = extract_adapter_state(adapter_index, lora_params, adapter_config.rank)
-
```
raulchen, line 219: we used to call extract_adapter_state and pass the result to save_safetensors. 
This is broken because extract_adapter_state returns a GraphState rather than a Module, which doesn't support the ArrayRef write through.

skyrl-tx/tests/models/test_deepseekv3_lora_training.py
```diff
@@ -70,7 +39,7 @@ def test_lora_training_moe_rank_normalized():
         init_lora_adapter(model, adapter_index=0, lora_config=LoraConfig(rank=16, alpha=16, seed=0))
         init_lora_adapter(model, adapter_index=1, lora_config=LoraConfig(rank=8, alpha=8, seed=1))
 
-        optimizer = nnx.Optimizer(model, optax.adamw(1e-4), wrt=model.is_lora_param)
+        optimizer = nnx.Optimizer(model, optax.adamw(1e-3), wrt=model.is_lora_param)
```
raulchen, line 42: Stacked layers use scan which changes LoRA initialization order,
    producing slightly different he_uniform values. With bf16 precision,
    1e-4 learning rate needs >10 steps to register a loss change.
    Bump to 1e-3 for reliable convergence within 10 steps.

raulchen:
closing in favor of https://github.com/NovaSky-AI/SkyRL/pull/1083 and stacked PRs

PR #997, opened 2026-01-31
[tx] Make sharding explicit in LoRA constructors

pcmoritz:
This is in preparation for merging https://github.com/NovaSky-AI/SkyRL/pull/996, so we don't need to depend on the jax tracer. It is also slightly cleaner this way and the assert is not needed any more, since the error is "defined away".

It also adds the FSDP sharding for llama3.

### DIFF
--- a/skyrl-tx/tests/models/test_models_common.py
+++ b/skyrl-tx/tests/models/test_models_common.py
@@ -13,7 +13,7 @@ from tx.models.qwen3 import Qwen3ForCausalLM
 from tx.utils.models import load_safetensors
 
 MODEL_PARAMS = [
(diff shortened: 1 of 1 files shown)


raulchen (approved):
we can define constants for the shardings, as they are repeated.

pcmoritz:
Do you mean across models? Within models there is generally not a lot of repetition, only a little bit, but then it is also nice that each tensor has its sharding explicitly directly next to it. I'm going to merge this for now so we can unblock the other PR, but we can think about whether there is a good way to structure this in a better way to get rid of repetition (and forgetting certain sharding) across models. Let me know if you have some ideas :)

PR #999, opened 2026-01-31
Add sampling support for Tinker SkyRL backend

tyler-griggs:
## Summary

  Implements sampling/inference support for the SkyRL Tinker backend so we can start enabling RL training workflows via the Tinker API.

  ## Key Changes

  - **Implements `sample()` method** in `SkyRLTrainBackend` using existing `InferenceEngineClient`
  - **Adds weight synchronization** via `save_weights_for_sampler()`, which just passes it along to the `WorkerDispatch`

  ## Limitations

  - Prompt logprobs not yet supported (returns `None` with warning)
  - Checkpoint paths in requests ignored (always uses current weights)
  - Single model/checkpoint at a time

### DIFF
--- a/claude/tinker-skyrl-quickstart.md
+++ b/claude/tinker-skyrl-quickstart.md
@@ -1,0 +1,146 @@
+# Running Tinker Cookbook on SkyRL
+
+A quick guide to running [tinker-cookbook](https://github.com/thinkingmachines/tinker-cookbook) recipes using SkyRL as the backend.
+
+## Prerequisites
+
+- Linux machine with NVIDIA GPUs (tested on 4xL4)
+- Python 3.12
+- [uv](https://github.com/astral-sh/uv) package manager
+
+## Setup
+
+### 1. Clone the repositories
+
+```bash
+# Clone SkyRL
+git clone https://github.com/NovaSky-AI/SkyRL.git
+cd SkyRL
+git checkout tyler/tinker-sampling-main  # or main once PR #999 is merged
+
+# Clone tinker-cookbook (in a separate directory)
+cd ~
+git clone https://github.com/thinkingmachines/tinker-cookbook.git
+```
+
+### 2. Start the Tinker API Server
+
+```bash
+cd ~/SkyRL/skyrl-tx
+
+# Clean any previous state
+rm -f tx/tinker/tinker.db
+
+# Start the server
+uv run --extra skyrl_train --extra tinker -m tx.tinker.api \
+    --base-model "Qwen/Qwen3-0.6B" \
+    --backend skyrl_train
+```
+
+The server takes ~2 minutes to initialize. Wait until you see:
+```
+INFO:     Uvicorn running on http://0.0.0.0:8000
+```
+
+### 3. Run a Tinker Cookbook Recipe
+
+In a new terminal:
+
+```bash
+cd ~/tinker-cookbook
+
+TINKER_API_KEY=tml-test uv run --with tinker --with datasets --with torch \
+    python -m tinker_cookbook.recipes.rl_loop \
+    base_url=http://localhost:8000 \
+    model_name="Qwen/Qwen3-0.6B" \
+    batch_size=8 \
+    group_size=4 \
+    lora_rank=32 \
+    max_tokens=128 \
+    save_every=5 \
+    log_path="/tmp/tinker-rl-test"
+```
+
+## Supported Models
+
+Use models from the [Qwen3 family](https://huggingface.co/Qwen):
+- `Qwen/Qwen3-0.6B` - Small, fast for testing
+- `Qwen/Qwen3-1.7B` - Medium
+- `Qwen/Qwen3-4B` - Larger, needs more VRAM
+- `Qwen/Qwen3-8B` - Requires 4+ GPUs
+
+## Configuration Options
+
+### Server options
+
+| Option | Description | Default |
+|--------|-------------|---------|
+| `--base-model` | HuggingFace model ID | Required |
+| `--backend` | Backend type (`skyrl_train` or `jax`) | `jax` |
+| `--checkpoints-base` | Checkpoint storage path | `/tmp/tx_checkpoints` |
+
+### rl_loop.py options
+
+| Option | Description | Default |
+|--------|-------------|---------|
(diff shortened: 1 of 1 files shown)


skyrl-tx/tx/tinker/backends/skyrl_train.py
```diff
@@ -57,9 +63,52 @@ def _build_config(base_model: str, config: SkyRLTrainBackendConfig, lora_config:
     cfg.trainer.policy.optimizer_config.scheduler = "constant"
     cfg.trainer.policy.optimizer_config.num_warmup_steps = 0
 
+    # Workaround: Qwen2.5-0.5B has 14 attention heads, use tensor_parallel_size=2
+    if "Qwen2.5-0.5B" in base_model:
```
pcmoritz: Should we remove this before merging? It would probably be better to expose the `.yaml` so it can be set in there, right? You could add an argument to `SkyRLTrainBackendConfig` which specifies the path to the .yaml and then load it from here -- and then it can be used like `--backend-config '{"config_path": "/path/to/yaml"}' or similar.

skyrl-tx/tx/tinker/backends/skyrl_train.py
```diff
@@ -57,9 +63,52 @@ def _build_config(base_model: str, config: SkyRLTrainBackendConfig, lora_config:
...
+    tokenizer,
+    cfg,
+    shared_pg,
+):
+    """Create inference engines for sampling using SkyRL default config.
+
+    Reads configuration from cfg.generator.* (SkyRL's config system).
+    TODO: Allow overrides via backend_config once config management is implemented.
+    """
+    from skyrl_train.inference_engines.ray_wrapped_inference_engine import create_ray_wrapped_inference_engines
```
pcmoritz: I realize trying to avoid importing `skyrl_train` will get uglier and uglier over time. We need to figure out a better way, we can make this a function https://github.com/NovaSky-AI/SkyRL/blob/afa60b715c0c8be7e972563a2634a3c0d1d2ee2d/skyrl-tx/tx/tinker/engine.py#L133 that returns the available backends, and it will only import skyrl_train if the dependencies are installed, so everything is in one centralized place. E.g.

```python
def get_backends() -> dict[str, tuple[...]]:
    backends = {}
    try:
        from tx.tinker.backends.skyrl_train import SkyRLTrainBackend, SkyRLTrainBackendConfig
        backends["skyrl_train"] = (SkyRLTrainBackend, SkyRLTrainBackendConfig)
    except ImportError:
        # print warning

    try:
        from tx.tinker.backends.jax import JaxBackend, JaxBackendConfig
        backends["jax"] = (JaxBackend, JaxBackendConfig)
    except:
        # print warning

    return backends
```

Maybe this can be made even nicer...

skyrl-tx/tx/tinker/backends/skyrl_train.py
```diff
@@ -57,9 +63,52 @@ def _build_config(base_model: str, config: SkyRLTrainBackendConfig, lora_config:
...
+    from skyrl_train.inference_engines.inference_engine_client import InferenceEngineClient
+
+    colocate_all = cfg.trainer.placement.colocate_all if hasattr(cfg.trainer.placement, "colocate_all") else False
+
+    engine_kwargs = {
```
pcmoritz: Could this code be structured such that `cfg.generator` is mostly passed through? That would make it easier going forward e.g. when new fields are added. These fields are currently implementation details that this file should better not be concerned with (the more minimal we can keep this file, the better).

skyrl-tx/tx/tinker/backends/skyrl_train.py
```diff
@@ -222,7 +284,159 @@ def sample(
...
+        if self._inference_engine_client is None:
+            error = types.ErrorResponse(
+                error="Sampling not enabled. Inference engines were not initialized (num_inference_engines=0 in SkyRL config).",
+                status="error",
+            )
+            return {req_id: error for req_id, _, _, _, _ in prepared_batch.request_batch_slices}
+
+        # 2. Validate single model
+        unique_models = set(prepared_batch.all_model_ids)
+        if len(unique_models) != 1 or list(unique_models)[0] != self._model_id:
```
pcmoritz: I think this can be slightly simplified to `unique_models != {self._model_id}`

skyrl-tx/tx/tinker/backends/skyrl_train.py
```diff
@@ -222,7 +284,159 @@ def sample(
...
+                sampling_params = prepared_batch.all_sampling_params[i]
+
+                # Convert to InferenceEngineClient format
+                params_dict = {
+                    "temperature": sampling_params.temperature,
+                    "max_tokens": sampling_params.max_tokens,
+                    "seed": sampling_params.seed,
+                }
+
+                # Handle top_k: convert -1 to None (vLLM expects None for no limit)
```
pcmoritz: It is worth checking if these are actually needed (e.g. https://github.com/vllm-project/vllm/blob/8869cd8ec1b2586df96e6556f2d2e3b3760bfb93/vllm/sampling_params.py#L159). In `external_inference.py` we are just passing it through and it has been working well afaik.

skyrl-tx/tx/tinker/backends/skyrl_train.py
```diff
@@ -222,7 +284,159 @@ def sample(
...
+                if sampling_params.top_k is not None and sampling_params.top_k != -1:
+                    params_dict["top_k"] = sampling_params.top_k
+
+                if sampling_params.top_p is not None:
+                    params_dict["top_p"] = sampling_params.top_p
+
+                # Handle stop conditions: combine stop_tokens and stop_strings
```
pcmoritz: Do these need to be handled differently? E.g. `stop` vs `stop_token_ids` https://github.com/vllm-project/vllm/blob/8869cd8ec1b2586df96e6556f2d2e3b3760bfb93/vllm/sampling_params.py#L171

skyrl-tx/tx/tinker/backends/skyrl_train.py
tyler-griggs: Yeah I'll remove! I'm hoping to rely on Sumanth's changes to use pythonic configs very soon..

PR #1001, opened 2026-01-31
[train] Pythonic Configs 1/N - Introduce configuration dataclasses for the current YAML and migrate tests

SumanthRH:
# What does this PR do?

1/2 PRs for introducing pythonic configurations in SkyRL-train. 

A short design doc is here: https://docs.google.com/document/d/1YzLPCmVZpBauRyh4l-17AUhXiXALmOX-N1gXZkA9slA

This is going to be lightweight in terms of configuration hierarchy changes to keep user experience similar. 

**Summary**: 

- Introduce  dataclasses corresponding to the current configuration hierarchy - `PolicyConfig`, `ModelConfig`, `AlgorithmConfig`, `PlacementConfig`, etc.
- Preserves backwards compatibility - internal config structure matches YAML 1:1
- Entrypoint scripts continue to use generic `DictConfig` - only migrate tests to new dataclasses. 
- Existing user scripts don’t break

### DIFF
--- a/skyrl-train/tests/gpu/gpu_ci/test_megatron_worker.py
+++ b/skyrl-train/tests/gpu/gpu_ci/test_megatron_worker.py
@@ -643,7 +643,7 @@ async def test_megatron_dp(ray_init_fixture, worker_type, tp, pp, gpus_per_node)
     print("\n\n")
     print("megatron results dp: ", results_megatron_dp)
 
-    keys_to_compare = ["policy_loss", "policy_lr", "ppo_clip_ratio", "policy_entropy", "policy_kl", "raw_grad_norm"]
+    keys_to_compare = ["policy_loss", "policy_lr", "ppo_clip_ratio", "policy_entropy", "policy_kl"]
     for i, result in enumerate(results_megatron_dp):
         for k in keys_to_compare:
             assert isinstance(result[k], (int, float)), f"{k} should be an int or float"


skyrl-train/skyrl_train/config/config.py
```diff
@@ -0,0 +1,535 @@
...
+from skyrl_gym.envs.search.env import SearchEnvConfig
+from skyrl_gym.envs.sql.env import Text2SQLEnvConfig
+
+# ---------------------------------------------------------------------------
+# Data
+# ---------------------------------------------------------------------------
+
+
+@dataclass
+class BaseConfig(ABC):
```
tyler-griggs, line 27: Do we even need `BaseConfig`? If it just provides the method `from_dict_config`, can we just make that a helper method? Would be nice if all configs didn't have to inherit it to get the one method.

skyrl-gym/skyrl_gym/envs/searchcode/env.py
```diff
@@ -4,15 +4,14 @@
...
-from omegaconf import DictConfig
 
 
 class SearchCodeEnv(BaseTextEnv):
     """
     Environment that calls multiple tools
     """
 
-    def __init__(self, env_config: DictConfig, extras: Dict[str, Any] = {}):
+    def __init__(self, env_config: Any, extras: Dict[str, Any] = {}):
```
tyler-griggs: I noticed some of these were moved to `Any` while others actually had the Union of the two possible types (e.g., the file above: `Union[SearchEnvConfig, DictConfig]`). It would be nice to avoid `Any` since we know what config it should take.

tyler-griggs (approved):
thanks!

skyrl-gym/skyrl_gym/envs/searchcode/env.py
SumanthRH: The class doesn't even use the `env_config` paramter, which is why I'd used Any. I can change the type hint to be consistent with other envs.

skyrl-train/skyrl_train/config/config.py
SumanthRH: No harm in keeping a base class here. In the next PR, I also plan to have another method for overrides, and there's possibly more functionality one could add later.

skyrl-gym/skyrl_gym/envs/searchcode/env.py
SumanthRH: I kep the type hint Any - and added default None to indicate that it's optional

Issue #998, opened 2026-01-31
[tinkerification] Remove scheduler from inside the workers and default to use set_lr() and get_lr() in training loop


erictang000:
After #978, we should remove the scheduler logic in SkyRL-Train workers and just rely on the tinker convention of setting learning rate for each `optim_step()` call

cc: @tyler-griggs

jonahsamost:
I can take this

PR #1002, opened 2026-02-01
Remove LR scheduler for workers

jonahsamost:
Referencing #998 

Remove internal learning rate scheduler logic from the skyRL train workers. 

Since #978 added set_lr() support, Tinker can now control the LR externally. Since the workers's LRs could conflict with Tinker's LR, we just remove the ability for workers to set their own LRs

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1002" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/skyrl/backends/skyrl_train/distributed/fsdp_strategy.py
+++ b/skyrl/backends/skyrl_train/distributed/fsdp_strategy.py
@@ -17,6 +17,7 @@ from torch import distributed as dist
 from torch import optim
 from torch.distributed.fsdp import CPUOffload, MixedPrecision
 from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
+
 from skyrl.backends.skyrl_train.distributed.fsdp_utils import (
     CPUOffloadPolicy,
     MixedPrecisionPolicy,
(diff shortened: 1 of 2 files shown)


pcmoritz:
@erictang000 Do you think this makes sense to do / when would be the right time to merge it? As we fully standardize on the Tinker API, I feel like this should be removed, right?

SumanthRH:
@jonahsamost Thanks for the PR and apologies for the delay - we've been a bit busy completing the migration to the new `skyrl` package. 


Can you please rebase on `main` and move all your changes to the new `skyrl` package? Also, please ensure that the non-tinker path where we interact only with RayPPOTrainer + Worker is not broken - you would need to set the LR per step in the RayPPOTrainer. Let's try to make sure that the base class sets this in a way that the subclasses (AsyncRayPPOTrainer) can easily re-use.

jonahsamost:
Sounds good. Will ping you when I make the changes

jonahsamost:
@SumanthRH 
Rebased to main and applied the changes. 
Removed the scheduler from workers/strategies so LR is now controlled externally via `set_lr` and `get_lr`. Added lr scheduling to `RayPPOTrainer` using transformer's `get_scheduler` similar to how it was done in fsdp_strategy.py for the non-tinker path. All the fsdp tests pass locally. Let me know what you think

PR #1005, opened 2026-02-02
Fix tests when JAX is not installed

tyler-griggs:
# Fix tests when JAX is not installed

Fixes test failures from #1004 when running without the `[jax]` extra.

## Changes

- Make JAX import optional in `tests/conftest.py`
- Only configure JAX devices if JAX is available
- Allows tests to run with `--extra tinker --extra dev` (without `--extra jax`)

## Test plan

```bash
uv run --extra tinker --extra dev pytest --forked -s tests --ignore=tests/gpu
```

Should now pass for tests that don't require JAX.

### DIFF
--- a/skyrl-tx/tests/conftest.py
+++ b/skyrl-tx/tests/conftest.py
@@ -1,9 +1,15 @@
-import jax
 import pytest
+
+try:
+    import jax
+
+    HAS_JAX = True
+except ImportError:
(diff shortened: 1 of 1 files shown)


tyler-griggs:
Closing in favor of proper fix - updating CI to include --extra jax instead

Issue #952, opened 2026-01-25
[tx] Implement mHC


pcmoritz:
The SkyRL tx JaxBackend is designed to be simple/hackable and make research easy. One technique that would be interesting to implement in this spirit is https://arxiv.org/pdf/2512.24880, and the fact that there is a common code base for inference and training should make it particularly suited for this task.

tanmaysachan:
Implemented mHC in torch recently - taking up this issue
Thanks!

PR #1008, opened 2026-02-02
[tx] General implementation of trainable Hyper Connections

tanmaysachan:
Addresses https://github.com/NovaSky-AI/SkyRL/issues/952

This PR is a general implementation of Hyper connections.

This is supposed to be an extension like Lora, where the default case mimics a standard residual connection with identity mappings.

### Default case - Trainable is false. Expansion rate is 1.
[edit] we now bypass this case entirely for a regular residual network.
1. H_res is a single value matrix [1]
2. H_pre and H_post are vectors of [1, 1, 1, ...] that result in no-op matmuls

### For expansion rate > 1
1. H_res is initialized as identity of size nxn (n is the expansion rate)
2. H_pre is [1/n, 1/n, ...]
3. H_post is [1, 1, 1, ...]

These matrices preserve identity mapping. So expansion rate > 1 but untrainable still results in the the **same** outputs.

### Todos
- [x] simplify rms integration - added elementwise_affine as a flag
- [x] Benchmark/ensure no regression for expansion_rate = 1 - minimal difference in step time when expansion rate is 1 and untrainable.

### Future work
- Fine tune on custom data with mHC + LoRA to see perf gains

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1008" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/skyrl-tx/tx/models/deepseekv3.py
+++ b/skyrl-tx/tx/models/deepseekv3.py
@@ -436,14 +436,14 @@ class DeepseekV3DecoderLayer(nnx.Module):
 
         self.attn_connector = LoRAConnector(
             config.hidden_size,
-            config.expansion_rate,
+            config.mhc_expansion_rate,
             max_lora_adapters=config.max_lora_adapters,
             dtype=dtype,
             rngs=rngs,
         )
         self.mlp_connector = LoRAConnector(
             config.hidden_size,
-            config.expansion_rate,
+            config.mhc_expansion_rate,
             max_lora_adapters=config.max_lora_adapters,
             dtype=dtype,
             rngs=rngs,
@@ -530,7 +530,7 @@ class DeepseekV3Model(nnx.Module):
         )
 
         hidden_states = self.embed_tokens(input_ids, adapter_indices=adapter_indices)
-        hidden_states = jnp.repeat(hidden_states[..., None, :], self.config.expansion_rate, axis=-2)
+        hidden_states = jnp.repeat(hidden_states[..., None, :], self.config.mhc_expansion_rate, axis=-2)
 
         # Forward through all layers
         hidden_states, all_hidden_states, kv_cache = self.layers(
--- a/skyrl-tx/tx/utils/models.py
+++ b/skyrl-tx/tx/utils/models.py
@@ -265,7 +265,7 @@ def load_lora_checkpoint(
             prefix="base_model.model.",
             filter_fn=lambda path: (
                 (("lora_A" in path or "lora_B" in path) and filter_lora(adapter_config, path))
-                or (model.config.expansion_rate > 1 and is_connector_path(path))
+                or (model.config.mhc_expansion_rate > 1 and is_connector_path(path))
             ),
             adapter_index=adapter_index,
             rank=adapter_config.rank,
@@ -300,7 +300,7 @@ def save_lora_checkpoint(
             prefix="base_model.model.",
             filter_fn=lambda path: (
                 (("lora_A" in path or "lora_B" in path) and filter_lora(adapter_config, path))
-                or (model.config.expansion_rate > 1 and is_connector_path(path))
+                or (model.config.mhc_expansion_rate > 1 and is_connector_path(path))
             ),
             adapter_index=adapter_index,
             rank=adapter_config.rank,
--- a/skyrl-tx/tests/layers/test_connectors.py
+++ b/skyrl-tx/tests/layers/test_connectors.py
@@ -119,8 +119,8 @@ def test_deepseek_connector_identity_expansion_rate():
     )
     config_e1 = DeepseekV3Config(base_config, max_lora_adapters=4, max_lora_rank=8, shard_attention_heads=True)
     config_e4 = DeepseekV3Config(base_config, max_lora_adapters=4, max_lora_rank=8, shard_attention_heads=True)
-    config_e1.expansion_rate = 1
-    config_e4.expansion_rate = 4
+    config_e1.mhc_expansion_rate = 1
+    config_e4.mhc_expansion_rate = 4
 
     input_ids = np.array([[1, 2, 3, 4, 5], [6, 7, 8, 9, 10]], dtype=np.int32)
     attention_mask = np.ones_like(input_ids, dtype=np.int32)
@@ -151,8 +151,8 @@ def test_deepseek_connector_identity_expansion_rate():
 
 
 class _TinyConfig:
-    def __init__(self, expansion_rate: int):
-        self.expansion_rate = expansion_rate
+    def __init__(self, mhc_expansion_rate: int):
+        self.mhc_expansion_rate = mhc_expansion_rate
 
     def get_num_experts(self):
         return 0
@@ -171,8 +171,8 @@ class _TinyConnector(nnx.Module):
 
 
 class _TinyModel(nnx.Module, ModelForCausalLM):
-    def __init__(self, expansion_rate: int, max_adapters: int = 3, max_rank: int = 4):
-        self.config = _TinyConfig(expansion_rate=expansion_rate)
+    def __init__(self, mhc_expansion_rate: int, max_adapters: int = 3, max_rank: int = 4):
+        self.config = _TinyConfig(mhc_expansion_rate=mhc_expansion_rate)
         self.self_attn = _TinyLoRA(max_adapters=max_adapters, max_rank=max_rank)
         self.attn_connector = _TinyConnector(max_adapters=max_adapters)
 
@@ -213,8 +213,8 @@ def _adapter_filter(path: tuple) -> bool:
 
 
 def test_connector_adapter_slice_save_load_safetensors(tmp_path):
-    src = _TinyModel(expansion_rate=4)
-    dst = _TinyModel(expansion_rate=4)
+    src = _TinyModel(mhc_expansion_rate=4)
+    dst = _TinyModel(mhc_expansion_rate=4)
     _fill_tiny_model(src)
     _zero_tiny_model(dst)
 
@@ -256,12 +256,12 @@ def test_connector_extract_insert_adapter_state_roundtrip():
     rank = 2
     adapter_index = 1
 
-    src = _TinyModel(expansion_rate=4)
+    src = _TinyModel(mhc_expansion_rate=4)
     _fill_tiny_model(src)
     _, src_lora_params, _ = nnx.split(src, src.is_lora_param, ...)
     extracted = extract_adapter_state(adapter_index, src_lora_params, rank)
 
-    dst = _TinyModel(expansion_rate=4)
+    dst = _TinyModel(mhc_expansion_rate=4)
     _zero_tiny_model(dst)
     _, dst_lora_params, _ = nnx.split(dst, dst.is_lora_param, ...)
     insert_adapter_state(adapter_index, dst_lora_params, extracted, rank)
@@ -292,7 +292,7 @@ def test_connector_extract_insert_adapter_state_roundtrip():
 def test_lora_checkpoint_includes_connectors_when_trainable(tmp_path):
     adapter_cfg = LoraConfig(rank=2, alpha=4, seed=0)
 
-    model_true = _TinyModel(expansion_rate=4)
+    model_true = _TinyModel(mhc_expansion_rate=4)
     _fill_tiny_model(model_true)
     ckpt_true = tmp_path / "with_connectors.tar.gz"
     save_lora_checkpoint(model_true, "dummy/base", adapter_cfg, adapter_index=1, output_path=ckpt_true)
@@ -301,7 +301,7 @@ def test_lora_checkpoint_includes_connectors_when_trainable(tmp_path):
         tensors = safetensors.numpy.load_file(extracted_dir / "adapter_model.safetensors")
         assert any("attn_connector" in key for key in tensors.keys())
 
-    loaded_true = _TinyModel(expansion_rate=4)
+    loaded_true = _TinyModel(mhc_expansion_rate=4)
     _zero_tiny_model(loaded_true)
     load_lora_checkpoint(loaded_true, adapter_cfg, adapter_index=1, checkpoint_path=ckpt_true)
     np.testing.assert_allclose(
@@ -309,7 +309,7 @@ def test_lora_checkpoint_includes_connectors_when_trainable(tmp_path):
         np.asarray(model_true.attn_connector.alpha_pre[...][1]),
     )
 
-    model_false = _TinyModel(expansion_rate=1)
+    model_false = _TinyModel(mhc_expansion_rate=1)
     _fill_tiny_model(model_false)
     ckpt_false = tmp_path / "without_connectors.tar.gz"
     save_lora_checkpoint(model_false, "dummy/base", adapter_cfg, adapter_index=1, output_path=ckpt_false)
--- a/skyrl-tx/tests/tinker/test_jax_backend.py
+++ b/skyrl-tx/tests/tinker/test_jax_backend.py
@@ -102,7 +102,7 @@ def test_max_adapters_after_delete():
 
 def test_clear_lora_adapter():
     """Test that clear_lora_adapter zeros out adapter state."""
-    backend = create_backend(expansion_rate=4)
+    backend = create_backend(mhc_expansion_rate=4)
     model_id = "test_model"
     adapter_idx = create_model(backend, model_id)
 
@@ -357,7 +357,7 @@ def test_process_optim_step_hyperparams_behavior():
 
 def test_optim_step_returns_metrics():
     """optim_step should return learning rate and grad norm metrics."""
-    config = JaxBackendConfig(max_lora_adapters=8, max_lora_rank=32, expansion_rate=4)
+    config = JaxBackendConfig(max_lora_adapters=8, max_lora_rank=32, mhc_expansion_rate=4)
     backend = JaxBackend(BASE_MODEL, config)
 
     model_id = "adapter_metrics"
--- a/skyrl-tx/tx/models/configs.py
+++ b/skyrl-tx/tx/models/configs.py
@@ -16,7 +16,7 @@ class ModelConfig(PretrainedConfig):
         shard_attention_heads: Whether to shard attention across tensor parallel devices
         loss_chunk_size: Chunk size for cross-entropy loss computation (0 = no chunking)
         gradient_checkpointing: Recompute activations during backward to save memory
-        expansion_rate: mHC expansion rate. Connectors are trainable when this is > 1.
+        mhc_expansion_rate: mHC expansion rate. Connectors are trainable when this is > 1.
     """
 
     # Type hints for config attributes
@@ -25,7 +25,7 @@ class ModelConfig(PretrainedConfig):
     shard_attention_heads: bool
     loss_chunk_size: int
     gradient_checkpointing: bool
-    expansion_rate: int
+    mhc_expansion_rate: int
 
     def __init__(
         self,
@@ -36,7 +36,7 @@ class ModelConfig(PretrainedConfig):
         shard_attention_heads: bool,
         loss_chunk_size: int = 0,
         gradient_checkpointing: bool = False,
-        expansion_rate: int = 1,
+        mhc_expansion_rate: int = 1,
     ):
         # Copy all attributes from the base config
         super().__init__(**config.to_dict())
@@ -47,7 +47,7 @@ class ModelConfig(PretrainedConfig):
         self.shard_attention_heads = shard_attention_heads
         self.loss_chunk_size = loss_chunk_size
         self.gradient_checkpointing = gradient_checkpointing
-        self.expansion_rate = expansion_rate
+        self.mhc_expansion_rate = mhc_expansion_rate
 
     def get_num_experts(self):
         return getattr(self, "num_experts", None) or getattr(self, "n_routed_experts", None)
--- a/skyrl-tx/tx/models/qwen3.py
+++ b/skyrl-tx/tx/models/qwen3.py
@@ -289,14 +289,14 @@ class Qwen3DecoderLayer(nnx.Module):
 
         self.attn_connector = LoRAConnector(
             config.hidden_size,
-            config.expansion_rate,
+            config.mhc_expansion_rate,
             max_lora_adapters=config.max_lora_adapters,
             dtype=dtype,
             rngs=rngs,
         )
         self.mlp_connector = LoRAConnector(
             config.hidden_size,
-            config.expansion_rate,
+            config.mhc_expansion_rate,
             max_lora_adapters=config.max_lora_adapters,
             dtype=dtype,
             rngs=rngs,
@@ -373,7 +373,7 @@ class Qwen3Model(nnx.Module):
         )
 
         hidden_states = self.embed_tokens(input_ids, adapter_indices=adapter_indices)
-        hidden_states = jnp.repeat(hidden_states[..., None, :], self.config.expansion_rate, axis=-2)
+        hidden_states = jnp.repeat(hidden_states[..., None, :], self.config.mhc_expansion_rate, axis=-2)
 
         hidden_states, all_hidden_states, new_kv_cache = self.layers(
             hidden_states,
--- a/skyrl-tx/tx/models/types.py
+++ b/skyrl-tx/tx/models/types.py
@@ -19,7 +19,7 @@ class ModelForCausalLM:
     def is_lora_param(self, path: tuple, _value) -> bool:
         """Return True if a parameter path corresponds to trainable LoRA/connector weights."""
         is_lora = any(name in path for name in ("lora_A", "lora_B"))
-        is_connector = self.config.expansion_rate > 1 and any(
+        is_connector = self.config.mhc_expansion_rate > 1 and any(
             name in path for name in ("attn_connector", "mlp_connector")
         )
         return is_lora or is_connector
--- a/skyrl-tx/tx/tinker/backends/jax.py
+++ b/skyrl-tx/tx/tinker/backends/jax.py
@@ -89,7 +89,7 @@ class JaxBackendConfig(BaseModel, extra="forbid"):
         default=False,
         description="Per-layer activation checkpointing: recompute activations during backward to save memory",
     )
-    expansion_rate: int = Field(
+    mhc_expansion_rate: int = Field(
         default=1,
         ge=1,
         description=(
@@ -209,7 +209,7 @@ class JaxBackendImpl(AbstractBackend):
             shard_attention_heads=config.shard_attention_heads,
             loss_chunk_size=config.loss_chunk_size,
             gradient_checkpointing=config.gradient_checkpointing,
-            expansion_rate=config.expansion_rate,
+            mhc_expansion_rate=config.mhc_expansion_rate,
         )
 
         model_class = get_model_class(self.model_config)
@@ -517,7 +517,7 @@ class JaxBackendImpl(AbstractBackend):
             mean_grads = accumulated_grads.get_mean(adapter_index)
             grad_norm = optax.global_norm(mean_grads)
             mhc_gradient_norm = None
-            if self.config.expansion_rate > 1:
+            if self.config.mhc_expansion_rate > 1:
                 mhc_grads = jax.tree.map_with_path(
                     lambda path, g: g if is_connector_path(path) else jnp.zeros_like(g),
                     mean_grads,


skyrl-tx/tx/layers/layernorm.py
```diff
@@ -16,7 +16,7 @@ class RMSNorm(nnx.Module):
     def __init__(self, size: int, *, eps: float = 1e-6, dtype: jnp.dtype, rngs: nnx.Rngs) -> None:
         self.eps = eps
         self.weight = Param(
-            size, dtype=dtype, kernel_init=nnx.with_partitioning(nnx.initializers.normal(), jax.P(None)), rngs=rngs
+            size, dtype=dtype, kernel_init=nnx.with_partitioning(nnx.initializers.ones_init(), jax.P(None)), rngs=rngs
```
tanmaysachan: Temporary, testing

tanmaysachan: https://docs.pytorch.org/docs/stable/generated/torch.nn.modules.normalization.RMSNorm.html

Torch also initalizes to one by default

pcmoritz:
This looks very elegant, thanks a lot for putting it together! Have you tried to do any end-to-end runs yet / studied the performance, both in terms of learning dynamics / accuracy, as well as how much slowdown it incurs :)

tanmaysachan:
Just waiting for the weekend to give it a spin 😅

I'll give Qwen0.6B a shot on an A/H100

pcmoritz:
Sounds great! I'm putting together the 0.3.0 release at the moment, so it will probably need to wait then, but 0.3.1 should come relatively soon thereafter, so it is not a problem. I'll put a callout in the release blog anyways, if somebody wants to try it out, they can just apply the diff themselves given how simple this is :)

tanmaysachan:
Did some analysis on the step times for each on Qwen 0.6B (on a 5060Ti)

Expansion rate as 1 does cause a hit to the average step time (about 0.3s slower, baseline has a step time of 2.1s vs 2.4s). An easy fix would be to just short circuit the entire thing for expansion rate = 1.

For expansion rate = 4, the step time was around 3.17s, so about 46% slower.

tanmaysachan:
qwen_expansion_4_loss

Loss plot for Qwen0.6B with an expansion rate = 4 max_lora_adapters=2, max_lora_rank=1.




(some more analysis todo)

tanmaysachan:
qwen_loss_comparison

1.7B Qwen, without expansion rate and with rate = 4 (roughly identical loss plots)
mHC times with training are about 93% higher than regular per step.

tanmaysachan:
The loss differences are in a similar scale as to what is observed in the mHC paper.
image

Ground truth mHC analysis - 
image

skyrl-tx/tx/models/deepseekv3.py
```diff
@@ -525,6 +546,10 @@ def __call__(
             is_training=is_training,
         )
 
+        hidden_states = hidden_states.sum(axis=-2)
```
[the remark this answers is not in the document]
tanmaysachan: In the end, we do an rmsnorm, which cancels this out. There are already parity tests in place for deepseek.

skyrl-tx/tx/layers/util.py
```diff
@@ -63,6 +63,16 @@ def Param(*shape: int, dtype: jnp.dtype, kernel_init: nnx.Initializer, rngs: nnx
     return nnx.Param(kernel_init(rngs.param(), shape, dtype))
 
 
+def sinkhorn_knopp(M: jax.Array, iters: int = 20) -> jax.Array:
```
pcmoritz: Let's move this function into connectors.py, since it is not used outside of the mHC implementation

skyrl-tx/tx/utils/models.py
```diff
@@ -164,6 +165,9 @@ def load_safetensors(
         # Skip LoRA parameters if requested
         if skip_lora and ("lora_A" in path or "lora_B" in path or "lora_scaling" in path or "lora_ranks" in path):
             continue
+        # Skip connector parameters
+        if is_connector_path(path):
```
pcmoritz: I believe you only want to skip loading the connector path if `skip_lora` is true, right? We only want to skip loading these parameters from the base model file, but for the LoRA adapters, we actually want to load them -- otherwise the won't become part of the sampling weights after each training iteration (you might not have run into this problem if you mainly / only did SFT experiments)

skyrl-tx/tx/layers/connectors.py
```diff
@@ -0,0 +1,138 @@
...
+        self.alpha_post = nnx.Param(jnp.zeros((max_lora_adapters,), dtype=dtype))
+        self.alpha_res = nnx.Param(jnp.zeros((max_lora_adapters,), dtype=dtype))
+
+    def _get_adapter_indices(self, batch_size: int, adapter_indices: jax.Array | None) -> jax.Array:
+        if adapter_indices is None:
+            return jnp.zeros((batch_size,), dtype=jnp.int32)
+        return adapter_indices.astype(jnp.int32)
+
+    def _get_params(self, adapter_indices: jax.Array):
+        """Stop gradients when the connectors are not trainable."""
```
pcmoritz: Do you think that's needed? If we are doing sampling, JAX will not compute gradients (https://github.com/NovaSky-AI/SkyRL/blob/main/skyrl-tx/tx/tinker/backends/jax.py#L822), the gradients are only computed for `forward_backward` (https://github.com/NovaSky-AI/SkyRL/blob/main/skyrl-tx/tx/tinker/backends/jax.py#L359) so it shouldn't be needed :)

skyrl-tx/tx/layers/connectors.py
```diff
@@ -0,0 +1,138 @@
...
+            kernel_init=nnx.initializers.normal(stddev=0.02),
+            rngs=rngs,
+        )
+
+        target_h_pre = jnp.array(1.0 / n, dtype=dtype)
+        clamped = jnp.clip(target_h_pre, 1e-6, 1.0 - 1e-6)
+        inv_sigmoid = jnp.log(clamped) - jnp.log(1.0 - clamped)
+        self.b_pre = nnx.Param(jnp.full((max_lora_adapters, n), inv_sigmoid, dtype=dtype))
+        self.b_post = nnx.Param(jnp.zeros((max_lora_adapters, n), dtype=dtype))
+        self.b_res = nnx.Param(jnp.broadcast_to(10.0 * jnp.eye(n, dtype=dtype), (max_lora_adapters, n, n)))
```
pcmoritz: Where does this factor of 10.0 come from?

tanmaysachan: This is a smoke test to check for exploding values if sinkhorn isn't working properly. This matrix just becomes identity once it passes through sinkhorn op.

skyrl-tx/tx/layers/connectors.py
tanmaysachan: Ah yes, the connectors are only a part of lora_params when they are trainable already. Thanks, this is redundant

skyrl-tx/tx/layers/connectors.py
pcmoritz: Hmm, if I understand the code correctly, it will change the initializations though -- if you add a heavy bias towards the identity matrix like this, the off-diagonal entries of H_res will be regularized towards zero much more aggressively than without this. So my guess is this does more than just a smoke test. I haven't studied it in full detail yet though.

Do you have a reference on how you chose the different initializations (e.g. the 0.02 above or using the identity for b_res and the inv_sigmoid above? The paper is a little light on these details it looks like :) And even if it weren't, it is not clear that the initializations chosen for pre-training are also good for post-training :)

tanmaysachan: **For sinkhorn**
That is what we want though, right? The initialization of this matrix should be identity to preserve identity-mapping with the residual stream - a large identity forces is strongly towards identity after sinkhorn. Perhaps smoke test isn't the correct term - for example a regular identity matrix would fail here. We would want a large eye.


**For initialization**

The 0.02 figure was taken from the baseline implementation here - https://github.com/AndreSlavescu/mHC.cu/blob/main/src/python/benchmarks/bench_layer.py . I'm not sure of the rationale, the paper doesn't seem to mention it.

There's really no literature to guide if that's optimal or not either

pcmoritz: I think I understand this a little better now -- what setting self.b_res to a large multiple of the identity matrix will do is to make sure the initial model is close to the base model, right? So this is kind of desired behavior (kind of similar to only initializing one of the LoRA adapters to nonzero for LoRA training) and kind of crucial. I wonder if this is the best way to do this, since it also impacts the initializations a bunch like I said above.

This stuff is worth understanding and documenting pretty thoroughly. I'll also run some experiments later and share what I find out :)

pcmoritz: Hehe looks like we posted at the same time -- you are correct!

skyrl-tx/tx/utils/models.py
tanmaysachan: I'll fix this, the checkpointing logic is also skipping connector params. I'll try to do this cleanly and push out a fix in a bit

skyrl-tx/tx/layers/connectors.py
tanmaysachan: Yes! I'll clean this up and leave thorough comments.

skyrl-tx/tx/utils/models.py
tanmaysachan: Fixed this issue. We detect and treat connector weights as LoRA like, but separate from lora.

skyrl-tx/tx/layers/connectors.py
tanmaysachan: Added comments for mHC initialization - we try to keep the matrices in a way that mHC is essentially a no-op

The matrices are free to drift away afterwards, just like LoRA

pcmoritz:
Btw, at the moment, the only way to activate mHC is by editing the code, right? We should probably expose it in `--backend-config` for now. It is slightly unfortunate that it needs to be configured on the backend, but for now that's ok I think. In the docstring of the field in JaxBackendConfig we should note that this is currently experimental (e.g. if we can configure it per request going forward that would be better, but for now what you currently have is good, certainly simpler, also the behavior will likely still change a bunch).

tanmaysachan:
> Btw, at the moment, the only way to activate mHC is by editing the code, right? We should probably expose it in `--backend-config` for now. It is slightly unfortunate that it needs to be configured on the backend, but for now that's ok I think. In the docstring of the field in JaxBackendConfig we should note that this is currently experimental (e.g. if we can configure it per request going forward that would be better, but for now what you currently have is good, certainly simpler, also the behavior will likely still change a bunch).

Addressed

skyrl-tx/tx/layers/lora.py
```diff
@@ -379,9 +403,11 @@ def clear_lora_adapter(model: ModelForCausalLM, adapter_index: int):
 
     def clear_adapter(path, value):
         key = path[-2].key
+        idx = get_adapter_idx(path, adapter_index)
+        if is_connector_path(path):
+            return value.at[idx].set(0.0)
```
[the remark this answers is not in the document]
tanmaysachan: Resolved, clear_lora_adapter now resets to identity semantics instead of zeroing it out

skyrl-tx/tx/layers/layernorm.py
tanmaysachan: Due to adapter indexing, ended up re-implementing norm in the connector layer itself - this change can be removed. But considering torch as the baseline, ones_init fits better still

pcmoritz:
I just merged in main, so we have https://github.com/NovaSky-AI/SkyRL/pull/1142 and can have a look at the gradient norm. I can also make a patch to add the gradient norm of the mHC parameters vs. everything else so we can monitor that as well. Maybe it is best to keep the patch out of this PR for now and first refactor main a little more to make this easier :)

pcmoritz:
This should be the diff to include the mHC gradient norm, however the norm is currently zero for me, so either this diff is wrong or the gradient norm is actually zero for some reason :)

```diff
diff --git a/skyrl-tx/tx/tinker/backends/jax.py b/skyrl-tx/tx/tinker/backends/jax.py
index f836026e..c15299ad 100644
--- a/skyrl-tx/tx/tinker/backends/jax.py
+++ b/skyrl-tx/tx/tinker/backends/jax.py
@@ -37,6 +37,7 @@ from pydantic import BaseModel, Field, TypeAdapter
 from transformers import AutoTokenizer, PretrainedConfig
 
 from tx.models.configs import Qwen3Config
+from tx.layers.connectors import is_connector_path
 from tx.layers.lora import clear_lora_adapter, init_lora_adapter
 from tx.tinker import types
 from tx.tinker.backends.backend import AbstractBackend
@@ -498,12 +499,17 @@ class JaxBackendImpl(AbstractBackend):
             lora_params: nnx.State,
             optimizer: nnx.Optimizer,
             adapter_index: jax.Array,
-        ) -> tuple[AccumulatedGradients, jax.Array]:
+        ) -> tuple[AccumulatedGradients, jax.Array, jax.Array]:
             """Compute full gradients, apply optimizer update, and reset accumulated grads."""
             mean_grads = accumulated_grads.get_mean(adapter_index)
             grad_norm = optax.global_norm(mean_grads)
+            mhc_grads = jax.tree.map_with_path(
+                lambda path, g: g if is_connector_path(path) else jnp.zeros_like(g),
+                mean_grads,
+            )
+            mhc_grad_norm = optax.global_norm(mhc_grads)
             optimizer.update(lora_params, mean_grads)
-            return accumulated_grads.reset_adapter(adapter_index), grad_norm
+            return accumulated_grads.reset_adapter(adapter_index), grad_norm, mhc_grad_norm
 
         if self.config.enforce_eager:
             self._compute_grads_and_update = compute_grads_and_update
@@ -766,16 +772,27 @@ class JaxBackendImpl(AbstractBackend):
 
         # JIT-compiled: compute full gradients, apply optimizer update, and reset accumulated grads
         with jax.set_mesh(self.mesh):
-            self.accumulated_grads, grad_norm = self._compute_grads_and_update(
+            self.accumulated_grads, grad_norm, mhc_grad_norm = self._compute_grads_and_update(
                 self.accumulated_grads,
                 self.lora_params,
                 optimizer,
                 jnp.int32(adapter_index),
             )
 
-        grad_norm = float(jax.device_get(grad_norm))
-        logger.info(f"Applied optimizer step for model {model_id} (adapter {adapter_index}), grad_norm={grad_norm}")
-        return types.OptimStepOutput(metrics={"skyrl.ai/grad_norm": grad_norm, "skyrl.ai/learning_rate": learning_rate})
+        grad_norm, mhc_grad_norm = jax.device_get((grad_norm, mhc_grad_norm))
+        grad_norm = float(grad_norm)
+        mhc_grad_norm = float(mhc_grad_norm)
+        logger.info(
+            f"Applied optimizer step for model {model_id} "
+            f"(adapter {adapter_index}), grad_norm={grad_norm}, mhc_grad_norm={mhc_grad_norm}"
+        )
+        return types.OptimStepOutput(
+            metrics={
+                "skyrl.ai/grad_norm": grad_norm,
+                "skyrl.ai/mhc_gradient_norm": mhc_grad_norm,
+                "skyrl.ai/learning_rate": learning_rate,
+            }
+        )
 
     def sample(
         self,
```

pcmoritz:
Ah, the problem was actually on my local setup, I hadn't incorporated your JaxBackendConfig change into my workflow yet so my code change got overwritten -- the mHC gradients are flowing for me now :)

pcmoritz:
These are the best settings I have found so far (one thing it does is break the symmetry between the different streams), but the gains are still relatively small, we can likely find something better. Ideally there would be an initialization that is a little more similar to what we do with LoRA (one adapter zero, the other random, so the model is the same as the base initially, but also there is gradient signal). Choosing b_res a large multiple of the identity matrix feels a little more forced. If you have any thoughts let me know :)

[the FP32 changes are not needed I think, I just wanted to make sure there is no problem with the numerics]

```diff
diff --git a/skyrl-tx/tx/layers/connectors.py b/skyrl-tx/tx/layers/connectors.py
index 013cfe53..892052b2 100644
--- a/skyrl-tx/tx/layers/connectors.py
+++ b/skyrl-tx/tx/layers/connectors.py
@@ -12,11 +12,32 @@ def is_connector_path(path: tuple[Any, ...]) -> bool:
     return any(name in normalized_path for name in ("attn_connector", "mlp_connector"))
 
 
+def _logit(x: jax.Array) -> jax.Array:
+    """Inverse sigmoid: logit(x) = log(x / (1-x))."""
+    x = jnp.clip(x, 1e-6, 1.0 - 1e-6)
+    return jnp.log(x) - jnp.log(1.0 - x)
+
+
+def default_b_pre(n: int, dtype: jnp.dtype = jnp.float32) -> jax.Array:
+    """H_pre = sigmoid(b_pre) = 1/n: uniform aggregation across streams."""
+    return _logit(jnp.array(1.0 / n, dtype=dtype))
+
+
+def default_b_post(n: int, dtype: jnp.dtype = jnp.float32) -> jax.Array:
+    """H_post spectrum from 0 to 2: stream 0 is pure memory, stream n-1 is full update.
+
+    Mean H_post = 1.0, preserving standard residual behavior on average.
+    """
+    return _logit(jnp.linspace(0.0, 1.0, n, dtype=dtype))
+
+
 class LoRAConnector(nnx.Module):
     """
     Implementation of Manifold constrained HyperConnections (https://arxiv.org/pdf/2512.24880)
 
-    Weights initialized with identity mapping; Default behaviour equates to residual networks.
+    Streams are initialized with a spectrum of roles: stream 0 acts as pure memory
+    (preserving the residual), stream n-1 carries the full sublayer update, and
+    intermediate streams carry various mixtures. Mean behavior equals standard residual.
     """
 
     def __init__(
@@ -40,41 +61,32 @@ class LoRAConnector(nnx.Module):
         n = expansion_rate
         C = hidden_dim
 
-        self.input_norm_weight = nnx.Param(jnp.ones((max_lora_adapters, n * C), dtype=dtype))
+        # All connector parameters are stored in FP32 for numerical stability.
+        # Phi matrices are zero-initialized so that alpha * x @ 0 + bias = bias at init.
+        self.input_norm_weight = nnx.Param(jnp.ones((max_lora_adapters, n * C), dtype=jnp.float32))
         self.phi_pre = Param(
-            max_lora_adapters, n * C, n, dtype=dtype, kernel_init=nnx.initializers.normal(stddev=0.02), rngs=rngs
+            max_lora_adapters, n * C, n, dtype=jnp.float32, kernel_init=nnx.initializers.zeros_init(), rngs=rngs
         )
         self.phi_post = Param(
-            max_lora_adapters, n * C, n, dtype=dtype, kernel_init=nnx.initializers.normal(stddev=0.02), rngs=rngs
+            max_lora_adapters, n * C, n, dtype=jnp.float32, kernel_init=nnx.initializers.zeros_init(), rngs=rngs
         )
         self.phi_res = Param(
-            max_lora_adapters,
-            n * C,
-            n * n,
-            dtype=dtype,
-            kernel_init=nnx.initializers.normal(stddev=0.02),
-            rngs=rngs,
+            max_lora_adapters, n * C, n * n, dtype=jnp.float32, kernel_init=nnx.initializers.zeros_init(), rngs=rngs,
         )
 
-        # Initialize biases for identity-like behavior:
-        # H_pre = 1/n (uniform aggregation), H_post = 1 (full distribution), M = I (identity mixing)
-
-        # H_pre = sigmoid(b_pre) = 1/n  =>  b_pre = logit(1/n)
-        target_h_pre = jnp.array(1.0 / n, dtype=dtype)
-        clamped = jnp.clip(target_h_pre, 1e-6, 1.0 - 1e-6)
-        inv_sigmoid = jnp.log(clamped) - jnp.log(1.0 - clamped)
-        self.b_pre = nnx.Param(jnp.full((max_lora_adapters, n), inv_sigmoid, dtype=dtype))
+        # H_pre = sigmoid(b_pre) = 1/n: uniform aggregation across streams
+        self.b_pre = nnx.Param(jnp.full((max_lora_adapters, n), default_b_pre(n), dtype=jnp.float32))
 
-        # H_post = 2 * sigmoid(b_post) = 1  =>  b_post = 0
-        self.b_post = nnx.Param(jnp.zeros((max_lora_adapters, n), dtype=dtype))
+        # H_post = 2 * sigmoid(b_post): spectrum from 0 to 2, creating stream diversity.
+        # Stream 0 = pure memory, stream n-1 = full update, mean = 1 (standard residual).
+        self.b_post = nnx.Param(jnp.broadcast_to(default_b_post(n), (max_lora_adapters, n)))
 
-        # Large identity matrix -> heavily biases Sinkhorn() to product an identity matrix
-        self.b_res = nnx.Param(jnp.broadcast_to(10.0 * jnp.eye(n, dtype=dtype), (max_lora_adapters, n, n)))
+        # M ≈ I: identity mixing via Sinkhorn
+        self.b_res = nnx.Param(jnp.broadcast_to(3.0 * jnp.eye(n, dtype=jnp.float32), (max_lora_adapters, n, n)))
 
-        # Alpha = 0 so phi matrices don't contribute initially
-        self.alpha_pre = nnx.Param(jnp.zeros((max_lora_adapters,), dtype=dtype))
-        self.alpha_post = nnx.Param(jnp.zeros((max_lora_adapters,), dtype=dtype))
-        self.alpha_res = nnx.Param(jnp.zeros((max_lora_adapters,), dtype=dtype))
+        self.alpha_pre = nnx.Param(jnp.full((max_lora_adapters,), 0.01, dtype=jnp.float32))
+        self.alpha_post = nnx.Param(jnp.full((max_lora_adapters,), 0.01, dtype=jnp.float32))
+        self.alpha_res = nnx.Param(jnp.full((max_lora_adapters,), 0.01, dtype=jnp.float32))
 
     def _get_adapter_indices(self, batch_size: int, adapter_indices: jax.Array | None) -> jax.Array:
         if adapter_indices is None:
@@ -92,10 +104,11 @@ class LoRAConnector(nnx.Module):
         return jax.lax.fori_loop(0, iters, step, M)
 
     def _norm(self, x_flat: jax.Array, adapter_indices: jax.Array) -> jax.Array:
-        """Separate norm from layernorm.RMSNorm due to adapter indexing and trainability"""
+        """Separate norm from layernorm.RMSNorm due to adapter indexing and trainability."""
         input_norm_weight = self.input_norm_weight[adapter_indices]
-        rms = jnp.sqrt(jnp.mean(x_flat**2, axis=-1, keepdims=True) + self.eps)
-        return (input_norm_weight[:, None, :] * x_flat) / rms
+        x_float = x_flat.astype(jnp.float32)
+        rms = jnp.sqrt(jnp.mean(x_float**2, axis=-1, keepdims=True) + self.eps)
+        return (input_norm_weight[:, None, :] * x_float) / rms
 
     def pre(self, x: jax.Array, adapter_indices: jax.Array | None = None) -> tuple[jax.Array, jax.Array]:
         B, T, n, C = x.shape
@@ -105,7 +118,7 @@ class LoRAConnector(nnx.Module):
 
         adapter_indices = self._get_adapter_indices(B, adapter_indices)
         x_flat = x.reshape(B, T, n * C)
-        x_norm = self._norm(x_flat, adapter_indices)
+        x_norm = self._norm(x_flat, adapter_indices)  # float32
 
         alpha_pre = self.alpha_pre[adapter_indices]
         phi_pre = self.phi_pre[adapter_indices]
@@ -114,10 +127,10 @@ class LoRAConnector(nnx.Module):
         tilde_H_pre = alpha_pre[:, None, None] * pre_logits + b_pre[:, None, :]
 
         H_pre = jax.nn.sigmoid(tilde_H_pre)
-        x_agg = (H_pre[..., None] * x).sum(axis=-2)
+        x_agg = (H_pre[..., None] * x.astype(jnp.float32)).sum(axis=-2)
 
         # Return residual norm for future use by post()
-        return x_agg, x_norm
+        return x_agg.astype(x.dtype), x_norm
 
     def post(
         self,
@@ -148,6 +161,6 @@ class LoRAConnector(nnx.Module):
         H_post = 2.0 * jax.nn.sigmoid(tilde_H_post)
         M = self._sinkhorn_knopp(tilde_H_res, self.sinkhorn_iters)
 
-        y_dist = H_post[..., None] * output[..., None, :]
-        x_mixed = M @ residual
-        return x_mixed + y_dist
+        y_dist = H_post[..., None] * output.astype(jnp.float32)[..., None, :]
+        x_mixed = M @ residual.astype(jnp.float32)
+        return (x_mixed + y_dist).astype(residual.dtype)
diff --git a/skyrl-tx/tx/layers/lora.py b/skyrl-tx/tx/layers/lora.py
index d9660bfd..83170086 100644
--- a/skyrl-tx/tx/layers/lora.py
+++ b/skyrl-tx/tx/layers/lora.py
@@ -3,7 +3,7 @@ import jax
 from jax import numpy as jnp
 
 from tx.utils.models import filter_lora, get_adapter_idx
-from tx.layers.connectors import is_connector_path
+from tx.layers.connectors import default_b_post, default_b_pre, is_connector_path
 from tx.layers.util import Param, prepare_routing, ragged_dot
 from tx.models.types import ModelForCausalLM
 from tx.tinker.types import LoraConfig
@@ -354,23 +354,20 @@ def init_lora_adapter(model: ModelForCausalLM, adapter_index: int, lora_config:
         if is_connector_path(path):
             connector_slot = value[idx]
             if key_name in {"alpha_pre", "alpha_post", "alpha_res"}:
-                return value.at[idx].set(jnp.zeros_like(connector_slot))
+                return value.at[idx].set(jnp.full_like(connector_slot, 0.01))
             if key_name == "input_norm_weight":
                 return value.at[idx].set(jnp.ones_like(connector_slot))
             if key_name in {"phi_pre", "phi_post", "phi_res"}:
-                new_phi = nnx.initializers.normal(stddev=0.02)(rngs.params(), connector_slot.shape, value.dtype)
-                return value.at[idx].set(new_phi)
+                return value.at[idx].set(jnp.zeros_like(connector_slot))
             if key_name == "b_pre":
                 n = connector_slot.shape[-1]
-                target_h_pre = jnp.array(1.0 / n, dtype=value.dtype)
-                clamped = jnp.clip(target_h_pre, 1e-6, 1.0 - 1e-6)
-                inv_sigmoid = jnp.log(clamped) - jnp.log(1.0 - clamped)
-                return value.at[idx].set(jnp.full(connector_slot.shape, inv_sigmoid, dtype=value.dtype))
+                return value.at[idx].set(jnp.full(connector_slot.shape, default_b_pre(n, value.dtype), dtype=value.dtype))
             if key_name == "b_post":
-                return value.at[idx].set(jnp.zeros_like(connector_slot))
+                n = connector_slot.shape[-1]
+                return value.at[idx].set(jnp.broadcast_to(default_b_post(n, value.dtype), connector_slot.shape))
             if key_name == "b_res":
                 n = connector_slot.shape[-1]
-                eye = 10.0 * jnp.eye(n, dtype=value.dtype)
+                eye = 3.0 * jnp.eye(n, dtype=value.dtype)
                 return value.at[idx].set(jnp.broadcast_to(eye, connector_slot.shape))
             return value
 
@@ -409,22 +406,21 @@ def clear_lora_adapter(model: ModelForCausalLM, adapter_index: int):
         # remains behaviorally neutral for mHC before being reinitialized.
         if is_connector_path(path):
             connector_slot = value[idx]
-            if key in {"alpha_pre", "alpha_post", "alpha_res", "b_post"}:
+            if key in {"alpha_pre", "alpha_post", "alpha_res"}:
                 return value.at[idx].set(jnp.zeros_like(connector_slot))
             if key in {"phi_pre", "phi_post", "phi_res"}:
-                # Keep clear deterministic and neutral: alpha=0 makes phi inactive.
                 return value.at[idx].set(jnp.zeros_like(connector_slot))
             if key == "input_norm_weight":
                 return value.at[idx].set(jnp.ones_like(connector_slot))
             if key == "b_pre":
                 n = connector_slot.shape[-1]
-                target_h_pre = jnp.array(1.0 / n, dtype=value.dtype)
-                clamped = jnp.clip(target_h_pre, 1e-6, 1.0 - 1e-6)
-                inv_sigmoid = jnp.log(clamped) - jnp.log(1.0 - clamped)
-                return value.at[idx].set(jnp.full(connector_slot.shape, inv_sigmoid, dtype=value.dtype))
+                return value.at[idx].set(jnp.full(connector_slot.shape, default_b_pre(n, value.dtype), dtype=value.dtype))
+            if key == "b_post":
+                n = connector_slot.shape[-1]
+                return value.at[idx].set(jnp.broadcast_to(default_b_post(n, value.dtype), connector_slot.shape))
             if key == "b_res":
                 n = connector_slot.shape[-1]
-                eye = 10.0 * jnp.eye(n, dtype=value.dtype)
+                eye = 3.0 * jnp.eye(n, dtype=value.dtype)
                 return value.at[idx].set(jnp.broadcast_to(eye, connector_slot.shape))
             return value.at[idx].set(0.0)
         if key not in ("lora_ranks", "lora_scaling", "lora_A", "lora_B"):
diff --git a/skyrl-tx/tx/tinker/backends/jax.py b/skyrl-tx/tx/tinker/backends/jax.py
index f836026e..c15299ad 100644
--- a/skyrl-tx/tx/tinker/backends/jax.py
+++ b/skyrl-tx/tx/tinker/backends/jax.py
@@ -37,6 +37,7 @@ from pydantic import BaseModel, Field, TypeAdapter
 from transformers import AutoTokenizer, PretrainedConfig
 
 from tx.models.configs import Qwen3Config
+from tx.layers.connectors import is_connector_path
 from tx.layers.lora import clear_lora_adapter, init_lora_adapter
 from tx.tinker import types
 from tx.tinker.backends.backend import AbstractBackend
@@ -498,12 +499,17 @@ class JaxBackendImpl(AbstractBackend):
             lora_params: nnx.State,
             optimizer: nnx.Optimizer,
             adapter_index: jax.Array,
-        ) -> tuple[AccumulatedGradients, jax.Array]:
+        ) -> tuple[AccumulatedGradients, jax.Array, jax.Array]:
             """Compute full gradients, apply optimizer update, and reset accumulated grads."""
             mean_grads = accumulated_grads.get_mean(adapter_index)
             grad_norm = optax.global_norm(mean_grads)
+            mhc_grads = jax.tree.map_with_path(
+                lambda path, g: g if is_connector_path(path) else jnp.zeros_like(g),
+                mean_grads,
+            )
+            mhc_grad_norm = optax.global_norm(mhc_grads)
             optimizer.update(lora_params, mean_grads)
-            return accumulated_grads.reset_adapter(adapter_index), grad_norm
+            return accumulated_grads.reset_adapter(adapter_index), grad_norm, mhc_grad_norm
 
         if self.config.enforce_eager:
             self._compute_grads_and_update = compute_grads_and_update
@@ -766,16 +772,27 @@ class JaxBackendImpl(AbstractBackend):
 
         # JIT-compiled: compute full gradients, apply optimizer update, and reset accumulated grads
         with jax.set_mesh(self.mesh):
-            self.accumulated_grads, grad_norm = self._compute_grads_and_update(
+            self.accumulated_grads, grad_norm, mhc_grad_norm = self._compute_grads_and_update(
                 self.accumulated_grads,
                 self.lora_params,
                 optimizer,
                 jnp.int32(adapter_index),
             )
 
-        grad_norm = float(jax.device_get(grad_norm))
-        logger.info(f"Applied optimizer step for model {model_id} (adapter {adapter_index}), grad_norm={grad_norm}")
-        return types.OptimStepOutput(metrics={"skyrl.ai/grad_norm": grad_norm, "skyrl.ai/learning_rate": learning_rate})
+        grad_norm, mhc_grad_norm = jax.device_get((grad_norm, mhc_grad_norm))
+        grad_norm = float(grad_norm)
+        mhc_grad_norm = float(mhc_grad_norm)
+        logger.info(
+            f"Applied optimizer step for model {model_id} "
+            f"(adapter {adapter_index}), grad_norm={grad_norm}, mhc_grad_norm={mhc_grad_norm}"
+        )
+        return types.OptimStepOutput(
+            metrics={
+                "skyrl.ai/grad_norm": grad_norm,
+                "skyrl.ai/mhc_gradient_norm": mhc_grad_norm,
+                "skyrl.ai/learning_rate": learning_rate,
+            }
+        )
 
     def sample(
         self,
```

tanmaysachan:
@pcmoritz Thanks for running experiments with it!
I'm pretty compute restricted so can't play around with it much to run experiments 😓 
I'll try to do it on some small variants

I was thinking something along the lines of havine b_res initialized as product of 2 randomly initialized square matrices A and A^-1 followed by sinkhorn. Since b_res is quite small, this would be negligible parameter count wise, but this would be a departure from core mHC

tanmaysachan:
For clarity - should clear_adapter and init_adapter methods be implemented as static by the adapter classes themselves? I can make the change

pcmoritz:
@tanmaysachan Yes, that would be much better, there are a few more small changes I'd like to try (like shifting `is_lora_param` into the ModelForCausalLM base class so it won't get repeated. Why don't you go ahead and make the change about shifting clear_adapter and init_adapter as static into LoRAConnector, and then afterwards I'll have another pass to make the other cleanups :)

pcmoritz:
Thanks a lot for all the updates, I'll do the rest (already merged a PR that cleans things up a little https://github.com/NovaSky-AI/SkyRL/pull/1191) :)

skyrl-tx/tx/layers/connectors.py
```diff
@@ -0,0 +1,188 @@
...
+        self.expansion_rate = expansion_rate
+        self.max_lora_adapters = max_lora_adapters
+        self.trainable = trainable
+        self.sinkhorn_iters = sinkhorn_iters
+        self.eps = eps
+        n = expansion_rate
+        C = hidden_dim
+
+        # Phi matrices are zero-initialized so that alpha * x @ 0 + bias = bias at init.
+        self.input_norm_weight = nnx.Param(jnp.ones((max_lora_adapters, n * C), dtype=dtype))
```
pcmoritz: I'm curious, why did you make the RMSNorm per adapter and trainable? That seems wrong, we should probably just use the RMSNorm from the base model :) [I don't think any of the LoRA codes out there make the RMSNorm trainable]

pcmoritz: Actually I think I misunderstood the code and you are doing the right thing :)

pcmoritz: Sorry for going back on forth on this, but I think the actually correct implementation would be to pass the input norm parameters from the model to the constructor of `LoRAConnector` and use it for the normalization below and keep it non-trainable. It will be slightly redundant to apply the norm twice, but I think for code clarity that's fine for now (there is more optimizations to do anyways). Let me know about your thoughts, I'll give that a shot :)

tanmaysachan: Thanks for the change!
My layernorm change was around making the entire block including the norms trainable - but yeah if its something like lora that shouldn't be the case

skyrl-tx/tx/layers/connectors.py
```diff
@@ -0,0 +1,182 @@
...
+        input_norm: Callable[[jax.Array], jax.Array],
+        adapter_indices: jax.Array | None = None,
+    ) -> tuple[jax.Array, jax.Array]:
+        B, T, n, C = x.shape
+        if self.expansion_rate == 1:
+            # Single-stream fast path: pre is identity on the residual stream.
+            return x[..., 0, :], x.reshape(B, T, n * C)
+
+        adapter_indices = self._get_adapter_indices(B, adapter_indices)
+        # Apply input_norm independently to each of the n streams.
```
pcmoritz, line 140: The paper is not super clear on whether this is the right way to do it -- below equation (5) it says RMSNorm is applied to the last dimension C. In equation (7) it looks more like the RMSNorm is applied on the full n * C dimension. I chose the interpretation according to equation (5) since it is slightly more elegant and doesn't require changing the definition of the RMSNorm. Once a DeepSeek model is released that supports mHC, we can revisit this.

pcmoritz: I think HC applies the RMSNorm to the last dimension C and mHC applies it to n * C, this becomes pretty apparent from equations (15) and (16). We should switch as soon as we have a model that natively supports mHC.

tanmaysachan: Did you observe anything off with trainable norms for n*C? Just curious as to why that wont fit

pcmoritz: The actual difference in performance is pretty small, and once we have a model that is trained with mHC, it will have an RMSNorm weight of size n * C, so it will be very easy to adapt the current code to it (and there is no need for a trainable norm in the LoRA setting), so I feel like that's the better solution for now :)

tanmaysachan: Sounds good. For already pretrained models, the original HC paper is probably a better fit than mHC as it does have experimentations with HC as an augmentation (unlike mHC).

So RMS over C is probably better for the general case.

Issue #845, opened 2026-01-06
[skyrl-train][RFC] Unified InferenceHandle Abstraction


kouroshHakha:
## TL;DR

Replace SkyRL's fragmented inference abstractions (`InferenceEngineClient`, `RayWrappedInferenceEngine`, `RemoteInferenceEngine`) with a single `InferenceHandle` interface that standardizes on HTTP endpoints. This enables seamless integration with frameworks like Ray serve LLM for easy deployments on WideEP, and Prefill-Decode disaggregation.

### Current Problems

| Problem | Impact |
|---------|--------|
| **Dual code paths** | `run_engines_locally` creates two completely different flows (Ray actors vs HTTP), doubling maintenance burden |
| **Scattered weight sync** | Trainer manually orchestrates NCCL groups, rank offsets, IPC handles—complexity that should be server-side |
| **Backend-specific logic** | `if engine_backend == "vllm"` / `"sglang"` checks scattered across 5+ files |
| **No clear API boundary** | Generator and Trainer both reach into `InferenceEngineClient` internals |

### Why Now?

Ray Serve LLM already provides multi-replica orchestration, fault tolerance, and OpenAI-compatible endpoints. Extending it with weight sync and lifecycle APIs lets SkyRL inherit these capabilities without reimplementing them.

### Single Abstraction

```python
class InferenceHandle(ABC):
    """Unified interface for inference interactions."""
    
    # Data Plane (Generator)
    async def chat_completion(request_payload: Dict) -> Dict
    async def completion(request_payload: Dict) -> Dict
    async def generate(input_batch: InferenceEngineInput) -> InferenceEngineOutput
    
    # Control Plane (Trainer)  
    async def init_weight_transfer(master_addr, master_port, rank_offset, world_size) -> Dict
    async def collective_rpc(model, method, args, kwargs) -> Dict
    async def pause() / resume() / sleep() / wakeup()
    async def server_info() -> Dict
```

### Client Implementation

All server patterns use the same client: `RemoteInferenceHandle` (HTTP). For in-process Ray Serve, `RayServeInferenceHandle` bypasses HTTP overhead (optional)

### Two Server Patterns

| Pattern | Server | Features | Best For |
|---------|--------|----------|----------|
| **Ray Serve LLM** | `SkyRLIngress` via `serve.run()` | WideEP, PD, fault tolerance, sleep/wake | Production training |
[...]

PR #1016, opened 2026-02-03
[train] Fix issues with new HTTP inference codepath and add weight sync tests

SumanthRH:
# What does this PR do?

Fixes some issues with the new HTTP inference codepath (see RFC https://github.com/NovaSky-AI/SkyRL/issues/845) and adds basic weight sync tests

1. CLI Parsing: The GSM8K example fails with `_SKYRL_USE_NEW_INFERENCE=1` with a `AttributeError`:

```bash
(pid=3677738, ip=10.0.153.235, actor_id=66beaa83e5a5ee29a61a169213010000, repr=<skyrl_train.inference_servers.vllm_server_actor.VLLMServerActor object at 0x7b0ea4afa720>)
[36m(skyrl_entrypoint pid=3675949)[0m   File "/home/ray/anaconda3/lib/python3.12/concurrent/futures/_base.py", line 449, in result
[36m(skyrl_entrypoint pid=3675949)[0m     return self.__get_result()
[36m(skyrl_entrypoint pid=3675949)[0m            ^^^^^^^^^^^^^^^^^^^
[36m(skyrl_entrypoint pid=3675949)[0m   File "/home/ray/anaconda3/lib/python3.12/concurrent/futures/_base.py", line 401, in __get_result
[36m(skyrl_entrypoint pid=3675949)[0m     raise self._exception
[36m(skyrl_entrypoint pid=3675949)[0m            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[36m(skyrl_entrypoint pid=3675949)[0m   File "/tmp/ray/session_2026-01-28_12-40-25_893362_5766/runtime_resources/working_dir_files/_ray_pkg_825ce54be2c8b337/skyrl_train/inference_servers/vllm_server_actor.py", line 231, in start
[36m(skyrl_entrypoint pid=3675949)[0m     await self._wait_until_healthy()
[36m(skyrl_entrypoint pid=3675949)[0m            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[36m(skyrl_entrypoint pid=3675949)[0m   File "/tmp/ray/session_2026-01-28_12-40-25_893362_5766/runtime_resources/working_dir_files/_ray_pkg_825ce54be2c8b337/skyrl_train/inference_servers/vllm_server_actor.py", line 246, in _wait_until_healthy
[36m(skyrl_entrypoint pid=3675949)[0m     raise exc
[36m(skyrl_entrypoint pid=3675949)[0m            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[36m(skyrl_entrypoint pid=3675949)[0m   File "/tmp/ray/session_2026-01-28_12-40-25_893362_5766/runtime_resources/working_dir_files/_ray_pkg_825ce54be2c8b337/skyrl_train/inference_servers/vllm_server_actor.py", line 266, in _run_server
[36m(skyrl_entrypoint pid=3675949)[0m     app = build_app(self._cli_args)
[36m(skyrl_entrypoint pid=3675949)[0m           ^^^^^^^^^^^^^^^^^^^^^^^^^
[36m(skyrl_entrypoint pid=3675949)[0m   File "/home/ray/.cache/uv/builds-v0/.tmpCvhl19/lib/python3.12/site-packages/vllm/entrypoints/openai/api_server.py", line 961, in build_app
[36m(skyrl_entrypoint pid=3675949)[0m     if args.disable_fastapi_docs:
[36m(skyrl_entrypoint pid=3675949)[0m        ^^^^^^^^^^^^^^^^^^^^^^^^^
[36m(skyrl_entrypoint pid=3675949)[0m AttributeError: 'Namespace' object has no attribute 'disable_fastapi_docs'
```

The core issue is that we construct an ad-hoc namespace object while vLLM expects an object containing all supported fields - this includes fields in vLLM's `FrontendArgs`  and `AsyncEngineArgs`. This PR uses the same argument parsing logic as in vLLM :  https://github.com/vllm-project/vllm/blob/e346e2d056a66bb84287e4fea049bde9a37bd72b/vllm/entrypoints/openai/cli_args.py#L276-L313

and then overrides with SkyRL's configuration. 

2. Session re-use in `RemoteInferenceClient`: In some tests in GPU CI, we create multiple event loops in the same test. `RemoteInferenceClient` creates an `aiohttp.ClientSession` object which is valid for the original event loop only. The fix is to check if the event loop has changed and re-create the session

3. vLLM server placement for colocated training: For colocated training, the new HTTP inference codepath diverged from the existing Ray-actor-based codepath ([here](https://github.com/NovaSky-AI/SkyRL/blob/d1e1b297717195f9b54cb36f68c79d6aa406baed/skyrl-train/skyrl_train/inference_engines/ray_wrapped_inference_engine.py#L81)) in placement - the vllm servers were created with the default resource of 1 GPU per vLLM rank, which will not work for colocated training. 

4. Unit tests: Added unit testing for the new codepath for weight syncing for the colocated and the non-colocated case. Tests pass after the above fixes.

Misc: Also moved registration of `VLLMServerActor` with GCS to be global, one time

### DIFF
--- a/skyrl-train/skyrl_train/inference_servers/server_group.py
+++ b/skyrl-train/skyrl_train/inference_servers/server_group.py
@@ -113,12 +113,7 @@ class ServerGroup:
 
     def _create_actor_class(self, pg: PlacementGroup, start_bundle_idx: int) -> Any:
         """Create actor class with scheduling constraints for a specific bundle."""
-        from skyrl_train.inference_servers.vllm_server_actor import VLLMServerActorRay, VLLMServerActor
-
-        ray_wrapped_actor_cls = (
-            VLLMServerActorRay if self._server_actor_cls == VLLMServerActor else ray.remote(self._server_actor_cls)
-        )
-        return ray_wrapped_actor_cls.options(
+        return ray.remote(self._server_actor_cls).options(
             num_gpus=0,  # GPU allocation managed by placement group
             num_cpus=COLOCATED_ACTOR_CPU_FRACTION,
             scheduling_strategy=PlacementGroupSchedulingStrategy(
--- a/skyrl-train/skyrl_train/inference_servers/vllm_server_actor.py
+++ b/skyrl-train/skyrl_train/inference_servers/vllm_server_actor.py
@@ -2,7 +2,6 @@
 vLLM Server Actor - Ray actor running a vLLM OpenAI-compatible API server.
 """
 
-import ray
 import asyncio
 import logging
 import os
@@ -389,6 +388,3 @@ class VLLMServerActor(ServerActorProtocol):
                 await self._server_task
             except asyncio.CancelledError:
                 pass
-
-
-VLLMServerActorRay = ray.remote(VLLMServerActor)


skyrl-train/skyrl_train/utils/utils.py
```diff
@@ -868,3 +868,49 @@ def get_free_port():
     port = s.getsockname()[1]
     s.close()
     return port
+
+
+# TODO: Add a test for validation
+def build_vllm_cli_args(cfg) -> Namespace:
```
kouroshHakha: could we move this to inference_servers/utils.py?

skyrl-train/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -143,7 +143,8 @@ class RemoteInferenceClient:
 
     async def _get_session(self) -> aiohttp.ClientSession:
         """Get or create the aiohttp session."""
-        if self._session is None or self._session.closed:
+        current_loop = asyncio.get_event_loop()
```
kouroshHakha: can you add a note in comments on why we need this?

skyrl-train/skyrl_train/inference_servers/server_group.py
```diff
@@ -107,14 +107,15 @@ def _get_placement_group(self) -> PlacementGroup:
 
     def _create_actor_class(self, pg: PlacementGroup, start_bundle_idx: int) -> Any:
         """Create actor class with scheduling constraints for a specific bundle."""
-        return ray.remote(self._server_actor_cls).options(
+        from skyrl_train.inference_servers.vllm_server_actor import VLLMServerActorRay, VLLMServerActor
+
+        ray_wrapped_actor_cls = (
+            VLLMServerActorRay if self._server_actor_cls == VLLMServerActor else ray.remote(self._server_actor_cls)
+        )
+        return ray_wrapped_actor_cls.options(
```
kouroshHakha: why do we do this? The previous code was cleaner and achieve the same thing? We don't need to define a `VLLMServerActorRay` right?

skyrl-train/skyrl_train/inference_servers/server_group.py
```diff
@@ -107,14 +107,15 @@ def _get_placement_group(self) -> PlacementGroup:
...
             num_gpus=0,  # GPU allocation managed by placement group
-            num_cpus=1,
-            scheduling_strategy=PlacementGroupSchedulingStrategy(
-                placement_group=pg,
-                placement_group_capture_child_tasks=True,
-                placement_group_bundle_index=start_bundle_idx,
-            ),
+            num_cpus=0.2,
+            scheduling_strategy=PlacementGroupSchedulingStrategy(pg, start_bundle_idx),
```
kouroshHakha: we need child=True as well. otherwise vllm ray actors won't be able to get scheduler on parent's placement group

skyrl-train/skyrl_train/inference_servers/server_group.py
```diff
@@ -107,14 +107,15 @@ def _get_placement_group(self) -> PlacementGroup:
...
+        )
+        return ray_wrapped_actor_cls.options(
             num_gpus=0,  # GPU allocation managed by placement group
-            num_cpus=1,
-            scheduling_strategy=PlacementGroupSchedulingStrategy(
-                placement_group=pg,
-                placement_group_capture_child_tasks=True,
-                placement_group_bundle_index=start_bundle_idx,
-            ),
+            num_cpus=0.2,
```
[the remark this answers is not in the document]
kouroshHakha: +1

skyrl-train/skyrl_train/inference_servers/vllm_server_actor.py
```diff
@@ -101,6 +104,10 @@ def __init__(
         # Ensure SkyRL's custom worker extension is used for weight sync
         self._ensure_worker_extension()
 
+        # Ensure vLLM sleep endpoints are enabled by using dev mode
+        os.environ["VLLM_SERVER_DEV_MODE"] = "1"
+        os.environ["VLLM_RAY_PER_WORKER_GPUS"] = str(0.2 if colocated_training else 1.0)
```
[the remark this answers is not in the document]
kouroshHakha: Let's not leak colocated_training into the actor class because we just need to set the env var. We can set the env var from where we create the class. Basically: 

```

ServerActorCls.options(
    runtime_env={
        "env_vars": {
            "VLLM_...": 0.2 if self.external_pg is not None else 1.0
            "VLLM_SERVER_DEV_MODE": "1",
        }
    }
).remote(....)

```

skyrl-train/tests/gpu/gpu_ci/test_policy_local_engines_e2e.py
```diff
@@ -75,8 +87,10 @@ def test_policy_local_engines_e2e(ray_init_fixture, colocate_all, weight_sync_ba
         cfg.generator.backend = backend
         cfg.generator.inference_engine_tensor_parallel_size = tp_size
 
+        tokenizer = AutoTokenizer.from_pretrained(MODEL)
```
kouroshHakha, line 73: why did we not have tokenizer before and now we do?

skyrl-train/skyrl_train/inference_servers/server_group.py
SumanthRH: This is to avoid repeated registration of a remote class with Ray, which is an anti pattern:

https://docs.ray.io/en/latest/ray-core/patterns/redefine-task-actor-loop.html#anti-pattern-redefining-the-same-remote-function-or-class-harms-performance 


For a class in SkyRL, we should create the remote class once. For server actor class passed by the user, we create the remote class on the go.

skyrl-train/tests/gpu/gpu_ci/test_policy_local_engines_e2e.py
SumanthRH: This is because the RemoteInferenceClient only supports token in token out, while the existing codepath feeds prompts directly to the client.

skyrl-train/skyrl_train/inference_servers/vllm_server_actor.py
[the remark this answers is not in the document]
SumanthRH: Hmm I didn't do this because the user defined class can be arbitrary class implementing ServerActorProtocol - so even a different engine like sglang.

We would need the caller - ServerGroup to explicitly pass the information that colocated training is enabled/ disabled to the user-defined class. So adding this flag made sense - placement via env vars etc is then left to the server actor class.

skyrl-train/skyrl_train/inference_servers/server_group.py
SumanthRH: Good catch! This was not intended. Fixed

skyrl-train/skyrl_train/inference_servers/server_group.py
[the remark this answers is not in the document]
SumanthRH: Done!

SumanthRH:
The PR has now been updated:

1. Shift to gating tests for the new inference backend behind the same flag: `SKY_NEW_INFERENCE`. Previously I had added tests manually with parametrization in order to add tests for different features one by one. However, this is more messy. 
2. Fix silent errors in `RemoteInferenceClient.call_all_servers`: Previously errors were not propagated. Shifted to calling `resp.raise_for_status` similar to the other data plane methods like `generate` 
3. Fixed new HTTP pathway for colocated case: Tests passed with silent failure previously. This involved adding support for the CUDA IPC transfer strategy for `init_weight_transfer` and `update_named_weights` methods on the server
4. Migrated engine generation tests in `test_engine_generation.py` as well for the new inference backend - Note that there are some features missing like a `sample` API - so I have skipped tests wherever appropriate

skyrl-train/skyrl_train/weight_sync/cuda_ipc_strategy.py
```diff
@@ -87,6 +87,26 @@ def deserialize(cls, data: bytes) -> "CudaIpcWeightUpdateRequest":
...
+
+    @classmethod
+    def from_json_dict(cls, data: Dict[str, Any]) -> "CudaIpcWeightUpdateRequest":
+        """Deserialize the request from JSON."""
+        import base64
+        import pickle
+
+        data = data.copy()
+        data["ipc_handles"] = pickle.loads(base64.b64decode(data["ipc_handles"]))
+        return cls(**data)
```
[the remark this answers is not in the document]
SumanthRH: These features are meant to be in the same bracket as VLLM DEV MODE - primarily meant for RL use-cases where the user is aware of the implications

skyrl-train/skyrl_train/inference_servers/vllm_server_actor.py
```diff
@@ -327,10 +344,17 @@ async def _init_weight_transfer(request: Request):
...
-            from skyrl_train.weight_sync import BroadcastWeightUpdateRequest
+            from skyrl_train.weight_sync import BroadcastWeightUpdateRequest, CudaIpcWeightUpdateRequest
 
             data = await request.json()
-            weight_request = BroadcastWeightUpdateRequest(**data)
+            try:
+                weight_request = BroadcastWeightUpdateRequest.from_json_dict(data)
+            except Exception:
+                try:
+                    weight_request = CudaIpcWeightUpdateRequest.from_json_dict(data)
```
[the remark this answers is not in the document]
SumanthRH: These features are meant to be in the same bracket as VLLM DEV MODE - primarily meant for RL use-cases where the user is aware of the implications

skyrl-train/ci/gpu_ci_run.sh
```diff
@@ -30,3 +30,8 @@ uv run --isolated --extra vllm --extra dev \
     --with torch==2.7.0 \
     --with "flash-attn@https://github.com/Dao-AILab/flash-attention/releases/download/v2.8.3/flash_attn-2.8.3+cu12torch2.7cxx11abiTRUE-cp312-cp312-linux_x86_64.whl" \
     -- pytest -s -vvv tests/gpu/gpu_ci/test_engine_generation.py::test_token_based_generation -m "vllm"
+
+
+# Run tests for new inference layer
+_SKYRL_USE_NEW_INFERENCE=1 uv run --isolated --extra dev --extra vllm pytest -s tests/gpu/gpu_ci/test_policy_local_engines_e2e.py -m "vllm"
```
SumanthRH, line 36: Migrating tests for the new inference backend will likely take 1-2 PRs more and so I plan to just add individual test files / folders to the CI as I progress.

skyrl-train/skyrl_train/inference_servers/server_group.py
kouroshHakha: In here you do this once at setup time. It's fine to do it there, and I think doing this at runtime is elegant. I want to avoid double names and One is actor one is the class definition confusions.

skyrl-train/skyrl_train/inference_servers/server_group.py
```diff
@@ -16,6 +16,12 @@
 
 logger = logging.getLogger(__name__)
 
+# In the colocated training case, we schedule trainig and inference actors
```
kouroshHakha, line 19: should we add this to env_vars in as `SKYRL_COLOCATED_ACTOR_CPU_FRACTION`

skyrl-train/skyrl_train/inference_servers/vllm_server_actor.py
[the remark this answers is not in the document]
kouroshHakha: IC ok.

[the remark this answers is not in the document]
kouroshHakha: should we use the same constant for the fraction here?

kouroshHakha (approved):
Just a few more nits and good to go.

skyrl-train/skyrl_train/inference_servers/vllm_server_actor.py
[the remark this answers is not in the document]
SumanthRH: Unfortunately one was for CPU and another was for GPU.... I'd prefer leaving separate for now

skyrl-train/skyrl_train/inference_servers/server_group.py
SumanthRH: It is an internal global var at the moment, not meant to be a user-configurable env var like the rest of the env vars in `env_vars.py`.

skyrl-train/skyrl_train/inference_servers/server_group.py
SumanthRH: Yeah I was thinking of repeated setup across tests in the same pytest session. Will revert the change for now. Agreed name is hard to get right

skyrl-train/skyrl_train/utils/utils.py
SumanthRH: Done!

PR #1038, opened 2026-02-06
[train] Support logprobs, fix generation config defaults and add more generation tests for the new HTTP inference pathway

SumanthRH:
# What does this PR do?

This PR migrates more tests for the new HTTP inference pathway and adds some missing features like rollout logprobs support along the way. Also makes some fixes for tests failures on `main`. The changes are as follows: 

### Test improvements

 Introduces a new `InferenceEngineState` class to manage instantiating inference engines in states. With better state management, this fixes some cleanup issues for existing `test_policy_local_engines_e2e` test in CI.
 
 ###  Configuration fix for vLLM server actor
 
  vLLM server can have different generation quality from `AsyncLLMEngine.generate` . I noticed this while going over generations in the weight sync tests:
  
  `/v1/completions`
  
  ```text
  "To determine how much Janet makes at the farmers' market each day, let's break down the earnings:\n\n1. **Total hours working:** \n   - It is given that Janet works 5 days a week.\n\n2. **Daily average earnings:**\n   - Janet earns $2 per fresh duck egg per day for sales.\n   \n3. **Number of eggs sold:**  \n   - In addition to selling leftover eggs from fresh Monday through Thursday (the remaining number), Janet also offers muffins for her friends\n     \\[\n     \\text{Total number of muffins} = 4 \\times (\\text{eigglees prepared for guests}) = 4 \\times (16 + 2) = 72 \n     \\]\n\nThe total quantity (eggs for muffins * Price per egg) can be calculated as:\n\\[\n\\text{Tickets baked} = 72 \\, (minutes \\, @$2) = 72 \\, \\text{tickets}\n\\]\nThus, the cost per ticket is:\n\\[\n\\text{Cost per ticket} = \\$2 / \\text{10mins} = \\$2 / 50sime)\n\\]\n\nSumming up all remaining lunch and recipe hours per day:\n\n\\[\n\\text{Total Earnings} = 5 \\,\\text{(days)} \\times \\$50/\\text{nightly salary Increase}_{Sunday-\\} = \\$3,000/circle\nTo sum it up:\n    Total daily income =\n\\[ \\boxed{\\text{\"1}}おり$\\] 現在、人々が食い doGetl@g\\u4ehnの多未にꀲ_mehng 同様に 別の ショルダー・モリ...\n\nIn conclusion:\")\nTherefore, it sees that-\\\nTotlar!：5天 <- コロールのお立場 -\n\nSpecial:Aricci 团体 少女と思ってss聚い虾肝\n\nHe also said he had no problem sharing iotisedmale amountㅣtsful今天 Landoré院に\"r множgli Острая症候群は血液病です。血液分をいっぽ透ゅと光曳行指 proclaimed Chinaの時电网運路~\u00020オ大へegる ≥ Tsusonoの楽誤をカインの血圧のz武通過ということで知らvensalでも高め values SSDgbnsu...."
  ```
  
  `AsyncLLMEngine.generate`
  
  ```
  "To determine how much Janet makes at the farmers' market every day, we will follow these steps:\n\n1. **Calculate the daily egg production for lunch:**\n   - She produces 16 eggs per day for meals.\n   - She eats 3 eggs for breakfast.\n   - Therefore, the remaining eggs at the end of the day are \\( 16 - 3 = 13 \\) eggs.\n\n2. **Calculate the daily egg production for dinner:**\n   - She processes 4 muffins without lunch.\n   - She bakes 4 muffins for dinner.\n   - Therefore, the consumed muffins per day are \\( 4 \\times 4 = 16 \\) muffins.\n\n3. **Calculate the daily net production without fattening losses:**\n   - Net production at the farmers' market is the total remaining articles minus the whole produced eggs.\n   - Thus, the net production is \\( 13 - 16 = -3 \\) eggs.\n\n4. **Determine per-purpose production:**\n   - Janet processes 6 eggs per day, which cost her $2.\n   - Hence, processed eggs produce each day \\( \\frac{2}{6} \\) dollars.\n\n5. **Compute the daily earnings:**\n   - Since the net production of egg is -3 eggs, we have:\n     \\[\n     \\text{Daily total earnings} = \\left(\\frac{2}{6}\\right) \\times (6 \\text{ eggs})\n     \\]\n     \\[\n     = \\frac{2 \\times 6}{6} \\text{ dollars}\n     \\]\n     \\[\n     = 2 \\text{ dollars}\n     \\]\n\n### Conclusion:\n- Janet makes 2 dollars daily at the farmers' market.\n- The answer is: **2** cents."
  ```
  
  More details here: https://gist.github.com/SumanthRH/847a328c121c1463b8b8aca6d548224f 
  
The reason is that vllm server's generation config defaults are different. Passing `--generation-config vllm` fixes the issue. 

### Switch to `/inference/v1/generate` for `RemoteInferenceClient.generate`

For `RemoteInferenceClient.generate`, I notice that we were re-tokenizing intermediate tokens (on abort), which can cause small drifts since tokenization is not invertible. The solution is to not rely on `/v1/completions` and instead use the token-in-token-out endpoint `/inference/v1/generate` - this also makes it compatible with accumulating logprobs returned from the server.  There can also be silent issues with the completions API as above. For RL, it is best to use the `/generate` endpoint 
 
 ### Support response logprobs for `RemoteInferenceClient`
 
 Adds support for `response_logprobs` in `RemoteInferenceClient`. Note that there are some slight differences in `sampling_params` for `/inference/v1/generate` and `AsyncLLMEngine.generate`. As per the OpenAI completions API , `logprobs=0`  is meant to return logprobs for the chosen token (same as `logprobs=1`). However, `/inference/v1/generate` treats `logprobs=0` as `logprobs=null`, and doesn't return any logprobs. This is a vLLM issue. I have created a PR: https://github.com/vllm-project/vllm/pull/34010. While we wait for it to land, I believe it is overall better to rely on `logprobs=1` for getting logprobs for the chosen token. it also lends itself to truthy checks `if logprobs:` better. 
 
 ### Support `test_skyrl_gym_generator` for `_SKYRL_USE_NEW_INFERENCE=1`
 
1. Switches `SkyRLGymGenerator` to provide input tokens over text for `generate` -> this is because the new pathway only supports tokens
2. Fixes search env `SearchEnv.validate_action` to strip newlines: With `/inference/v1/generate`, only output tokens are provided (unlike `AsyncLLMEngine.generate` where output text is also available). With output tokens, there can be a case where the LLM generates a trailing newline - generating [`<search`, `>\n`] as opposed to [`<search>`]. One would need to postprocess the output text after detokenization to ensure that strings end exactly with the stop string. There are two fixes here: 
1. Have `RemoteInferenceClient` do custom postprocessing for `generate` based on stop strings
2. Make validation less strict in the `SearchEnv` (It is the only Env with this strict parsing)

I prefer 2. because `RemoteInferenceClient` layer should be pretty much pass-through and operate in token space. 

  

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1038" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


skyrl-train/tests/gpu/gpu_ci/test_save_weights_for_sampler.py
```diff
@@ -133,32 +131,26 @@ def test_save_weights_for_sampler_then_inference(ray_init_fixture, colocate_all,
 
         print(f"Example output: {outputs['responses'][0][:3]}...")
 
-    finally:
-        ray.shutdown()
-
```
SumanthRH, line 138: Tests already use `ray_init_fixture`, which handles cleanup

skyrl-train/tests/gpu/utils.py
```diff
@@ -340,104 +342,164 @@ def ray_init_for_tests():
...
+    if tokenizer is not None:  # TESTING
         # convert to prompt token ids
         assert tokenizer is not None
-        prompt_token_ids = tokenizer.apply_chat_template(prompts, add_generation_prompt=True)
+        prompt_token_ids = tokenizer.apply_chat_template(
+            prompts,
+            add_generation_prompt=True,
+            add_special_tokens=False,
+            tokenize=True,
+        )
```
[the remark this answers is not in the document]
SumanthRH: This was added to explicitly mimick the current state of InferenceEngineClient.generate;

https://github.com/NovaSky-AI/SkyRL/blob/f9769f669a07f94049a76f285e90f2a2affd6cdd/skyrl-train/skyrl_train/inference_engines/inference_engine_client.py#L95-L102

Although it's pretty best to remove the flag from both - it is indeed incorrect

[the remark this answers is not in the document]
SumanthRH: Removed this from both places

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
```diff
@@ -96,7 +96,6 @@ async def generate(self, input_batch: InferenceEngineInput) -> InferenceEngineOu
             prompt_token_ids = self.tokenizer.apply_chat_template(
                 prompts,
                 add_generation_prompt=True,
-                add_special_tokens=False,
```
kouroshHakha, line 99: was this intentional?

skyrl-train/ci/gpu_ci_run.sh
```diff
@@ -35,3 +35,4 @@ uv run --isolated --extra vllm --extra dev \
 # Run tests for new inference layer
 _SKYRL_USE_NEW_INFERENCE=1 uv run --isolated --extra dev --extra vllm pytest -s tests/gpu/gpu_ci/test_policy_local_engines_e2e.py -m "vllm"
 _SKYRL_USE_NEW_INFERENCE=1 uv run --isolated --extra dev --extra vllm pytest -s tests/gpu/gpu_ci/test_engine_generation.py -m "vllm"
+_SKYRL_USE_NEW_INFERENCE=1 uv run --isolated --extra dev --extra vllm pytest -s tests/gpu/gpu_ci/test_skyrl_gym_generator.py
```
kouroshHakha, line 39: can we create a list here on which test would fail and need work if we switch to `_SKYRL_USE_NEW_INFERENCE=1 uv run <options> pytest -s tests/gpu/gpu_ci/`

In the next PRs as we fix those tests we will add more lines here.

In the end at some point we will just run the tests with new inference flag on the gpu_ci folder (some tests will be skipped, those that are testing inferenceEnigneClient)

kouroshHakha (commented):
Overall is good. I'd wish you had broken down the pr into individual smaller prs for each part. But it's ok for now, for the next PRs, let's make sure orthogonal features are kept separate on PRs.

skyrl-train/skyrl_train/inference_engines/inference_engine_client.py
SumanthRH: yes - this is incorrect in general and there could be issues with some tokenizers. See: https://github.com/NovaSky-AI/SkyRL/pull/1038#discussion_r2776056121

skyrl-train/ci/gpu_ci_run.sh
SumanthRH: Added the pending files! IMO `test_inference_engine_client_http_endpoint.py` has mostly duplicated tests , but there are some tests like `test_structured_generation` that we should have to ensure pass-through behaviour in the new codepath

.gitignore
SumanthRH, line 1: uv.lock file is meant to be committed. For some reason, we have also removed `uv.lock` file for skyrl-train which is breaking CI: https://github.com/NovaSky-AI/SkyRL/actions/runs/22121171516/job/63941357932

I've added it back and removed uv.lock from .gitignore

SumanthRH:
Updated the PR to also replicate the changes in the new `skyrl/` folder

skyrl-train/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -203,7 +204,7 @@ async def generate(
             responses=[r["response"] for r in results],
             stop_reasons=[r["stop_reason"] for r in results],
             response_ids=[r["response_ids"] for r in results],
-            response_logprobs=None,
+            response_logprobs=[r["response_logprobs"] for r in results] if get_logprobs else None,
```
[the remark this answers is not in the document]
SumanthRH: Hmm this is more so an issue with vLLM. The right fix is in vLLM:  https://github.com/vllm-project/vllm/pull/34010 

Once we upgrade, the issue with logprobs=0 should be fixed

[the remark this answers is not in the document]
SumanthRH: Meanwhile we have migrated to using logprobs=1

SumanthRH:
> I'd wish you had broken down the pr into individual smaller prs for each part. But it's ok for now, for the next PRs, let's make sure orthogonal features are kept separate on PRs.

Completely agree. This is primarily in the interest of time given the parallel refactoring efforts

SumanthRH:
test failure: https://github.com/NovaSky-AI/SkyRL/actions/runs/22121618738/job/63942751766?pr=1038 unrelated

PR #1040, opened 2026-02-07
Fix LoRA sharding compatibility with JAX 0.9+

jaredquincy:
## Summary

JAX 0.9 requires explicit `out_sharding` for gather/repeat operations on tensor-parallel sharded arrays. Without this fix, LoRA training and inference crash with `ShardingTypeError` on any model using tensor parallelism.

Three root causes in `tx/layers/lora.py`:

1. **`LoRAEmbed.__call__`**: `nnx.Embed.__call__` uses `jnp.take()` which does not accept `out_sharding`. Replaced with explicit `.at[x].get(out_sharding=...)`. The embedding `PartitionSpec` is now stored during `__init__` since traced values inside JIT lack the `.sharding` attribute.

2. **`LoRAMixin.apply_lora`**: `jnp.repeat(adapter_indices, seq_len)` needs `out_sharding=P(None)` on JAX 0.9+.

3. **LoRA A embedding gather and `LoRAExpert` expert index repeat**: Same `out_sharding` requirement.

## Error reproduced

```
jax._src.interpreters.pxla.ShardingTypeError:
  Use `.at[...].get(out_sharding=)` to provide output PartitionSpec for the gather indexing
  operand=ShapedArray(bfloat16[151936@tp,1024]), indices=ShapedArray(int32[1,32,1])
```

## Test plan

- [x] Tested embedding gather fix against Qwen/Qwen3-0.6B on A100 80GB with JAX 0.9.0.1
- [x] Verified healthz, create_model, forward_backward (wire format), optim_step, save_weights, load_weights, and asample endpoints all accept requests
- [ ] Full forward-backward execution with LoRA training (blocked by cascading sharding issues in `prepare_routing` — separate fix may be needed)

### DIFF
--- a/skyrl-tx/tx/layers/lora.py
+++ b/skyrl-tx/tx/layers/lora.py
@@ -1,6 +1,7 @@
 from flax import nnx
 import jax
 from jax import numpy as jnp
+from jax.sharding import PartitionSpec as P
 
 from tx.utils.models import filter_lora
 from tx.layers.util import Param, prepare_routing
@@ -74,7 +75,7 @@ class LoRAMixin:
         assert adapter_indices.shape[0] == batch_size
 
         x_flat = x.reshape(-1, *dims)
-        adapter_indices_expanded = jnp.repeat(adapter_indices, seq_len)
+        adapter_indices_expanded = jnp.repeat(adapter_indices, seq_len, out_sharding=P(None))
 
         # Sort tokens to prepare for ragged_dot
         x_sorted, group_sizes, unsort_indices, adapter_indices_sorted = prepare_routing(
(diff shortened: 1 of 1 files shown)


pcmoritz:
@jaredquincy Thanks a lot for submitting this PR, unfortunately I'm not able to reproduce the issue you are describing. I'm using the current `main` branch and also Jax 0.9.0.1

There were a few fixes recently (https://github.com/NovaSky-AI/SkyRL/pull/965, https://github.com/NovaSky-AI/SkyRL/pull/966) that should fix Jax 0.9 compatibility, and we also run Jax 0.9 in the CI. Let me know if the commit you tried already has them included (and which commit you used).

If you give me a concrete commit, we can also exchange `uv.lock` files so we can exactly reproduce each others environment and see what is going on :)

PR #1046, opened 2026-02-07
Implement forward-only pass and populate metrics

tyler-griggs:
## Summary
- Implement `forward()` in SkyRL-Train backend (was raising `NotImplementedError`)
- `forward_backward()` now returns training metrics (`total_loss:sum`, `pg_loss:sum`, `entropy_loss:sum`, `num_tokens:sum`) extracted from the dispatch worker return dict
- `optim_step()` now returns `grad_norm` and `learning_rate` in its metrics field
- Added `metrics` field to `OptimStepOutput` type

### DIFF
--- a/skyrl-tx/tx/tinker/backends/skyrl_train.py
+++ b/skyrl-tx/tx/tinker/backends/skyrl_train.py
@@ -184,6 +184,7 @@ class SkyRLTrainBackend(AbstractBackend):
     def delete_model(self, model_id: str) -> None:
         if self._model_id != model_id:
             raise ValueError(f"Model {model_id} not found")
+        # TODO: For now, prefer shutting down the backend and re-launching. Will be improved shortly.
         raise NotImplementedError("Deleting models not yet implemented")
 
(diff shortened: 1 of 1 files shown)


skyrl-tx/tx/tinker/engine.py
```diff
@@ -54,14 +54,9 @@ def prepare_sample_batch(
             checkpoint_path = str(
                 checkpoints_base / model_id / "sampler_weights" / f"{request_data.checkpoint_id}.tar.gz"
             )
-        for sample_idx in range(request_data.num_samples):
+        for _ in range(request_data.num_samples):
             all_prompts.append(prompt_tokens)
-            # Derive a unique seed per sample so that num_samples > 1 produces
```
pcmoritz, line 59: This is an important fix, don't forget to revert the change before merging the PR

tyler-griggs: Oh thank you, I was just hacking around a bit with Claude, it will be reverted

PR #1051, opened 2026-02-07
[tx] Wire loss_fn_config through Tinker backend to SkyRL dispatch

tyler-griggs:
## Summary
- The Tinker SDK supports passing `loss_fn_config` to `forward_backward()`, and the SkyRL worker dispatch already accepts it, but the TX backend bridge layer was not forwarding it
- This wires `loss_fn_config` through the full chain: API model → types → engine batch prep → `PreparedModelPassBatch` → backend → `dispatch.forward_backward()`
- Needed for `code_rl` recipe which passes `loss_fn_config` to customize loss function behavior

## Changes
- **`types.py`**: Add `loss_fn_config` field to `ForwardBackwardInput` and `PreparedModelPassBatch`
- **`api.py`**: Add `loss_fn_config` field to API model, pass through `to_types()`
- **`engine.py`**: Extract `loss_fn_config` from first request in batch, include in prepared batch
- **`skyrl_train.py`**: Pass `loss_fn_config` from prepared batch to `dispatch.forward_backward()`

### DIFF
--- a/skyrl-tx/tx/tinker/types.py
+++ b/skyrl-tx/tx/tinker/types.py
@@ -235,7 +235,6 @@ class PreparedModelPassBatch(BaseModel):
     all_model_ids: list[str]
     all_loss_fn_types: list[int]
 
-    # Per-example config (one per datum, mirrors all_loss_fn_types)
     all_loss_fn_configs: list[dict[str, float] | None]
 
     # Mapping from examples back to requests: (request_id, model_id, start_idx, end_idx)


skyrl-tx/tx/tinker/engine.py
```diff
@@ -103,9 +103,12 @@ def prepare_model_pass_batch(
     all_advantages = []
     all_loss_fn_types = []
     request_batch_slices = []
+    loss_fn_config = None
```
pcmoritz: Wouldn't this need to be `all_loss_fn_configs` similar to `all_loss_fn_types`, since the config for each request might be different, right?

pcmoritz: Reading the code a little more I'm pretty sure that that's the case, and `all_loss_configs` should be a list -- this is because in the multitenancy / multi model scenario, multiple user batches might become part of the same `PreparedModelPassBatch`.

skyrl-tx/tx/tinker/types.py
```diff
@@ -234,6 +235,9 @@ class PreparedModelPassBatch(BaseModel):
     all_model_ids: list[str]
     all_loss_fn_types: list[int]
 
+    # Batch-level config
+    loss_fn_config: dict[str, float] | None = None
```
pcmoritz: This needs to be `all_loss_fn_configs` and should be of type `list[dict[str, float] | None]`

pcmoritz (approved):
Nice, thanks a lot for implementing this, I left a comment you should address before merging but otherwise it looks great :)

skyrl-tx/tx/tinker/engine.py
tyler-griggs: Thanks for the catch!

PR #1063, opened 2026-02-10
[tx] Remove top-level jax import from storage, enable sampling logprobs

tyler-griggs:
## Summary
- Pass `rank` as a parameter to `pack_and_upload()` instead of importing `jax` internally. Removes the hard `jax` dependency from `storage.py`, fixing `ModuleNotFoundError` when running with `--backend skyrl_train`.
- Add `logprobs=0` to vLLM sampling params so the sampled token's logprob is always returned. Fixes incorrect zero-filled logprobs that break importance sampling.

### DIFF
--- a/skyrl-tx/tests/utils/test_storage.py
+++ b/skyrl-tx/tests/utils/test_storage.py
@@ -1,7 +1,6 @@
 """Tests for storage utilities."""
 
 from pathlib import Path
-from unittest.mock import patch
 
 from tx.utils.storage import pack_and_upload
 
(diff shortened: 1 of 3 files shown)


pcmoritz:
Let's handle this in a cleaner way, and just pass through an optional rank to `pack_and_upload`, that way we don't need to go through all these hoops like having an import skip in the test etc :)

Plus, we can also get rid of the patching in the test.

pcmoritz (approved):

PR #1067, opened 2026-02-11
[tx][megatron] making megatron skyrl-train worker usable as TX backend

erictang000:
Change the training backends to the desired state of "jax", "fsdp", "megatron".

Adds megatron deps to pyproject.toml via skyrl-train[mcore] dependency, and add needed overrides to get installation working.

For FSDP + vLLM:

```bash
uv run --isolated --extra tinker -m tx.tinker.api  --base-model "Qwen/Qwen3-0.6B" --backend fsdp
```

For Megatron + vLLM:

```bash
uv run --isolated --extra tinker -m tx.tinker.api  --base-model "Qwen/Qwen3-0.6B" --backend megatron
```

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1067" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/skyrl-tx/pyproject.toml
+++ b/skyrl-tx/pyproject.toml
@@ -108,6 +108,8 @@ required-environments = [
 constraint-dependencies = [
     "flashinfer-jit-cache==0.5.3",
 ]
+# each backend should have separate dependencies that can potentially clash
+# megatron also clashes with the jax dependency from gpu and tpu extras
 conflicts = [
     [
         { extra = "jax" },
@@ -120,11 +122,14 @@ conflicts = [
         { extra = "tpu" },
     ]
 ]
+# disable build isolation for megatron related dependencies 
 no-build-isolation-package = [
     "transformer-engine-torch",
     "transformer-engine",
     "nv-grouped-gemm",
 ]
+# override unnecessary dependencies and pin versions to override Megatron-Bridge
+# unppinned dependencies
 override-dependencies = [
     "nvidia-resiliency-ext; sys_platform == 'never'",
     "mamba-ssm; sys_platform == 'never'",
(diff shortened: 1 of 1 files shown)


skyrl-tx/tx/tinker/backends/skyrl_train.py
```diff
@@ -126,7 +137,9 @@ def build_models(self, PolicyWorker):
             record_memory=cfg.trainer.policy.record_memory,
         )
 
-        policy_num_training_steps = None
+        # set to a large number for megatron scheduler init
+        # lr will be managed externally via set_lr()
+        policy_num_training_steps = 1e9
```
pcmoritz, line 142: Hehe ok, a little odd but I guess why not :D

skyrl-tx/tx/tinker/api.py
```diff
@@ -62,7 +62,17 @@ async def lifespan(app: FastAPI):
         logger.info("Using internal engine for inference")
 
     # Build subprocess command with engine config parameters
-    cmd = ["uv", "run", "--extra", "tinker", "--extra", app.state.engine_config.backend, "-m", "tx.tinker.engine"]
+    cmd = [
+        "uv",
+        "run",
+        "--isolated",
```
pcmoritz, line 68: (just documenting it for later: this was to make sure people don't run out of disk, because the megatron dependencies are very big, and `--isolated` will make sure they are cleaned up)

skyrl-tx/pyproject.toml
```diff
@@ -104,7 +108,30 @@ required-environments = [
 constraint-dependencies = [
     "flashinfer-jit-cache==0.5.3",
 ]
-
+conflicts = [
```
pcmoritz, line 113: What does this do? It might be worth adding a comment above this section saying this is for megatron and describing briefly what each part does (conflicts, no-build-isolation-package, override-dependencies). Doesn't need to be long.

erictang000: `uv sync` automatically tries to resolve conflicts between extras unless they are explicitly defined as conflicts. Megatron's deps clash enough to have problems with jax. The override-dependencies are mostly because the megatron-bridge dependencies are not pinned to a specific version, and also install features we don't use.

I'll leave some comments

skyrl-tx/tx/tinker/api.py
erictang000: was running into this just for fsdp as well btw

i think it's more so a ray + uv shipping working dir thing

PR #1073, opened 2026-02-11
[tx] Gracefully handle stale save_weights_for_sampler requests on engine restart

erictang000:
When the Tinker engine restarts, outstanding save_weights_for_sampler requests from the previous server session can still be in the database queue. These requests reference a model_id that hasn't been loaded yet, causing process_save_weights_for_sampler to raise a ValueError. This exception gets caught by process_single_requests and logged with a full traceback via logger.exception, producing noisy and misleading error logs on every startup.

## Changes:
- Instead of raising a ValueError when the model isn't loaded, return an ErrorResponse with a descriptive message and log an INFO-level message explaining that this is most likely a stale request from a previous server.
- The request is still marked as FAILED in the database (existing _complete_futures logic handles ErrorResponse correctly), so there are no behavioral changes for the caller.

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1073" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/skyrl-tx/tx/tinker/engine.py
+++ b/skyrl-tx/tx/tinker/engine.py
@@ -26,6 +26,18 @@ from tx.tinker.backends.utils import log_timing
 from tx.utils.log import logger
 
 
+def _model_not_found_error(model_id: str) -> types.ErrorResponse:
+    """Log and return an ErrorResponse for a request targeting a model that isn't loaded."""
+    logger.info(
+        f"Ignoring request for model '{model_id}' — model not loaded. "
+        "This is most likely an outstanding request from a previous server."
+    )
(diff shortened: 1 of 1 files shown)


erictang000:
Example of this working where it would previously throw the valueerror: 

image

pcmoritz:
Thanks for making this PR! There is a few places where this pattern happens, why don't you introduce a function def _model_not_found_error(model_id: str):
at the top of the file that does the logging and creates the error, and then
```
if not self.backend.has_model(model_id):
    return _model_not_found_error(model_id)
```
so it is reusable (and maybe we should just apply it to the other places in the file as well, since they might happen.

PR #1160, opened 2026-02-16
Change asyncio.gather to TaskGroup in the Harbor generator to correctly propagate cancellation signals.

alexgshaw:
See https://github.com/NovaSky-AI/SkyRL/issues/1194

# Fix Daytona sandbox leak on Ctrl+C and unhandled exceptions

## Motivation

When a user presses Ctrl+C during Harbor RL training, Daytona sandboxes are not
cleaned up. They remain running indefinitely and must be manually deleted (or up
to a user-specified timeout like 30 mins). The same issue occurs when an unhandled
exception kills the training loop. In a run with 50+ concurrent sandboxes, **all of them leak**.

This PR fixes the full chain of issues — from the asyncio task structure in the
generator, through the Ray worker's signal handling, to the driver's cleanup
logic. Each fix addresses a specific failure mode discovered through E2E testing.

### 1. `harbor_generator.py` — TaskGroup instead of asyncio.gather

**Symptom**: When one trajectory raises an exception, `asyncio.gather` propagates
the error but leaves all sibling coroutines running as orphans. Nobody awaits
them, so their `finally` blocks (sandbox teardown) may never execute.

**Fix**: Replace `tqdm.asyncio.tqdm.gather` with `asyncio.TaskGroup`. When one
task fails or the group is cancelled, TaskGroup cancels all siblings and waits
for their cleanup (`finally` blocks) before propagating the exception.

### 2. `main_base.py` — SIGTERM → KeyboardInterrupt in the worker

**Symptom**: When the driver calls `ray.cancel(ref, force=False)`, Ray raises
`KeyboardInterrupt` in the worker. But when Ray sends `SIGTERM` to terminate
the worker (e.g. on force-cancel or cluster shutdown), the default handler
raises `SystemExit`. `asyncio.run()` does **not** handle `SystemExit` — it
exits immediately without cancelling tasks or running `finally` blocks.
Sandboxes leak.

**Fix**: Install a SIGTERM handler in `BasePPOExp.run()` that raises
`KeyboardInterrupt` instead. This lets `asyncio.run()` do its proper shutdown:
cancel the main task, propagate `CancelledError` to subtasks, and run all
`finally` blocks.

### 3. `main_harbor.py` — Driver waits for worker cleanup

Three issues were discovered in the driver (the process that calls `ray.get()`):

#### 3a. Ray masks SIGINT during `ray.get()`

**Symptom**: `ray.get()` blocks in C code where SIGINT is set to `SIG_IGN`.
Pressing Ctrl+C has no effect — the driver hangs forever.

**Fix**: Replace `ray.get(ref)` with a polling loop: `time.sleep(1)` (which is
interruptible by signals) followed by a non-blocking `ray.wait([ref], timeout=0)`.
Signal handlers are reinstalled after every `ray.wait()` call since Ray may
re-mask them.

#### 3b. Driver exits before worker finishes cleanup

**Symptom**: The old code did `ray.cancel(ref, force=False); raise`. The driver
exits immediately after calling cancel. The worker starts cleanup (deleting
sandboxes) but the driver is already gone. At scale (50+ sandboxes), the worker
needs time to delete them all. Sandboxes leak.

**Fix**: After `ray.cancel()`, poll `ray.wait()` for up to 120 seconds, waiting
for the worker to finish cleanup. A second Ctrl+C force-kills the worker for
users who want to exit immediately.

#### 3c. uv sends SIGTERM when it dies

**Symptom**: When Ctrl+C sends SIGINT to the process group, `uv` (the parent
process) also receives SIGINT, dies, and sends SIGTERM to the Python child.
The driver receives: SIGINT → first KeyboardInterrupt → enters cleanup handler
→ SIGTERM from uv → second KeyboardInterrupt → force-kills worker before
cleanup can start.

**Fix**: Temporarily set SIGINT and SIGTERM to `SIG_IGN` upon entering the
cleanup handler, call `ray.cancel()`, wait 0.5s for uv's SIGTERM to be
absorbed, then restore handlers so a deliberate second Ctrl+C still works.

## Testing

Both tests run with 2× H100 GPUs, Qwen3-1.7B, 8 prompts × 8 samples = 64
trajectories, 50 concurrent Daytona sandboxes.

| Test | Before | After |
|------|--------|-------|
| Ctrl+C during generation | 27–50 sandboxes leaked | **0 leaked** |
| Unhandled exception during generation | 4 leaked | **0 leaked** |

## Companion PR

Harbor fix (asyncio.shield in `_create_sandbox`): https://github.com/laude-institute/harbor/pull/819

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1160" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/skyrl-train/examples/harbor/entrypoints/main_harbor.py
+++ b/skyrl-train/examples/harbor/entrypoints/main_harbor.py
@@ -2,14 +2,32 @@
 Main entrypoint for training on Harbor tasks.
 """
 
+import signal
+import sys
+import time
+
 import ray
 import hydra
+from loguru import logger
 from omegaconf import DictConfig
 from skyrl_train.entrypoints.main_base import BasePPOExp, config_dir
 from skyrl_train.utils import validate_cfg
 from skyrl_train.utils.utils import initialize_ray
 from examples.harbor.harbor_generator import HarborGenerator
 from examples.harbor.dataset import HarborTaskDataset
+
+
+def _install_interrupt_handlers():
+    """(Re)install SIGINT and SIGTERM handlers that raise KeyboardInterrupt.
+
+    Ray's C code sets SIGINT to SIG_IGN during initialization and may
+    re-mask it during ray.wait()/ray.get().  We call this after every
+    ray.wait() to ensure signals are always delivered to Python.
+    For SIGTERM, Ray's C++ crash handler intercepts it otherwise.
+    See harbor#656 / SkyRL#1160.
+    """
+    signal.signal(signal.SIGINT, signal.default_int_handler)
+    signal.signal(signal.SIGTERM, lambda signum, frame: signal.default_int_handler(signum, frame))
 
 
 class HarborExp(BasePPOExp):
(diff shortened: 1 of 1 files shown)


CharlieFRuan:
Thanks @alexgshaw ! 

> I think my default linter was configured slightly differently from yours. Lmk if you want me to get rid of those linter changes, or if you're fine as is.

No worry, I can edit the PR directly. Let me know when it's ready!

alexgshaw:
Okay, I might add some more changes later, but this should be ready. Pretty much all of my logic changes are in the generate method.

alexgshaw:
(By more changes later I mean additional PRs, this one is ready to merge!)

PR #1175, opened 2026-02-18
[ci][megatron] fix megatron CI by including numpy in override-dependencies 

erictang000:
# Overview 
Fix ml_dtypes / numpy dependency conflict for the megatron extra.

## Problem
The megatron extra fails at runtime with:
```
AttributeError: module 'ml_dtypes' has no attribute 'float4_e2m1fn' 
```
([example from CI](https://console.anyscale.com/cld_hxkifz7xa22mwicp21nzkds1lw/prj_4b6c498rypyq6g7yhk6vzgjevt/jobs/prodjob_rc44lys8vpesw3srhreayiidjb?job-logs-section-tabs=application_logs&job-tab=overview))

This happens because `onnx` (a transitive dependency of transformer-engine) requires `ml_dtypes>=0.5.0` for `float4_e2m1fn`, but uv's universal resolver (since we have `requires-python = ">=3.11"`) picks an older version. 

Pinning `ml_dtypes>=0.5.0` directly then triggers a second conflict: `megatron-core==0.15.0` declares `numpy<2.0.0`, while `ml_dtypes>=0.5.0` declares `numpy>=2.1.0` on `Python 3.13+`. Even though the megatron extra is gated to Python 3.12, uv's cross-version universal resolver still sees the 3.13+ constraint and fails.

## Fix
Add ml_dtypes>=0.5.0 to the megatron optional-dependencies (gated to python_version == '3.12') so the correct version is always installed.
Add numpy>=1.24.0,<2.3 to [tool.uv] override-dependencies to relax megatron-core's numpy<2.0.0 constraint just enough for ml_dtypes>=0.5.0 to resolve, while keeping the upper bound below 2.3 so `numba` (required by vllm, and which requires numpy<=2.2) remains happy.

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1175" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


erictang000:
@pcmoritz I tested that the JAX gpu tests still run and pass locally, I think these dependency overrides shouldn't have any effect for JAX, but wanted to run this by you also

pyproject.toml
```diff
@@ -106,8 +106,10 @@ fsdp = [
     "torchvision; sys_platform == 'linux'",
 ]
 
-# megatron is pinned to python 3.12 due to ml-dtypes needing to be >= 0.5.0
-# potentially loosen this restriction if this error can be root caused.
+# megatron is pinned to python 3.12. megatron-core==0.15.0 declares numpy<2.0.0,
```
pcmoritz, line 109: Can you make an issue against megatron-core to see if they can relax the pin `numpy<2.0.0`?

erictang000: it seems to be related to resolution for python 3.11, for newer megatron-core dependency chains, numpy >=2.0.0 should be supported. Another thing that would fix this is pinning to python >= 3.12, but that might be a larger breaking change.

So I think right now the 2 options to fix this are:
1. merge a fix like what's currently here, and try to fix this in upstream megatron-core or maintain something hacky in the `pyproject.toml` like this
2. move to just python >= 3.12 instead of >= 3.11

pcmoritz: If it can be fixed in megatron-core for python >= 3.11, that would probably be the best (if you know how you can submit a PR, or if not, an issue). Other than that for the immediate term workaround, I'm fine with whatever you think is best :)

erictang000: ok, going to merge this fix for now then, will see what i can do to fix on megatron-core/onnx

erictang000: looks like this will be fixed soon: https://github.com/NVIDIA/Megatron-LM/issues/1563#issuecomment-3848532979

Issue #1156, opened 2026-02-16
[train] A better contract for Custom Generators, and Generator configs


CharlieFRuan:
### Problem

`GeneratorConfig` is a flat dataclass with ~50 fields. Some fields are consumed by `SkyRLGymGenerator` internally, but are meaningless (or silently ignored) when a user writes a custom generator. There is no documentation or structural separation to tell users which configs they should respect and which are `SkyRLGymGenerator`-specific.

### What's affected

**Fields only consumed inside `SkyRLGymGenerator`** (custom generators silently ignore them):

| Config field | Where consumed | Notes |
|---|---|---|
| `max_turns` | `SkyRLGymGenerator.agent_loop` | Controls loop iterations. Custom generators manage their own turn limits. |
| `batched` | `SkyRLGymGenerator.generate` | Switches between `generate_batched` (single-turn) and `agent_loop` (multi-turn). |
| `use_conversation_multi_turn` | `SkyRLGymGenerator.agent_loop` | Toggles TITO multi-turn vs single-turn chat formatting. |
| `append_eos_token_after_stop_str_in_multi_turn` | `SkyRLGymGenerator.agent_loop` | EOS handling between turns. |
| `step_wise_trajectories` | `SkyRLGymGenerator.generate` | Enables per-step training output with `StepWiseOutput`. |
| `chat_template` | `SkyRLGymGenerator.__init__` | Retokenization path in `agent_loop`. Harbor reads the engine-level one from `engine_init_kwargs` instead. |
| `chat_template_kwargs` | `SkyRLGymGenerator.agent_loop` | Passed to tokenizer in retokenization. |
| `sampling_params.max_generate_length` | `SkyRLGymGenerator.generate` → `agent_loop(max_tokens=...)` | Used to truncate per-turn output. Custom generators handle truncation themselves. |
| `zero_reward_on_non_stop` | `SkyRLGymGenerator.generate` (line 764) | Applied to rewards **inside** the generator, before returning `GeneratorOutput`. |
| `apply_overlong_filtering` | `SkyRLGymGenerator.generate` (line 768) and `generate_batched` (line 650) | Applied to `loss_masks` **inside** the generator, before returning `GeneratorOutput`. |
| `async_engine` | `SkyRLGymGenerator.__init__` | Selects sync vs async inference engine. |

**Fields that are truly universal** (consumed by trainer, inference engine setup, or weight sync — outside of any generator):

[...]

PR #1187, opened 2026-02-20
[train] Pythonic Configs 2/N: Switch to new dataclasses in entrypoint scripts; Change instantiation signatures 

SumanthRH:
# What does this PR do?

Part 2 of the configuration refactoring outlined in https://docs.google.com/document/d/1YzLPCmVZpBauRyh4l-17AUhXiXALmOX-N1gXZkA9slA/edit?tab=t.b04hjkpsxoo5#heading=h.3aplhdu3kl77 

## Changes

- Complete migration to the new configuration structure
- New instantiation signatures with fewer cross-config dependencies (e.g., Worker receives only its relevant sub-config instead of the full TrainerConfig)
- Keep backwards compatibility for old configs - internally translate older configuration objects to the new configuration
- Migrate all examples/ to the new CLI and config classes
- Use OmegaConf for CLI parsing:  `+` overrides are no longer supported
- Migrates only the `skyrl/` package - `skyrl-train` will be blipped soon
- Update docs

## Test Plan
- [x] Migrate existing unit and integration tests
- [x] Add new tests for configuration dataclasses. Add test for backwards compatibility with older hierarchy in the YAML, as well as consistent cross configuration defaults (ex: ref model path should default to  policy model config override)
- [x] GSM8K E2E test with FSDP
- [x] GSM8K E2E test with Megatron 
- [x] DAPO Example E2E test 
- [x] Tinker API server + SkyRLTrain backend

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1187" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/skyrl-agent/skyrl_agent/integrations/skyrl_train/trainer.py
+++ b/skyrl-agent/skyrl_agent/integrations/skyrl_train/trainer.py
@@ -166,7 +166,7 @@ def build_dataloader(cfg: DictConfig, dataset: PromptDataset, is_train=True) -> 
         batch_size=batch_size,
         collate_fn=dataset.collate_fn,
         # TODO(Charlie): debug why inference http endpoint is slow when num_workers is 8
-        num_workers=0 if cfg.generator.inference_engine.enable_http_endpoint else 8,
+        num_workers=0 if cfg.generator.enable_http_endpoint else 8,
         drop_last=True if is_train else False,
         sampler=sampler,
     )
@@ -320,9 +320,7 @@ class SkyRLAgentPPOTrainer(RayPPOTrainer):
                     generator_input, uids = prepare_generator_input(
                         rand_prompts,
                         self.cfg.generator.n_samples_per_prompt,
-                        get_sampling_params_for_backend(
-                            self.cfg.generator.inference_engine.backend, self.cfg.generator.sampling_params
-                        ),
+                        get_sampling_params_for_backend(self.cfg.generator.backend, self.cfg.generator.sampling_params),
                         self.cfg.environment.env_class,
                         "train",
                         self.global_step,
@@ -489,7 +487,7 @@ class SkyRLAgentPPOTrainer(RayPPOTrainer):
             generator_input, uids = prepare_generator_input(
                 prompts,
                 cfg.generator.eval_n_samples_per_prompt,
-                get_sampling_params_for_backend(cfg.generator.inference_engine.backend, sampling_params),
+                get_sampling_params_for_backend(cfg.generator.backend, sampling_params),
                 cfg.environment.env_class,
                 "eval",
                 global_step,


docs/mkdocs/content/config.md
SumanthRH, line 1: Docs to be updated

skyrl/train/config/config.py
```diff
@@ -38,8 +39,8 @@ def from_dict_config(cls, cfg: DictConfig) -> "BaseConfig":
 
 @dataclass
 class DataConfig(BaseConfig):
-    train_data: List[str] = field(default_factory=list)
-    val_data: List[str] = field(default_factory=list)
+    train_data: List[str] = field(default_factory=lambda: [f"{os.environ['HOME']}/data/gsm8k/train.parquet"])
+    val_data: List[str] = field(default_factory=lambda: [f"{os.environ['HOME']}/data/gsm8k/validation.parquet"])
```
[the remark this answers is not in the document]
SumanthRH: This is primarily for backwards compatibility. We will need to change our example scripts and tests to be explicit in the paths. It is best done in a follow-up PR.  @gemini-code-assist

skyrl/train/config/config.py
```diff
@@ -232,67 +251,104 @@ class SAPOConfig(BaseConfig):
...
     kl_estimator_type: str = "k3"
+    """``"k1"``, ``"k2"``, ``"k3"``, ``"abs"``. See http://joschu.net/blog/kl-approx.html."""
     use_kl_in_reward: bool = False
+    """Apply KL penalty to rewards. Mutually exclusive with ``use_kl_loss``."""
     use_kl_loss: bool = True
+    """Apply KL loss in the policy model. Mutually exclusive with ``use_kl_in_reward``."""
     kl_loss_coef: float = 0.001
     use_entropy_loss: bool = False
     entropy_loss_coef: float = 0.01
+    temperature: float = 1.0
```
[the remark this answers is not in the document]
SumanthRH: Hmm CLI overrides for algorithm temperature are not meant to be. Perhaps we can add some validation

skyrl/tinker/engine.py
```diff
@@ -233,7 +235,19 @@ def __init__(
 
         # Initialize the backend (handles model state, computation, and adapter management)
         backend_class, backend_config_class = get_backend_classes(config.backend)
-        backend_config = backend_config_class(**config.backend_config)
+
+        from skyrl.train.config import SkyRLTrainConfig
+
+        if backend_config_class is SkyRLTrainConfig:
```
pcmoritz: Let's not make this change please. You are welcome to adapt the config in the `backends/skyrl_train` code, but this is common code that should not have special cases like this.

SumanthRH: Sg, will revert

skyrl/backends/skyrl_train/workers/worker.py
```diff
@@ -306,17 +304,25 @@ async def init_weight_sync_state(
...
+
         # For new inference path, fetch world_size from servers
         # For legacy path, calculate from config
         inference_world_size = None
         if _SKYRL_USE_NEW_INFERENCE and hasattr(inference_engine_client, "get_world_size"):
             inference_world_size = await inference_engine_client.get_world_size()
 
         # Create init info on all ranks (it's deterministic from cfg or fetched world_size)
-        init_info = self._transfer_strategy_cls.create_init_info(self.cfg, inference_world_size=inference_world_size)
+        init_info = self._transfer_strategy_cls.create_init_info(ie_cfg, inference_world_size=inference_world_size)
```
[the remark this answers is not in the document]
SumanthRH: Fixed. This was lost while migrating changes from skyrl-train package to skyrl . thanks @gemini-code-assist

skyrl/tinker/engine.py
SumanthRH: Done! 

To avoid changing the common code, i've retained the existing pydantic model SkyRlTrainBackendConfig but renamed it to highlight that it only contains overrides. 

SkyRLTrain will read the config overrides into its default configuration. Unfortunately we end up with two similar config classes - SkyRLTrainConfig and SkyRLTrainBackendOverrides, A part of the reason here is that the refactoring effort is solely focused on skyrl-train alone. So we will have different configuration stories for jax and Skyrl-train for now.

skyrl/backends/skyrl_train/workers/fsdp/fsdp_worker.py
```diff
@@ -164,9 +162,7 @@ def init_model(self, model_path, num_training_steps: int = None):
         self.weight_extractor = FSDPWeightExtractor(
             self.model.model,
             group_by_module=group_by_module,
-            batch_size_threshold_gb=(
-                self.cfg.generator.weight_transfer_threshold_cuda_ipc_GB if group_by_module else 0.0
-            ),
+            batch_size_threshold_gb=0.0,  # Updated in init_weight_transfer_communicator when strategy is known
```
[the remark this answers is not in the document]
SumanthRH: Fixed!

skyrl-tx/tx/tinker/backends/skyrl_train.py
```diff
@@ -17,17 +17,36 @@
...
-from skyrl_train.training_batch import TrainingInputBatch
-from skyrl_train.workers.worker import PPORayActorGroup
-from skyrl_train.workers.worker_dispatch import WorkerDispatch
-from skyrl_train.utils.utils import initialize_ray
-from skyrl_train.utils import get_ray_pg_ready_with_timeout
-from skyrl_train.config.utils import get_default_config
-from skyrl_train.env_vars import SKYRL_RAY_PG_TIMEOUT_IN_S
-from skyrl_train.inference_engines.ray_wrapped_inference_engine import create_ray_wrapped_inference_engines
-from skyrl_train.inference_engines.inference_engine_client import InferenceEngineClient
+try:  # Optional dependency: keep other backends importable without ray/skyrl-train.
```
pcmoritz: This looks like an accidental change that should be reverted

examples/train_integrations/harbor/entrypoints/main_harbor.py
CharlieFRuan, line 1: do we need to update the harbor `.sh` scripts

docs/mkdocs/content/config.md
CharlieFRuan: In a separate PR I assume?

examples/train_integrations/openenv/README.md
```diff
@@ -1,3 +1,5 @@
+> **Note:** SkyRL configuration has transitioned to Python-based dataclasses. The canonical config reference is now `skyrl_train/config/config.py`.
```
CharlieFRuan: --> `skyrl/train/config/config.py`

examples/train_integrations/openenv/README.md
```diff
@@ -77,7 +79,7 @@ LOGGER="wandb"
 
 All training parameters can be modified in `run_openenv.sh`, such as the model choice (`trainer.policy.model.path`), GRPO group size (`generator.n_samples_per_prompt`), or training batch size (`trainer.train_batch_size`).
 
-See all available training configuration parameters in `ppo_base_config.yaml`.
+See all available training configuration parameters in `skyrl_train/config/config.py` (Python dataclass definitions).
```
CharlieFRuan: --> skyrl/train/config/config.py

examples/train_integrations/verifiers/README.md
```diff
@@ -1,3 +1,5 @@
+> **Note:** SkyRL configuration has transitioned to Python-based dataclasses. The canonical config reference is now `skyrl_train/config/config.py`.
```
CharlieFRuan: --> `skyrl/train/config/config.py`

examples/train_integrations/verifiers/README.md
```diff
@@ -61,7 +63,7 @@ bash integrations/verifiers/run_verifiers.sh
 
 All training parameters can be modified in `run_verifiers.sh`, such as the model choice (`trainer.policy.model.path`), GRPO group size (`generator.n_samples_per_prompt`), or training batch size (`trainer.train_batch_size`).
 
-See all available training configuration parameters in `ppo_base_config.yaml`.
+See all available training configuration parameters in `skyrl_train/config/config.py` (Python dataclass definitions).
```
CharlieFRuan: --> `skyrl/train/config/config.py`

skyrl-train/examples/terminal_bench/run_codecontest.sh
CharlieFRuan, line 1: I see you updated this script. But I renamed `terminal_bench` -> `harbor`. Could you do it for `harbor` instead? Especially the `examples/train_integration`. Thank you!

skyrl-tx/tx/tinker/backends/skyrl_train.py
SumanthRH: Yes! This was WIP. Done!

skyrl/backends/skyrl_train_backend.py
```diff
@@ -16,54 +16,56 @@
...
         lora_config: LoRA configuration if using LoRA
     """
-    from omegaconf import OmegaConf
 
-    cfg = get_default_config()
+    # Apply user overrides from backend_config
+    user_overrides = dict(overrides.model_extra)
+    cfg = SkyRLTrainConfig.from_cli_overrides(user_overrides)
+
     cfg.trainer.policy.model.path = base_model
```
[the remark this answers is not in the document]
SumanthRH: This is actually a problem with the old code - the same overriding behaviour would carry through even with OmegaConf

skyrl/backends/skyrl_train/inference_engines/vllm/vllm_engine.py
```diff
@@ -510,6 +510,9 @@ async def _handle_openai_request(self, request_payload: Dict[str, Any], endpoint
         body = request_payload.get("json", {})
         headers = request_payload.get("headers", {})
 
+        print(f"Handling OpenAI request: {body}", flush=True)
```
CharlieFRuan: seems to be unintented?

skyrl/backends/skyrl_train/inference_engines/vllm/vllm_engine.py
```diff
@@ -562,6 +565,8 @@ async def _handle_openai_request(self, request_payload: Dict[str, Any], endpoint
                 "maximum context length" in error_message or "maximum model length" in error_message
             )
 
+            print(f"Error message in the vllm engine: {e}, {error_message}", flush=True)
```
CharlieFRuan: seems to be unintented?

docs/mkdocs/content/config.md
SumanthRH: This comment should be resolved. The original file referenced `skyrl_train`, while I wanted to update it to `skyrl.train`. 

First, I updated the other API reference files here: https://github.com/NovaSky-AI/SkyRL/pull/1192 

And then made the update to this PR.

skyrl/backends/skyrl_train/inference_engines/vllm/vllm_engine.py
SumanthRH: Yes. Sorry this was pushed just now. Fixing

skyrl/backends/skyrl_train/inference_engines/inference_engine_client.py
```diff
@@ -46,28 +45,33 @@ def __init__(
         self,
         engines: List[InferenceEngineInterface],
         tokenizer: PreTrainedTokenizerBase,
-        full_config: Union[SkyRLConfig, DictConfig],
+        model_path: str,
+        lora_cfg: SkyRLLoraConfig,
```
CharlieFRuan, line 49: I don't think the lora config is used anywhere

skyrl/backends/skyrl_train/workers/worker.py
```diff
@@ -796,10 +805,10 @@ def _forward_backward_micro(
                 sequences,
                 num_actions,
                 attention_mask=attention_mask,
-                temperature=self.cfg.generator.sampling_params.temperature,
+                temperature=self.cfg.algorithm.temperature,
```
CharlieFRuan, line 811: this is really subtle and really important I feel like... I totally see why you used algorithm, since for HTTP entrypoint, our sampling_params does not control this. Is there someway of highlight it

skyrl/train/config/config.py
```diff
@@ -556,15 +664,160 @@ def build_nested_dataclass(datacls: Type[T], d: dict) -> T:
...
+                    translated = translate_legacy_config(merged_dict)
+                    return build_nested_dataclass(cls, translated)
+            except Exception:
+                pass  # Legacy translation failed, re-raise original error
+
+            # Re-raise original error if not a legacy config issue
+            raise
+
+
+def make_config(
```
CharlieFRuan, line 775: this seems a bit specific, `algorithm`, `trainer`, and `generator`. Why cannot we override anything? I know algorithm is one level below, but that should still be doable if user handles it right?

SumanthRH: This is a utility function for easilly constructing a new SkyRLTrainConfig subclass when the user only wants to customize the algorithm, trainer or generator configuration. 

As such, the user can always define their own subclass of SkyRLTrainConfig and build a new class themselves. But, for common patterns, I believe this will reduce some boilerplate.

For example, this: 

```python

        @dataclass
        class MyAlgorithmConfig(AlgorithmConfig):
            my_param: int = 42

        @dataclass
        class MyTrainerConfig(TrainerConfig):
            algorithm: MyAlgorithmConfig = field(default_factory=MyAlgorithmConfig)
         
         
         @dataclass
          class MyConfig(SkyRLConfig): 
            trainer: MyTrainerConfig = field(default_factory=MyTrainerConfig)  
  
        
        cfg = MyConfig.from_cli_overrides(sys.argv[1:])
```

can be simplified to

```python

        @dataclass
        class MyAlgorithmConfig(AlgorithmConfig):
            my_param: int = 42

        MyConfig = make_config(algorithm_cls=MyAlgorithmConfig)
        cfg = MyConfig.from_cli_overrides(sys.argv[1:])
```    

The user is free to use either way for changing the config class

skyrl-train/examples/terminal_bench/run_codecontest.sh
SumanthRH: On it!

skyrl/train/config/config.py
CharlieFRuan: Ah makes sense. Thanks for the explanation!

skyrl/backends/skyrl_train/inference_engines/inference_engine_client.py
SumanthRH: That is correct. Currently it is a signature change - InferenceEngineClient needs to have the full configuration for the model being used, the lora configuration and the inference engine configuration, so all of these are parameters .

skyrl/backends/skyrl_train/workers/worker.py
SumanthRH: Currently, from the user's perspective, they would control temperature via generator_config.sampling_params.temperature ....internally, while building SkyRLConfig, we set `trainer.algorithm.temperature` to be this value. 

Since it's not exposed to the user directly right now (unless they use parts of the trainer directly), I think just having the docstring is enough?

```python
    temperature: float = 1.0
    """Temperature for scaling logits in policy loss computation.
    Typically set from ``generator.sampling_params.temperature`` during config validation."""
```

examples/train_integrations/harbor/entrypoints/main_harbor.py
SumanthRH: Yes this is needed! I just updated all the scripts for the new configuration dataclasses. 

I've tested that the entrypoint script can run upto config validation for Harbor

skyrl/backends/skyrl_train/workers/worker.py
CharlieFRuan: >  it's not exposed to the user directly right now

Ah I think we should set a TODO here, it should be exposed. Otherwise it'd be incorrect when users use the HTTP codepath, setting a non-1.0 temperature, without using the samplings param

SumanthRH: Oh sorry, a correction. ~It can be set directly by the user even now:~

```bash
.....
     trainer.algorithm.temperature=1.0 
```

It's rather not well documented I guess!

CharlieFRuan: @SumanthRH Could we add a note in the doc that, if the user is using HTTP and `/chat/completion`, they need to explicitly set this config?

examples/train_integrations/harbor/harbor_trial_config/default.yaml
```diff
@@ -51,7 +51,7 @@ agent:
     temperature: 1.0
 
     # Model info for token budgeting
-    # NOTE: max_input_tokens should match +generator.engine_init_kwargs.max_model_len
+    # NOTE: max_input_tokens should match generator.inference_engine.engine_init_kwargs.max_model_len
```
CharlieFRuan, line 54: There are still a couple of places in the repo where `+generator.` can be found (mainly comments / docs). Could you fix them as well please? Thank you!

SumanthRH: Yes will do!

skyrl/backends/skyrl_train/workers/worker.py
SumanthRH: Hmm currently there is an unconditional overwrite for this value because it's hard to distinguish when a user has provided a configuration for this:

```python
        # Copy temperature from generator sampling params to algorithm config
        # so workers can access it without needing the generator config
        self.trainer.algorithm.temperature = self.generator.sampling_params.temperature
```

So the docstring needs some update

skyrl/train/config/config.py
```diff
@@ -556,15 +672,160 @@ def build_nested_dataclass(datacls: Type[T], d: dict) -> T:
...
+            self.generator.max_input_length = self.trainer.max_prompt_length
+
+        # generator rope params default to trainer rope params
+        if self.generator.rope_scaling is None and self.trainer.rope_scaling is not None:
+            self.generator.rope_scaling = self.trainer.rope_scaling
+        if self.generator.rope_theta is None and self.trainer.rope_theta is not None:
+            self.generator.rope_theta = self.trainer.rope_theta
+        # Copy temperature from generator sampling params to algorithm config
+        # so workers can access it without needing the generator config
+        self.trainer.algorithm.temperature = self.generator.sampling_params.temperature
```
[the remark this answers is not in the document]
SumanthRH: This is an issue with the docstring. I believe the current method of setting the default is not ideal but works

skyrl/backends/skyrl_train/workers/worker.py
CharlieFRuan: hmm so does it mean that we should follow gemini's suggestions, make it `Optional` and resolve it in post_init. i.e. only use `sampling_params` when it is not provided

skyrl/train/config/config.py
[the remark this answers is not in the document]
SumanthRH: The alternative is to have a algorithm config temperature default of `None`, but then instantiating the AlgorithmConfig individually will be broken now.

skyrl/backends/skyrl_train/workers/worker.py
SumanthRH: Or just ask the user to set `generator.sampling_params.temperature` ? I've updated the docstring. There are tradeoffs either way - if you keep it as `None` then code that tries to init the TrainerConfig or the AlgorithmConfig directly will hit a surprise when the default `temperature` doesn't work. 

And we do expect these two temperature values to match in any case

skyrl/train/utils/utils.py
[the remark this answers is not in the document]
SumanthRH: This block of code is deprecated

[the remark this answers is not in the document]
SumanthRH: Sglang backend is deprecated. Let me update this codeblock in any case.

examples/train_integrations/harbor/harbor_generator.py
```diff
@@ -57,24 +57,24 @@ def __init__(
...
+        ie_cfg = generator_cfg.inference_engine
+        self.base_url = f"http://{ie_cfg.http_endpoint_host}:{ie_cfg.http_endpoint_port}"
         self.generator_cfg = generator_cfg
         self.tokenizer = tokenizer
         self.max_seq_len = max_seq_len
-        self.model_name = generator_cfg.model_name
 
         # Harbor config template - users can specify any Harbor TrialConfig options in YAML or command line.
         # SkyRL injects: model_name and api_base (once at init), task.path and session_id (per trial)
         self._harbor_trial_config_template = OmegaConf.to_container(harbor_cfg, resolve=True)
```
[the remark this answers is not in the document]
SumanthRH: Nice!

skyrl/backends/skyrl_train/workers/worker.py
CharlieFRuan: > And we do expect these two temperature values to match in any case

Hmm then why do we have two configs then, we should just keep a single `generator.sampling_params.temperature` right

CharlieFRuan: > code that tries to init the TrainerConfig or the AlgorithmConfig directly

could you elaborate on this a bit? Like overriding them from SkyRLTrainConfig post-init?

CharlieFRuan: Another motivation of prioritizing `algorithm.temperature` if specified is that, currently the `sampling_params` are only used in `SkyRLGymGenerator` and I kind of want to keep it that way. As noted in this issue (claude-coded up): https://github.com/NovaSky-AI/SkyRL/issues/1156, some generator configs are SkyRLGymGenerator-specific. I wish the sampling params is one of them. This PR already solves this issue partially by introducing the `InferenceEngineConfig`. One could imagine `SkyRLGymGeneratorConfig` introduced under `GenreatorConfig`

SumanthRH: Given that I've already used default `None` for `algorithm.max_seq_len` (+ set at post init), I believe it's best to do the same for `algorithm.temperature`

Updated the PR now

PR #1211, opened 2026-02-25
[train][tests] Add tests for `RemoteInferenceClient`

SumanthRH:
# What does this PR do?

Adds generation, error handling and chat template related tests for the new inference servers codepath . With this PR, we should have test coverage for all the features around basic full param training with the inference servers (with E2E tests on the same in progress). 

## Changes

1. Error handling: Previously, if errors were raised by vllm servers, the error message was not propagated to the client because we didn't read the body. This PR makes a change so that `RemoteInferenceClient` will read the error message details before raising a HTTP error. 
2. Served model name support: Propagated `cfg.generator.served_model_name` to vLLM args for the new inference server codepath
3. Add generation, error handling and chat template tests: Added all relevant tests from the  `test_inference_engine_client_http_endpoint.py` for the new inference server codepath. 

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1211" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -648,3 +650,21 @@ def __setstate__(self, state: dict) -> None:
...
+
+
+def raise_for_status(resp: aiohttp.ClientResponse, body: Optional[Any] = None) -> None:
+    """Modified version of resp.raise_for_status() that reads the body for the error message.
+
+    Raises aiohttp.ClientResponseError with the error message from the body if there is an error
+    """
+    if resp.status >= 300 and body is not None:
+        error_detail = body.get("error", {})
+        detail_msg = error_detail.get("message", resp.reason) if isinstance(error_detail, dict) else resp.reason
```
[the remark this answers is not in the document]
SumanthRH: The current implementation matches vLLM's (and OpenAI's) error format.

For confirmation you can spin up a vllm server:

```bash
uv run --isolated --extra fsdp -- vllm serve Qwen/Qwen2.5-0.5B-Instruct --port 8000 --max-model-len 50
```

And then submit a large prompt. The error message is

```bash
{"error":{"message":"This model's maximum context length is 40 tokens. However, your request has 609 input tokens. Please reduce the length of the input messages.","type":"BadRequestError","param":null,"code":400}}
```

tests/backends/skyrl_train/gpu/utils.py
```diff
@@ -464,7 +467,7 @@ def create(
         # Return both router and server group if created to keep references alive
         router = None
         server_group = None
-        if _SKYRL_USE_NEW_INFERENCE:
+        if use_new_inference_servers or _SKYRL_USE_NEW_INFERENCE:
```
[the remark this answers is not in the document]
SumanthRH: Ah. it doesn't override when False, only when True.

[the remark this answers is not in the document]
SumanthRH: Fixed!

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_chat_template.py
```diff
@@ -0,0 +1,106 @@
...
+    cfg.generator.num_inference_engines = num_inference_engines
+    cfg.generator.inference_engine_tensor_parallel_size = TP_SIZE
+    cfg.generator.run_engines_locally = True
+    cfg.generator.sampling_params.max_generate_length = 256
+    return cfg
+
+
+@pytest.mark.vllm
+@pytest.mark.parametrize("use_custom_template", [False, True])
+def test_custom_chat_template(ray_init_fixture, use_custom_template: bool):
```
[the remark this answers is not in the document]
SumanthRH: This is nice in theory but doesn't work because vLLMServerActor doesn't do cleanup properly right now. I've opened an issue: https://github.com/NovaSky-AI/SkyRL/issues/1216

Until then, it is better to have test-level ray init/shutdown @gemini-code-assist

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -648,3 +650,21 @@ def __setstate__(self, state: dict) -> None:
         """Restore state after unpickling."""
         self.__dict__.update(state)
         self._session = None
+
+
+def raise_for_status(resp: aiohttp.ClientResponse, body: Optional[Any] = None) -> None:
+    """Modified version of resp.raise_for_status() that reads the body for the error message.
+
+    Raises aiohttp.ClientResponseError with the error message from the body if there is an error
```
kouroshHakha, line 658: Add this extra explaination: 

"""
The standard raise_for_status() only uses the HTTP reason phrase (e.g. "Bad Request"), which is often unhelpful. APIs typically put more descriptive error details in the response body. This function bridges that gap by surfacing the body's error message in the exception.
"""

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -648,3 +650,21 @@ def __setstate__(self, state: dict) -> None:
...
+    """
+    if resp.status >= 300 and body is not None:
```
kouroshHakha: shouldn't this be >=400? 3xx are usually re-direct, which should not happen here???

skyrl/backends/skyrl_train/inference_servers/utils.py
```diff
@@ -33,6 +33,8 @@ def build_vllm_cli_args(cfg: "Union[SkyRLConfig, DictConfig]") -> Namespace:
         enable_sleep_mode=cfg.trainer.placement.colocate_all,
         # NOTE (sumanthrh): We set generation config to be vLLM so that the generation behaviour of the server is same as using the vLLM Engine APIs directly
         generation_config="vllm",
+        # NOTE: vllm expects a list entry for served_model_name
+        served_model_name=[cfg.generator.served_model_name] if cfg.generator.served_model_name else None,
```
kouroshHakha: At some point we should just createa  pass through for engine_kwargs that will be augemented with other parts of the input yaml (model_name, etc). so that we don't have to do much here anymore.

tests/backends/skyrl_train/gpu/conftest.py
```diff
@@ -11,3 +11,13 @@ def ray_init_fixture():
     yield
     # call ray shutdown after a test regardless
     ray.shutdown()
+
+
+@pytest.fixture(scope="module")
```
kouroshHakha, line 16: why do you need this? and how is it used?

ci/gpu_ci_run_skyrl_train.sh
```diff
@@ -34,7 +34,7 @@ uv run --isolated --extra fsdp --extra dev \
     -- pytest -s -vvv tests/backends/skyrl_train/gpu/gpu_ci/test_engine_generation.py::test_token_based_generation -m "vllm"
 
 # Run tests for new inference layer
-# TODO (sumanthrh): Migrate the remaining tests: test_verifiers_generator.py , test_pause_and_continue_generation.py and the relevant tests in test_inference_engine_client_http_endpoint.py
+# TODO (sumanthrh): Migrate the remaining tests: test_verifiers_generator.py , test_pause_and_continue_generation.py
```
kouroshHakha, line 33: hmm I am not following the real delta on the tests? Is there a new test that is getting run on the new inference stack? 

The change in `test_pause_and_continue_generation.py` is just a docstring update the change in `conftest.py` is a new fixture, which I am not sure how it is used? I don't find its presence in any other place and `utils.py` changes are also just an extra api definition.

tests/backends/skyrl_train/gpu/utils.py
```diff
@@ -486,7 +489,9 @@ def create(
                 f"proxy_url={proxy_url}, server_urls={server_urls}, colocated={cfg.trainer.placement.colocate_all}"
             )
             client = RemoteInferenceClient(
-                proxy_url=proxy_url, server_urls=server_urls, model_name=cfg.trainer.policy.model.path
+                proxy_url=proxy_url,
+                server_urls=server_urls,
+                model_name=served_model_name if served_model_name else "default",
```
kouroshHakha: else it should be the model itself right? not default?

tests/backends/skyrl_train/gpu/utils.py
```diff
@@ -464,7 +467,7 @@ def create(
         # Return both router and server group if created to keep references alive
         router = None
         server_group = None
-        if _SKYRL_USE_NEW_INFERENCE:
+        if use_new_inference_servers or (use_new_inference_servers is None and _SKYRL_USE_NEW_INFERENCE):
```
kouroshHakha, line 471: why not keep the _SKYRL_USE_NEW_INFERENCE as the gate and control it in the tests? 

or if you want to change this part, I would just do 

```
if use_new_inference_servers is True: ...
```

Then in the caller of those other places I would do 
```
InferenceEngineState.create(..., use_new_inference_servers=os.environ.get("_SKYRL_USE_NEW_INFERENCE", False))
```

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_chat_template.py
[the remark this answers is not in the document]
kouroshHakha: see my comment on https://github.com/NovaSky-AI/SkyRL/issues/1216#issuecomment-3961551376

ci/gpu_ci_run_skyrl_train.sh
kouroshHakha: oh I think I reviewed an older commit. There are new files I did not review. just give me a few more mins.

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_chat_template.py
```diff
@@ -0,0 +1,106 @@
...
+        cfg = get_test_actor_config(num_inference_engines=1, model=MODEL_QWEN3)
+        cfg.trainer.placement.colocate_all = True
+        engines = InferenceEngineState.create(
+            cfg=cfg,
+            use_local=True,
+            backend="vllm",
+            model=MODEL_QWEN3,
+            sleep_level=1,
+            engine_init_kwargs={"chat_template": TEMPLATE_PATH} if use_custom_template else None,
+            use_new_inference_servers=True,
```
kouroshHakha, line 54: do we need this? Can we just run all the inference_servers/ tests with the env var?

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_chat_template.py
```diff
@@ -0,0 +1,106 @@
...
+
+import skyrl
+from skyrl.train.config import SkyRLConfig
+from tests.backends.skyrl_train.gpu.utils import InferenceEngineState
+from transformers import AutoTokenizer
+
+MODEL_QWEN3 = "Qwen/Qwen3-0.6B"
+TP_SIZE = 1
+
+TEMPLATE_PATH = str(Path(skyrl.train.utils.__file__).parent / "templates/qwen3_acc_thinking.jinja2")
```
kouroshHakha, line 22: I am not sure how this would be shipped to work_dir in a remote node case? This test runs on a head GPU node so everything remains local. We should test a cpu head node + gpu worker nodes as well.

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_generation.py
```diff
@@ -0,0 +1,304 @@
...
+    cfg.generator.served_model_name = SERVED_MODEL_NAME
+    cfg.generator.sampling_params.max_generate_length = 256
+    return cfg
+
+
+@pytest.fixture(scope="module")
+def vllm_server(module_scoped_ray_init_fixture):
+    """Single vLLM server + router + RemoteInferenceClient."""
+    cfg = get_test_actor_config(num_inference_engines=1, model=MODEL_QWEN2_5)
+    cfg.trainer.placement.colocate_all = True
```
kouroshHakha: why don't you make collocate_all true in the get_test_actor_config function?

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_chat_template.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+@pytest.mark.vllm
+@pytest.mark.parametrize("use_custom_template", [False, True])
+def test_custom_chat_template(ray_init_fixture, use_custom_template: bool):
+    """Test custom chat template via RemoteInferenceClient."""
+    engines = None
+    try:
+        cfg = get_test_actor_config(num_inference_engines=1, model=MODEL_QWEN3)
+        cfg.trainer.placement.colocate_all = True
```
kouroshHakha: why not do this in get_test_actor_config?

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_generation.py
```diff
@@ -0,0 +1,304 @@
...
+    engines = InferenceEngineState.create(
```
kouroshHakha: yo uneed to do a try finally pattern here I think to be robust

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_generation.py
```diff
@@ -0,0 +1,304 @@
...
+    yield engines
+    engines.close()
+
+
+def _check_chat_completions_outputs(outputs: List[Dict], test_type: str, num_samples: int, backend: str):
+    for output in outputs:
+        assert not ("error" in output or output.get("object", "") == "error"), f"Error in output: {output}"
+    assert len(outputs) == num_samples
+    for response_data in outputs:
+        if test_type == "litellm":
```
kouroshHakha: when do we need litellm test_type?

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_generation.py
```diff
@@ -0,0 +1,304 @@
+"""
+Generation and error handling tests for `RemoteInferenceClient` (new inference path)
```
kouroshHakha: All of this test can be deleted when we remove generation APIs of remote_inference_client right?

kouroshHakha: if so, can we add a note on what should be done here, with a special tag that we can find later across the project to find all the todos associated with that action item?

kouroshHakha: the reason I am saying we can remove this, is that if we standardaize on passing through vllm server API, then those are tested on vllm side so we don't have to worry about it in here.

tests/backends/skyrl_train/gpu/utils.py
SumanthRH: These are tests specifically written for the new inference codepath. I don't think it makes sense to plan for both possible values of  `_SKYRL_USE_NEW_INFERENCE` . 

That is the main reason why the tests use the new inference codepath by default and are placed in `inference_servers/`

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_chat_template.py
SumanthRH: >  Can we just run all the inference_servers/ tests with the env var?

`inference_servers/` tests are meant to test the new codepath in any case. They are written specifically for the new codepath, so I don't think it makes sense to have any gating with the env var.

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_generation.py
SumanthRH: I don't think this is needed. This code lives in a module-scoped fixture - cleanup happens on finishing all tests in this module regardless of tests passing/failng. Further, we use the `module_scoped_ray_init_fixture` for ray init / shutdown

tests/backends/skyrl_train/gpu/conftest.py
SumanthRH: Please see 

https://github.com/NovaSky-AI/SkyRL/blob/920157be259b4c0852cdd6aecca4ce01d1443873/tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_generation.py#L52-L55

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_chat_template.py
SumanthRH: Even if it's tests running on the remote node, the `skyrl` package path will be the local package path on that node. Global variable init happens per process , so for the ray task / actor running on that node will use the local jinja file for the `skyrl` package on that node. As long as the jinja files are shipped as a part of the package - which it is : 

https://github.com/NovaSky-AI/SkyRL/blob/920157be259b4c0852cdd6aecca4ce01d1443873/pyproject.toml#L160-L161

It should be fine?

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_generation.py
SumanthRH: Ah, this is because I am reusing the setup code from `test_inference_engine_http_endpoint.py` . Let me improve this in both places.

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_generation.py
SumanthRH: These are basic tests for ensuring that completion, chat completion methods as well as error handling work correctly for `RemoteInferenceClient`. I believe we will retain them even after we upgrade vLLM and migrate to the new keep mode.

skyrl/backends/skyrl_train/inference_servers/utils.py
```diff
@@ -32,6 +32,8 @@ def build_vllm_cli_args(cfg: SkyRLTrainConfig) -> Namespace:
...
```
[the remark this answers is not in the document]
SumanthRH: Resolving issues after merge from main. Devin is jumpy

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_generation.py
SumanthRH: Synced offline. We will be moving away from using the remote inference client and transition to interacting with the open ai compatible API server from the router. 

I initially added a simple TODO indicating the same.

However, now I believe all the tests here can be refactored and improved. We can as such separate out some tests to interact directly with the APi server via OpenAI / litellm client and then have additional tests for Remote inference client (which is needed for now)

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -270,8 +270,8 @@ async def _generate_single(
                 headers["X-Session-ID"] = str(session_id)
 
             async with session.post(url, json=payload, headers=headers) as resp:
-                resp.raise_for_status()
                 response = await resp.json()
+                raise_for_status(resp, response)
```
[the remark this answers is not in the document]
SumanthRH: We interact with vllm servers directly, and these return JSON. I believe this can be revisited later if needed (i.e we add a new router that does more post processing). Further, the generation APIs in remote inference client are temporary -so it is unlikely there is a new path before these methods are deleted

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
SumanthRH: 300 errors won't follow the openai format for error messages, you're right. Fixed!

tests/backends/skyrl_train/gpu/utils.py
SumanthRH: Fixed!

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_chat_template.py
SumanthRH: Fixed!

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_generation.py
SumanthRH: Please see the new tests. The helper here is from the older tests for inference engine client's http endpoint.

I've ported over the litellm tests as well now, which makes this codepath useful in the new test file

SumanthRH:
I've now refactored the test file `test_new_inference_generation.py` into two groups of tests:


Group A: Generation and error handling tests that interact directly with the router's OpenAI-compatible endpoints via `requests` or LiteLLM.

We use Litellm for passing some sampling parameters that are not supported by OpenAI's chat completion client. 

Group B: Generation and error handling tests that use the `RemoteInferenceClient`.

Both of these are important right now. 

Once we make progress in the migration  #845 , we can remove the generation APIs in the remote inference client and have Generators interact directly with the router's HTTP endpoints. I've added a TODO for this case.

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_generation.py
```diff
@@ -0,0 +1,304 @@
...
+
+@pytest.mark.vllm
+def test_context_length_error_returns_400(vllm_server):
+    """Test that context length errors return HTTP 400."""
+    client = vllm_server.client
+
+    # Oversized prompt (max_model_len=1024 in fixture)
+    messages_oversized = [{"role": "user", "content": "hello " * 1500}]
+
+    with pytest.raises(Exception) as exc_info:
```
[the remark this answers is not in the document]
SumanthRH: Fixed

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_generation.py
```diff
@@ -0,0 +1,304 @@
...
+# --- Group B: Error handling ---
+
+
+@pytest.mark.vllm
+def test_error_handling(vllm_server):
+    """Test error handling via RemoteInferenceClient."""
+    client = vllm_server.client
+
+    # Missing required field (messages)
+    with pytest.raises(Exception) as exc_info:
```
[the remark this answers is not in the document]
SumanthRH: Fixed!

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_generation.py
```diff
@@ -0,0 +1,304 @@
...
+                    "max_tokens": 50,
+                }
+            }
+        )
+    )
+    assert "choices" in result and len(result["choices"]) > 0
+    assert result["choices"][0]["message"]["content"] is not None
+
+    # Request with model path should fail (model name mismatch)
+    with pytest.raises(Exception):
```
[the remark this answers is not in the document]
SumanthRH: Fixed

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_generation.py
```diff
@@ -0,0 +1,304 @@
...
+    """Test error handling via RemoteInferenceClient."""
+    client = vllm_server.client
+
+    # Missing required field (messages)
+    with pytest.raises(Exception) as exc_info:
+        asyncio.run(client.chat_completion({"json": {"model": SERVED_MODEL_NAME}}))
+    assert exc_info.value is not None
+
+    # Wrong model name
+    with pytest.raises(Exception) as exc_info:
```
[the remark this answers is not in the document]
SumanthRH: Fixed

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
SumanthRH: Actually, 300 errors won't even be seen by aiohttp's client, so this code was functionally correct even before. I'll still retain the edit to 400 and above

skyrl/backends/skyrl_train/inference_servers/utils.py
```diff
@@ -32,6 +32,12 @@ def build_vllm_cli_args(cfg: SkyRLTrainConfig) -> Namespace:
         enable_sleep_mode=cfg.trainer.placement.colocate_all,
         # NOTE (sumanthrh): We set generation config to be vLLM so that the generation behaviour of the server is same as using the vLLM Engine APIs directly
         generation_config="vllm",
+        # NOTE: vllm expects a list entry for served_model_name
+        served_model_name=(
```
CharlieFRuan, line 36: ideally this should be a pass-through in the future right, rather than us having to explicitly support this. Just like say `enable_ray_prometheus_stats`.

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_new_inference_generation.py
```diff
@@ -0,0 +1,501 @@
...
+Group A: Generation and error handling tests that interact directly with the router's OpenAI-compatible
+endpoints via `requests` or LiteLLM.
+
+Group B: Generation and error handling tests that use the `RemoteInferenceClient`.
+
+# Run with:
+uv run --isolated --extra dev --extra fsdp pytest tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_new_inference_generation.py -m vllm -v
+"""
+
+# TODO (sumanthrh) (RemoteInferenceClient data-plane-deprecation): Remove the tests in Group B once we migrate all generation interactions to the router's HTTP API.
```
CharlieFRuan, line 15: Does this mean `SkyRLGymGenerator` would post HTTP requests as well?

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_new_inference_generation.py
```diff
@@ -0,0 +1,501 @@
...
+
+    response = requests.post(f"{base_url}/chat/completions", json=payload)
+    assert response.status_code == HTTPStatus.OK
+    result = response.json()
+    text = result["choices"][0]["message"]["content"]
+    assert json.loads(text) is not None, f"Output is not valid JSON: {text}"
+
+
+@pytest.mark.vllm
+def test_error_handling(vllm_server: InferenceEngineState):
```
CharlieFRuan, line 303: Comparing agaainst the `test_http_endpoint_error_handling` in `tests/backends/skyrl_train/gpu/gpu_ci/test_inference_engine_client_http_endpoint.py`:
- We guard `stream==False` in the old path. Do we / should we support streaming in the new codepath?

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_new_inference_generation.py
SumanthRH: Yes that is correct! This would make things more natural for users to also bring in any generators

skyrl/backends/skyrl_train/inference_servers/utils.py
SumanthRH: Yes that's right! We would simply parse user's input into a list, because that is what vLLM excepts (matches vLLM CLI's argument parsing)

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_new_inference_generation.py
SumanthRH: There's no reason why streaming shouldn't work for the router's HTTP endpoint. Let me add a test for this to be sure!

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_generation.py
kouroshHakha: it feels irrelevant to skyrl to test this in general. As long as you test openAI compatibility, isn't litellm compatible with that?

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_generation.py
kouroshHakha: I realized what aI said might not be applicable here.

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_chat_template.py
kouroshHakha: it's more like a nit. but either way you'd want to change these tests after this path becomes the default, because this component in particular is shared between the old and new path. So the question is which one is simpler. I think all the inference_servers/ tests should have the env var set automatically so the underlying modules change behavior based on the env var and you don't do anything special at the interfaces.

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_generation.py
SumanthRH: litellm supports additional sampling parameters that vllm supports, and some of our first party integrations like Harbor rely on this. So I believe it is good to have the tests here.

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_new_inference_generation.py
SumanthRH: Added a basic test for streaming!

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_generation.py
kouroshHakha: ok can you add the context / comment to why keep litellm tests here?

SumanthRH: Already done!

https://github.com/NovaSky-AI/SkyRL/blob/438f72f255c469c656cf132498c84587d63115d3/tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_new_inference_generation.py#L424-L426

https://github.com/NovaSky-AI/SkyRL/blob/438f72f255c469c656cf132498c84587d63115d3/tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_new_inference_generation.py#L364-L365

PR #1220, opened 2026-02-26
[train] Remove deprecated SGLang backend from `skyrl/`

SumanthRH:
# What does this PR do?

Removes the deprecated SGLang backend in the new `skyrl/` package. 

## Summary
- Removes the SGLang backend in `inference_engines/` and all references 
- SkyRL still retains the `generator.inference_engine.backend` parameter so that we can revisit this in the future and allow users forking SkyRL to easily experiment with other backends. 
- Removes `pytest.mark.vllm` used in tests to demarcate vllm specific tests. SkyRL now runs only one inference engine so this is not needed. 

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1220" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/tests/backends/skyrl_train/gpu/gpu_ci/test_pause_and_continue_generation.py
+++ b/tests/backends/skyrl_train/gpu/gpu_ci/test_pause_and_continue_generation.py
@@ -1,7 +1,7 @@
 """
 Test pause and continue generation with inference engine client HTTP endpoint.
 
-uv run --isolated --extra dev --extra vllm pytest tests/gpu/gpu_ci/test_pause_and_continue_generation.py -m "vllm"
+uv run --isolated --extra dev --extra vllm pytest tests/gpu/gpu_ci/test_pause_and_continue_generation.py
 """
 
 import asyncio


skyrl/backends/skyrl_train/inference_engines/ray_wrapped_inference_engine.py
```diff
@@ -259,80 +256,13 @@ def create_ray_wrapped_inference_engines(
                     **other_kwargs,
                 )
                 inference_engine_actors.append(engine)
-        elif backend == "sglang":
```
CharlieFRuan, line 262: we should add an `else` and error out here. I know we didn't have it before, but that was probably a mistake

SumanthRH: @CharlieFRuan If the backend is not vllm, this function will error out before hitting this line:

https://github.com/NovaSky-AI/SkyRL/blob/54f4d9c64801394a5a3ddf8337809014ff401128/skyrl/backends/skyrl_train/inference_engines/ray_wrapped_inference_engine.py#L123-L136

PR #1232, opened 2026-02-27
[train] Fix issue with unset `pad_token_id`

### LINKED ISSUE #1231: SkyRLTrainBackend pads with None → crash in _to_training_batch (meta-llama/Llama-3.2-1B, --backend fsdp)
### Summary
When running the `skyrl.tinker.api` entrypoint with `meta-llama/Llama-3.2-1B`, training proceeds for some time, then crashes once a batch requires padding. Stack trace ends at:
```bash
/work/SkyRL/skyrl/backends/skyrl_train_backend.py:285 in _to_training_batch    
sequences_tensor = torch.tensor(sequences, dtype=torch.long)
TypeError: 'NoneType' object cannot be interpreted as an integer
```

This appears to be caused by SkyRLTrainBackend using a HuggingFace tokenizer whose `pad_token_id` is `None` (common for Llama tokenizers), and then using that value for sequence padding.

### Repro command
I am launching the API like:
```bash
uv run --extra tinker --extra fsdp -m skyrl.tinker.api \  --base-model meta-llama/Llama-3.2-1B \  --backend fsdp \  --port "$PORT" \  --checkpoints-base "$CHECKPOINTS_DIR"
```

### In more detail
In SkyRLTrainBackend._to_training_batch, sequences are left-padded via:

```python
skyrl_train_backend.py (Lines 277-286)

pad_len = max_seq_len - len(seq)
sequences.append([self._tokenizer.pad_token_id] * pad_len + list(seq))
...
sequences_tensor = torch.tensor(sequences, dtype=torch.long)
```
If `self._tokenizer.pad_token_id` is None, any non-zero `pad_len` inserts None values into sequences, which then causes `torch.tensor(..., dtype=torch.long)` to throw `TypeError: 'NoneType' object cannot be interpreted as an integer.`

SumanthRH:
# What does this PR do?

Fixes #1231 

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1232" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/tests/backends/skyrl_train/gpu/utils.py
+++ b/tests/backends/skyrl_train/gpu/utils.py
@@ -35,6 +35,7 @@ from skyrl.backends.skyrl_train.env_vars import SKYRL_PYTHONPATH_EXPORT, _SKYRL_
 from skyrl.backends.skyrl_train.inference_servers.remote_inference_client import RemoteInferenceClient
 from skyrl.backends.skyrl_train.inference_servers.server_group import ServerGroup
 from skyrl.backends.skyrl_train.inference_servers.router import InferenceRouter
+from skyrl.utils.tok import get_tokenizer
 
 TEST_DATA_PATH = os.path.expanduser("~/data/gsm8k/validation.parquet")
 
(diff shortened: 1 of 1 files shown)


SumanthRH:
With this PR, it is still not possible to successfully train with `meta-llama/Llama-3.2-1B` with `--fsdp`.

The issue is that that the repo for `meta-llama/Llama-3.2-1B` on Huggingface is gated, and we need to propagate `HF_TOKEN` env var to the FSDP workers. However, the current recommended path for doing so with `--env-file .env` does not work. I've opened another issue: #1234 . 

If I manually set the env var in the workers, I am able to run SFT with `meta-llama/Llama-3.2-1B`

PR #1244, opened 2026-03-02
[tpu] set local_device_ids in jax.distributed.initialize to avoid auto init

andrewsykim:
[jax.distributed.initialize()](https://github.com/jax-ml/jax/blob/main/jax/_src/distributed.py#L86-L95) calls auto_detect_unset_distributed_params if either `coordinator_address`, `num_processes`, `process_id` or  `local_device_ids` is not set. 

For TPUs I suepect the call to auto_detect_unset_distributed_params reassigns JAX process IDs which breaks the current examples that assume process ID is set from `--process-id`. This PR also sets `local_device_ids` to skip  reinitialization from JAX.

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1244" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/skyrl/backends/jax.py
+++ b/skyrl/backends/jax.py
@@ -1073,14 +1073,13 @@ class JaxBackend(JaxBackendImpl):
             tpu_worker_id = int(os.environ.get("TPU_WORKER_ID"))
 
             jax.distributed.initialize(
-                cluster_detection_method="deactivate",
                 coordinator_address=config.coordinator_address,
                 num_processes=config.num_processes,
-                process_id=0,
-                local_device_ids=range(config.tensor_parallel_size),
+                # process_id=0,
+                # local_device_ids=range(config.tensor_parallel_size),
             )
             logger.info(
-                f"JAX distributed initialized: input_process_id=0, "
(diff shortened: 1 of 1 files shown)


skyrl/backends/jax.py
```diff
@@ -1191,9 +1193,15 @@ def main():
         required=True,
         help="This process's ID (must be > 0 for workers)",
     )
+    parser.add_argument(
+        "--tensor-parallel-size",
```
andrewsykim, line 1204: Is it okay to require this? If not can we just default it to 1 for the GPU case?

andrewsykim: (updated to not required but default to 1)

andrewsykim:
(converting to draft until I run more tests)

skyrl/backends/jax.py
pcmoritz: How about we call this `--num-local-devices`? I don't think we want to necessarily couple it to the `--tensor-parallel-size` (even though in practice these will often be the same). And we can document that this will be translated to Jax's `local_device_ids`.

andrewsykim:
(closing for now while I run more tests)

PR #1251, opened 2026-03-02
[train] Fix NUMA and checkpoint serialization for optimizer CPU offload

tyler-griggs:
## Summary

Three runtime patches discovered during GPU validation of GLM-4.7-Flash GRPO training on 8x A100-80GB with optimizer CPU offload.

**Depends on:** #1241 (transformers 5.x compatibility) -- GLM-4.7-Flash requires transformers>=5.0.0, and #1241 adds the necessary return_dict=False fixes to all apply_chat_template call sites.

### Changes

- **NUMA memory interleave**: use numa_set_interleave_mask instead of numa_set_membind. With optimizer CPU offload enabled, binding all workers to a single NUMA node causes system OOM during checkpoint serialization. Interleaving spreads memory pressure across all nodes. Falls back to the original membind if unavailable.

- **Skip optimizer state in checkpoint save/load**: when optimizer_cpu_offload is enabled on the Megatron optimizer config, set _skip_optimizer_checkpoint = True. This causes PolicyWorkerBase and CriticWorkerBase to pass None for optimizer/scheduler during checkpoint save, avoiding serialization OOM and mcore 0.16.0 flattened_range errors. On load, optimizer/scheduler states are skipped and re-initialized fresh. Reduces checkpoint size from about 120 GiB to 56 GiB (model-only).

### Files changed (2)

- skyrl/backends/skyrl_train/workers/worker.py -- NUMA fix + save/load checkpoint patches
- skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py -- set skip flag when CPU offload enabled

### Validated on

- 8x A100-SXM4-80GB, CUDA 12.9, mcore 0.16.0, vLLM 0.16.0
- GLM-4.7-Flash (30B MoE, TP=1/EP=8) with optimizer CPU offload (optimizer_offload_fraction=1.0)
- Checkpoint save (56 GiB, 292s) + resume from checkpoint verified end-to-end

### DIFF
--- a/skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
+++ b/skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
@@ -555,6 +555,11 @@ class MegatronPolicyWorkerBase(MegatronWorker, PolicyWorkerBase):
         )
         self.optimizer = get_megatron_optimizer(self.actor_module, optim_config)
 
+        # Skip optimizer state in checkpoints when using CPU offload to avoid
+        # system OOM during serialization.
+        if optim_config.optimizer_cpu_offload:
+            self._skip_optimizer_checkpoint = True
+
         # create scheduler
         self.scheduler = get_megatron_optimizer_param_scheduler(
             optimizer=self.optimizer,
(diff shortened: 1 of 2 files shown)


erictang000:
btw https://github.com/NovaSky-AI/SkyRL/pull/1268 should fix the flattened_range related optimizer checkpoint issues

PR #1255, opened 2026-03-03
[tinker] Propagate all `uv run` args from server startup command to backend workers

### LINKED ISSUE #1234: [train] `uv` environment variables not propagated to the backend with tinker API server
# Summary

Setting environment variables at server launch via `uv run --env-file .env` does not propagate the env vars to the backed worker processes with the Skyrl-Train backend

## To reproduce

Start the tinker API server with the skyrl-train backend:

```bash
uv run --env-file .env  --extra tinker --extra fsdp -m skyrl.tinker.api  --base-model Qwen/Qwen2-0.5B-Instruct   --backend fsdp  --port 8001
```

Set the following env vars in the `.env` file:

```env
TEST_ENV_VAR=test # variable we want to propagate
SKYRL_DUMP_INFRA_LOG_TO_STDOUT=1 # log to stdout
```

Add a print statement to inspect the environment variable in  `FSDPPolicyWorkerBase.init_model`:

https://github.com/NovaSky-AI/SkyRL/blob/83d2bd71d1d24267556d6ecb9eda0d5df8e5a6e0/skyrl/backends/skyrl_train/workers/fsdp/fsdp_worker.py#L104-L106

 For example:

```python
import os
from loguru import logger
logger.info(f"test env var {os.environ.get("TEST_ENV_VAR", None)}")
```

Run SFT in another terminal:

```bash
TINKER_API_KEY=tml-dummy uv run --with tinker --with datasets     python -m tinker_cookbook.recipes.sl_loop     base_url=http://localhost:8001     model_name="meta-llama/Llama-3.2-1B"
```

### Current behaviour

I see the following the logs:

```bash
2026-02-27 23:06:19.287 | INFO     | skyrl.backends.skyrl_train.workers.fsdp.fsdp_worker:init_model:108 - test env var None
```

### Expected behaviour

I expect the environment variable to be propagated so the printed value should be `test`

SumanthRH:
# What does this PR do?

Fixes https://github.com/NovaSky-AI/SkyRL/issues/1234

After the fixes in this PR, I can successfully train with `meta-llama/Llama-3.2-1B` by passing an env-file to the server startup command :

```bash
 uv run --extra tinker --extra fsdp --env-file skyrl-train/.env.example -m skyrl.tinker.api  --base-mod
el meta-llama/Llama-3.2-1B   --backend fsdp 
```

(For context: See #1231 )

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1255" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/skyrl/tinker/api.py
+++ b/skyrl/tinker/api.py
@@ -45,42 +45,45 @@ API_SERVER_STARTUP_ARGS = ["-m", "skyrl.tinker.api"]
 SHUTDOWN_TIMEOUT_SECONDS = 10
 
 
-def _get_parent_uv_run_args() -> list[str]:
-    """Extract parent `uv run <uv run args>` flags for the engine launch.
+def _get_parent_uv_run_args(parent_cmd: list[str]) -> list[str]:
+    """Extract parent `uv run <uv run args>` flags for the engine launch given the parent process's startup command
 
     `uv run` starts this Python API process as a child. To recover the original
     `uv run <uv run args> ...` flags, we inspect the parent process command line
     and extract all the uv run args before the script argument.
     """
-    tokens = psutil.Process(os.getppid()).cmdline()
     # the API server startup command can be
     # uv run <uv run args> -m skyrl.tinker.api
     # or uv run <uv run args> python -m skyrl.tinker.api
     # or uv run <uv run args> -- python -m skyrl.tinker.api
     stop_strings = ["--", "python"]
     detected = False
-    for i in range(len(tokens) - 1):
-        if tokens[i] in stop_strings or (
-            tokens[i] == API_SERVER_STARTUP_ARGS[0] and tokens[i + 1] == API_SERVER_STARTUP_ARGS[1]
-        ):
+    for i in range(len(parent_cmd) - 1):
+        if parent_cmd[i] in stop_strings or parent_cmd[i : i + len(API_SERVER_STARTUP_ARGS)] == API_SERVER_STARTUP_ARGS:
             detected = True
             break
     if not detected or i < 2:
         raise ValueError(
-            f"Unable to parse tinker API server startup command: {tokens}. "
+            f"Unable to parse tinker API server startup command: {parent_cmd}. "
             "Ensure that the tinker API server was started with `uv run <uv run args> -m skyrl.tinker.api`"
         )
-    tokens = tokens[2:i]  # ignore uv run
-    return tokens
+    parent_cmd = parent_cmd[2:i]  # ignore uv run
+    return parent_cmd
 
 
-def _build_uv_run_cmd_engine(engine_config: BaseModel) -> list[str]:
+def _build_uv_run_cmd_engine(parent_cmd: list[str], engine_config: BaseModel) -> list[str]:
+    """Builds uv run command for the engine
+
+    Args:
+        parent_cmd: The command for the parent process starting the engine
+        engine_config: Engine configuration
+    Returns:
+        cmd: The uv run command for the tinker engine
+    """
     cmd = ["uv", "run"]
-    parent_flags = _get_parent_uv_run_args()
+    parent_flags = _get_parent_uv_run_args(parent_cmd)
     logger.debug(f"Detected API server uv run flags: {parent_flags}")
     cmd.extend(parent_flags)
-    if "--isolated" not in cmd:
-        cmd.extend(["--isolated"])
     # NOTE: uv deduplicates extras so we can unconditionally add the tinker extra
     cmd.extend(["--extra", "tinker", "--extra", engine_config.backend])
     cmd.extend(["-m", "skyrl.tinker.engine"])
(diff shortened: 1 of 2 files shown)


skyrl/tinker/api.py
```diff
@@ -40,23 +39,53 @@
...
+            "Ensure that the tinker API server was started with `uv run <uv run args> -m skyrl.tinker.api`"
+        )
+    tokens = tokens[2:i]  # ignore uv run
+    return tokens
+
+
+def _build_uv_run_cmd_engine(engine_config: BaseModel) -> list[str]:
+    cmd = ["uv", "run"]
+    parent_flags = _get_parent_uv_run_args()
+    logger.debug(f"Detected API server uv run flags: {parent_flags}")
```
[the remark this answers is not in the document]
SumanthRH: This doesn't actually log the contents of the env file, just the path.

skyrl/tinker/api.py
```diff
@@ -40,23 +39,53 @@
...
+    cmd.extend(parent_flags)
+    if "--isolated" not in cmd:
```
pcmoritz: I think given we just pass the arguments through now, we should not modify them like this. If the user wants `--isolated`, they should specify it on their command line (and we should add it to the examples where neede)

tests/tinker/test_api.py
```diff
@@ -430,3 +433,168 @@ def cleanup_logs_found():
         return "Auto-unloaded stale model" in log_output and "Deleted model" in log_output
 
     assert wait_for_condition(cleanup_logs_found, timeout_sec=10, poll_interval_sec=1), "Cleanup logs not found"
+
+
+@patch("skyrl.tinker.api.psutil.Process")
```
pcmoritz: Instead of mocking this, we should modify `_build_uv_run_cmd_engine` to take a `parent_cmdline` variable, which in `lifespan` will be set to `psutil.Process(os.getppid()).cmdline()`. This way we can just test the `_build_uv_run_cmd_engine` function properly without the mocking.

skyrl/tinker/api.py
```diff
@@ -40,23 +39,53 @@
...
-    return list({extra for extra in parsed.extra if extra in {"gpu", "tpu"}})
+    # the API server startup command can be
+    # uv run <uv run args> -m skyrl.tinker.api
+    # or uv run <uv run args> python -m skyrl.tinker.api
+    # or uv run <uv run args> -- python -m skyrl.tinker.api
+    stop_strings = ["--", "python"]
+    detected = False
+    for i in range(len(tokens) - 1):
+        if tokens[i] in stop_strings or (
+            tokens[i] == API_SERVER_STARTUP_ARGS[0] and tokens[i + 1] == API_SERVER_STARTUP_ARGS[1]
```
pcmoritz: This can be cleaner done as `tokens[i : i + len(API_SERVER_STARTUP_ARGS)] == API_SERVER_STARTUP_ARGS`

skyrl/tinker/api.py
SumanthRH: Makes sense! Given that the skyrl package is root level now and that `.venv` is in `.gitignore` I believe `--isolated` need not be the default path (because ray will not package it in working dir by default. Previously we did this because users cd'ed into the skyrl-train folder)

tests/tinker/test_api.py
SumanthRH: Cleaner! done!

skyrl/tinker/api.py
SumanthRH: Done!

skyrl/tinker/api.py
```diff
@@ -40,23 +39,56 @@
...
+    # the API server startup command can be
+    # uv run <uv run args> -m skyrl.tinker.api
+    # or uv run <uv run args> python -m skyrl.tinker.api
+    # or uv run <uv run args> -- python -m skyrl.tinker.api
+    stop_strings = ["--", "python"]
+    detected = False
+    for i in range(len(parent_cmd) - 1):
+        if parent_cmd[i] in stop_strings or parent_cmd[i : i + len(API_SERVER_STARTUP_ARGS)] == API_SERVER_STARTUP_ARGS:
+            detected = True
+            break
```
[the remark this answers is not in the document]
SumanthRH: The case of `uv run --with python -m skyrl.tinker.api` is not handled since it is somewhat nonsensical.


> The same issue applies if any flag has -- as its value (e.g. an unusual --env-file path).

That is not the string matching check though? We are looking for exact string match with `--` or `python`

skyrl/tinker/api.py
pcmoritz: It actually is still needed in some cases I think, e.g. @erictang000 was running into the environment not being cleaned up and therefor the disk getting full (e.g. with the megatron backend). We should probably just add it to the relevant examples going forward :)

erictang000: hmm yeah i haven't seen whether this is fixed now for the case (i.e. on Anyscale) where the uv cache is not on the same disk as the ray runtime and the "failed to hardlink" message comes up and triggers a copy. would be good to test

PR #1271, opened 2026-03-04
[train][1/N] Native Weight Sync API: NCCL

hao-aaron:
# Use vLLM Native Weight Sync APIs for NCCL Weight Transfer

## Motivation

SkyRL's weight synchronization between trainer and inference engines previously relied on custom HTTP endpoints (`/init_weight_transfer`, `/update_weights`, `/get_server_info`) and a `WorkerWrap` extension injected into vLLM workers. 

As of https://github.com/NovaSky-AI/SkyRL/pull/1240, vLLM now provides native weight sync endpoints (`/init_weight_transfer_engine`, `/update_weights`, `/get_world_size`) via the RLHF router (enabled with `VLLM_SERVER_DEV_MODE=1`). These use vLLM's `NCCLWeightTransferEngine` internally.

## Design Decisions

- **`_SKYRL_USE_NEW_INFERENCE=1` env var gates the new path**: When unset, all behavior is unchanged (legacy path).
- **Dict-based API payloads**: The new path uses plain dicts for all vLLM API calls, matching vLLM's native endpoint contracts. This avoids coupling to SkyRL-specific dataclasses on the vLLM side.
- **Per-server rank offsets computed client-side**: Each inference server gets a different NCCL `rank_offset` in its init payload, computed from `world_size_per_server`. This replaces the old `for_engine()` method that computed offsets inside the server.
- We choose to defer IPC for a future PR, in the mean time ipc path with _SKYRL_USE_NEW_INFERENCE=1 will be broken.

### Dual-path note

The legacy inference path (without `_SKYRL_USE_NEW_INFERENCE`) still uses SkyRL's custom endpoints, `WorkerWrap`, and the per-chunk broadcast flow. Because of this, the codebase currently supports both the old custom endpoints and the new vLLM native APIs side-by-side, gated by the environment variable. This dual-path requirement introduces some unavoidable code complexity (e.g. `_send_chunks_legacy` vs `_send_chunks_vllm_native`, branching in `worker.py` and `create_sender`, using `BroadcastInitInfo.update_rank_offset and BroadcastInitInfo.to_api_payload`).

### Core Architecture

**`broadcast_strategy.py`** — New vLLM native send path alongside legacy broadcast:
- Added `BroadcastInitInfo.update_rank_offset(world_size_per_server)` to expand a single init info into per-server payloads with correct NCCL rank offsets. SkyRL API used to address this in the custom endpoint to support multiple servers, but native api does not.
- Added `BroadcastInitInfo.to_api_payload()` to convert init info into the dict format expected by vLLM's native API.
- New `_send_chunks_vllm_native()` method: sends weight metadata via `/update_weights`, then streams tensors via `NCCLWeightTransferEngine.trainer_send_weights` in packed mode.
- `create_sender()` now initializes `NCCLWeightTransferEngine.trainer_init` on rank 0 when using the new path, instead of creating a custom `torch.distributed` process group.

**`remote_inference_client.py`** — Switched to vLLM native endpoints:
- `get_world_size()` now uses `/get_world_size` (was `/get_server_info`) and returns `Tuple[int, List[int]]` (total + per-server).
- `init_weight_update_communicator()` now accepts `List[Dict]` (one per server) and calls `/init_weight_transfer_engine` on each server with its specific payload.
- `update_named_weights()` accepts a plain `Dict[str, Any]` and forwards to `/update_weights` (no more `WeightUpdateRequest` objects).
- Removed unused `WeightUpdateRequest` and `TYPE_CHECKING` imports.

**`vllm_server_actor.py`** — Removed custom endpoints and worker extension:
- Removed `_ensure_worker_extension()` and the `WorkerWrap` dependency.
- Removed custom `/get_server_info`, `/init_weight_transfer`, `/update_weights`, and `/finalize_weight_update` FastAPI endpoints. vLLM's RLHF router (enabled via `VLLM_SERVER_DEV_MODE=1`) now handles these natively.
- Removed `_get_extended_server_info()` helper.

### Test Plan
- [x] `test_policy_local_engines_e2e.py`
- [x] `inference_servers/test_weight_sync.py`
- [x] `test_engine_generation.py`
- [x] E2E GSM8K run: https://wandb.ai/sky-posttraining-uc-berkeley/gsm8k_fully_async_ci_native_api/runs/4mwsqjim?nw=nwuserahao
image

## Future Work

- **CUDA IPC support on the new path**: Colocated setups currently fall back to the legacy `CudaIpcTransferStrategy`. The dict-based API is designed to accommodate CUDA IPC payloads once vLLM adds native support.
- **Remove `WorkerWrap` entirely**: Once CUDA IPC is also handled natively, the `vllm_worker.py` extension can be fully removed.
- **Remove legacy broadcast path**: Once the new path is validated at scale, the `_send_chunks_legacy` path and its per-chunk HTTP calls can be removed.
- **Megatron metadata without materialization**: The current Megatron `get_weight_metadata` still runs `export_hf_weights` once (with caching). A lighter-weight approach to discover HF parameter names from the Megatron model structure would be ideal.

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1271" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


skyrl/backends/skyrl_train/workers/worker.py
```diff
@@ -345,7 +346,14 @@ async def init_weight_sync_state(
         # NOTE: For broadcast strategy, sender and receiver init must run concurrently
         # because both need to join the same process group to avoid deadlock
         if torch.distributed.get_rank() == 0:
-            tasks.append(inference_engine_client.init_weight_update_communicator(init_info))
+            if _SKYRL_USE_NEW_INFERENCE:
```
kouroshHakha: it would be better for separation of concerns if the remote inference client does this offset management inside and the contract between inference client and trainer will be this:

On rank 0, trainer calls inference_client.init_weight_update_communitactor(init_info) with offset being 1. The client know about the server / the world_size per server, etc. so it can adjust the offset internally when calling into the endpoints.

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -545,77 +542,109 @@ async def reset_prefix_cache(
...
+            async with session.post(
+                f"{url}/init_weight_transfer_engine",
+                json={"init_info": init_dict},
+                timeout=timeout,
+            ) as resp:
+                body = await resp.json() if resp.content_length else None
+                raise_for_status(resp, body)
+                return url, {"status": resp.status, "body": body}
+
+        results = await asyncio.gather(*[post_init(url, d) for url, d in zip(self.server_urls, init_info_list)])
```
kouroshHakha: instead of repeating the session.post code here, please extend `_call_all_servers` to take either a list of json or a single json. If a list is passed in it will assume index 0 of that list corresponds to server_url[0] and so on.

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -545,77 +542,109 @@ async def reset_prefix_cache(
...
         """
         if self._world_size is not None:
             return self._world_size
 
-        results = await self._call_all_servers("/get_server_info", {}, method="GET")
+        results = await self._call_all_servers("/get_world_size", {}, method="GET")
 
         total_world_size = 0
-        for server_url, resp in results.items():
+        world_size_per_server: List[int] = []
```
kouroshHakha: is there any reason we want to allow world_size per server to be heterogenous? That over complicates things. We should just assume servers have the same world_size per server. Let's do this: assert if all(results) != results[0] and return results[0] * len(server_urls).

We can reconsider this assertion if use-cases come up later.

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -545,77 +542,109 @@ async def reset_prefix_cache(
...
         Returns:
             Dict mapping server_url to response.
         """
-        return await self._call_all_servers("/update_weights_skyrl", request.to_json_dict())
+        return await self._call_all_servers(
+            "/update_weights",
+            {"update_info": update_info},
+        )
 
     async def finalize_weight_update(self) -> Dict[str, Any]:
```
kouroshHakha: I don't think anyone uses this API. We should delete instead?

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
kouroshHakha: We don't need to return world_size_per_server either if we do what I said about separation of concerns and managing rank offsets inside the remote inference client.

skyrl/backends/skyrl_train/workers/fsdp/fsdp_worker.py
```diff
@@ -226,7 +251,11 @@ async def broadcast_to_inference_engines(self, inference_engine_client, inferenc
             return
 
         # Extract and send weights using the sender created at init time
-        await self._weight_transfer_sender.send_chunks(self.weight_extractor.extract_weights(generator_dtype))
+        weight_metadata = self.weight_extractor.get_weight_metadata(generator_dtype)
+        await self._weight_transfer_sender.send_chunks(
+            self.weight_extractor.extract_weights(generator_dtype),
+            weight_metadata=weight_metadata,
+        )
```
kouroshHakha, line 259: cleaner:

```
weight_metadata = self.weight_extractor.get_weight_metadata(generator_dtype)
weight_iterator =  self.weight_extractor.extract_weights(generator_dtype)
await self._weight_transfer_sender.send_chunks(weight_iterator, weight_metadata)
```

skyrl/backends/skyrl_train/weight_sync/broadcast_strategy.py
```diff
@@ -71,55 +113,111 @@ class BroadcastWeightUpdateRequest(WeightUpdateRequest):
...
             init_info: BroadcastInitInfo containing all config-derived args.
-            model_update_group: Process group for broadcast operations (None on non-rank-0 training workers).
+            model_update_group: Process group for broadcast (legacy path); None when vllm_group set.
             inference_client: Client for coordinating with inference engines.
+            vllm_group: vLLM NCCL communicator from NCCLWeightTransferEngine.trainer_init (new path).
         """
         self._init_info = init_info
         self._model_update_group = model_update_group
         self._inference_client = inference_client
+        self._vllm_group = vllm_group
```
kouroshHakha: should we raise an error if the env var is set but vllm_group is not set?

kouroshHakha: I would avoid using vllm in the name. This is really the inference_process_group? 
And its type should something special? Like PyNCCL... blah? if you cannot put it in typechecking we should comment on it.

skyrl/backends/skyrl_train/weight_sync/broadcast_strategy.py
```diff
@@ -71,55 +113,111 @@ class BroadcastWeightUpdateRequest(WeightUpdateRequest):
...
+        def do_gather_and_send() -> None:
+            if rank == 0 and self._vllm_group is not None:
+                NCCLWeightTransferEngine.trainer_send_weights(
+                    iterator=weight_iterator(),
+                    group=self._vllm_group,
+                    packed=True,
+                )
+            else:
+                for _ in weight_iterator():
+                    pass
```
kouroshHakha: I don't think the else is needed here? Can you explain it?

skyrl/backends/skyrl_train/weight_sync/broadcast_strategy.py
```diff
@@ -71,55 +113,111 @@ class BroadcastWeightUpdateRequest(WeightUpdateRequest):
...
+        chunks: Iterable[WeightChunk],
+        weight_metadata: Optional[Dict[str, list]] = None,
+    ) -> None:
+        """Batched path: one update_weights call + trainer_send_weights (vLLM native).
+
+        All ranks must evaluate the chunks iterator (extract_weights uses
+        collective all-gather internally). Only rank 0 sends the gathered
+        tensors to vLLM via the NCCL weight transfer engine.
+        """
+        from vllm.distributed.weight_transfer.nccl_engine import NCCLWeightTransferEngine
```
kouroshHakha: could this method be implemented much simpler? 

```
# no-op if you are not on rank zero
# the weights are already gathered? 
if rank != 0:
    return 

assert self._vllm_group is not None
assert len(chunk.names) == len(chunk.tensors)

update_task = asyncio.create_task(inference_client.update_named_weights(update_info))
NCCLWeightTransferEngine.trainer_send_weights(
    iterator=zip(chunk.names, chunk.tensors),
    group=self._vllm_group,
    packed=True,
)

await update_task
torch.distributed.barrier()
```

skyrl/backends/skyrl_train/weight_sync/broadcast_strategy.py
kouroshHakha: wait, actually why is `model_update_group` different than `vllm_group`?

skyrl/backends/skyrl_train/weight_sync/broadcast_strategy.py
```diff
@@ -255,31 +350,43 @@ def create_sender(
...
+                )
+        else:
+            if rank == 0:
+                model_update_group = init_custom_process_group(
+                    backend=init_info.backend,
+                    init_method=get_tcp_url(init_info.master_addr, init_info.master_port),
+                    world_size=init_info.world_size,
+                    rank=0,
+                    group_name=init_info.group_name,
+                )
```
kouroshHakha: why are we not consolidating vllm_group and model_update_group? I don't understand.

skyrl/backends/skyrl_train/inference_servers/utils.py
```diff
@@ -30,6 +31,7 @@ def build_vllm_cli_args(cfg: SkyRLTrainConfig) -> Namespace:
         max_num_batched_tokens=ie_cfg.max_num_batched_tokens,
         max_num_seqs=ie_cfg.max_num_seqs,
         enable_sleep_mode=cfg.trainer.placement.colocate_all,
+        weight_transfer_config=WeightTransferConfig(backend="nccl"),
```
kouroshHakha: We should determine this based on the cfg object. If colocated use ipc (raise error for now), otherwise use nccl?

skyrl/backends/skyrl_train/inference_servers/vllm_server_actor.py
```diff
@@ -100,9 +96,6 @@ def __init__(
         self._server_idx = server_idx
         self._num_gpus_per_server = self.compute_num_gpus_per_server(vllm_cli_args)
 
-        # Ensure SkyRL's custom worker extension is used for weight sync
-        self._ensure_worker_extension()
-
         # Ensure vLLM sleep endpoints are enabled by using dev mode
         os.environ["VLLM_SERVER_DEV_MODE"] = "1"
         os.environ["VLLM_RAY_PER_WORKER_GPUS"] = str(0.2 if colocated_training else 1.0)
```
kouroshHakha, line 101: we need to also set `VLLM_ALLOW_INSECURE_SERIALIZATION`

tests/backends/skyrl_train/gpu/gpu_ci/test_policy_local_engines_e2e.py
```diff
@@ -39,19 +39,22 @@ def get_test_actor_config() -> SkyRLTrainConfig:
...
+        # colocated uses CUDA IPC strategy, not yet supported on new inference path
+        # pytest.param(True, "nccl", "fsdp2", 2),
     ],
     ids=[
         "no_colocate_nccl_fsdp_vllm",
-        "colocate_nccl_fsdp_vllm",
-        "no_colocate_gloo_fsdp_vllm",
-        "colocate_gloo_fsdp_vllm",
+        # "colocate_nccl_fsdp_vllm",
+        # "no_colocate_gloo_fsdp_vllm",
```
kouroshHakha: uncomment?

skyrl/backends/skyrl_train/weight_sync/broadcast_strategy.py
hao-aaron: weight iterator does all gather which requires all ranks to participate, i can add a comment explaining

tests/backends/skyrl_train/gpu/gpu_ci/test_policy_local_engines_e2e.py
hao-aaron: colocate_nccl_fsdp_vllm actually uses ipc, and gloo is not yet supported

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
hao-aaron: i think keeping the return for both keeps things simpler, since its they are cached values that may not exist so i prefer to access it with function even from inside the class, but i will make world size per server a single int

skyrl/backends/skyrl_train/weight_sync/broadcast_strategy.py
hao-aaron: it should be the same mb

skyrl/backends/skyrl_train/weight_sync/broadcast_strategy.py
hao-aaron: the complexity comes from the fact that chunks is an iterator that may contain batched params, so we have to put this in a format that trainer_send_weights expects, which is just iterating over each parameter, and we have to use to_thread on trainer_send_weights so we don't block update_named_weights. I can clean up the code paths though.

skyrl/backends/skyrl_train/inference_servers/utils.py
SumanthRH: +1

tests/backends/skyrl_train/gpu/gpu_ci/test_policy_local_engines_e2e.py
SumanthRH: commenting the parameters here is not good because this means that these tests are disabled now for the old inference codepath as well. We should instead keep all configurations here, and then select the ones that are valid for the new inference codepath in CI:

https://github.com/NovaSky-AI/SkyRL/blob/ed3305f18793d408563dd8056f5ea9a621be8399/ci/gpu_ci_run_skyrl_train.sh#L34

We can reenable the tests once we support the IPC backend

skyrl/backends/skyrl_train/weight_sync/broadcast_strategy.py
```diff
@@ -59,6 +59,49 @@ def for_engine(self, engine_index: int, tp_size: int, pp_size: int) -> "Broadcas
         cumulative_offset = engine_index * tp_size * pp_size
         return replace(self, rank_offset=self.rank_offset + cumulative_offset)
 
+    # TODO (Aaron): native weight sync only needs the following params:
+    #     master_address, master_port, rank_offset, world_size
+    # so we need a new method (to_api_payload) to return the payload for the native weight sync.
+    # Also we need a new method (update_rank_offset) to update the rank_offset for the native weight
+    # sync, since this is done automatically in the legacy weight sync.
+
+    def update_rank_offset(self, world_size_per_server: int, num_servers: int) -> List["BroadcastInitInfo"]:
```
SumanthRH:
```suggestion
    def for_servers(self, world_size_per_server: int, num_servers: int) -> List["BroadcastInitInfo"]:
```

`update_rank_offset` is not the right name for this - it sounds like we are updating the original object in some way

skyrl/backends/skyrl_train/weight_sync/broadcast_strategy.py
```diff
@@ -71,55 +114,105 @@ class BroadcastWeightUpdateRequest(WeightUpdateRequest):
...
+            for _ in weight_iterator():
+                pass
+
+        torch.distributed.barrier()
+
+    async def _send_chunks_legacy(self, chunks: Iterable[WeightChunk]) -> None:
+        """Per-chunk HTTP + torch.distributed.broadcast (legacy path)."""
         rank = torch.distributed.get_rank()
 
-        # Rank 0 must have a process group to broadcast to inference engines
```
SumanthRH, line 110: Nit: Why are the older comments deleted?

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -427,30 +427,38 @@ async def detokenize(
     async def _call_all_servers(
         self,
         endpoint: str,
-        json: Dict[str, Any],
+        json: Union[Dict[str, Any], List[Dict[str, Any]]],
         method: str = "POST",
```
SumanthRH: Nit: While this change is okay, it is better if you keep `_call_all_servers` as is and just make a separate `_call_server` helper method;

```python
async def _call_server(self, server_url: str, endpoint: str, json: Dict[str, Any], method: str = "POST")
```


`_call_all_servers` can just do fan-out as before.

skyrl/backends/skyrl_train/weight_sync/broadcast_strategy.py
hao-aaron: oops claude

skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
```diff
@@ -189,6 +189,34 @@ def calculate_size_in_bytes(param, tp_size, ep_size):
             self.param_buckets[-1].append(task)
             curr_size += size
 
+    def get_weight_metadata(self, dtype: torch.dtype) -> dict:
```
SumanthRH, line 192: @hao-aaron have you run weight syncing test with the megatron backend as well ?

hao-aaron: pushed changes for megatron test and verified passing

Issue #815, opened 2025-12-30
[skyrl-train] Enable routing replay in SkyRL


erictang000:
We should enable routing replay for deterministic training with MoE models once this is upstreamed to vLLM

VLLM PR: https://github.com/vllm-project/vllm/pull/28284

erictang000:
router replay data transfer:

Router replay can quickly make the size of the data batch to transfer between trainer and inference quite large. See example below where 32GB gets transferred per node for GLM-4.7-Flash, with batch size=128, n_samples_per_prompt=8, num_layers=47, top_k_experts=4, and max_context_length=20000. 

This quickly results in ~16 GiB of expert indices, and since we are doing tp=4, we have 2 dp ranks per node materializing the data, resulting in 32GB per node

<img width="636" height="156" alt="Image" src="https://github.com/user-attachments/assets/09758ce9-450d-4e1c-ace4-a6e1629beb8b" />

<img width="1170" height="1078" alt="Image" src="https://github.com/user-attachments/assets/87e0ec68-c9fe-4361-b215-1883251aff47" />

This can be a bottleneck for scaling to larger models with more layers and longer context length. We should take the following steps to improve this

- [x] use uint8/int16 for expert indices (GLM-5 uses 256 routed experts, Kimi-2.5 uses 384, which needs more than uint8...) https://github.com/NovaSky-AI/SkyRL/pull/1378
- [x] remove unnecessary full batch data transfer to each node: https://github.com/NovaSky-AI/SkyRL/pull/1376
- [ ] remove padding tokens for expert indices
- [ ] ensure that all trainer ranks block on retrieving data to avoid hanging NCCL communication if different DP ranks take different amounts of time to retrieve from ray object store.
- [ ] Work on gpu based transfer for routed expert indices
- [ ] Time object serialization/deserialization from driver -> train workers

sandys:
@erictang000 will this work with the tinker api ?

erictang000:
hey @sandys - not currently, added that ask a task to track!

sandys:
@erictang000 thank you - super appreciate it !

jamesbraza:
Can I request FSDP-trainer support for R3 on the roadmap? 🤓

PR #1273, opened 2026-03-04
R3 PR: Rollout Routing Replay

erictang000:
# Overview
This PR adds support for Rollout Routing Replay (R3) from ([See Paper](https://arxiv.org/abs/2510.11370)).

See #815 for tracking of future tasks to fully support routing replay in all settings.

We add the following flags to enable R3:

```
cfg.generator.inference_engine.enable_return_routed_experts=True
cfg.trainer.policy.megatron_config.moe_enable_routing_replay=True
```
`cfg.generator.inference_engine.enable_return_routed_experts=True` is a pass through argument to vLLM, which records expert router indices (returning a list of dimension `(batch_size, seq_len, num_layers, top_k)`.

We then pass this list `rollout_expert_indices` list through to Megatron's native `RouterReplay` feature ([link](https://github.com/NVIDIA/Megatron-LM/blob/main/docs/api-guide/router_replay.md)).

When `cfg.trainer.policy.megatron_config.moe_enable_routing_replay` is set to `true`, Megatron initializes an instance of `RouterReplay` on each training worker rank. `RouterReplay.set_replay_data(per_layer_data)` can be used to set router decisions, and `RouterReplay.set_global_router_replay_action(RouterReplayAction.REPLAY_FORWARD)` and `RouterReplay.set_global_router_replay_action(RouterReplayAction.REPLAY_BACKWARD)` can be used to set the routing mode to be forward or backward.

# Results
GSM8K Training on `moonlight16b-a3b` shows R3 improves training stability - this can be seen both in logprob diffs as well as in `clip_ratio`, `grad_norm`, and `loss`, which otherwise explode and collapse training.
image
image

# Supported Settings
Router Replay is supported for the following settings:
#### Generator Settings
- `use_conversation_multi_turn=True` and `use_conversation_multi_turn=False`
- `batched=False` and `batched=True`
- `async_engine=True` and `async_engine=False`
- NOT `retokenize_chat_history` mode - i.e. ` self.use_conversation_multi_turn and self.custom_chat_template`
- NOT `self.generator_cfg.step_wise_trajectories` - there are some question marks about how to support this when using step wise training and not strictly appending (what should the routing look like for per turn obs that the inference engine doesn't see? - do we need to disable routing overrides for those tokens?)
- fully_async training - technically should work but not tested in this PR. Tracking in #815 
#### Inference Engine Settings
- TP and EP and DP should be supported from the vLLM side
- NOTE: `_SKYRL_USE_NEW_INFERENCE` is not supported - this will be added in a follow up PR
- NOTE: `cfg.generator.distributed_executor_backend` must be set to `mp` - hanging related to a Ray Compiled Graph issue occurs when using the default `ray` vLLM distributed executor backend. (see https://github.com/vllm-project/vllm/issues/36237 for details on the error that comes up)
- NOTE: The above use of `mp` also means that serving must be single node per engine, until we add support for using the mp backend with multi-node serving - progress tracked here: https://github.com/NovaSky-AI/SkyRL/issues/1309

#### Trainer Settings
- TP, EP, DP are all supported. CP is in progress in this PR but needs more testing. CP + PP will be added in a follow up PR.

#### Custom Generator support
- Custom generators using SkyRL's inference engine should just plumb through 

# Tests
Adds `test_router_replay.py`, which includes:
- `test_logprobs` - integration test that runs a training batch through vllm, and through megatron with and without R3, to verify that logprob diffs are lower with routing replay
- `test_forward_backward` - unit test for `forward_backward` that verifies that a training step can complete successfully when routing replay indices are passed in

Adds `test_generator_multi_turn_gsm8k_router_replay` to `test_skyrl_gym_generator` to verify that the `SkyRLGymGenerator` plumbs through the router indices in an expected format.

# Rollout Routing Replay 

image

Relevant resources: 
vLLM PR: https://github.com/vllm-project/vllm/pull/28284
Verl PR: https://github.com/volcengine/verl/pull/4101
Mindlab blog: https://macaron.im/mindlab/research/router-replay-r3-why-it-failed-and-how-we-fixed-it
Megatron-LM API guide: https://github.com/NVIDIA/Megatron-LM/blob/main/docs/api-guide/router_replay.md

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1273" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/examples/train/router_replay/run_dapo_moonlight_16b_a3b.sh
+++ b/examples/train/router_replay/run_dapo_moonlight_16b_a3b.sh
@@ -1,0 +1,129 @@
+set -x
+
+# Colocated DAPO training+generation for Moonlight-16B-A3B on DAPO with Megatron with router replay.
+# Should run on 2 node of 8xH100s
+
+# bash examples/train/algorithms/dapo/prepare_dapo_data.sh
+# bash examples/train/router_replay/run_dapo_moonlight_16b_a3b.sh
+
+MODEL_NAME="moonshotai/Moonlight-16B-A3B-Instruct"
+DATA_DIR="$HOME/data/dapo"
+TRAIN_FILE="$DATA_DIR/dapo-math-17k-cleaned.parquet"
+TEST_FILE="$DATA_DIR/aime-2024-cleaned.parquet"
+NUM_NODES=2
+NUM_GPUS_PER_NODE=8
+NUM_INFERENCE_ENGINES=2
+INFERENCE_ENGINE_TENSOR_PARALLEL_SIZE=8
+LOGGER="wandb"  # change to "console" to print to stdout
+
+# flash attention off
+FLASH_ATTN=false
+
+CLIP_RATIO_LOW=0.2
+CLIP_RATIO_HIGH=0.28
+# use token mean loss reduction
+LOSS_REDUCTION="token_mean"
+# applies overlong filtering (but not soft overlong punishment)
+APPLY_OVERLONG_FILTERING=true
+# apply soft overlong punishment with custom trainer impl in main_dapo.py
+OVERLONG_BUFFER_LEN=$((1024 * 4))
+OVERLONG_BUFFER_PENALTY_FACTOR=1.0
+
+# other DAPO parameters
+USE_KL_LOSS=false
+TEMPERATURE=1.0
+TOP_P=1.0
+EVAL_TOP_P=0.7
+CLIP_RATIO_C=10.0
+MAX_PROMPT_LENGTH=$((1024 * 2))
+MAX_RESPONSE_LENGTH=$((1024 * 8))
+
+# repro run parameters
+TRAIN_BATCH_SIZE=128
+MINI_BATCH_SIZE=32
+N_SAMPLES_PER_PROMPT=16
+EVAL_N_SAMPLES_PER_PROMPT=32
+ENFORCE_EAGER=true # cuda graphs can cause some instability
+LR=1e-6
+
+# megatron config
+MEGATRON_TP=4
+MEGATRON_PP=1
+MEGATRON_CP=1
+MEGATRON_EP=8
+MEGATRON_ETP=1
+
+
+# Router replay (r3)
+ROUTER_REPLAY=true
+DISTRIBUTED_EXECUTION_BACKEND="mp"
+
+SKYRL_RAY_PG_TIMEOUT_IN_S=300 uv run --isolated --extra megatron -m examples.train.algorithms.dapo.main_dapo \
+  data.train_data="['$TRAIN_FILE']" \
+  data.val_data="['$TEST_FILE']" \
+  trainer.algorithm.advantage_estimator="grpo" \
+  trainer.algorithm.policy_loss_type="dual_clip" \
+  trainer.algorithm.overlong_buffer_len=$OVERLONG_BUFFER_LEN \
+  trainer.algorithm.overlong_buffer_penalty_factor=$OVERLONG_BUFFER_PENALTY_FACTOR \
+  trainer.algorithm.loss_reduction=$LOSS_REDUCTION \
+  generator.inference_engine.enforce_eager=$ENFORCE_EAGER \
+  generator.apply_overlong_filtering=$APPLY_OVERLONG_FILTERING \
+  generator.sampling_params.temperature=$TEMPERATURE \
+  generator.sampling_params.top_p=$TOP_P \
+  generator.eval_sampling_params.top_p=$EVAL_TOP_P \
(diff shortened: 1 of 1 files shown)


erictang000:
Forward pass with router replay showing lower logprob diff!

image

erictang000:
Current State:

For small scale tests, routing replay seems to be working as shown above - tested only for TP=8 serving and TP=4 and EP=8 training. 
image

What's not working:

- [x] Deepseek style models (like moonlight-16b-a3b) - the average logprobs for both with and without routing replay seem to be way off from what is expected from the inference engine (like ~8 average logprobs vs 0.2 average logprobs) - maybe this is specific to moonlight + some megatron config? not sure.  Reproducible by running the `test_logprobs` tests
- [ ] Running a real training batch runs into hanging for all models - see https://github.com/vllm-project/vllm/issues/36237 for the relevant stack trace. This seems to happen pretty much only when router replay is enabled on the vllm end. We could try upgrading vllm to 0.17.0 to see if anything has been pushed that could help fix this. Reproducible by running a large enough training batch with router replay on the `test_logprobs` tests

TODOs:
- [x] fix above bugs
- [x] clean up code
- [x] add checks for settings where r3 won't be supported for now (async RL, batched generator, `use_conversation_multi_turn=false`)
- [x] run full end to end test showing r3 minimizes drift on DAPO (with at least one model family - others just need to pass tests.

erictang000:
>  Deepseek style models (like moonlight-16b-a3b) - the average logprobs for both with and without routing replay seem to be way off from what is expected from the inference engine (like ~8 average logprobs vs 0.2 average logprobs) - maybe this is specific to moonlight + some megatron config? not sure. Reproducible by running the test_logprobs tests

Solved! the issue was that in our test we were setting `cfg.trainer.use_sample_packing=False`, and also setting `NVTE_FUSED_ATTN=0`. This was causing flash-attn to be used without sample packing for the moonlight forward pass - flash attn produces incorrect logprobs that differ greatly from vLLM:

```bash
vLLM logprobs     - mean: -2.564655
Megatron (replay) - mean: -9.231623
Megatron (no rep) - mean: -9.647593
```

after setting `cfg.trainer.use_sample_packing=True` and `NVTE_FUSED_ATTN=1` (to allow setting `transformer_config_kwargs["attention_backend"]` to "fused" correctly

```bash
vLLM logprobs     - mean: -0.223607, std: 0.674102
Megatron (replay) - mean: -0.223626, std: 0.674850
Megatron (no rep) - mean: -0.224379, std: 0.677036
With replay    - logprob diff mean: 0.006648, std: 0.021737
Without replay - logprob diff mean: 0.011115, std: 0.035957
```

erictang000:
>  Running a real training batch runs into hanging for all models - see https://github.com/vllm-project/vllm/issues/36237 for the relevant stack trace. This seems to happen pretty much only when router replay is enabled on the vllm end. We could try upgrading vllm to 0.17.0 to see if anything has been pushed that could help fix this. Reproducible by running a large enough training batch with router replay on the test_logprobs tests


verified that cherry picking the changes from #1300 to use the mp backend allow us to work around the compiled graph timeout 

image

tests/backends/skyrl_train/gpu/gpu_ci/test_router_replay.py
```diff
@@ -0,0 +1,371 @@
...
+            "advantages": torch.zeros((batch_size, num_actions), dtype=torch.float32),
+            "action_mask": response_mask.to(dtype=torch.int64),
+        }
+    )
+    training_input.metadata = {"response_length": num_actions}
+    return training_input
+
+
+@pytest.mark.megatron
+@pytest.mark.skip(reason="Skipping router replay test for now due to size constraints")
```
erictang000, line 108: skipping these for now - need to test some smaller models out - tracking as a follow up task in #815, will do ASAP along with PP + CP support

tests/backends/skyrl_train/gpu/utils.py
```diff
@@ -289,6 +290,20 @@ def get_test_prompts(model: str, num_samples: int = 20) -> List[ConversationType
     return prompts
 
 
+def _ensure_chat_template(tokenizer):
```
erictang000, line 293: this is needed for "allenai/OLMoE-1B-7B-0924", which is used in the skyrl_gym_generator test - i plan to port over this model to the other router replay tests (and maybe megatron moe tests in general) since it's supported in Megatron-Bridge and is a 7B with 1B activated MoE that makes it nice for CI.

erictang000:
additional curves for DAPO for posterity (green is DAPO on moonlight w/ TIS, blue is same but w/o TIS + router replay)
image
image
image
image


Time for fwd/train is slightly higher with R3:
image
image

SumanthRH (approved):

Issue #1286, opened 2026-03-06
[tinker] Sample API in new inference server


nithinvc:
Currently, the new inference server does not support the `sample` API, which is required before integrating with tinker.

nithinvc:
Done in #1396

PR #1287, opened 2026-03-06
[tinker][SkyRL] Add sample and renderer API to new inference client

nithinvc:
### Summary

This PR adds the `sample()` and `render_chat_completion()` methods to RemoteInferenceClient and wires the new HTTP-based inference pathway into SkyRLTrainBackend, gated behind the `_SKYRL_USE_NEW_INFERENCE` env var.

Issues addressed: #1286 #1288

### Changes
  - `RemoteInferenceClient` — Added `sample()` (fires `num_samples` parallel `_generate_single` calls and aggregates into a single `InferenceEngineOutput`) and `render_chat_completion()` (calls `/v1/chat/completions/render` to tokenize chat messages without generating). Tested against vllm `0.16.0`.
  - `SkyRLTrainBackend` — Added `_create_remote_inference_client()` with the same 4-way branching logic as `main_base.py` (external proxy + servers, proxy only, servers only, fully internal `ServerGroup` + `InferenceRouter`). `_ensure_inference_engines()` now branches on `_SKYRL_USE_NEW_INFERENCE` to use the new HTTP client path.
  - `test_engine_generation.py` — Removed the guard on the sample API test. The new sample API passes.
  - `test_save_weights_for_sampler.py` — Fixed GPU test to pass tokenizer to `run_inference` and added `gpu_memory_utilization=0.5` for colocated placement to avoid OOM when running on L4 GPUs.
  - CPU tests — Added mock `/v1/chat/completions/render endpoint` and `test_render_chat_completion` to `test_remote_inference_client.py`.
  - GPU tests — Added `test_client_render_chat_completion` to `test_new_inference_generation.py`.

### Testing

  - end-to-end on 2x L4 GPUs using the tinker cookbook RL loop (requires fix PR #1302 ):
 ```
  TINKER_API_KEY=tml-dummy uv run --with tinker --with datasets --with torch \
      python -m tinker_cookbook.recipes.rl_loop \
      base_url=http://localhost:8000 \
      model_name="Qwen/Qwen3-0.6B" \
      lora_rank=0
```
  - CPU unit tests pass for:
      -  `test_remote_inference_client.py` - no regressions + render API tests
      -  `test_engine_generation.py` - Sample API test
  - GPU unit tests pass for: 
      - `test_save_weights_for_sampler.py` - GPU inference + weight syncing tests

### Limitations

  - LoRA is not supported. `lora_rank=0` is required. LoRA support will be added once we move to the native VLLM weight sync API.

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1287" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
+++ b/skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
@@ -447,40 +447,6 @@ class RemoteInferenceClient:
             raise_for_status(resp, response)
             return response
 
-    async def render_chat_completion(
-        self,
-        request_payload: Dict[str, Any],
-    ) -> List[Any]:
-        """
-        Render chat messages into a tokenized prompt via /v1/chat/completions/render.
-
-        Applies the model's chat template and tokenizes without generating text.
-
-        Args:
-            request_payload: Dict with {"json": <request-body>, "headers": <headers-dict>}.
-                The request body should contain messages and optional chat template params.
-                session_id can be included in json for consistent routing.
-
-        Returns:
-            List of [conversation, engine_prompts] where engine_prompts contains
-            dicts with "prompt" and "prompt_token_ids".
-        """
-        body = request_payload.get("json", {})
-
-        session_id = body.pop("session_id", None)
-
-        headers = {"Content-Type": "application/json"}
-        if session_id:
-            headers["X-Session-ID"] = str(session_id)
-
-        session = await self._get_session()
-        url = f"{self.proxy_url}/v1/chat/completions/render"
-
-        async with session.post(url, json=body, headers=headers) as resp:
-            response = await resp.json()
-            raise_for_status(resp, response)
-            return response
-
     async def completion(
         self,
         request_payload: Dict[str, Any],
--- a/tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_new_inference_generation.py
+++ b/tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_new_inference_generation.py
@@ -34,7 +34,6 @@ from skyrl.backends.skyrl_train.inference_engines.base import (
 from skyrl.backends.skyrl_train.inference_engines.utils import (
     get_sampling_params_for_backend,
 )
-from skyrl.env_vars import _SKYRL_USE_NEW_INFERENCE
 from skyrl.train.config import SkyRLTrainConfig
 from tests.backends.skyrl_train.gpu.utils import InferenceEngineState, get_test_prompts
 
@@ -592,28 +591,3 @@ def test_client_tokenize_detokenize_roundtrip(vllm_server: InferenceEngineState)
 
     decoded = asyncio.run(client.detokenize([token_ids]))[0]
     assert decoded == text
-
-
-@pytest.mark.vllm
-@pytest.mark.skipif(not _SKYRL_USE_NEW_INFERENCE, reason="Render API only supported with new inference client")
-def test_client_render_chat_completion(vllm_server: InferenceEngineState):
-    """Test render_chat_completion via RemoteInferenceClient against real vLLM."""
-    client = vllm_server.client
-    messages = [{"role": "user", "content": "Hello world!"}]
-    request_payload = {
-        "json": {
-            "messages": messages,
-        },
-        "headers": {},
-    }
-    result = asyncio.run(client.render_chat_completion(request_payload))
-    # vLLM returns [conversation, engine_prompts]
-    assert isinstance(result, list)
-    assert len(result) == 2
-    conversation, engine_prompts = result
-    # engine_prompts should have prompt_token_ids matching local tokenizer output
-    assert len(engine_prompts) > 0
-    assert "prompt_token_ids" in engine_prompts[0]
-    tokenizer = AutoTokenizer.from_pretrained(MODEL_QWEN2_5)
-    expected_token_ids = tokenizer.apply_chat_template(messages, add_generation_prompt=True)
-    assert engine_prompts[0]["prompt_token_ids"] == expected_token_ids
--- a/tests/backends/skyrl_train/inference_servers/test_remote_inference_client.py
+++ b/tests/backends/skyrl_train/inference_servers/test_remote_inference_client.py
@@ -17,7 +17,6 @@ from skyrl.backends.skyrl_train.inference_servers.remote_inference_client import
     PauseMode,
     RemoteInferenceClient,
 )
-from skyrl.env_vars import _SKYRL_USE_NEW_INFERENCE
 
 
 def create_mock_vllm_server(server_id: int) -> FastAPI:
@@ -64,15 +63,6 @@ def create_mock_vllm_server(server_id: int) -> FastAPI:
     @app.post("/v1/chat/completions")
     async def chat_completions(request: Request):
         return {"choices": [{"message": {"content": f"Chat from server {server_id}"}}]}
-
-    @app.post("/v1/chat/completions/render")
-    async def render_chat_completion(request: Request):
-        body = await request.json()
-        messages = body.get("messages", [])
-        return [
-            messages,  # conversation (echo back)
-            [{"prompt": "rendered prompt", "prompt_token_ids": [1, 2, 3]}],  # engine_prompts
-        ]
 
     @app.post("/tokenize")
     async def tokenize(request: Request):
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: skyrl/backends/skyrl_train/inference_servers/common.py [hunk 1]
```diff
@@ -72,3 +73,28 @@ def get_open_port(start_port: int | None = None) -> int:
     with socket.socket(socket.AF_INET6, socket.SOCK_STREAM) as s:
         s.bind(("", 0))
         return s.getsockname()[1]
+
+
+def find_and_reserve_port(start_port: int) -> Tuple[int, socket.socket]:
```
pcmoritz: I think this would be best done in a separate PR (making separate PRs is e.g. very useful if one of the changes needs to be reverted, and also helps reviewing).

The same goes for the tests associated to this change of course.

nithinvc: Sounds good! I can split this into two? One for the sample / render and one for the port collision.

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -325,6 +368,48 @@ async def chat_completion(
...
 
+    async def render_chat_completion(
+        self,
+        messages: List[Dict[str, Any]],
+        add_generation_prompt: bool = True,
+        continue_final_message: bool = False,
+        session_id: Optional[Union[str, int]] = None,
+    ) -> Dict[str, Any]:
+        """
+        Render chat messages into a tokenized prompt via /v1/chat/completions/render.
```
pcmoritz: This is very nice! I hadn't know about this endpoint, looks very useful. For client side rendering, I have found https://github.com/thinking-machines-lab/tinker-cookbook/tree/main/tinker_cookbook/renderers useful, but you are right to use the vllm endpoint for this PR, it will make it much easier to keep everything consistent :)

tests/backends/skyrl_train/inference_servers/test_remote_inference_client.py
```diff
@@ -243,6 +253,19 @@ async def test_completion(self, client):
         result = await client.completion(request_payload)
         assert "choices" in result
 
+    @pytest.mark.asyncio
+    @pytest.mark.skipif(not _SKYRL_USE_NEW_INFERENCE, reason="Render API only supported with new inference client")
+    async def test_render_chat_completion(self, client):
```
pcmoritz: For this we should also test if the result is actually correct / is what we expect.

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_new_inference_generation.py
```diff
@@ -585,3 +586,20 @@ def test_client_tokenize_detokenize_roundtrip(vllm_server: InferenceEngineState)
...
+
+
+@pytest.mark.vllm
+@pytest.mark.skipif(not _SKYRL_USE_NEW_INFERENCE, reason="Render API only supported with new inference client")
+def test_client_render_chat_completion(vllm_server: InferenceEngineState):
+    """Test render_chat_completion via RemoteInferenceClient against real vLLM."""
+    client = vllm_server.client
+    messages = [{"role": "user", "content": "Hello"}]
+    result = asyncio.run(client.render_chat_completion(messages=messages))
+    # vLLM returns [conversation, engine_prompts]
```
pcmoritz: Again it would be good to have stricter checks here

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -293,6 +293,49 @@ async def _generate_single(
             "response_logprobs": response_logprobs if len(response_logprobs) > 0 else None,
         }
 
+    async def sample(
```
pcmoritz: The signature looks roughly good to me, but I think for LoRA support we will need to pass in a model_id or lora adapter information, right? Also I would prefer if sampling_params was typed better instead of being a Dict[str, Any].

cc @kouroshHakha

skyrl/backends/skyrl_train_backend.py
```diff
@@ -184,19 +186,89 @@ def init_weight_sync_state(self):
         self._dispatch.init_weight_sync_state(self._inference_engine_client)
         logger.info("Initialized weight sync state for policy model and inference engines.")
 
+    def _create_remote_inference_client(self):
+        """Create a RemoteInferenceClient using HTTP endpoints.
+
+        Mirrors main_base.py._get_new_inference_client() with the same 4-way
+        branching on external_proxy_url / external_server_urls.
+        """
+        from skyrl.backends.skyrl_train.inference_servers.remote_inference_client import RemoteInferenceClient
```
pcmoritz: You can put these inputs to the top of the file I think :)

pcmoritz:
It looks good to me overall, maybe @kouroshHakha has some more feedback :)

One thing that seems a little strange is that we need all of {sample, render_chat_completion, chat_completion, completion, tokenize, detokenize}. It seems to me we would always go through `sample`, and since we pretty much always want token in token out, it seems less useful to have endpoints that do both tokenization and processing like completion and chat_completion.

cc @CharlieFRuan since he has been thinking a lot about token in token out :)

skyrl/backends/skyrl_train_backend.py
nithinvc: moved up!

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_new_inference_generation.py
nithinvc: Added a local tokenizer check for the prompt ids

tests/backends/skyrl_train/inference_servers/test_remote_inference_client.py
nithinvc: Added a check for the prompt and prompt token ids returned by the mock

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
nithinvc: Added a `lora_id` field which default to `None` for now as a stand in. Eventually I think we would pass in IDs to VLLM's [LoRA request](https://docs.vllm.ai/en/stable/api/vllm/lora/request/#vllm.lora.request.LoRARequest).

Also added typing for `sampling_params` but not sure if `Dict[str, Union[str, int, float, bool, List, Dict]]` is much better since there are multiple value types

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
nithinvc: I saw this earlier! It's very useful - I think it might make sense to model the training side changes in a similar way so there's little drift between the client renderer <> training backend

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -293,6 +293,50 @@ async def _generate_single(
...
+        self,
```
kouroshHakha: I would make the interface of this consistent with how chat/completion are implemented. i.e. `request_body` goes in and then it is parsed out. The request_body should have the same API as the tinker sample api. 

Re lora_id / model_id I think we should do this:

InferenceClient has a field called `model_name` that should be used for the apis. I looked at sample API spec from tinker and they don't need to define that there because the client carries the model name information. We should do a similar thing here.

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_backend_weight_sync.py
```diff
@@ -0,0 +1,175 @@
...
+            resp = httpx.get(f"{url}/health", timeout=5.0)
+            if resp.status_code == 200:
+                return True
+        except httpx.RequestError:
+            time.sleep(2.0)
+    return False
+
+
+@pytest.fixture(scope="class")
+def ray_env_with_new_inference():
```
kouroshHakha: don't we already have a conftest for this?

nithinvc: Yes, in fact this test was redundant with some the existing tests so I removed it

nithinvc:
@kouroshHakha @pcmoritz I modified the sample method to follow the `request_payload`, similar to the completions method. The input request is expected to be the args to a tinker sampling client and the output is a dict version of [SampleResponse](https://github.com/thinking-machines-lab/tinker/blob/35a77e79fe873fb4b92504684723de0ac9eae7c9/src/tinker/types/sample_response.py). Could you guys take a look? Thank you!

PR #1291, opened 2026-03-07
[train][2/N] Native Weight Syncing APIs: IPC 

hao-aaron:
follow up to https://github.com/NovaSky-AI/SkyRL/pull/1271,
introduces ipc for colocated case.

Since cuda ipc is not natively supported in 0.16 (will be available in next version), I copied the implementation and registered as a custom engine. This will eventually be removed in favor of the built in version. 

Tests:
- [x] enable colocated tests in `test_policy_local_engines_e2e.py`
- [x] new weight sync unit test in `test_weight_sync.py`
- [x] e2e CI run:

### DIFF
--- a/skyrl/backends/skyrl_train/inference_servers/vllm_server_actor.py
+++ b/skyrl/backends/skyrl_train/inference_servers/vllm_server_actor.py
@@ -13,6 +13,7 @@ import httpx
 import uvicorn
 import vllm.envs as envs
 from fastapi import Request
+from ray.util.placement_group import PlacementGroup
 from vllm.engine.arg_utils import AsyncEngineArgs
 from vllm.engine.async_llm_engine import AsyncLLMEngine
 from vllm.entrypoints.openai.api_server import (
@@ -63,6 +64,21 @@ class VLLMServerActor(ServerActorProtocol):
         """
         return vllm_cli_args.tensor_parallel_size * vllm_cli_args.pipeline_parallel_size
 
+    @staticmethod
+    def prepare_server_kwargs(
+        pg: PlacementGroup,
+        start_bundle_idx: int,
+        num_gpus_per_server: int,
+        **kwargs,
+    ) -> dict:
+        if kwargs.get("distributed_executor_backend") == "mp":
+            from skyrl.train.utils.utils import get_gpu_ids_for_pg_bundles
+
+            bundle_indices = list(range(start_bundle_idx, start_bundle_idx + num_gpus_per_server))
+            gpu_ids = get_gpu_ids_for_pg_bundles(pg, bundle_indices)
+            kwargs["mp_cuda_visible_devices"] = ",".join(str(g) for g in gpu_ids)
+        return kwargs
+
     def __init__(
         self,
         vllm_cli_args: Namespace,
@@ -76,6 +92,8 @@ class VLLMServerActor(ServerActorProtocol):
         enable_pd: bool = False,
         nixl_side_channel_base: int = 5600,
         colocated_training: bool = False,
+        distributed_executor_backend: str = "ray",
+        mp_cuda_visible_devices: Optional[str] = None,
     ):
         """
         Initialize the vLLM server actor.
@@ -93,12 +111,21 @@ class VLLMServerActor(ServerActorProtocol):
             enable_pd: Enable prefill-decode disaggregation
             nixl_side_channel_base: Base port for NIXL side channel
             colocated_training: Whether the server is colocated with training workers
+            distributed_executor_backend: vLLM distributed executor backend.
+                ``"ray"`` spawns TP/PP workers as Ray tasks (default).
+                ``"mp"`` spawns workers as local processes using
+                CUDA_VISIBLE_DEVICES.
+            mp_cuda_visible_devices: Comma-separated physical GPU IDs for the
+                ``"mp"`` backend. Pre-computed by ServerGroup from the
+                per-server placement group. Only used when
+                ``distributed_executor_backend="mp"`` and TP*PP > 1.
         """
         self._cli_args = vllm_cli_args
         self._ip = get_node_ip()
         self._port = get_open_port(start_port)
         self._server_idx = server_idx
         self._num_gpus_per_server = self.compute_num_gpus_per_server(vllm_cli_args)
+        self._use_mp_backend = distributed_executor_backend == "mp"
 
         # Ensure vLLM sleep endpoints are enabled by using dev mode
         os.environ["VLLM_SERVER_DEV_MODE"] = "1"
@@ -106,8 +133,8 @@ class VLLMServerActor(ServerActorProtocol):
         # TODO (aaron): once native ipc stops needing this, remove
         os.environ["VLLM_ALLOW_INSECURE_SERIALIZATION"] = "1"
 
-        # Ensure Ray executor is used (required for GPU inheritance in placement groups)
-        self._ensure_ray_executor()
+        # Configure the distributed executor backend
+        self._cli_args.distributed_executor_backend = distributed_executor_backend
 
         # Update args with our assigned host/port
         self._cli_args.host = "0.0.0.0"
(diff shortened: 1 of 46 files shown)


skyrl/backends/skyrl_train/weight_sync/cuda_ipc_strategy.py
```diff
@@ -39,6 +42,14 @@ def strategy_type() -> type:
         """Return the strategy class for this init info type."""
         return CudaIpcTransferStrategy
 
+    def for_servers(self, world_size_per_server: int, num_servers: int) -> List["CudaIpcInitInfo"]:
+        """IPC init is a no-op, so return identical copies for each server."""
+        return [self] * num_servers
```
SumanthRH: These are not copies but reference the same object. Can you make a copy with `deepcopy` here?

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -446,7 +446,12 @@ async def _call_server(
         session = await self._get_session()
         url = f"{server_url}{endpoint}"
         async with session.request(method, url, json=json) as resp:
-            body = await resp.json() if resp.content_length else None
+            body = None
+            if resp.content_length:
+                try:
+                    body = await resp.json()
+                except Exception:
+                    body = {"error": {"message": await resp.text()}}
```
SumanthRH: Could you explain why this was needed? As such the error message in `body` will be in openai compatible format `error : message: "..."` with vllm errors

skyrl/backends/skyrl_train/weight_sync/vllm_ipc_engine.py
```diff
@@ -0,0 +1,183 @@
...
+            weight = func(*list_args)  # type: ignore
+            weights.append((name, weight))
+
+        load_weights(weights)
+
+    def shutdown(self) -> None:
+        pass
+
+
+# TODO (aaron): remove this once WeightTransferConfig accepts arbitrary backends
```
SumanthRH, line 141: Oh...so the WeighTransferConfig object in vllm 0.16.0 is broken? You can't register a custom backend effectively?

pyproject.toml
```diff
@@ -161,6 +161,9 @@ include-package-data = true
 [tool.setuptools.dynamic]
 version = {attr = "skyrl.__version__"}
 
+[project.entry-points."vllm.general_plugins"]
```
SumanthRH, line 164: Could you add a TODO here as well with a tag for easy lookup that this is because of the weight transfer config related issue with 0.16.0 not required after upgrading to vllm 0.17.0 ?

skyrl/backends/skyrl_train/inference_servers/vllm_server_actor.py
```diff
@@ -102,6 +102,11 @@ def __init__(
         # TODO (aaron): once native ipc stops needing this, remove
         os.environ["VLLM_ALLOW_INSECURE_SERIALIZATION"] = "1"
 
+        # Register custom weight transfer engines not yet in this vLLM version
+        from skyrl.backends.skyrl_train.weight_sync.vllm_ipc_engine import register_ipc_engine
+
+        register_ipc_engine()
+
```
SumanthRH: Why is this needed if we are registering the ipc engine using vllm's plugin system in `pyproject.toml` ?

skyrl/backends/skyrl_train/inference_servers/utils.py
```diff
@@ -10,6 +11,10 @@ def build_vllm_cli_args(cfg: SkyRLTrainConfig) -> Namespace:
     from vllm.utils.argparse_utils import FlexibleArgumentParser
     from vllm.config import WeightTransferConfig
 
+    from skyrl.backends.skyrl_train.weight_sync.vllm_ipc_engine import _patch_weight_transfer_config
+
+    _patch_weight_transfer_config()
+
```
SumanthRH: Again I'm not sure why this is needed if we're using vllm's plugin system

skyrl/backends/skyrl_train/weight_sync/__init__.py
```diff
@@ -43,11 +43,22 @@ def get_transfer_strategy_cls(weight_sync_backend: str, colocate_all: bool) -> T
     Returns:
         The strategy class (CudaIpcTransferStrategy or BroadcastTransferStrategy).
     """
-    if weight_sync_backend == "nccl" and colocate_all:
+    # if weight_sync_backend == "nccl" and colocate_all:
+    #     return CudaIpcTransferStrategy
+    # return BroadcastTransferStrategy
```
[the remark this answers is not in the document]
SumanthRH: +1

skyrl/backends/skyrl_train/weight_sync/cuda_ipc_strategy.py
```diff
@@ -133,31 +145,99 @@ async def send_chunks(
...
+        torch.cuda.synchronize()
+
+        if rank == 0:
+            # Merge per-parameter handle dicts across ranks so each
+            # parameter's dict maps every GPU UUID to its handle.
+            merged_handles: List[Dict[str, IpcHandle]] = []
+            for param_idx in range(len(per_param_handles)):
+                merged: Dict[str, IpcHandle] = {}
+                for rank_handles in gathered:
+                    if rank_handles is not None:
```
SumanthRH: Why is this guard present here? Shouldn't we check the format of `gathered` right after the `all_gather_object` call and make sure there's no `None` entry?

skyrl/backends/skyrl_train/weight_sync/cuda_ipc_strategy.py
```diff
@@ -133,31 +145,99 @@ async def send_chunks(
...
+            update_info: Dict[str, Any] = {
+                "names": names,
+                "dtype_names": dtype_names,
+                "shapes": shapes,
+                "ipc_handles_pickled": pickled,
+            }
+            await self._inference_client.update_named_weights(update_info)
+
+        torch.cuda.ipc_collect()
+        torch.distributed.barrier()
```
SumanthRH: The order here is best reversed right? I guess because the previous code has a reference to the ipc handles, it is okay, but in general we should wait for rank 0 to complete - i.e wait until the inference engine has updated the weights on all ranks - until we run ipc_collect on all ranks

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_weight_sync.py
```diff
@@ -344,3 +347,206 @@ async def test_update_weights_flow(self, weight_update_env):
...
+        return {
+            "names": names,
+            "dtype_names": dtype_names,
+            "shapes": shapes,
+            "ipc_handles_pickled": pickled,
+        }
+
+
+@pytest_asyncio.fixture(scope="class")
+async def ipc_weight_update_env(ray_init_fixture):
```
SumanthRH, line 333: Is there a way we can re-use the utiility class `InferenecEngineState` here?

https://github.com/NovaSky-AI/SkyRL/blob/05ae3d016aea8c4f15d5d4defe084bd4e2c07b08/tests/backends/skyrl_train/gpu/utils.py#L376

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_weight_sync.py
```diff
@@ -344,3 +347,206 @@ async def test_update_weights_flow(self, weight_update_env):
...
+    router_port = get_open_port()
+    router = InferenceRouter(server_urls, host="0.0.0.0", port=router_port)
+    router_url = router.start()
+    assert wait_for_url(router_url), "Router failed to start"
+
+    client = RemoteInferenceClient(
+        proxy_url=router_url,
+        server_urls=server_urls,
+        model_name=MODEL,
+    )
```
SumanthRH: This code seems common to both `ipc_weight_update_env` and `weight_update_env` and can be put in a helper function in this test

tests/backends/skyrl_train/gpu/gpu_ci/test_policy_local_engines_e2e.py
SumanthRH, line 1: Can you run this with the old and the new inference codepath and make sure tests pass?

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
hao-aaron: it was just useful for me debugging but i can remove it if we don't want it

skyrl/backends/skyrl_train/weight_sync/vllm_ipc_engine.py
hao-aaron: yes, the weight transfer config only accepts "nccl" string, will be changed to accept any arbitrary string

pyproject.toml
hao-aaron: this is needed to register the custom engine, even once the config issue is fixed, so it won't change in 0.17

skyrl/backends/skyrl_train/inference_servers/vllm_server_actor.py
hao-aaron: removing

skyrl/backends/skyrl_train/weight_sync/cuda_ipc_strategy.py
hao-aaron: removed the guard, we shouldn't ever have none and if we for some reason we do it will error

pyproject.toml
SumanthRH: Understood. I mistook this for the `patch`, which is called inside the `register_ipc_engine` function, which is the function to be removed later?

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_weight_sync.py
SumanthRH: This can get awkward given the custom vllm cli args

hao-aaron: refactored the test to use it

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_weight_sync.py
hao-aaron: cleaned up by using InferenceEngineState

tests/backends/skyrl_train/gpu/gpu_ci/test_policy_local_engines_e2e.py
hao-aaron: ran both old and new and passing.

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
SumanthRH: Can you remove it?

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_weight_sync.py
SumanthRH: Looks like you've not pushed the updates yet?

PR #1296, opened 2026-03-09
[BREAKING][skyrl-train] Implement loss reduction via advantage normalization and fix `token_mean` reduction strategy

justinvyu:
**This is a breaking change for the default `token_mean` loss behavior, as well as observed `policy_loss` metrics!** See the "Differences in reported loss metric magnitudes" section below.

## Summary

* Change `reduce_loss()` to always returns a simple masked sum (`(loss * mask).sum()`). To achieve different reduction strategies, we pre-scale the advantages before they enter the loss function, which also aligns with how Tinker's API handles it.
  * Scales the loss by the DP size before calling `backward()` to counteract the default data parallel mean gradient all-reduce across workers to do a sum instead.
* Fixes the `token_mean` loss reduction method to take a mean across all tokens in the minibatch rather than averaging across microbatches. Allows running with the old loss reduction with the `token_mean_legacy` config.

## Loss reduction strategies 
                                                                                                                                                                                           
* **Option 1: token_mean**
  * Average loss per token across the entire mini-batch.                                                                                                                          
  * **This is the fixed version where the denominator is the total token count across the full mini-batch, so the gradient is independent of how the minibbatch is split into micro-batches.**
                                                                                                                                                                                           
* **Option 1b: token_mean_legacy**
  * Compute token-mean loss within each micro-batch, then average across micro-batches.                                                                                                                                                                                                                                       
  * This reproduces the `token_mean` behavior before this PR.
  * **The problem: if micro-batches have different token counts, the effective weighting differs from a true global token mean. This is also less usable since changing micro batch size affects the loss and the training dynamics.**
  * Kept as a fallback in case of performance regressions — we should remove this down the line.
                                                
* **Option 2: sequence_mean**
  * Compute per-token loss within each sequence, average across sequences.
  * This is unchanged and is just implemented via advantage normalization instead.
                                                                                                                                                                                           
* **Option 3: seq_mean_token_sum_norm**
  * Dr. GRPO style — normalize by a fixed constant to avoid any length-dependent weighting.                                                                                       
  * This is unchanged and is just implemented via advantage normalization instead.

## Mean all-reduce -> sum all-reduce

We need the loss to be summed across microbatches and data parallel workers:
* DDP/FSDP defaults to a mean all-reduce for gradients across workers. This PR counteracts this by multiplying by the DP world size in order to keep the loss sum across data parallel groups.
* Megatron also does a similar mean reduction across microbatches and workers, so we counteract this by multiplying by num microbatches and DP size to achieve the sum.

## Difference in reported loss metric magnitudes

You will observe that the loss metric reported has a different magnitude compared to your older runs. This is beacuse the old token_mean implementation was somewhere between a true token mean and a sequence mean due to per-micro-batch normalization (ex: micro_batch_size=1 was equivalent to sequence mean).

The new `token_mean` is a proper minibatch token mean, while `sequence_mean` weights every sequence equally regardless of length. When comparing the loss produced by different reduction methods computed on the same advantages, from a real run:

```
  token_mean (new):  0.322   — every token weighted equally across the mini-batch
  token_mean (old):  0.065   — mean of per-micro-batch token means, where micro_batch_size=4
  sequence_mean:     0.00098 — every sequence weighted equally
```
                                                                                    
The old token_mean gave each micro-batch equal weight rather than each token, so its scale depended on how advantages were distributed across micro-batches. The new implementation is invariant to micro-batch size. 

Note that `token_mean_legacy` reports the old metrics still, and the `sequence_mean` and `seq_mean_token_sum_norm` modes also match exactly. See this [comment](https://github.com/NovaSky-AI/SkyRL/pull/1296#issuecomment-4145207966) for more details.

## Tinker compatibility

Here was the first attempt at fixing the loss reduction across microbatches: https://github.com/NovaSky-AI/SkyRL/pull/909

This method was to track total tokens and then do one big normalization at the `optim_step` in order to get an average per-token loss. But, we decided to align with Tinker's way of just summing up the loss at the end, and pushing any loss normalization to the user's advantage calculation.

The benefit is that users have full control of customizing their loss reduction strategy, rather than having it happen in our opaque `forward_backward`, `optim_step` implementation which would require some configuration argument that diverges from tinker's API. For example, we would need to add a config somewhere to determine how to average/sum the loss:

```python
client.forward_backward(...)
client.optim_step(..., loss_reduction="token_mean")  # no longer tinker compatible
```

The current PR aligns with [Tinker semantics](https://tinker-docs.thinkingmachines.ai/losses#proximal-policy-optimization-ppo):

> Notice that for all objectives we sum the token-level losses over the sequence length unlike some other loss implementations. If you would like to explore different aggregation schemes, you can include that in the advantage tensor computation.

Example for `loss_reduction="token_mean"`:
* Move the `1/num_minibatch_tokens` normalization into the advantage: `loss = sum( -advantage_i * ratio_i for i in range(num_minibatch_tokens) ) / num_minibatch_tokens`
* -> `sum( -(advantage_i / num_minibatch_tokens) * ratio_i for i in range(num_minibatch_tokens) )`

## Learning curve comparisons before/after the PR

### FSDP ([wandb](https://wandb.ai/justinvyu/dapo_aime/reports/FSDP-experiments--VmlldzoxNjIyNDExMQ))

<img width="817" height="664" alt="Screenshot 2026-03-20 at 3 29 10 PM" src="https://github.com/user-attachments/assets/be3ac108-b5b6-4408-a270-884b7fd4bce3" />

### Megatron ([wandb](https://wandb.ai/justinvyu/dapo_aime/reports/Megatron-loss-reduction--VmlldzoxNjIyNDE2Ng))

#### 1.7B:
<img width="1340" height="637" alt="Screenshot 2026-03-20 at 3 29 40 PM" src="https://github.com/user-attachments/assets/632d0c21-5b9c-48f3-aba9-d65ed00b13df" />

#### 30B lora:

master baseline:
<img width="1629" height="698" alt="Screenshot 2026-03-20 at 3 33 29 PM" src="https://github.com/user-attachments/assets/54c07117-9e40-4551-91c0-78ae5f25e347" />

`token_mean_legacy` + fixed `token_mean`:

<img width="1149" height="660" alt="Screenshot 2026-03-24 at 11 16 56 AM" src="https://github.com/user-attachments/assets/ef91b399-5f15-4db2-81ea-1e14c5932406" />


<!-- devin-review-badge-begin -->

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1296" target="_blank">
  <picture>
    <img src="https://static.devin.ai/assets/gh-open-in-devin-review-light.svg?v=1" alt="Open with Devin">
  </picture>
</a>
<!-- devin-review-badge-end -->

### DIFF
--- a/skyrl/backends/skyrl_train/workers/megatron/megatron_model_wrapper.py
+++ b/skyrl/backends/skyrl_train/workers/megatron/megatron_model_wrapper.py
@@ -318,15 +318,6 @@ class MegatronModelWrapper:
                 }
                 return loss, metrics
 
-            # Policy losses are pre-scaled to achieve the correct loss_reduction when summing across the entire minibatch
-            # (see `apply_loss_reduction_to_advantages_minibatch`).
-            # Megatron divides loss by num_microbatches
-            # (https://github.com/NVIDIA/Megatron-LM/blob/core_v0.15.2/megatron/core/pipeline_parallel/schedules.py#L248)
-            # and the data parallel all-reduce averages gradients across dp_size
-            # (https://github.com/NVIDIA/Megatron-LM/blob/core_v0.15.2/megatron/core/distributed/distributed_data_parallel.py#L285)
-            # so we multiply by both factors to recover the correct sum reduction.
-            grad_sum_correction_factor = num_microbatches * dp_size
-
             # RL path: add optional KL/entropy terms
             # entropy loss
             with torch.set_grad_enabled(loss_config.use_entropy_loss):
@@ -351,6 +342,17 @@ class MegatronModelWrapper:
                 kl_loss = torch.tensor(0.0)
             kl_loss_term = kl_loss * loss_config.kl_loss_coef
 
+            # Policy losses are pre-scaled to achieve the correct loss_reduction
+            # when summing across the entire minibatch (see `apply_loss_reduction_to_advantages_minibatch`).
+            # Megatron divides loss by num_microbatches
+            # (https://github.com/NVIDIA/Megatron-LM/blob/core_v0.15.2/megatron/core/pipeline_parallel/schedules.py#L248)
+            # and the data parallel all-reduce averages gradients across dp_size
+            # (https://github.com/NVIDIA/Megatron-LM/blob/core_v0.15.2/megatron/core/distributed/distributed_data_parallel.py#L285)
+            # so we multiply by both factors to recover the correct sum reduction.
+            grad_sum_correction_factor = num_microbatches * dp_size
+
+            # NOTE: The KL and entropy loss terms are not pre-scaled,
+            # so we just average them across microbatches and DP workers.
             loss = policy_loss * grad_sum_correction_factor + kl_loss_term - entropy_loss_term
             unscaled_loss = loss / grad_sum_correction_factor
 
--- a/skyrl/backends/skyrl_train/workers/worker.py
+++ b/skyrl/backends/skyrl_train/workers/worker.py
@@ -854,11 +854,6 @@ class PolicyWorkerBase(Worker):
                 "loss_fn_outputs": loss_fn_outputs,
             }
         else:
-            # DP all-reduce averages gradients, but policy losses are pre-scaled sums
-            # (see `apply_loss_reduction_to_advantages_minibatch`), so we multiply by
-            # dp_size to recover the correct sum reduction across workers.
-            grad_sum_correction_factor = self.mesh_rank.dp_size
-
             # RL path: add optional KL/entropy terms
             # entropy loss
             with torch.set_grad_enabled(self.cfg.algorithm.use_entropy_loss):
@@ -885,7 +880,13 @@ class PolicyWorkerBase(Worker):
                 kl_loss = torch.tensor(0.0)
             kl_loss_term = kl_loss * self.cfg.algorithm.kl_loss_coef
 
-            # TODO: The KL and entropy loss terms still need to be averaged across microbatches and workers.
+            # DP all-reduce averages gradients, but policy losses are pre-scaled sums
+            # (see `apply_loss_reduction_to_advantages_minibatch`), so we multiply by
+            # dp_size to recover the correct sum reduction across workers.
+            grad_sum_correction_factor = self.mesh_rank.dp_size
+
+            # NOTE: The KL and entropy loss terms are not pre-scaled,
+            # so we just average them across microbatches and DP workers.
             loss = policy_loss * grad_sum_correction_factor + (kl_loss_term - entropy_loss_term) * microbatch_weight
             unscaled_loss = loss / grad_sum_correction_factor
             self.strategy.backward(loss, self.model, self.optimizer)


skyrl/train/trainer.py
```diff
@@ -1037,6 +1033,65 @@ def apply_reward_kl_penalty(
...
+        loss_mask = data["loss_mask"]
+        response_mask = data["response_mask"]
+
+        # Step 1: Z-score normalization (if enabled)
+        if self.cfg.trainer.algorithm.advantage_batch_normalize:
+            num_actions = response_mask.sum()
+            mean = advantages.mean()
+            std = ((advantages - mean).pow(2) * response_mask).sum()
+            rstd = (std / num_actions).clamp(min=1e-8).rsqrt()
+            advantages = (advantages - mean) * rstd
```
justinvyu: note: this is a bit different from before -- we're taking a per-minibatch z-score, compared to before it's using the epoch-level mean/std

skyrl/train/trainer.py
```diff
@@ -1037,6 +1033,65 @@ def apply_reward_kl_penalty(
...
+            for i in range(num_micro_batches):
+                start_idx = i * micro_batch_size
+                end_idx = (i + 1) * micro_batch_size
+                microbatch_advantages = advantages[start_idx:end_idx]
+                microbatch_loss_mask = loss_mask[start_idx:end_idx]
+                # Compute token-mean within each microbatch
+                microbatch_advantages = microbatch_advantages / microbatch_loss_mask.sum().clamp(min=1)
+                # Average across microbatches
+                microbatch_advantages /= num_micro_batches
+                data["advantages"][start_idx:end_idx] = microbatch_advantages
```
justinvyu: baseline mean across minibatch token mean implementation

skyrl/train/trainer.py
justinvyu: fixed this to revert back to the old epoch-level z-score

skyrl/backends/skyrl_train/utils/ppo_utils.py
```diff
@@ -1040,30 +992,60 @@ def importance_sampling_loss(
...
+
+    # Option 1: token mean
     if loss_reduction == "token_mean":
-        # sum over *all* valid tokens, divide by total valid-token count
-        loss = masked_mean(loss, loss_mask)
+        normalized_advantages = advantages / loss_mask.sum().clamp(min=1)
+
+    # Option 1b: legacy token-mean that normalizes per-microbatch then averages across microbatches.
+    elif loss_reduction == "token_mean_legacy":
+        num_micro_batches = batch_size // micro_batch_size
```
erictang000, line 1025: oh wow this is pretty clever!

after staring at it for a bit I believe that it's right now, but just want to confirm my understanding

i thought it was wrong b/c it wasn't taking dp size into account when calculating `num_micro_batches`, but this implicitly works by normalizing within each `micro_batch_per_gpu` (which is what happened with the token mean before), and then additionally dividing by `num_micro_batches`, and since

`num_micro_batches = num_accumulation_steps * micro_batch_size_per_gpu * dp_size`

this is equivalent to before since `num_accumulation_steps * micro_batch_size_per_gpu` is like the old `1/num_accumulation_steps` term, and `micro_batch_size_per_gpu` is making it take a mean. And then we additionally scale by `dp_size` now so that term is canceled too

skyrl/backends/skyrl_train/workers/megatron/megatron_model_wrapper.py
```diff
@@ -356,8 +368,8 @@ def loss_func(logits, data):
                 )
 
             metrics = {
-                "final_loss": loss.detach().item(),
-                "policy_loss": policy_loss.detach().item(),
+                "final_loss": unscaled_loss.detach().item() * dp_size,
+                "policy_loss": policy_loss.detach().item() * dp_size,
```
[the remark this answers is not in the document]
erictang000: @justinvyu i think i remember i made these changes to get these metrics matching and to be invariant to dp size... but just want to check - were the metric scales roughly similar for FSDP vs megatron on your runs?

for the 1.7b runs could you paste these metrics for megatron vs fsdp?

skyrl/backends/skyrl_train/workers/megatron/megatron_model_wrapper.py
```diff
@@ -368,8 +368,8 @@ def loss_func(logits, data):
                 )
 
             metrics = {
-                "final_loss": unscaled_loss.detach().item() * dp_size,
-                "policy_loss": policy_loss.detach().item() * dp_size,
+                "final_loss": unscaled_loss.detach().item(),
+                "policy_loss": policy_loss.detach().item(),
```
justinvyu, line 381: Metrics fix 1: remove `dp_size` multiplier in reported metrics, since there's no average that we need to correct for, since `reduce_microbatch_metrics` and `all_reduce_metrics` both do sums for `*_loss` metrics.

skyrl/train/trainer.py
```diff
@@ -1129,7 +1129,7 @@ def _execute_training_step(self, model: str, data: TrainingInputBatch) -> Dict[s
         # Reduce metrics across all mini-batches and epochs
         # pop out loss_fn_outputs since it's not a scalar metric and to avoid logging it
         all_metrics.pop("loss_fn_outputs", None)
-        reduced_metrics = reduce_metrics(all_metrics)
+        reduced_metrics = reduce_metrics_across_minibatches(all_metrics)
```
justinvyu: Metrics fix 2: Take an average across minibatches instead of still summing. This is because the loss reduction normalization happens at the minibatch level. Across different minibatches we should just average, otherwise we'll increase the reported loss scale by ~`num_minibatches`

skyrl/backends/skyrl_train/workers/worker.py
```diff
@@ -1079,7 +1070,7 @@ def forward_backward(self, data: TrainingInputBatch) -> Dict[str, float]:
             for k, v in metrics.items():
                 all_metrics[k].append(v)
 
-        return reduce_metrics(dict(all_metrics))
+        return reduce_metrics_across_microbatches(dict(all_metrics))
```
justinvyu: critic codepath may need to be reverted since it doesn't use the advantages?

justinvyu: reverted critic codepath to do mean reductions for metrics

skyrl/backends/skyrl_train/workers/megatron/megatron_model_wrapper.py
```diff
@@ -279,13 +290,15 @@ def loss_func(logits, data):
 
             # SFT path: cross_entropy loss (negative log likelihood)
             if resolved_loss_name == "cross_entropy":
-                loss = policy_loss
+                unscaled_loss = policy_loss
+                loss = unscaled_loss * grad_sum_correction_factor
```
justinvyu: Q: should this affect the SFT case? SFT doesn't look at the normalized advantages either, similar to the critic loss case.

Before the PR, the SFT case does a sum across the negative log likelihoods within a microbatch, but still averaged over microbatches and dp workers.
Now, we are summing negative log likelihood across the entire minibatch. What's the desired behavior here?

justinvyu:
To sanity check the difference in loss metric magnitudes, I dumped the raw advantages on the first step and manually calculated the loss with the different reduction methods on the same dumped data.

Using dumped advantage tensors from a real GRPO run to compare old vs. new:

**With `micro_batch_size=4` (128 micro-batches), old and new differ by ~5x** — matching what's observed in the actual run:
```
Average: old=0.065  new=0.322  ratio=4.92
```

**With `micro_batch_size=1`, the old `token_mean` reduces to `sequence_mean`** (each sequence weighted equally). The old values match `sequence_mean` exactly:
```
token_mean old:  Average=-0.024
sequence_mean:   Average=-0.024  ratio=1.0000
```

**With `micro_batch_size=512` (1 micro-batch = full mini-batch), old and new converge:**
```
Average: old=0.322  new=0.322  ratio=1.0000
```

The new `token_mean` value (`0.322`) is the same regardless of `micro_batch_size` — which is the correct behavior. The old value varied between `-0.024` (at `micro_batch_size=1`, i.e. `sequence_mean`) and `0.322` (at `micro_batch_size=mini_batch_size=512`, i.e. fixed `token_mean`) depending on how micro-batches were formed.

`token_mean_legacy` reproduces the old behavior. Runs using `token_mean` won't be directly comparable to before, but the difference is analogous to comparing `token_mean` vs. `sequence_mean` — a different weighting, not a bug.

skyrl/backends/skyrl_train/workers/megatron/megatron_model_wrapper.py
justinvyu: Reverting to the old behavior for now and we can tackle it in a follow-up. SFT loss reduction is broken already due to taking a sum within the microbatch but then a mean across microbatches/workers. The current behavior does not align with this comment: https://github.com/justinvyu/SkyRL/blob/c5feb83b38f4635c7fc705c2bb192a7d6ad16947/skyrl/backends/skyrl_train/utils/ppo_utils.py#L917

skyrl/backends/skyrl_train/workers/worker.py
```diff
@@ -714,7 +715,16 @@ def forward_backward(
             for k, v in metrics.items():
                 all_metrics[k].append(v)
 
-        result = reduce_metrics(dict(all_metrics))
+        # TODO: SFT path still averages metrics across microbatches and workers.
+        # This needs to be unified with the RL path which sums.
+        resolved_loss_name = loss_fn or self.cfg.algorithm.policy_loss_type
+        sum_loss_metrics = resolved_loss_name != "cross_entropy"
```
justinvyu, line 721: need to followup to unify the codepaths -- SFT loss is wrong right now

skyrl/backends/skyrl_train/workers/worker.py
```diff
@@ -869,7 +880,15 @@ def _forward_backward_micro(
...
-            loss = policy_loss + kl_loss_term - entropy_loss_term
+            # DP all-reduce averages gradients, but policy losses are pre-scaled sums
+            # (see `apply_loss_reduction_to_advantages_minibatch`), so we multiply by
+            # dp_size to recover the correct sum reduction across workers.
+            grad_sum_correction_factor = self.mesh_rank.dp_size
+
+            # NOTE: The KL and entropy loss terms are not pre-scaled,
+            # so we just average them across microbatches and DP workers.
+            loss = policy_loss * grad_sum_correction_factor + (kl_loss_term - entropy_loss_term) * microbatch_weight
+            unscaled_loss = loss / grad_sum_correction_factor
```
justinvyu, line 891: This part is a bit complicated to maintain kl/entropy loss parity:
* Previously, the kl/entropy terms are per-token averages within the microbatch (see the `masked_mean` above). Then, we took the average across microbatches and DP workers (same as the old loss).
* We can't just sum them because they were computed on the worker and we didn't pre-scale them in the same way we scaled the advantages. 
* So, to maintain the average behavior, we divide the terms by the microbatch weight (`1/num_microbatches`), and we don't apply the grad sum correction factor to keep the all-reduce as a mean across DP workers.

skyrl/backends/skyrl_train/workers/megatron/megatron_model_wrapper.py
```diff
@@ -340,7 +342,19 @@ def loss_func(logits, data):
...
+            # (https://github.com/NVIDIA/Megatron-LM/blob/core_v0.15.2/megatron/core/pipeline_parallel/schedules.py#L248)
+            # and the data parallel all-reduce averages gradients across dp_size
+            # (https://github.com/NVIDIA/Megatron-LM/blob/core_v0.15.2/megatron/core/distributed/distributed_data_parallel.py#L285)
+            # so we multiply by both factors to recover the correct sum reduction.
+            grad_sum_correction_factor = num_microbatches * dp_size
+
+            # NOTE: The KL and entropy loss terms are not pre-scaled,
+            # so we just average them across microbatches and DP workers.
+            loss = policy_loss * grad_sum_correction_factor + kl_loss_term - entropy_loss_term
+            unscaled_loss = loss / grad_sum_correction_factor
```
justinvyu, line 357: This is similar to the FSDP case, except megatron already divides by num_microbatches and dp_size internally (so no need to divide by num_microbatches here).

Issue #1299, opened 2026-03-09
[SkyRL][inference] Duplicated port assignment when one or more ports are assigned


nithinvc:
When two or more `VLLMServerActor` instances are created concurrently (via `ServerGroup`), they can silently be assigned the same port due to a race condition in port allocation. This causes all HTTP control-plane calls (weight sync init, weight updates, pause/resume) to reach only one server, leaving the other server's NCCL communicator uninitialized. The subsequent NCCL broadcast then hangs indefinitely. This only manifests in the tinker service since port 8000 is allocated by tinker, causing the first inference server to use port 8001 instead of 8000.

In the python entrypoint + pytests: two inference servers are allocated to ports 8000 and 8001.
In tinker, since port 8000 is occupied, inference server 1 tries 8000 and sees an allocated port then takes 8001. Inference server 2 is already assigned to 8001, causing port collision (`get_open_port`). 

To reproduce:
Allocate port 8000 then run the python entrypoint with the new inference server `_SKYRL_USE_NEW_INFERENCE=1`
```
nc -lk 10.0.26.4 8000
```
```
_SKYRL_USE_NEW_INFERENCE=1 bash examples/train/gsm8k/run_gsm8k.sh trainer.placement.colocate_all=false 
```

SumanthRH:
Closing this after #1302

PR #1302, opened 2026-03-09
[skyrl][inference] Fix port collision when ports are allocated.

nithinvc:
### Summary
Addresses issue #1299 

  - Replace `get_open_port()` (test-then-release) with `find_and_reserve_port()` which holds the socket open via `SO_REUSEADDR` to prevent port race. condition between concurrent `VLLMServerActor` instances when one or more ports have already been reserved (e.g., in tinker).
  - Add unit tests for port reservation behavior

  Test plan
  - Unit tests for `find_and_reserve_port` (concurrent allocation, sequential search)

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1302" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/skyrl/backends/skyrl_train/inference_servers/vllm_server_actor.py
+++ b/skyrl/backends/skyrl_train/inference_servers/vllm_server_actor.py
@@ -26,16 +26,14 @@ from vllm.utils.system_utils import set_ulimit
 
 from skyrl.backends.skyrl_train.inference_servers.common import (
     ServerInfo,
+    find_and_reserve_port,
     get_node_ip,
-    get_open_port,
 )
 from skyrl.backends.skyrl_train.inference_servers.protocols import ServerActorProtocol
 from skyrl.env_vars import (
     SKYRL_VLLM_DP_PORT_OFFSET,
     SKYRL_WAIT_UNTIL_INFERENCE_SERVER_HEALTHY_TIMEOUT_S,
 )
-from skyrl.backends.skyrl_train.inference_servers.common import ServerInfo, find_and_reserve_port, get_node_ip
-from skyrl.backends.skyrl_train.inference_servers.protocols import ServerActorProtocol
 
 logger = logging.getLogger(__name__)
 


skyrl/backends/skyrl_train/inference_servers/vllm_server_actor.py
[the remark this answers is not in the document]
nithinvc: I think this is okay? `_run_server()` should always be called

skyrl/backends/skyrl_train/inference_servers/common.py
```diff
@@ -72,3 +73,29 @@ def get_open_port(start_port: int | None = None) -> int:
     with socket.socket(socket.AF_INET6, socket.SOCK_STREAM) as s:
         s.bind(("", 0))
         return s.getsockname()[1]
+
+
+def find_and_reserve_port(start_port: int) -> Tuple[int, socket.socket]:
```
kouroshHakha, line 78: isn't this already what _get_port from vllm does? 
https://github.com/vllm-project/vllm/blob/f85b4eda3a22fedd885ef31650c825d56867587e/vllm/utils/network_utils.py#L190

nithinvc: I think it's slightly different since they're closing the socket once the context manager for the socket exits, while the proposed method doesn't close the socket till all ports are assigned and `start` is called on the server pool. If one of the `VLLMServerActors` isn't holding the socket while the other searches for ports (likely since the context manger exits once a port is found), the other actor will take the same port.

For the multi-replica case you've tested, was it the [launch_multiple_remote_servers.py](https://github.com/NovaSky-AI/SkyRL/blob/main/examples/train_scripts/launch_multiple_remote_servers.py) script? I think the reason this script doesn't have issues is because the port selection [here](https://github.com/NovaSky-AI/SkyRL/blob/main/examples/train_scripts/launch_multiple_remote_servers.py#L32) isn't pre-determined so it's unlikely to collide. 

This is also specific to the case where port 8000 (or the starting port) is occupied. I think usually the `self._start_port + server_idx` [here](https://github.com/NovaSky-AI/SkyRL/blob/main/skyrl/backends/skyrl_train/inference_servers/server_group.py#L141) avoids this problem since ports 8000 => 8000 + num servers will be free and each actor gets the port on first try.

PR #1313, opened 2026-03-11
WIP Qwen 3.5 Dense FSDP

CharlieFRuan:


benrhodes26:
@CharlieFRuan : I am curious if there is a blocker here, or it's just low priority? I ask because I am very keen to use Qwen3.5 9B through the Tinker API :)

PR #1320, opened 2026-03-13
[train][fix] Fix concurrency limitations in the new inference codepath

### LINKED ISSUE #1307: [train] Training with the new inference codepath hangs
# Summary

Training with the new inference codepath hangs right now. Tested with GSM8K - training hangs during the generation phase

## Steps to reproduce

Use the following env file:

```bash
# .env.example
_SKYRL_USE_NEW_INFERENCE=1
```

and edit the GSM8K training script:

```diff
+ uv run --isolated --extra fsdp --env-file .env.example -m skyrl.train.entrypoints.main_base \
- uv run --isolated --extra fsdp -m skyrl.train.entrypoints.main_base \
```

And run training :

```bash
bash examples/train/gsm8k/run_gsm8k.sh
```

SumanthRH:
# What does this PR do?

Fixes #1307 . 

This PR fixes concurrency limitations in the new inference codepath. 

## Summary

Training with the GSM8K script at `examples/train/gsm8k/run_gsm8k.sh` hangs with the new inference codepath. To investigate the issue, I ran some load tests to concurrently fire a bunch of requests at different scales (100 -> 10k). Note that for the configuration in `examples/train/gsm8k/run_gsm8k.sh`, we actually end up running generate requests at the concurrency of ~ 5K during training, so this is a reasonable test. At scales of 50K, you'll start hitting OS limits of open ports, so the test sweeps concurrencies from 500-10,000. 

There are three parts in the new inference stack:

```
RemoteInferenceClient  -> InferenceEngineRouter -> vLLM Server
```

The load test tests concurrency limitations ablating on the different components. 

### Issues with`Router + vLLM Server` 

`httpcore.ReadError`: Transient errors fixed by adding in retries. 

### Issues with`RemoteInferenceClient + Router + vLLM Server`

Firstly, we use a single shared session for re-use but with the default connection limit of 100 for the `aiohttp.TCPConnector`. This will make generation extremely slow. Other pending tasks in the generation loop wait for a long time. This was fixed by raising the connection limit to 50K (very generous). 

Now, the load test script progressed faster, but there were failures even at low concurrency (~ 500). There were mainly two errors I saw:

1. `Connection reset by peer` error: There were two causes for this:
a. Closing stale connections from a previous event loop
a. Increasing keep alive timeouts on the `aiohttp` client
2. `ReadError` for different requests: These are transient errors fixed by adding in retries 

These fixes were enough to get the load test script to run without issues. I also added retries on the proxy to ensure that transient failures do not affect the router layer. 

## E2E testing
After load testing script ran successfully, I revisted the GSM8k training script again. I noticed that training now progressed but failed with a large number of connection errors (basically hitting 3/3 retries for a number of generation samples) after a 4 steps of training. I finally solved the issue by introducing a semaphore to guard against overwhelming the http server with too many concurrent requests. This solved the issue and training ran to completion for the GSM8K script. 

Screenshot 2026-03-12 at 11 54 33 PM

## Summary of Changes
- Introduces a new load testing script to load test concurrency at different layers of the new HTTP stack
- Introduces retries at the router and the `RemoteInferenceClient` layer to handle transient read / connection errors
- Removes stale connections from previous event loop in `RemoteInferenceClient`
- Raises the connection limit for the aiohttp session in `RemoteInferenceClient`
- Adds a concurrency limit per engine to avoid overwhelming the http server with too many concurrent requests. 

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1320" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


skyrl/benchmarks/load_test_concurrency.py
```diff
@@ -0,0 +1,363 @@
+#!/usr/bin/env python3
```
kouroshHakha, line 1: should we add this as part of gpu nightly runs?

kouroshHakha (commented):
LG, Just one q

skyrl/benchmarks/load_test_concurrency.py
SumanthRH: Yes this should be done! I would prefer to leave a TODO for now. The primary reason is that the new inference codepath will have more improvements soon and for the specific load test I think we should add it to CI once it is more feature complete (in terms of using speeding up the client with multiprocessing, etc).

kouroshHakha (approved):

PR #1325, opened 2026-03-14
[transformers] set return dict false for transformers v5 compatibility

erictang000:
# Overview
This PR makes 2 changes for transformers-v5 compatibility:
- Sets `return_dict=False` where needed for transformers-v5 compatibility - this can be merged prior to explicitly upgrading to transformers v5 in the pyproject.toml, since vllm still does not technically fully support it in the latest release. (This change should be backwards compatible with transformers v4.*, since the behavior prior to v5 was that the default value for `return_dict` was `None`, which was interpreted as False.)
- check if fsdp_transformer_layer_cls_to_wrap is a set for v5 compatibility while maintaining backwards compatibility

huggingface/transformers breaking PR `return_dict=false` PR for v5: https://github.com/huggingface/transformers/pull/42567

## Validation
Checked all CPU tests are passing both with transformers < 5.0.0 and for transformers==5.3.0, and checked that `examples/train/gsm8k/run_gsm8k` runs for both old/new transformers.

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1325" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/tests/backends/skyrl_train/gpu/gpu_ci/test_pause_and_continue_generation.py
+++ b/tests/backends/skyrl_train/gpu/gpu_ci/test_pause_and_continue_generation.py
@@ -225,7 +225,9 @@ def test_continue_generation_generate_vllm_engine_generation(ray_init_fixture):
         # 2. Prepare a single ConversationType prompt; each generate() call will be single-request
         messages: List[ConversationType] = get_test_prompts(MODEL, num_samples=1)[0]
         # Convert to prompt_token_ids to work with both InferenceEngineClient and RemoteInferenceClient
-        prompt_token_ids = tokenizer.apply_chat_template(messages, add_generation_prompt=True)
+        prompt_token_ids = tokenizer.apply_chat_template(
+            messages, add_generation_prompt=True, return_dict=False, tokenize=True
+        )
 
         # 3. Fire 6 concurrent client.generate() single-request calls, then pause/resume mid-flight
         async def run_requests_then_pause():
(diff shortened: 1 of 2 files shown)


tests/train/generators/test_utils.py
```diff
@@ -169,10 +169,15 @@ def tokenizer_w_dummy_template():
 )
 def test_encode_messages(messages, tokenizer_w_dummy_template):
     # For a simple chat template, the fixed base approach is expected to behave the same
-    # as `apply_chat_template`
-    expected_token_ids = tokenizer_w_dummy_template.apply_chat_template(messages)
+    # as `apply_chat_template`.  We compare decoded strings rather than raw token IDs
```
erictang000, line 172: this comparison between full chat template vs stripping base only happens in this test and not in the SkyRLGymGenerator

CharlieFRuan: is it broken by v5? I wonder why it doesn't surface until now

CharlieFRuan (approved):
Thank you!

tests/train/generators/test_utils.py
erictang000: yeah i asked claude it told me this was the relevant PR: https://github.com/huggingface/transformers/pull/40936 and that in that PR

```
LlamaTokenizer was rewritten to inherit from TokenizersBackend (Rust) instead of the old Python SentencePiece backend
legacy was changed from defaulting to True to defaulting to False
The _get_prepend_scheme helper was added to select "first" vs "always" based on the legacy flag
```

PR #1329, opened 2026-03-16
[train] Add support for LoRA in the new inference codepath

SumanthRH:
# What does this PR do?

Adds LoRA support in the new inference codepath.

## Summary
- Adds a new `/v1/load_lora_adapter` endpoint to `VLLMServerActor` to support loading lora weights in the vllm server
- Adds support for lora weights in `RemoteInferenceClient` with a new method `load_lora_from_disk`. The caller is expected to provide a `lora_path` argument pointing to the lora weights on disk.
- Improved concurrency in `generate` : While running E2E tests for lora in the new codepath, I noticed that the generation speeds where 10x worse in the new codepath compared to the old `inference_engines/` codepath in SkyRL. The root cause is that the currently we use a single semaphore for the generate + detokenize stage. This led to detokenize stage starving GPUs when max concurrency was reached. the solution is to use separate semaphores so that generation can proceed independently. Generation speed is now at parity with old codepath
- Adds lora weight sync test to CI 

## Test Plan
- [x] Unit tests:  `_SKYRL_USE_NEW_INFERENCE=1 uv run --isolated --extra dev --extra fsdp pytest -s tests/backends/skyrl_train/gpu/gpu_ci/test_lora.py` 
- [x]  E2E lora test

## Results

I ran the lora example `examples/train/lora/run_qwen2_5_0.5b_gsm8k_grpo_lora.sh` with  `_SKYRL_USE_NEW_INFERENCE=1` and compared it with the baseline

Some curves are below: 

Screenshot 2026-03-18 at 11 17 10 PM

Screenshot 2026-03-18 at 11 17 23 PM

image

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1329" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


skyrl/backends/skyrl_train/workers/fsdp/fsdp_worker.py
SumanthRH, line 296: Previously, we were not awaiting on the `cache_reset_task` for LoRA

hao-aaron (approved):
LGTM, but have a design question. Do we want lora weight syncing to be on the same codepath as normal weight syncing instead of a separate function? Thinking about the future lora weight sync api for vllm, is that also something we want to adopt?

SumanthRH:
Long term, we want `update_named_weights` to be the single interface for the trainer that supports in-memory weight sync for lora and full param. 

Actually, currently the caller will directly use `load_lora_adapter`, so the changes in `update_named_weights` are not needed. Let me get rid of that.

SumanthRH:
Manual test: I've run the GPU CI test `test_lora.py` again after the new changes. tests pass.

Issue #1200, opened 2026-02-24
[skyrl][RFC] VLM Training in SkyRL with the SkyRL train backend


nithinvc:
# VLM Training in SkyRL

This document outlines the changes required for SkyRL/skyrl to support vision inputs, similar to the official tinker implementation. The proposed changes are only applicable to the pytorch / skyrl-train backend. The jax backend is out of scope.

_Test Case_: The supervised vision language training example ([here](https://github.com/thinking-machines-lab/tinker-cookbook/tree/main/tinker_cookbook/recipes/vlm_classifier)) should run and converge. The VLM SFT example should work out of the box. The task is to classify CalTech101, a very standard image object recognition benchmark (expected accuracy >90%, especially post-SFT).

An example implementation is on this [branch](https://github.com/nithinvc/SkyRL/pull/2/changes).

## Control Flow Overview
The largest change is passing the list of `ModelInputChunks` to the SkyRL-train backend. This is to reconcile the fact that pytorch / huggingface expects sequences with image token padding while VLLM inserts image padding within the engine. The proposed workaround is to pass the `ImageChunk`s to SkyRL-train which decides, based on the method call, whether to add image pad tokens or not.

<img width="3469" height="1387" alt="Image" src="https://github.com/user-attachments/assets/e0cf06d4-1237-4def-9429-9fc7acd6379e" />

## Tokenization Responsibilities
The client handles tokenization of text and encoding images into bytes. The server then has two code paths: 
1) Sampling through vllm and 
2) Training with huggingface / fsdp2.

VLLM: skyrl-train will be responsible for converting the bytes64 into a pillow image. This will then be passed through a VLLM [renderer](https://docs.vllm.ai/en/v0.10.2/api/vllm/entrypoints/renderer.html).
 
HF / FSDP2 Training: Because the datum is a list of encoded chunks which need to get concatenated, the skyrl-train backend should:
1. Patchify the image
2. Compute the expected number of tokens (validate if provided by client)
3. Concatenate input chunks to create the model input

## Tasks Breakdown
There’s a lot of moving pieces to get to a full VLM SFT implementation. Instead, here are a set of incremental changes which should preserve backwards compatibility and interoperability with the Jax backend.

**1. Inference Engine Piping**
Changes localized around the `sample` workflow. We will only support the new inference stack (`inference_engine_client_http_endpoint.py`).
- [x] Inference engine client call for `/v1/chat/completion/render` + inclusion of multi-modal data.
[...]

PR #1331, opened 2026-03-17
[train] Multi-modal inputs support in FSDP2

nithinvc:
This PR adds support for variable-shape vision inputs (e.g. `pixel_values`, `image_grid_thw`) to the FSDP2 training backend, item 2 in #1200 .

### Summary

- **`TensorList`**: New container in `training_batch.py` for lists of tensors with variable shapes per batch element (e.g., each image may produce a different number of patches). Supports the same batch operations as regular tensors: slicing, chunking, concatenation, repeat, device transfer, and pickle serialization.
- **`TensorBatch` extended**: `TensorBatch` now accepts `TensorList` fields alongside regular tensors. All operations (chunk, slice, cat, repeat, pickle, equality, `__setitem__`) handle both types transparently.
- **`TrainingInput` schema and `Experience` dataclass**: Added optional `pixel_values` and `image_grid_thw` fields.
- **`BatchIterator.batch_to_experience`**: Propagates vision fields from `TrainingInputBatch` to `Experience`.
- **`HFModelWrapper.forward`**: Accepts `pixel_values` and `image_grid_thw` kwargs, converts `TensorList` to concatenated tensors for the HF model, and routes through a VLM-specific forward path (no position IDs, passes `pixel_values`/`image_grid_thw` directly).
- **`HFModelWrapper.__init__`**: Auto-detects VLMs via `AutoModelForCausalLM` config mapping; falls back to `AutoModelForVision2Seq` and sets `is_vlm=True`.
- **`PolicyWorkerBase`**: Both the training step and the initial policy log-prob computation now pass vision fields through to the model.

### Test plan

**CPU unit tests**:

- `test_train_batch.py` — `TensorList` unit tests + `TensorBatch` integration tests + bfloat16 pickle roundtrip.
- `test_vlm_data_plumbing.py` — Verifies `pixel_values`/`image_grid_thw` flow through `batch_to_experience` and `Experience.to_device`.

**GPU integration tests**:
New `test_vlm_model_wrapper.py` with the following tests:
- Tests VLM model loading (`is_vlm` flag).
- Forward pass with vision data and text-only fallback.
- Test vision affects output log probabilities.
- Verifies returned log-probs vs. manually computed log-probs correctness.
- Tests that different produce different log probs.
- Tests semantic color recognition. Given a solid color image, asserts the correct color has the highest log probability.
- Test batched forward pass and verifies batch permutation leads to the same log probs.

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1331" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/skyrl/backends/skyrl_train/training_batch.py
+++ b/skyrl/backends/skyrl_train/training_batch.py
@@ -309,7 +309,7 @@ class TensorBatch(dict, Generic[DictType]):
                 new_batch[key] = value.repeat(repeats)
             else:
                 assert isinstance(value, torch.Tensor), f"Field {key} must be a tensor, got {type(value)}"
-                new_batch[key] = value.repeat(repeats, *([1] * (value.dim() - 1)))
+                new_batch[key] = value.repeat(repeats)
         new_batch = self.__class__(new_batch)
         new_batch.metadata = self.metadata
         return new_batch
@@ -333,7 +333,7 @@ class TensorBatch(dict, Generic[DictType]):
                 new_batch[key] = value.repeat_interleave(repeats)
             else:
                 assert isinstance(value, torch.Tensor), f"Field {key} must be a tensor, got {type(value)}"
-                new_batch[key] = value.repeat_interleave(repeats, dim=0)
+                new_batch[key] = value.repeat_interleave(repeats)
         new_batch = self.__class__(new_batch)
         new_batch.metadata = self.metadata
         return new_batch
--- a/tests/backends/skyrl_train/test_train_batch.py
+++ b/tests/backends/skyrl_train/test_train_batch.py
@@ -405,7 +405,7 @@ def test_tensor_list_cat():
 
 def _make_mixed_batch(batch_size=4, seq_len=3):
     """Helper: create a TensorBatch with both Tensor and TensorList fields."""
-    sequences = torch.randn(batch_size, seq_len)
+    sequences = torch.randn(batch_size)
     pixel_values = TensorList([torch.randn(i + 1, 8) for i in range(batch_size)])
     return TensorBatch({"sequences": sequences, "pixel_values": pixel_values})
 


skyrl/backends/skyrl_train/training_batch.py
```diff
@@ -337,14 +410,17 @@ def __eq__(self, other: Any) -> bool:
...
-            if k not in other or not torch.equal(v, other[k]):
+            if k not in other:
+                return False
+            other_v = other[k]
+            if isinstance(v, TensorList) and not v == other_v:
+                return False
+            elif not torch.equal(v, other[k]):
                 return False
-        return True
+        return all(self._values_equal(v, other[k]) for k, v in self.items())
```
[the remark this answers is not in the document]
nithinvc: fixed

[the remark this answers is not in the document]
nithinvc: fixed

skyrl/backends/skyrl_train/training_batch.py
```diff
@@ -337,12 +410,17 @@ def __eq__(self, other: Any) -> bool:
...
+            if k not in other:
+                return False
+            other_v = other[k]
+            if isinstance(v, TensorList) and not v == other_v:
+                return False
+            elif isinstance(v, torch.Tensor) and isinstance(other_v, torch.Tensor) and not torch.equal(v, other_v):
+                return False
+            elif isinstance(v, torch.Tensor) and isinstance(other_v, TensorList):
                 return False
         return True
```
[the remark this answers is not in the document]
nithinvc: i think this is fair, there's no none checks right now and we allow none values.

skyrl/backends/skyrl_train/workers/model_wrapper.py
```diff
@@ -105,6 +113,16 @@ def __init__(
 
             model_config = AutoConfig.from_pretrained(pretrain_or_model, trust_remote_code=True, **model_config_kwargs)
 
+            # Fall back to AutoModelForVision2Seq for VLMs (e.g. Qwen3-VL)
```
pcmoritz: This seems very ad hoc (e.g. why are we always falling back to `AutoModelForVision2Seq` if the model config has no AutoModelForCausalLM mapping -- there might for example be other classes we need to fall back to), is there a better approach to make this work?

pcmoritz: Also the fact that `use_liger_kernel` and `type(model_config) not in AutoModelForCausalLM._model_mapping` determine if self.is_vlm is true or false seems off

skyrl/backends/skyrl_train/training_batch.py
```diff
@@ -337,12 +410,21 @@ def __eq__(self, other: Any) -> bool:
...
         if self.metadata != other.metadata:
             return False
-        if len(self) != len(other):
-            return False
-        if len(self.items()) != len(other.items()):
+        if set(self.keys()) != set(other.keys()):
             return False
         for k, v in self.items():
-            if k not in other or not torch.equal(v, other[k]):
+            if k not in other:
```
pcmoritz: We already tested this above with `if set(self.keys()) != set(other.keys()): return False`

skyrl/backends/skyrl_train/training_batch.py
```diff
@@ -337,12 +410,21 @@ def __eq__(self, other: Any) -> bool:
...
+                return False
+            other_v = other[k]
```
pcmoritz: You should check this, but it seems to me this can be simplified:

```python
if isinstance(v, torch.Tensor) and isinstance(other[k], torch.Tensor) and not torch.equal(v, other[k]):
    return False
if v != other[k]:
    return False
```

tests/backends/skyrl_train/gpu/gpu_ci/test_vlm_model_wrapper.py
```diff
@@ -0,0 +1,343 @@
...
+            num_actions,
+            inputs["attention_mask"],
+        )
+
+    assert action_log_probs.shape == (batch_size, num_actions)
+    assert torch.isfinite(action_log_probs).all()
+    assert (action_log_probs <= 0).all()
+
+
+def test_vlm_vision_affects_output(vlm_model, processor):
```
pcmoritz: I feel like `test_vlm_forward_with_vision_data`, `test_vlm_forward_text_only` and `test_vlm_vision_affects_output` can be consolidated into a single test, there is really no reason to have three separate ones that basically do the same thing.

skyrl/backends/skyrl_train/training_batch.py
```diff
@@ -337,12 +410,12 @@ def __eq__(self, other: Any) -> bool:
...
-            return False
-        if len(self.items()) != len(other.items()):
+        if set(self.keys()) != set(other.keys()):
             return False
         for k, v in self.items():
-            if k not in other or not torch.equal(v, other[k]):
+            if isinstance(v, torch.Tensor) and isinstance(other[k], torch.Tensor) and not torch.equal(v, other[k]):
+                return False
+            elif v != other[k]:
                 return False
```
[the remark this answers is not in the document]
nithinvc: Agree, added the nested if

skyrl/backends/skyrl_train/workers/model_wrapper.py
nithinvc: I removed the `use_liger_kernel` and `_model_mapping` check and instead use `"vision_config" in model_config` which should be a better indicator of a VLM. `AutoModelForImageTextToText` works for VLMs but in transformers 5.0.0+ all multi-modal models can use `AutoModelForMultimodalLM`.

Alternatively, we can fall back to the base `AutoModel` but that feels too broad.

skyrl/backends/skyrl_train/training_batch.py
nithinvc: Removed

skyrl/backends/skyrl_train/training_batch.py
nithinvc: I agree. I think the one other edge case is if `torch.equal(v, other[k])` is True, then the `v != other[k]` is run and can error. To fix this, I made the `torch.equal(v, other[k])` a nested if statement.

tests/backends/skyrl_train/gpu/gpu_ci/test_vlm_model_wrapper.py
nithinvc: That makes sense. All 3 (+ `test_vlm_different_images_diverge`) are actually redundant since `test_vlm_semantic_color_recognition` will test the forward pass logic and checks the vision inputs affects log probs. I updated the tests to only:
1. `test_vlm_log_probs_match_manual` - manual log prob calculation with the HF model
2. `test_vlm_semantic_color_recognition` - log prob calculation of different colors
3. `test_vlm_forward_batched_vision` - batched forward pass logic

skyrl/backends/skyrl_train/training_batch.py
```diff
@@ -214,6 +281,8 @@ def repeat(self, repeats: int):
         for key, value in self.items():
             if value is None:
                 new_batch[key] = value
+            elif isinstance(value, TensorList):
+                new_batch[key] = value.repeat(repeats)
             else:
                 assert isinstance(value, torch.Tensor), f"Field {key} must be a tensor, got {type(value)}"
                 new_batch[key] = value.repeat(repeats)
```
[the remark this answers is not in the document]
nithinvc: I think we should leave the existing repeat semantics, maybe change it in a future PR / issue

skyrl/backends/skyrl_train/workers/worker.py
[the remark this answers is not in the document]
nithinvc: this is a fair point, added the plumbing

skyrl/backends/skyrl_train/training_batch.py
```diff
@@ -11,6 +11,90 @@
...
+
+    Each element can have a different shape (e.g., pixel_values[i] has shape
+    [num_patches_i, dim]). Supports the same batch operations as tensors:
+    slicing, chunking, concatenation, device transfer, and serialization.
+    """
+
+    def __init__(self, tensors: list[torch.Tensor]):
+        if len(tensors) == 0:
+            raise ValueError("Cannot create a TensorList with no tensors.")
+        self.tensors = tensors
```
SumanthRH, line 60: It would be good to also assert that the tensors are all on the same device. 

It looks like you assume so based on `device` property implementation

skyrl/backends/skyrl_train/training_batch.py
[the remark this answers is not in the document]
SumanthRH: This is actually a good point, I would imagine the right thing to do with `TensorList.repeat` is to just call `.repeat` on all the individual tensors rather than the current approach. This makes it a natural Tensor extension rather than the current way

skyrl/backends/skyrl_train/training_batch.py
nithinvc: Done

skyrl/backends/skyrl_train/training_batch.py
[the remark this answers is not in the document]
nithinvc: Hmm do we expect `TensorList.repeat(2)` to repeat twice along the batch dimension? So with a tensorlist `[x1, x2, x3]`, I thought the expected semantics would be `[x1, x2, x3, x1, x2, x3]`, so repeating along the batch dim. But if we call `[t.repeat(2) for t in tensorlist]` this would create `[B, S*2]`?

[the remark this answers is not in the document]
SumanthRH: Discussed offline. the current semantics make sense

PR #1335, opened 2026-03-17
[train][2/N] Support for Megatron PP + CP for R3

devpatelio:
# Summary
Extending #1273, this PR provides support for pipeline parallelism and context parallelism for R3. See #815 for tracking of future tasks to fully support routing replay in all settings.

# Implementation

**Pipeline Parallelism**
For pipeline parallelism, we create a helper function `_get_current_pp_stage_layer_range(model_config)` which maps the current PP rank and its layers to the global layer offset across all the model layers so that we can use this offset to correctly select the corresponding replay instances from a `RouterReplay.global_router_replay_instances`. 

First, we get the number of pipeline stages from PP world size along with the total number of model layers. For models containing dense layers / unequal pipeline stages, megatron supports setting a customer number of layers for the first and last PP rank. Then, we capture these values from the model config and check to see if the remaining number of layers can be evenly distributed across the remaining PP ranks. Finally, we return the transformer-layer range owned by the current PP rank as s_p, n_p, where:
- s_p is the global starting layer index for rank p
- n_p is the number of transformer layers assigned to that stage

For an even partition with L total layers and P pipeline stages:
- next_n_pp_layers = L // P, start_index = next_n_pp_layers * pp_rank
- the offset should thus span (next_n_pp_layers * pp_rank) : (next_n_pp_layers * (pp_rank+1)

For uneven partitioning, if the first and/or last stages are assigned custom layer counts, we subtract those from $L$, split the remaining layers evenly among the remaining stages, and then shift the start index accordingly. This means we can support cases like Moonlight-16B models which have 27 layers, where we can pass `num_layers_in_first_pipeline_stage` as 13 for PP=2.  

**Context Parallelism**

When using sample_packing, our megatron worker pre-processes and post-processes packed sequences. When CP is enabled, it is split into CP*2 chunks, so each effective GPU gets 2 CP chunks of half the size. See https://github.com/NVIDIA/TransformerEngine/issues/1368. To account for this extra chunking, the `setup_per_microbatch_replay_forward` method is updated to so that the effective_seq_len accounts for cp_size * 2 (same as the alignment in preprocess_packed_seqs in megatron_utils.py) along with the seqlen_per_cp as seqlen_per_cp // 2. We then index the front and back halves of these CP chunks from the aligned indices across the CP ranks and then concatenate them. This ensures that the router replay indices see the correct tokens from this CP chunking for megatron.

**Testing**

You can test with CP and/or PP configs from the test_router_replay file.

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1327" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


skyrl/backends/skyrl_train/utils/replay_utils.py
```diff
@@ -228,26 +317,29 @@ def setup_per_microbatch_replay_forward(
...
         # (assumes dense layers precede MoE layers).
-        for i, router_instance in enumerate(instances):
+        for local_router_idx, router_instance in enumerate(instances):
             layer_number = getattr(router_instance, "layer_number", None)
             if layer_number is not None:
                 layer_idx = layer_number - 1  # layer_number is 1-based
             else:
-                layer_idx = i + (num_layers_in_data - num_instances)
-            if layer_idx < 0 or layer_idx >= num_layers_in_data:
+                layer_idx = local_layer_offset + local_router_idx
```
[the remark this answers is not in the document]
devpatelio: 👍🏼 local_num_layers for both paths in `_get_current_pp_stage_layer_range` does include dense layers so this makes sense. In the fallback, we should subtract num_instances from local_num_layers to get the actual number of MoE layers.

tests/backends/skyrl_train/gpu/gpu_ci/test_router_replay.py
```diff
@@ -267,105 +275,166 @@ def run_megatron_forward(enable_replay: bool) -> torch.Tensor:
...
+        loss_no_replay = metrics_no_replay["policy_loss"]
+        print(f"With replay    - loss: {loss_replay:.6f}")
+        print(f"Without replay - loss: {loss_no_replay:.6f}")
+        print(f"With replay metrics: {metrics_replay}")
+        print(f"Without replay metrics: {metrics_no_replay}")
+
+        diff = abs(loss_replay - loss_no_replay)
+        threshold = 0.5
+        print(f"Loss diff: {diff:.6f} (threshold: {threshold})")
+        assert diff < threshold, (
```
erictang000: this test always passes, because advantages are set to all zero above, and so loss will be all zero, and the diff in loss with and without replay will essentially be the same.  To really make this test work you have to also run a real forward pass to get non-zero logprobs.

I think it's non-trivial to have an actually good integration test for training here - I think we should treat  this as more of a unit test for `forward_backward` to make sure that training runs with issues with router replay on?

Can you revert all of the changes for this test except for making this run with tp=2, pp=2, ep=2? I think we can add cp=2 to that as well. We can remove the parameterization and just have a single setting for those parallelism strategies.

tests/backends/skyrl_train/gpu/gpu_ci/test_router_replay.py
```diff
@@ -246,15 +258,11 @@ def run_megatron_forward(enable_replay: bool) -> torch.Tensor:
         r3_logprobs = run_megatron_forward(enable_replay=True)
         no_r3_logprobs = run_megatron_forward(enable_replay=False)
 
-        mask = response_mask.bool()
```
erictang000, line 249: this masking is needed for correctness! please add it back

tests/backends/skyrl_train/gpu/gpu_ci/test_router_replay.py
```diff
@@ -217,16 +226,19 @@ def test_logprobs(ray_init_fixture):
...
+        cfg.trainer.policy.megatron_config.pipeline_model_parallel_size = pp
+        cfg.trainer.policy.megatron_config.context_parallel_size = cp
+        cfg.trainer.policy.megatron_config.expert_model_parallel_size = ep
+        cfg.trainer.policy.megatron_config.expert_tensor_parallel_size = etp
         cfg.trainer.micro_forward_batch_size_per_gpu = 2
         cfg.trainer.micro_train_batch_size_per_gpu = 2
 
         def run_megatron_forward(enable_replay: bool) -> torch.Tensor:
-            cfg.trainer.policy.megatron_config.moe_enable_routing_replay = enable_replay
+            cfg.trainer.policy.megatron_config.transformer_config_kwargs = {
```
erictang000: pls revert this! this was an old version of the test/flag - we have an explicit flag for it now

tests/backends/skyrl_train/gpu/gpu_ci/test_router_replay.py
```diff
@@ -105,8 +107,16 @@ def build_training_input_from_text_samples(
...
 @pytest.mark.megatron
-@pytest.mark.skip(reason="Skipping router replay test for now due to size constraints")
-def test_logprobs(ray_init_fixture):
+@pytest.mark.parametrize(
+    "tp,pp,cp,ep,etp,extra_tf_kwargs",
+    [
+        pytest.param(2, 1, 1, 2, 1, {}, id="baseline"),
+        pytest.param(2, 2, 1, 2, 1, {"num_layers_in_last_pipeline_stage": 13}, id="pp2"),
+        pytest.param(4, 1, 2, 8, 1, {}, id="cp2"),
+        pytest.param(2, 2, 2, 4, 1, {"num_layers_in_last_pipeline_stage": 13}, id="cp2_pp2"),
```
erictang000: let's just keep this one test with parallelism maxxed out to keep CI cost low

tests/backends/skyrl_train/gpu/gpu_ci/test_router_replay.py
```diff
@@ -105,8 +107,16 @@ def build_training_input_from_text_samples(
 
 
 @pytest.mark.megatron
-@pytest.mark.skip(reason="Skipping router replay test for now due to size constraints")
```
erictang000, line 108: let's keep this skip due to size constraints.

tests/backends/skyrl_train/gpu/gpu_ci/test_router_replay.py
```diff
@@ -267,105 +275,166 @@ def run_megatron_forward(enable_replay: bool) -> torch.Tensor:
...
```
erictang000, line 270: keep this skip for now

skyrl/train/utils/utils.py
```diff
@@ -206,12 +206,12 @@ def validate_megatron_cfg(cfg: SkyRLTrainConfig):
         assert (
             cfg.generator.inference_engine.enable_return_routed_experts
         ), "rollout router replay (r3) is only supported when enable_return_routed_experts is True"
-        assert (
-            cfg.trainer.policy.megatron_config.pipeline_model_parallel_size == 1
-        ), "pipeline parallel is not yet supported for router replay (r3) with megatron"
-        assert (
-            cfg.trainer.policy.megatron_config.context_parallel_size == 1
-        ), "context parallel is not yet supported for router replay (r3) with megatron"
+        # # assert (
```
erictang000: delete?

skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
```diff
@@ -372,7 +372,9 @@ def init_configs(
 
         self.strategy.hf_config = hf_config
         self.tokenizer = tokenizer
-        self.enable_router_replay = megatron_config.moe_enable_routing_replay
+        self.enable_router_replay = transformer_config_kwargs.get(
```
erictang000: delete?

skyrl/backends/skyrl_train/utils/replay_utils.py
```diff
@@ -164,48 +164,136 @@ def _pack_replay_indices(
...
+            out[dst_offset + half : dst_offset + seqlen_per_cp] = packed[back_start:back_end]
+            src_offset += seqlen_padded_i
+            dst_offset += seqlen_per_cp
+        packed = out
+
+    return packed.unsqueeze(0)  # [1, packed_len_per_cp, layers, topk]
+
+
+def _get_current_pp_stage_layer_range(model_config) -> tuple[int, int]:
+    """Return the current PP rank's transformer-layer range.
```
erictang000:
```suggestion
    """Return the current PP rank's transformer-layer range as (start_layer, end_layer)
```

skyrl/backends/skyrl_train/utils/replay_utils.py
```diff
@@ -164,48 +164,136 @@ def _pack_replay_indices(
...
+    Prefer Megatron's own helpers so replay indexing stays aligned with the
+    actual model partition, including embedding/loss pipeline accounting.
+    """
+    import megatron.core.parallel_state as mpu
+    from megatron.core.transformer.transformer_block import get_num_layers_to_build
+    from megatron.core.transformer.transformer_layer import get_transformer_layer_offset
+
+    pp_rank = mpu.get_pipeline_model_parallel_rank()
+
+    if get_num_layers_to_build is not None:
```
erictang000: when is this None? It looks like it's been in megatron-lm for a while

skyrl/backends/skyrl_train/utils/replay_utils.py
```diff
@@ -228,26 +317,29 @@ def setup_per_microbatch_replay_forward(
...
-    num_layers_in_data = len(per_layer_data)
+    global_num_layers_in_data = len(per_layer_data)
     instances = RouterReplay.global_router_replay_instances
     num_instances = len(instances)
 
-    if num_layers_in_data == num_instances:
-        RouterReplay.set_replay_data(per_layer_data)
+    local_layer_offset, local_num_layers = _get_current_pp_stage_layer_range(model_config)
+
+    if local_num_layers == num_instances:
```
erictang000, line 276: so this is if condition "local_num_layers == num_instances" is handling  pp=1 - but the previous if condition here was handling

```
 if num_layers_in_data == num_instances:
...
```

So are things still correct if pp=1 and `num_layers_in_data != num_instances`?

erictang000: should this condition still just be `if num_layers_in_data == num_instances:`, and we can just do `RouterReplay.set_replay_data(per_layer_data)`?

erictang000: essentially we can probably just revert this first if condition is what i'm thinking, but correct me if i'm misunderstanding

skyrl/backends/skyrl_train/utils/replay_utils.py
erictang000: if this is a stable API in megatron-core >= 0.16.0 can we just remove the excess code below?

erictang000 (commented):
some comments + questions 

particularly a little confused about how it works for MoE models with dense layers before moe layers and PP together.

tests/backends/skyrl_train/gpu/gpu_ci/test_router_replay.py
devpatelio: Ahh makes sense okay yes I think along the lines I might've just asked CC to write a test to make sure that the fwd/bwd pass replay piping was working without any errors. Would it be fine to keep the updated inference engine client and generation setup so that it matches more closely to the test_logprobs implementation and keeps things standardized? I'll revert to the original test w. added parallelism from #1273 in the meantime

skyrl/backends/skyrl_train/utils/replay_utils.py
devpatelio: Yep that condition was incorrect, I think it come from a bad tab accept it and then it technically wasn't failing so it went undetected. The logic below that was a fallback that verl implemented but that was for some old legacy tests I think it implemented so we can remove this (megatron helpers already do this)

skyrl/backends/skyrl_train/utils/replay_utils.py
devpatelio: So global_num_layers_in_data is in reference to the actual aligned data and the data entering the forward pass (not sharded). if global_num_layers_in_data == num_instances, it does tell us if PP=1 but our current setup also tells us if pp=1 and will work when num_layers_in_data != num_instances. 

Ie. if we are doing PP = 2 with Moonlight, the first pipeline stage is gonna see 13 layers but our num instances is only gonna show 12 instances. The mismatch means that we take the else path and map the non-dense layers to the correct instances for that rank. In the next rank, you're going to see 14 local layers and 14 instances so you can just set it directly with the offset. 

In the case we do PP = 1, the num instances will show 26 and the local num layers will be 27 (so num_layers != num instances) in which case we will go to the else branch and use the router's global indices to get the correct layers. So local_num_layers == num_instances is not uniquely handling PP = 1, it's operating per-rank when no dense layers exist

erictang000: i see, this makes sense!

erictang000:
screenshot from DMs:

image

skyrl/backends/skyrl_train/utils/replay_utils.py
```diff
@@ -220,34 +260,36 @@ def setup_per_microbatch_replay_forward(
...
+    global_num_layers_in_data = len(per_layer_data)
     instances = RouterReplay.global_router_replay_instances
     num_instances = len(instances)
+    local_layer_offset, local_num_layers = _get_current_pp_stage_layer_range(model_config)
 
-    if num_layers_in_data == num_instances:
-        RouterReplay.set_replay_data(per_layer_data)
+    if local_num_layers == num_instances:
+        local_per_layer_data = per_layer_data[local_layer_offset : local_layer_offset + local_num_layers]
+        RouterReplay.set_replay_data(local_per_layer_data)
```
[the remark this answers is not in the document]
erictang000: `per_layer_data` should always be total number of transformer layers. we can fix the test in a follow up PR.

PR #1352, opened 2026-03-19
[skyrl][tx] Move ModelInput token extraction to backends

nithinvc:
#  Move ModelInput token extraction to the SkyRL-train and jax backend

This PR is a precursor to adding multi-modal input chunks for VLM training (point 4, #1200 ). The tinker frontend passes to the backend lists of `ModelInput`s (for now only `EncodedTextChunk`s). 

### Changes

- Rename `ModelInputChunk` to `EncodedTextChunk` with an explicit `type: "encoded_text"` discriminator field (useful for future disambiguation of types). Moving forward, `ModelInputChunk` will be a type union of possible input types (e.g., `ImageChunk`).
- Replace `all_input_ids: list[list[int]]` with `all_input_chunks: list[ModelInput]` in `PreparedModelPassBatch`, and `all_prompts: list[list[int]]` with `all_prompts: list[ModelInput]` in `PreparedSampleBatch`
- Thread `ModelInput` objects through `prepare_model_pass_batch` and `prepare_sample_batch` in the engine, deferring token extraction to the backends
- Update Jax backend, SkyRL-Train backend, external inference client, and benchmarks to extract tokens from `ModelInput` chunks at the point of use

### Test Plan

```bash
# Tinker unit tests
uv run --extra jax pytest tests/tinker -v

# Jax backend
uv run --extra jax pytest tests/backends/test_jax_backend.py  -v

# SkyRL-Train inference engine + inference server + train batch tests
uv run --extra fsdp --extra dev pytest -v tests/backends/skyrl_train/inference_engines/ tests/backends/skyrl_train/inference_servers/ tests/backends/skyrl_train/test_train_batch.py -v

# GPU train batch test
uv run --extra dev --extra fsdp pytest tests/backends/skyrl_train/gpu/gpu_ci/test_train_batch.py -v

# GPU training step test
uv run --extra dev --extra fsdp pytest tests/backends/skyrl_train/gpu/gpu_ci/test_training_step.py -v
```

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1352" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/skyrl/tinker/types.py
+++ b/skyrl/tinker/types.py
@@ -232,7 +232,7 @@ class PreparedModelPassBatch(BaseModel):
     Engine extracts this from requests, backend converts to JAX arrays and computes.
     """
 
-    # Per-example data (list of lists)
+    # Per-example data
     all_model_inputs: list[ModelInput]
     all_targets: list[list[int]]
     all_token_weights: list[list[float]]


skyrl/tinker/api.py
```diff
@@ -295,11 +303,77 @@ class TrainingRun(BaseModel):
...
+    format: Literal["png", "jpeg"]
+    location: str
+    expected_tokens: int | None = None
+
+    @field_validator("location")
+    @classmethod
+    def validate_location(cls, v: str) -> str:
+        if not v:
+            raise ValueError("location must not be empty")
+        return v
```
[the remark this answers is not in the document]
nithinvc: I think this is okay for now

skyrl/tinker/api.py
```diff
@@ -295,11 +303,77 @@ class TrainingRun(BaseModel):
...
+def _get_model_chunk_type(v: Any) -> str:
+    if isinstance(v, dict):
+        if "type" in v:
+            return v["type"]
+        if "tokens" in v:
+            return "encoded_text"
+        if "location" in v:
+            return "image_asset_pointer"
+        if "data" in v:
+            return "image"
```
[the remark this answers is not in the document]
nithinvc: Agree it is brittle, but not sure of a better way since we aren't changing the SDK

skyrl/tinker/api.py
```diff
@@ -295,11 +303,77 @@ class TrainingRun(BaseModel):
...
+    type: Literal["image"] = "image"
+    data: str  # base64-encoded
+    format: Literal["png", "jpeg"]
+    expected_tokens: int | None = None
+
+    @field_validator("data")
+    @classmethod
+    def validate_base64(cls, v: str) -> str:
+        try:
+            base64.b64decode(v, validate=True)
```
pcmoritz: This means you will go through the data twice, once for the validation and once in `to_types`, which is costly. Is it possible to do this only once? E.g. try to use the native base64 decoder of pydantic like https://docs.pydantic.dev/2.0/usage/types/encoded/#base64-encoding-support

nithinvc: That makes sense. I replaced it with pydantic's `Base64Bytes` which does validation on construction:

```python
from pydantic import Base64Bytes, BaseModel
class Model(BaseModel):
    base64_bytes: Base64Bytes
Model(base64_bytes=b'undecodable') # errors
```
so we don't need to do the base64 checking ourselves.

nithinvc:
Slight scope change, this PR now pushes the token extraction logic to the backends.

skyrl/backends/jax.py
```diff
@@ -612,13 +612,16 @@ def _model_pass(
...
 
         results = {}
 
-        # Extract data from prepared batch
-        all_input_ids = prepared_batch.all_input_ids
+        # Extract token IDs from ModelInput chunks
+        all_input_ids = [
+            [tok for chunk in mi.chunks for tok in (chunk.tokens if hasattr(chunk, "tokens") else [])]
+            for mi in prepared_batch.all_input_chunks
+        ]
```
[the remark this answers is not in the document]
nithinvc: Explicitly not doing this since each of the current call sites need to handle different input chunk types differently.

skyrl/tinker/types.py
```diff
@@ -225,7 +233,7 @@ class PreparedModelPassBatch(BaseModel):
     """
 
     # Per-example data (list of lists)
-    all_input_ids: list[list[int]]
+    all_input_chunks: list[ModelInput]
```
pcmoritz: This should be `all_model_inputs`

skyrl/tinker/engine.py
```diff
@@ -114,7 +113,7 @@ def prepare_model_pass_batch(
     Returns:
         PreparedModelPassBatch with all data extracted from requests
     """
-    all_input_ids = []
+    all_input_chunks = []
```
pcmoritz: Similarly, this should be all_model_inputs

skyrl/tinker/types.py
```diff
@@ -247,7 +255,7 @@ class PreparedSampleBatch(BaseModel):
     """
 
     # Per-sample data
-    all_prompts: list[list[int]]
+    all_prompts: list[ModelInput]
```
pcmoritz: I suggest we also rename this to `all_model_inputs` to be consistent with `PreparedModelPassBatch`

skyrl/backends/skyrl_train_backend.py
```diff
@@ -534,7 +538,8 @@ def sample(
         async def sample_all():
             tasks = []
             for i in range(len(prepared_batch.all_prompts)):
-                prompt = prepared_batch.all_prompts[i]
+                model_input = prepared_batch.all_prompts[i]
+                prompt_token_ids = render_model_input([model_input])[0].prompt_ids
```
pcmoritz: why don't we do `all_prompt_ids = render_model_input(prepared_batch.all_prompts)` outside the loop and then `prompt_ids = all_prompt_ids[i]` in the loop? That's a little cleaner and also does the right batching :)

skyrl/tinker/types.py
nithinvc: updated all instances

skyrl/tinker/engine.py
nithinvc: updated

skyrl/tinker/types.py
nithinvc: updated to match `PreparedModelPassBatch`

skyrl/backends/skyrl_train_backend.py
nithinvc: Moved outside of `sample_all`

PR #1360, opened 2026-03-20
[train] LoRA-only weight syncing for Megatron 

### LINKED ISSUE #1336: [train] Support LoRA-only weight syncing for Megatron backend
# Summary

Currently, for weight syncing with LoRA, SkyRL's Megatron backend will fuse the weights and sync the merged weights with vLLM replicas. SkyRL should support syncing just the LoRA weights. 

This was previously attempted in #885 but it also requires vLLM changes. 

1. Change Megatron's weight extractor to only transfer trainable, LoRA weights. 
2. Fix vLLM's weight loading  in `LRUCacheWorkerLoRAManager` upstream to allow LoRA weight loading with in-memory LoRA weights
3. Upgrade SkyRL to use the vLLM release version with the above changes after 2. lands

hao-aaron:
closes https://github.com/NovaSky-AI/SkyRL/issues/1336
Uses megatron new `export_adapter_weights` to save loras to disk and load in vllm. Bumped megatron version for the new api. 

Added new megatron tests to `test_lora.py`, run test_lora with both `_USE_SKYRL_NEW_INFERENCE=0` and `1`.

Added some more robust shutdown that was causing issues with the test when using new inference

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1360" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
```diff
@@ -814,12 +867,16 @@ async def broadcast_to_inference_engines(
 
         torch.cuda.empty_cache()
 
-        # Extract and send weights using the sender created at init time
-        weight_metadata = self.weight_extractor.get_weight_metadata(generator_dtype)
-        await self._weight_transfer_sender.send_chunks(
-            self.weight_extractor.extract_weights(generator_dtype),
-            weight_metadata=weight_metadata,
-        )
+        if self._is_lora:
```
erictang000: hmm i don't think we want to get rid of the option to merge loras back into the base model to sync (since lora serving throughput is slower)

can we add a flag to gate this feature PR - something like `trainer.policy.megatron_config.lora_config.merge_lora` and have the default still be `True`?

then when we get to in memory lora weight sync we can replace the disk syncing?

erictang000:
also will this lora only weight sync mode only support `colocate_all=true` for now? let's make sure to add that to config validation

skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
SumanthRH: +1. 

We definitely want to support both modes. Default to merge for now

tests/backends/skyrl_train/gpu/gpu_ci/test_lora.py
```diff
@@ -1,6 +1,9 @@
 """
-# Run tests (requires fsdp extra):
-uv run --isolated --extra dev --extra fsdp pytest tests/backends/skyrl_train/gpu/gpu_ci/test_lora.py
+# Run FSDP tests:
+uv run --isolated --extra dev --extra fsdp pytest tests/backends/skyrl_train/gpu/gpu_ci/test_lora.py -k "fsdp"
+
+# Run Megatron tests:
+uv run --isolated --extra dev --extra megatron pytest tests/backends/skyrl_train/gpu/gpu_ci/test_lora.py -k "megatron"
```
SumanthRH, line 6: This should be `-m "not megatron"` and `-m megatron` right?

There is no `fsdp` marker

```suggestion
# Run FSDP tests:
uv run --isolated --extra dev --extra fsdp pytest tests/backends/skyrl_train/gpu/gpu_ci/test_lora.py -m "not megatron"

# Run Megatron tests:
uv run --isolated --extra dev --extra megatron pytest tests/backends/skyrl_train/gpu/gpu_ci/test_lora.py -m "megatron"
```

tests/backends/skyrl_train/gpu/gpu_ci/test_megatron_worker.py
SumanthRH, line 1: @hao-aaron can you add this to the megatron CI here and ensure that tests pass with `_SKYRL_USE_NEW_INFERENCE=1` ?

https://github.com/NovaSky-AI/SkyRL/blob/9d7bca91a505970211ec844f10bf5c072e921163/ci/gpu_ci_run_skyrl_train_megatron.sh#L8

tests/backends/skyrl_train/gpu/gpu_ci/test_lora.py
```diff
@@ -51,23 +67,29 @@ def get_test_actor_config(enable_lora: bool = False) -> SkyRLTrainConfig:
         pytest.param(True, "nccl", "fsdp", 2),
         pytest.param(False, "nccl", "fsdp2", 2),
         pytest.param(True, "nccl", "fsdp2", 2),
+        pytest.param(False, "nccl", "megatron", 2, marks=pytest.mark.megatron),
+        pytest.param(True, "nccl", "megatron", 2, marks=pytest.mark.megatron),
```
SumanthRH: Let's also parameterize saving to disk vs in-memory weight sync here for LoRA (after addressing @erictang000 's comment)

hao-aaron:
> also will this lora only weight sync mode only support colocate_all=true for now? let's make sure to add that to config validation

colocated and non colocated both work, i have tests for both in test_lora.py

tests/backends/skyrl_train/gpu/gpu_ci/test_lora.py
hao-aaron: it matches all test names with "fsdp" substring, so it gets fsdp and fsdp2

ci/gpu_ci_run_skyrl_train.sh
```diff
@@ -35,4 +35,4 @@ uv run --directory . --isolated --extra dev --extra fsdp pytest -s tests/backend
 _SKYRL_USE_NEW_INFERENCE=1 uv run --isolated --extra dev --extra fsdp pytest -s tests/backends/skyrl_train/gpu/gpu_ci/test_policy_local_engines_e2e.py
 _SKYRL_USE_NEW_INFERENCE=1 uv run --isolated --extra dev --extra fsdp pytest -s tests/backends/skyrl_train/gpu/gpu_ci/test_engine_generation.py
 _SKYRL_USE_NEW_INFERENCE=1 uv run --isolated --extra dev --extra fsdp pytest -s tests/backends/skyrl_train/gpu/gpu_ci/test_skyrl_gym_generator.py
-_SKYRL_USE_NEW_INFERENCE=1 uv run --isolated --extra dev --extra fsdp pytest -s tests/backends/skyrl_train/gpu/gpu_ci/test_lora.py
+_SKYRL_USE_NEW_INFERENCE=1 uv run --isolated --extra dev --extra fsdp pytest -s tests/backends/skyrl_train/gpu/gpu_ci/test_lora.py -k "fsdp"
```
SumanthRH: To ensure we don't miss out on other tests in this file in the future, it is best to use the `not megatron` marker


```suggestion
_SKYRL_USE_NEW_INFERENCE=1 uv run --isolated --extra dev --extra fsdp pytest -s tests/backends/skyrl_train/gpu/gpu_ci/test_lora.py -m "not megatron"
```

skyrl/backends/skyrl_train/inference_servers/router.py
```diff
@@ -242,9 +242,43 @@ def _wait_until_healthy(
...
-            self._server_thread.join(timeout=5)
+            self._server_thread.join(timeout=10)
+
+            if self._server_thread.is_alive():
+                logger.warning("Router thread did not exit gracefully, forcing server socket closure")
+                self._force_close_server_sockets()
+
+        if self._client:
+            try:
+                asyncio.get_event_loop().run_until_complete(self._client.aclose())
```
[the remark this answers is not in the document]
SumanthRH: @hao-aaron this can be a legit issue with tests like `test_weight_sync` which call `InferenceEngineState.close` and thus `router.shutdown` from a running event loop

https://github.com/NovaSky-AI/SkyRL/blob/58e3ab728c1b4e320d9d385b0717f3ce8e64dbf3/tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_weight_sync.py#L151-L153

For cleanup, can we actually add a shutdown method to the FastAPI app?

```python
@app.on_event("shutdown")
async def shutdown():
    await foo.aclose()
```
does this work? I guess the other problem with the router right now is that creation of the async client is done in the main thread while the app is run in a different thread - pretty bad IMO. But I think this shutdown should still work. Can you try this

SumanthRH (commented):
Only 2 minor comments left!

tests/backends/skyrl_train/gpu/utils.py
```diff
@@ -520,6 +519,11 @@ def create(
                 cli_args.enable_lora = True
                 if active_lora_name is None:
                     active_lora_name = "skyrl-lora"
+            else:
+                # Override build_vllm_cli_args which auto-enables LoRA based
+                # on lora rank in the config.  For merged weight sync the
+                # inference engine must NOT have LoRA wrapping enabled.
+                cli_args.enable_lora = False
```
SumanthRH: @hao-aaron actually the same fix is needed for `build_vllm_cli_args`? Otherwise this will fail outside of the tests? 

We need to unset `enable_lora` in the vllm cli if it's megatron backend and merge lora is true

SumanthRH (approved):

erictang000:
wait can we hold off on merging this?

the megatron-bridge bump might cause issues for deepseek-v3 models:
```bash
megatron-bridge = {git = "https://github.com/NVIDIA-NeMo/Megatron-Bridge", rev = "f78c65f9", marker = "sys_platform == 'linux'"}
```

erictang000 (changes requested):
megatron-bridge commit bump is a little scary for deepseek-v3 models - can we test that initializing glm-flash-4.7 still works at all?

https://github.com/NovaSky-AI/SkyRL/blob/main/examples/train/megatron/run_megatron_grpo_glm4_7_30b.sh

i saw this script not run anymore previously after bumping megatron-bridge to a newer commit


let me test this out

erictang000:
> colocated and non colocated both work, i have tests for both in test_lora.py

how does it work for multi-node + non-colocated if it's writing to disk? i guess you can write to shared storage but that's a little non-obvious - we should at least put up a warning if non-colocated + not merging lora?

erictang000:
@hao-aaron sorry for blocking this PR - finally got around to bumping megatron-bridge in https://github.com/NovaSky-AI/SkyRL/pull/1412 - can you pull in the latest from main into your branch nad make sure that mbridge==0.3.1 has the changes you need?

skyrl/train/config/config.py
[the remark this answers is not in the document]
erictang000: @hao-aaron i think this comment looks correct

erictang000 (approved):

PR #1376, opened 2026-03-24
[trainer] refactor `dispatch_from_staged` to individually serialize DP chunks to avoid materializing whole batch on all workers

erictang000:
## Summary

When training with large batches (e.g. with `rollout_expert_indices` inflating `TrainingInputBatch` to ~30GB), the old `stage_data` approach `ray.put` the **entire** batch as a single object and passed that `ObjectRef` to every worker. Each worker deserialized the full blob just to slice out its small portion. Workers on different nodes experienced minutes of deserialization skew — the fast worker would enter `forward_backward`, hit an NCCL `all_reduce` inside Megatron's `finalize_model_grads`, and timeout waiting for the slow worker (600s NCCL default), manifesting as a hang with one GPU at full utilization and the other at zero.

## Fix

Pre-stage **per-DP-rank, per-mini-batch chunks** in the object store before the training loop. Each worker now only deserializes its own small chunk (`batch_size / num_mini_batches / dp_size`), eliminating deserialization skew.

## Changes

- `MeshDispatch.stage_chunks`: new method that slices the full batch into mini-batches, chunks each by DP rank, and `ray.put`s each chunk upfront
- `MeshDispatch.dispatch_from_staged`: simplified to map pre-staged `chunk_refs[dp_rank]` → workers (no more slicing or `ray.put` on the hot path)
- `WorkerDispatch.stage_data`: calls `stage_chunks` with the model's DP size
- `WorkerDispatch.forward_backward_from_staged`: accepts `chunk_refs` instead of a full-batch `ObjectRef` + indices
- `trainer._execute_training_step`: calls `stage_data` once before the epoch loop, iterates over pre-staged chunk refs
- removes forward_backward_from_staged from `Worker` in `worker.py`, since each worker just gets its own data chunk.

## Test plan

- [x] `test_stage_chunks_and_dispatch_from_staged` updated to cover new two-phase API
- [x] Existing `test_mesh_dispatch` unchanged (non-staged path unaffected)
- [x] Verify no NCCL timeout on multi-node training with large `rollout_expert_indices`

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1376" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/skyrl/backends/skyrl_train/distributed/dispatch.py
+++ b/skyrl/backends/skyrl_train/distributed/dispatch.py
@@ -173,11 +173,10 @@ class MeshDispatch(Dispatch):
         mini_batch_size: int,
     ) -> List[List[ObjectRef]]:
         """
-        Pre-stage all mini-batch chunks in the object store before the training loop.
+        Pre-stage all mini-batch chunks into the object store.
 
         Slices the full batch into mini-batches, chunks each by DP rank, and
-        ``ray.put``s each chunk. This keeps all serialization off the critical
-        path so GPUs stay saturated during training.
+        ``ray.put``s each chunk.
 
         Args:
             dp_size: Number of data-parallel ranks
@@ -217,9 +216,8 @@ class MeshDispatch(Dispatch):
         """
         Dispatch pre-staged per-DP chunks to workers.
 
-        Each worker receives only its own small chunk (already in the object
-        store), avoiding large deserialization skew that can cause NCCL
-        collective timeouts.
+        Each worker receives only its own chunk (already in the object
+        store), avoiding unnecessary deserialization overhead.
 
         Args:
             actor_infos: List of actor info objects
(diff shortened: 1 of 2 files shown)


erictang000:
Confirmed no timing regression on `policy_train` (which includes serializing chunks separately) for basic GSM8K:

image

erictang000:
confirmed speedup and lower max object store memory on multi-node test

## Setup
I added the following code to `convert_prompts_responses_to_batch_tensors` in `preprocess.py` to add a dummy rollout_expert_indices tensor of ~16 GiB 

```
# len(prompts) is 1024 * 5 for run_gsm8k
# so this is 1024 * 5 * 1000 * 50 * 16 * 4 bytes = ~16 GiB
 rollout_expert_indices_tensor = torch.zeros(len(prompts), 1000, 50, 16, dtype=torch.int32)
```

I ran the basic gsm8k example with `Qwen2.5-1.5B-Instruct` on 16 gpus (2 nodes) to ensure that object store transfer would become a bottleneck. I ran this on main and on this PR to compare speedup.

## Results
We see a speedup of 120s -> 40s in this scenario, where the speedup is just from having to transfer less data across nodes for the object store

image

Confirmed the size of the data transfer is lower after the change

#### Before
~16 GiB per node during training
image

#### After
Only 8 GiB on the node where the driver isn't
image

PR #1379, opened 2026-03-24
[train] Fix double-serialization of TensorBatch in pickle

erictang000:
### Problem

`TensorBatch` extends `dict` and defines `__getstate__`/`__setstate__` for custom tensor serialization (numpy fast path with `torch.save` fallback). However, Python's pickle protocol for dict subclasses serializes data through **two** independent paths:

1. `__getstate__()` → custom numpy/`torch.save` serialization (the intended path)
2. `iter(self.items())` → pickle's default dict protocol, which pickles each raw tensor through `torch.storage._load_from_bytes`

This causes two issues:

- **Correctness**: Path 2 fails for dtypes backed by `UntypedStorage` (e.g. `uint16`, `uint32`, `uint64`) because `UntypedStorage` has no `.dtype` attribute, raising `AttributeError: type object 'torch.storage.UntypedStorage' has no attribute 'dtype'`.
- **Performance**: Every tensor in the batch is serialized and deserialized twice — the path 2 results are immediately overwritten by `__setstate__`, making that work entirely redundant.

### Fix

Override `__reduce__` on `TensorBatch` to return only the `__getstate__` output, bypassing the default dict-subclass behavior of separately pickling raw dict items. A module-level `_rebuild_tensor_batch` helper reconstructs the object from the serialized state.

def __reduce__(self):
    return (_rebuild_tensor_batch, (type(self), self.__getstate__()))

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1379" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/tests/backends/skyrl_train/test_train_batch.py
+++ b/tests/backends/skyrl_train/test_train_batch.py
@@ -227,8 +227,8 @@ def test_train_batch_pickle_uint_dtypes():
 
     Without the __reduce__ patch in training_batch.py, pickle's default dict-subclass protocol
     separately serializes raw dict items through torch's storage-level pickle,
-    which fails for uint16 (UntypedStorage has no .dtype attribute). This tests 
-    that we are correctly skipping the default dict-subclass protocol and 
+    which fails for uint16 (UntypedStorage has no .dtype attribute). This tests
+    that we are correctly skipping the default dict-subclass protocol and
     using the custom __reduce__ method to serialize the TensorBatch object.
     """
     batch_size = 3


erictang000:
noticeable speedup for forward pass, which calls put several times: 
image

SumanthRH (changes requested):
Great find! can you add a serialization test for uint8 dtype ? Tests should fail on main and pass with your fix

You can modify: 

https://github.com/NovaSky-AI/SkyRL/blob/50ebff5a01b8750cf670e948ded945cd3a956232/tests/backends/skyrl_train/test_train_batch.py#L157

erictang000:
added test, + ran all cpu tests locally, all passing, ready for review @SumanthRH 

image

PR #1384, opened 2026-03-25
[CI] Migrate MoE Tests to OlMoE

devpatelio:
Tentatively migrated all CI except 5 tests in `test_megatron_worker.py`

- Test_megatron_forward[tp4_pp1_cp1_ep4_etp1_policy_seq_packing]
- Test_megatron_lora_forward[tp4_pp1_cp1_ep4_etp1_policy]
- test_megatron_train[tp4_pp1_cp1_ep4_etp1_policy_seq_packing] 
- Test_megatron_train[tp4_pp1_cp1_ep4_etp1_policy_seq_packing_lora] 
- Test_megatron_offload_memory_and_correctness[policy] 

The Megatron Bridge does not support TP > 1 for EP > 1 for the q/k layernorm so this will require a patch. For now, OlMoE should reduce CI resources for majority of the tests. 

See https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/ffe05be084aaee1a2bccd005737329de9284ce91/src/megatron/bridge/models/olmoe/olmoe_provider.py#L137 for reference

### DIFF
--- a/tests/backends/skyrl_train/gpu/gpu_ci/test_megatron_worker.py
+++ b/tests/backends/skyrl_train/gpu/gpu_ci/test_megatron_worker.py
@@ -310,9 +310,7 @@ async def test_megatron_forward(
 
         return attention_mask.to("cpu").detach(), action_log_probs.to("cpu").detach(), num_actions
 
-    attention_mask, action_log_probs, num_actions = ray.get(
-        run_hf_forward.remote(batch, model_name)
-    )
+    attention_mask, action_log_probs, num_actions = ray.get(run_hf_forward.remote(batch, model_name))
 
     #### Compare results ####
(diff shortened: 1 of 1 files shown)


erictang000:
> The Megatron Bridge does not support TP > 1 for EP > 1 for the q/k layernorm so this will require a patch. For now, OlMoE should reduce CI resources for majority of the tests.

oh hmm ok i didn't realize this...  tp + ep seems like a prepreq for any of these tests... let's figure out another workaround for router replay tests, and we can keep megatron_worker tests just using fewer hidden layers

erictang000:
we could consider fixing any issue upstream in megatron-bridge, or implementing our own bridge for a small model like the small arcee moe models

PR #1388, opened 2026-03-25
log all trajectory repetitions per training step (not just first)

arteemg:
## Problem

During training, `log_example` was called only on `response_ids[0]` and `rewards[0]`, meaning only the first of `n_samples_per_prompt` trajectories was ever logged per training step. With `n_samples_per_prompt=8`, 7 out of 8 rollouts were invisible, making it impossible to debug agent behavior across repetitions.

## Change

Loop over all trajectories in the batch and log each one, labeled with its `instance_id` and `repetition_id` from `TrajectoryID` (when available), or a fallback index.

## Example
Run after the fix, with all rollouts for a single contest (4316) being correctly logged:

[code_contests_4316_all_reps.log](https://github.com/user-attachments/files/26254038/code_contests_4316_all_reps.log)


arteemg:
@SumanthRH could you take a look! thank you in advance!

PR #1407, opened 2026-03-30
Fix preprocess_packed_seqs crash with short sequences under CP > 1

CharlieFRuan:
## Summary

Fixes a `RuntimeError` crash in `preprocess_packed_seqs` when context parallelism (CP > 1) encounters sequences shorter than `align_size = TP * CP * 2`.

## Problem

When CP is enabled, `preprocess_packed_seqs` splits each sequence into `CP * 2` chunks across ranks. To do this it:

1. Pads the sequence length up to a multiple of `align_size = TP * CP * 2` (e.g. 16 for TP=4, CP=2)
2. Computes `half_seqlen = padded_len // CP // 2` from the **padded** length
3. Indexes into `d = input_ids[i, attention_mask[i]]` — the **unpadded** real tokens — using `half_seqlen`-based offsets

If a sequence has fewer real tokens than `align_size`, step 3 reads past the end of `d`:

```
align_size = TP(4) * CP(2) * 2 = 16
real tokens:    2  (e.g. a masked/failed placeholder sequence)
padded to:     16
half_seqlen:    4  (= 16 / 2 / 2)

d[0:4]  →  tries to read 4 elements from a 2-element tensor  →  RuntimeError
```

This consistently crashes runs whenever any batch contains short placeholder sequences from failed instances.

## Fix

Pad `d` to the aligned length with zeros before the CP chunk indexing (2 lines in `megatron_utils.py`). Since the output buffer `input_ids_rmpad` is already zero-initialized, this produces identical results — the zero-padded positions were going to be zeros anyway.

## Test plan

- [x] Unit test added (`test_preprocess_packed_seqs_cp.py`) — mocks `mpu` parallel state and calls `preprocess_packed_seqs` directly with short sequences across multiple TP/CP configs. Verified it fails without the fix (3/4 cases crash) and passes with it.

Run the unit test:
```bash
uv run --isolated --extra dev --extra megatron -- pytest -s tests/backends/skyrl_train/distributed/test_preprocess_packed_seqs_cp.py -v
```

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1407" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/tests/backends/skyrl_train/distributed/test_preprocess_packed_seqs_cp.py
+++ b/tests/backends/skyrl_train/distributed/test_preprocess_packed_seqs_cp.py
@@ -5,19 +5,62 @@ Regression test for the CP=2 tensor shape crash:
   size (2) at non-singleton dimension 0.
 
 Run with:
-  uv run --isolated --extra dev --extra megatron -- pytest -s tests/backends/skyrl_train/distributed/test_preprocess_packed_seqs_cp.py
+  uv run --isolated --extra dev -- pytest -s tests/backends/skyrl_train/distributed/test_preprocess_packed_seqs_cp.py
 """
 
-import importlib.util
-from unittest.mock import patch
+import sys
+from dataclasses import dataclass
+from types import ModuleType
+from typing import Any
+from unittest.mock import MagicMock, patch
 
 import pytest
 import torch
 
-_has_megatron = importlib.util.find_spec("megatron") is not None
+# ---------------------------------------------------------------------------
+# Inject lightweight megatron stubs into sys.modules so that megatron_utils
+# can be imported without megatron-core installed (CPU CI).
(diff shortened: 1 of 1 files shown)


tests/backends/skyrl_train/distributed/test_preprocess_packed_seqs_cp.py
```diff
@@ -0,0 +1,64 @@
+"""Test that preprocess_packed_seqs handles short sequences with CP > 1.
```
erictang000, line 1: this pr looks good, but is it possible to put this as a CPU test or is there too much patching needed to make this work without gpu

CharlieFRuan: Patched and moved. Made sure the tests fail without the fix

PR #1418, opened 2026-03-31
[tx] Add initial implementation of RayJaxBackend

### LINKED ISSUE #1393: [tinker] Support using Ray to manage tinker API server, engine and Jax workers
I recently spent some time trying to get SkyRL-tx working on a Ray cluster. I initially started with an approach similar to https://github.com/NovaSky-AI/SkyRL/pull/955, but I don't think it's quite what we want because it still requires running the `uv run` command on Ray nodes with GPUs / TPUs. Instead, it should be possible to run an entrypoint script on the head node that will then distribute Ray actors, similar to how skyrl-train works today.

This work may require some larger refactoring in `skyrl/tinker/api.py`, due to the way the engine parameters are extracted from the parent uv command used for the tinker API server, and because the tinker API server / engine need to run on the same Ray node as worker0. We may need to decouple how the arguments for API server / engine are set and add some additional logic to ensure the tinker apiserver / engine and worker 0 are co-located. I think the end result should be something similar to the existing [main entrypoint for skyrl-train](https://github.com/NovaSky-AI/SkyRL/blob/main/skyrl/train/entrypoints/main_base.py) but with some added logic for scheduling the tinker API server / engine. 

I'm happy to drive this but looking for some guidance on the overall approach.

andrewsykim:
Fixes https://github.com/NovaSky-AI/SkyRL/issues/1393

This PR introduces the initial implementation of `RayJaxBackend` and `RayJaxBackendImpl` to enable running skyrl-tx on a Ray cluster with a single `ray job submit` command. When enabled, the Tinker API and Engine run on the driver/head node, while the actual JAX backend operations are distributed across Ray actors running on worker nodes. This removes the need for manual multi-node orchestration for JAX distributed training.

Tested the changes by running the following command on my 4x4 v6e TPU cluster:
```sh
ray job submit --runtime-env-json '{"py_executable": "uv run"}' -- sh -c 'cd /home/ray/SkyRL && uv run --extra tpu --extra jax --extra tinker -m skyrl.tinker.api --base-model Qwen/Qwen3-0.6B --backend ray-jax  --backend-config '\''{"ray_pg_bundles": [{"CPU": 4, "TPU": 4},{"CPU": 4, "TPU": 4},{"CPU": 4, "TPU": 4},{"CPU": 4, "TPU": 4}], "tensor_parallel_size": 4, "sample_max_num_sequences": 256, "train_micro_batch_size": 8, "fully_sharded_data_parallel_size": 4, "num_processes": 4}'\'''

$ ray job submit -- sh -c $'cd /home/ray/SkyRL && uv run --extra tpu --extra jax --extra tinker -m skyrl.tinker.api --base-model Qwen/Qwen3-8B --backend ray-jax  --backend-config \'{"ray_actor_options": {"resources": {"TPU": 4}}, "ray_pg_bundles": [{"CPU": 4, "TPU": 4},{"CPU": 4, "TPU": 4},{"CPU": 4, "TPU": 4},{"CPU": 4, "TPU": 4}], "sample_max_num_sequences": 256, "train_micro_batch_size": 32, "tensor_parallel_size": 4, "fully_sharded_data_parallel_size": 4, "num_processes": 4}\''
```

Create a train script (copied from skyrl-tx examples)
```
import tinker
import numpy as np
from tinker import types

# Connect to the local server
service_client = tinker.ServiceClient(base_url="http://localhost:8000", api_key="tml-dummy")
training_client = service_client.create_lora_training_client(base_model="Qwen/Qwen3-8B")
tokenizer = training_client.get_tokenizer()

# Training examples
examples = [
    {"input": "banana split", "output": "anana-bay plit-say"},
    {"input": "quantum physics", "output": "uantum-qay ysics-phay"},
    {"input": "coding wizard", "output": "oding-cay izard-way"},
]

def process_example(example, tokenizer):
    prompt = f"English: {example['input']}\nPig Latin:"
    prompt_tokens = tokenizer.encode(prompt, add_special_tokens=True)
    completion_tokens = tokenizer.encode(f" {example['output']}\n\n", add_special_tokens=False)

    tokens = prompt_tokens + completion_tokens
    weights = [0] * len(prompt_tokens) + [1] * len(completion_tokens)

    return types.Datum(
        model_input=types.ModelInput.from_ints(tokens=tokens[:-1]),
        loss_fn_inputs=dict(weights=weights[1:], target_tokens=tokens[1:])
    )

processed = [process_example(ex, tokenizer) for ex in examples]

# Training loop
for _ in range(6):
    fwdbwd = training_client.forward_backward(processed, "cross_entropy").result()
    training_client.optim_step(types.AdamParams(learning_rate=1e-4)).result()

    logprobs = np.concatenate([o['logprobs'].tolist() for o in fwdbwd.loss_fn_outputs])
    weights = np.concatenate([e.loss_fn_inputs['weights'].tolist() for e in processed])
    print(f"Loss: {-np.dot(logprobs, weights) / weights.sum():.4f}")
```

Run this training script on the driver:
```
ray job submit --working-dir . -- sh -c 'cp rl_loop.py /home/ray/SkyRL/ && cd /home/ray/SkyRL && uv run rl_loop.py'
```


andrewsykim:
@pcmoritz looking for your high-level feedback on the approach before I test and polish the PR further.

skyrl/tinker/engine.py
```diff
@@ -175,9 +175,14 @@ def get_backend_classes(backend_name: str):
         )
 
         return SkyRLTrainBackend, MegatronBackendOverrides
+    elif backend_name == "ray-jax":
```
pcmoritz: From a user perspective, I feel like it would be most natural to not have a separate backend, since it is just the jax backend but running on top of Ray. You probably introduced a new backend to be able to nicely handle the dependencies, right?

Another way to do this would be to introduce a `ray` extra, so the user can run
```
uv run --extra ray --extra tinker -m skyrl.tinker.api --backend jax
```
And the ray backend will be chosen if a `use_ray` flag is set in the backend config.

I would have a slight preference for this, let me know what you think :)

The code organization of making a `ray_jax` file makes a lot of sense to me.

skyrl/backends/ray_jax.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def __init__(self, base_model: str, config: JaxBackendConfig, process_id: int):
+        self.base_model = base_model
+        self.config = config.model_copy()
+        self.process_id = process_id
+        self.backend = None
+
+        if process_id == 0:
+            self.node_ip = ray.util.get_node_ip_address()
+            self.port = 7777
```
pcmoritz: I like the fact that with the Ray backend the coordinator_address doesn't need to be configured. It would be better to not hard-code the port though. I could see why it is tricky to auto discover the port (jax doesn't seem to support `0` as the port). If we must hard-code it, it might be worth to introduce an environment variable to be able to override it (e.g. `SKYRL_JAX_COORDINATOR_PORT`).

skyrl/backends/ray_jax.py
```diff
@@ -0,0 +1,183 @@
...
+
+    This class runs in the driver program (along with Tinker API / Engine) and proxies
+    commands to all JAX workers running as Ray actors.
+    """
+
+    def __init__(self, base_model: str, config: JaxBackendConfig):
+        self.base_model = base_model
+        self.config = config.model_copy()
+
+        num_processes = self.config.num_processes or 1
```
pcmoritz: I think it makes most sense to remove the default `1` here and make `num_processes` required if the Ray backend is used (since it will be multi-node / multi process in most cases).

skyrl/tinker/engine.py
andrewsykim: makes sense, I added a use_ray to backend config and removed `ray-jax` backend

skyrl/backends/ray_jax.py
andrewsykim: Added a `_get_random_port` function to assign a random port, we can add the override if the random port doesn't work (can't see why it wouldn't though)

skyrl/backends/ray_jax.py
andrewsykim: updated to raise exception if num_processes is not set

skyrl/tinker/engine.py
```diff
@@ -155,12 +155,18 @@ def prepare_model_pass_batch(
     )
 
 
-def get_backend_classes(backend_name: str):
+def get_backend_classes(backend_name: str, use_ray: bool = False):
     """Lazy import backends to avoid importing unused backend dependencies (e.g., JAX, Ray)."""
     if backend_name == "jax":
-        from skyrl.backends.jax import JaxBackend, JaxBackendConfig
+        if use_ray:
```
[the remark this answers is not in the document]
andrewsykim: updated the error message

andrewsykim:
I was able to test this on TPUs using the following commands:

To start the tinker API server, engine and JAX workers
```
ray job submit -- sh -c $'uv run --python 3.12 --extra tpu --extra jax --extra tinker --extra ray -m skyrl.tinker.api --base-model Qwen/Qwen3-0.6B --backend jax  --backend-config \'{"use_ray": true, "ray_actor_options": {"resources": {"TPU": 4}}, "ray_pg_bundles": [{"CPU": 4, "TPU": 4},{"CPU": 4, "TPU": 4},{"CPU": 4, "TPU": 4},{"CPU": 4, "TPU": 4}], "sample_max_num_sequences": 256, "train_micro_batch_size": 32, "tensor_parallel_size": 4, "fully_sharded_data_parallel_size": 4, "num_processes": 4}\''
```

To run RL training:
```
ray job submit -- sh -c 'cd SkyRL && export TINKER_API_KEY=tml-dummy && uv run --with "tinker-cookbook[math-rl] @ git+https://github.com/thinking-machines-lab/tinker-cookbook.git@nightly" python -m tinker_cookbook.recipes.math_rl.train base_url=http://localhost:8000 model_name="Qwen/Qwen3-0.6B" group_size=4 groups_per_batch=100 learning_rate=1e-4 max_tokens=512 save_every=0'
```

skyrl/backends/ray_jax.py
```diff
@@ -0,0 +1,192 @@
...
+
+from skyrl.backends.backend import AbstractBackend
+from skyrl.backends.jax import JaxBackendConfig, JaxBackendImpl
+from skyrl.tinker import types
+from skyrl.utils.log import logger
+
+
+def _get_random_port() -> int:
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.bind(("", 0))
```
pcmoritz, line 14: This might cause a race condition if multiple processes do something like this at the same time (get a port, then release the port and re-use the port number). It might be more robust to either try a few random ports from a range and return a free one.

pcmoritz (approved):

skyrl/backends/ray_jax.py
andrewsykim: I don't think the race condition will happen because `_get_random_port` is only called by index 0 JAX worker during initialization.

pcmoritz: Ok, happy to merge it! If there is any other process on the machine that allocates ports, it can happen, but we can also fix it if it actually happens :)

andrewsykim: hmm good point, I didn't consider that. I can open a follow-up PR for this :)

andrewsykim: Opened https://github.com/NovaSky-AI/SkyRL/pull/1652

PR #1464, opened 2026-04-06
[skyrl] vLLM Renderer for rendering Multi-Modal ModelInputChunks for training backend

nithinvc:
### Summary

This PR introduces a vLLM renderer which used by the SkyRL backend to convert `ModelInputChunk`s into tokenized text, `pixel_values`, and `image_grid_thw`. The scope of this PR is limited to the actual renderer implementation, not adding to the backend yet.

- Add `VLLMRenderer`.
- Add `RenderedImage` NamedTuple return type from `_render_images`.
- Reuse `render_model_input` for the text-only fast path in `_render_single` instead of duplicating the token concatenation logic.
- Fix `RenderedModelInput.multi_modal_kwargs` type annotation from `dict[str, bytes]` to `dict[str, list[str]]` to match actual usage (list of base64-encoded strings per modality key).
- Add unit tests for `VLLMRenderer` with a mocked `RemoteInferenceClient` (text-only, image-only, mixed, error cases).
- Add GPU CI integration test (`test_vlm_renderer.py`) that exercises the full renderer against a real VLM via `InferenceEngineState`.
- Gate all VLM render tests behind `SKYRL_LOCAL_VLLM=1` since they depend on a local vLLM fork with `/v1/chat/completions/render` support that is not yet upstreamed.

### Test plan

- [x] Unit tests pass: `python -m pytest tests/backends/skyrl_train/test_renderer.py -v` (8 tests, no GPU required)
- [x] GPU CI integration tests: `SKYRL_LOCAL_VLLM=1 uv run --extra fsdp --extra dev --extra tinker pytest tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_vlm_inference_generation.py -m vllm -v`

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1464" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/skyrl/backends/renderer.py
+++ b/skyrl/backends/renderer.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import base64
-from typing import TYPE_CHECKING, NamedTuple, Tuple, Union
+from typing import TYPE_CHECKING, NamedTuple, Union
 
 import torch
 
@@ -11,6 +11,7 @@ from skyrl.tinker.types import (
     ImageAssetPointerChunk,
     ImageChunk,
     ModelInput,
+    MultiModalKwargs,
     MultiModalPlaceholder,
     RenderedModelInput,
 )
@@ -31,15 +32,18 @@ def render_model_input(model_inputs: list[ModelInput]) -> list[RenderedModelInpu
     ]
 
 
-def decode_mm_kwargs(rendered: RenderedModelInput) -> Tuple[torch.Tensor, torch.Tensor]:
-    """Decode a RenderedModelInput's multi_modal_kwargs into vision tensors.
+def decode_mm_kwargs(mm_kwargs: dict[str, list[str]] | None) -> MultiModalKwargs:
+    """Decode raw base64-encoded multimodal kwargs from vLLM into vision arrays.
+
+    Args:
+        mm_kwargs: Raw kwargs dict from vLLM's render endpoint, e.g. {"image": [b64_str, ...]}.
 
     Returns:
-        (pixel_values, image_grid_thw) — concatenated across all images in this
-        sample.  Returns empty tensors when the sample has no vision data.
+        MultiModalKwargs with decoded and concatenated pixel_values and image_grid_thw.
+        Values are None when no vision data is present.
     """
-    if not rendered.multi_modal_kwargs or "image" not in rendered.multi_modal_kwargs:
-        return torch.empty(0), torch.empty(0, 3, dtype=torch.long)
+    if not mm_kwargs or "image" not in mm_kwargs:
+        return MultiModalKwargs(pixel_values=None, image_grid_thw=None)
 
     from vllm.entrypoints.serve.disagg.mm_serde import (
         decode_mm_kwargs_item as _vllm_decode,
@@ -47,7 +51,7 @@ def decode_mm_kwargs(rendered: RenderedModelInput) -> Tuple[torch.Tensor, torch.
 
     pv_parts: list[torch.Tensor] = []
     thw_parts: list[torch.Tensor] = []
-    for b64_str in rendered.multi_modal_kwargs["image"]:
+    for b64_str in mm_kwargs["image"]:
         item = _vllm_decode(b64_str)
         data = item.get_data()
         if "pixel_values" in data and isinstance(data["pixel_values"], torch.Tensor):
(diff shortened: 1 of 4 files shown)


skyrl/backends/renderer.py
```diff
@@ -1,10 +1,171 @@
...
+    for b64_str in rendered.multi_modal_kwargs["image"]:
+        item = _vllm_decode(b64_str)
+        data = item.get_data()
+        if "pixel_values" in data and isinstance(data["pixel_values"], torch.Tensor):
+            pv_parts.append(data["pixel_values"])
+        if "image_grid_thw" in data and isinstance(data["image_grid_thw"], torch.Tensor):
+            thw_parts.append(data["image_grid_thw"])
+
+    pixel_values = torch.cat(pv_parts, dim=0) if pv_parts else torch.empty(0)
+    image_grid_thw = torch.stack(thw_parts, dim=0) if thw_parts else torch.empty(0, 3, dtype=torch.long)
```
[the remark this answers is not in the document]
nithinvc: Having tested this, vllm only returns shape [3]. However, this is a good call that we should standardize it / make this more robust.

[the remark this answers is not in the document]
nithinvc: standardized with a shape check before a cat along the first dim

tests/backends/skyrl_train/test_renderer.py
```diff
@@ -0,0 +1,142 @@
...
+    client.render_chat_completion = AsyncMock()
+    return client
+
+
+def _make_text_input(token_lists: list[list[int]]) -> ModelInput:
+    return ModelInput(chunks=[EncodedTextChunk(tokens=toks) for toks in token_lists])
+
+
+def _make_image_chunk(fmt: str = "jpeg") -> ImageChunk:
+    return ImageChunk(data=b"\xff\xd8\xff\xe0", format=fmt)
```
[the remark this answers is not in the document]
nithinvc: fair enough, changed

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_vlm_inference_generation.py
```diff
@@ -1,24 +1,38 @@
...
+import torch
 from PIL import Image
+from transformers import AutoTokenizer
 
+from skyrl.backends.renderer import VLLMRenderer, decode_mm_kwargs
+from skyrl.tinker.types import EncodedTextChunk, ImageChunk, ModelInput
 from skyrl.train.config import SkyRLTrainConfig
 from tests.backends.skyrl_train.gpu.utils import InferenceEngineState
 
+requires_local_vllm = pytest.mark.skipif(
```
nithinvc, line 28: All of the current vision language rendering and generation tests require a local vLLM install from main till the next vllm release

skyrl/tinker/types.py
```diff
@@ -130,7 +130,7 @@ class MultiModalPlaceholder(BaseModel):
 
 class RenderedModelInput(BaseModel):
     prompt_ids: list[int]
-    multi_modal_kwargs: dict[str, bytes] | None = None
+    multi_modal_kwargs: dict[str, list[str]] | None = None
```
pcmoritz: Would it make sense to make this a `TypedDict`? To make it easier to understand which keys can be there (even if they are optional)

skyrl/backends/renderer.py
```diff
@@ -1,10 +1,172 @@
...
+    """Text-only renderer. Concatenates token chunks."""
     return [
         RenderedModelInput(
             prompt_ids=[tok for chunk in mi.chunks for tok in (chunk.tokens if hasattr(chunk, "tokens") else [])]
         )
         for mi in model_inputs
     ]
+
+
+def decode_mm_kwargs(rendered: RenderedModelInput) -> Tuple[torch.Tensor, torch.Tensor]:
```
pcmoritz: It seems more natural to me to put only the multi_modal_kwargs as an argument (which will go well with the suggestion below to introduce a typeddict for it), since that's the only part of the renderedmodelinput that is used here

skyrl/tinker/types.py
nithinvc: That makes sense to me. Added it as a typed dict, but I kept the value typing as `any` instead of a `torch.Tensor` since `types.py` is also used by the jax backend.

skyrl/backends/renderer.py
nithinvc: Updated!

PR #1467, opened 2026-04-07
[feat][train] Add prefill-decode disaggregation support and refactor vLLMRouter

SumanthRH:
# What does this PR do?

1. Adds P/D disaggregation support for the new inference codepath, enabling separate prefill and decode server groups with NIXL-based KV transfer. Initial support at the `vLLMServerActor` layer was already present from #899 
2. Refactor `VLLMRouter` from subprocess CLI to use `vllm_router`'s python entrypoints (`RouterArgs` and `Router) directly.
3.  Fix ServerGroup creation to one group per engine (each with data_parallel_size servers). Previously, creation was incorrect - we were creating one ServerGroup with `num_engines` servers and ignoring data parallel size.

## Implementation choices

- Contiguous placement for prefill and decode: Prefill servers are placed first and then we move to decode servers. While our routing strategy is currently unaware of physical placement and just routes to the appropriate prefill / decode server based on consistent hashing, bundling the prefill and decode servers together is better practice to enable us to also run them as separate deployments in the future. 
- Initialization:  We add new configuration arguments `enable_pd`, `num_prefill` . For kv connector configuration, we simply make use of `engine_init_kwargs` to avoid introducing an additional intermediate dataclass. 
- Standardize on Nixl: Currently we only implement support for Nixl. Ideally I want to provide one happy path for users at the infrastructure layer and as much flexibility on the algorithmic layer. We can revisit this choice later. 

## Test Plan

- [x] Unit test `vLLMRouter` for supporting P/D
- [x] Unit test for engine generation with P/D in `test_engine_generation.py`
- [x] Unit test for weight syncing with P/D in `tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_weight_sync.py`
- [x]  E2E example with P/D enabled

### Test Results

GSM8K E2E test:

Screenshot 2026-04-06 at 11 07 56 PM
Screenshot 2026-04-06 at 11 08 19 PM

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1467" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/examples/train/gsm8k/run_gsm8k_pd.sh
+++ b/examples/train/gsm8k/run_gsm8k_pd.sh
@@ -7,7 +7,7 @@ set -x
 # export WANDB_API_KEY=<your_key_here>
 # bash examples/train/gsm8k/run_gsm8k_pd.sh
 
-# Minimum 4 GPUs: 2 prefill + 2 decode engines, non-colocated with training.
+# Requires 8 GPUs: 2 prefill + 2 decode engines, non-colocated with 4 training workers.
 
 : "${DATA_DIR:="$HOME/data/gsm8k"}"
 : "${NUM_GPUS:=4}"


skyrl/backends/skyrl_train/inference_servers/vllm_server_actor.py
CharlieFRuan, line 1: Line 24 should be `from vllm.inputs import TokensPrompt` instead of `from vllm.inputs.data import TokensPrompt`

Did a run of the run_gsm8k_pd.sh and it ran into

```
    from vllm.inputs.data import TokensPrompt
ModuleNotFoundError: No module named 'vllm.inputs.data'
```

examples/train/gsm8k/run_gsm8k_pd.sh
CharlieFRuan, line 1: On a 8xH100s, running

```bash
bash examples/train/gsm8k/run_gsm8k_pd.sh trainer.placement.colocate_all=false
```

runs into

```bash
  File "/tmp/ray/session_2026-04-08_06-53-26_977462_8912/runtime_resources/working_dir_files/_ray_pkg_390f81284f4569ee/skyrl/backends/skyrl_train/inference_servers/server_group.py", line 152, in _get_bundle_indices_for_server
    return [self._reordered_bundle_indices[logical_base + k] for k in range(gpus)]
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
IndexError: list index out of range
```

Can we add a GPU unit test for non-colocated setup? 1 P, 1D, 2 train, perhaps

skyrl/backends/skyrl_train/inference_servers/vllm_server_actor.py
SumanthRH: Fixed by #1482

examples/train/gsm8k/run_gsm8k_pd.sh
SumanthRH: Done! I've added non-colocated test for weight sync with P/D . 1P 1D and 1 train

PR #1477, opened 2026-04-08
feat: add max_tokens_per_microbatch config for token-based micro-batching

erictang000:
### Summary

Adds **token-based micro-batching**: instead of splitting a batch into fixed *sample-count* microbatches (`micro_*_batch_size_per_gpu`), microbatches are formed by **bin-packing samples up to a token budget** (`max_tokens_per_microbatch`). This balances real token counts across microbatches, improving throughput and memory utilization when sequence lengths are highly variable. It applies to both the forward (logprobs/values) and forward_backward (training) paths, on both the **FSDP** and **Megatron** backends.

Along the way it fixes several correctness issues that surface on the Megatron/THD path, and makes the pre-training "old logprobs" forward pack identically to the training step so the PPO ratio is exact at the first inner step.

### Config

- **`trainer.max_tokens_per_microbatch: int = -1`** — `> 0` enables token-based batching; `-1` keeps existing sample-count behavior. It is a *soft* cap: sequences are never split, so a single sequence longer than the cap gets its own (over-cap) microbatch. Documented in the config with guidance to set it `>= max_prompt_length + max_generate_length`.
- **`trainer.recompute_old_logprobs_per_minibatch: bool = True`** — compute the policy old-logprobs (and critic values) per mini-batch, matching the training step's mini-batch + DP partition (see Forward/backward packing consistency below).

### How it works

`worker_utils.py` gains a small iterator hierarchy behind `get_microbatch_iterator(...)`:

- `balanced_binpacking(token_counts, max_tokens)` — greedy least-loaded bin-packing over per-sample token counts (from `attention_mask`).
- `TokenBasedBatchIterator` — packs samples into microbatches; **synchronizes the microbatch count across DP ranks** by appending loss-neutral *padding microbatches* (so every rank runs the same number of forward passes, required by Megatron's pipeline schedule), and reorders/strips them on the way out.
- `SampleBasedBatchIterator` / `BatchIterator` — the existing fixed-count path, refactored under a shared `BaseBatchIterator`.

### Correctness fixes (mostly Megatron/THD-specific)

- **Multi-modal support:** `TensorList.__getitem__` now supports tensor/list fancy indexing, so `pixel_values` / `image_grid_thw` are gathered correctly when forming token-based microbatches.
- **Padding microbatch shape/device:** padding microbatches use the real `seq_len` (uniform across microbatches, as Megatron requires) and are built on the data's device — fixes a CPU/CUDA mismatch and an `IndexError` in `postprocess_packed_seqs`.
- **Single-token dummy rows:** rows added to equalize batch size within a microbatch use a one-hot `attention_mask` instead of all-ones — keeps both the packed (no zero-length `cu_seqlens` segment) and dense (no fully-masked-row NaN) paths valid while skipping nearly all of the dummy's compute.
- **`_pad_microbatch_to_size`** consolidated and hoisted to the base worker so ref/critic workers (which share `_forward_logprobs`) get it too.
- **PP > 1 forward:** the token-based reorder is skipped on non-last pipeline stages (which only return a placeholder), fixing an `IndexError`.
- **Metric dilution:** fully-padding microbatches are excluded from metric aggregation, so `loss_metrics/clip_ratio`, `policy_entropy`, etc. are no longer dragged toward `0`.
- **KL/entropy normalization:** the KL/entropy term is normalized by the count of *real* microbatches (`num_real_microbatches`) rather than the padded total, so regularization strength isn't diluted by padding microbatches at DP > 1.

### Forward/backward packing consistency

The pre-training forward (old logprobs) ran over the *full batch* while training runs *per mini-batch*, so the two used different DP partitions and THD packing — making `old_logprobs` differ from the epoch-0 recomputed logprobs (PPO ratio ≠ 1) on the packed path.

`recompute_old_logprobs_per_minibatch` makes the forward use the **same mini-batch + DP partition** as training via a new `WorkerDispatch.forward_from_staged` (mirroring `forward_backward_from_staged`, reusing `stage_data`), so packing — and thus logprobs — match exactly. The trainer helper is `_execute_forward_pass`.

### Observability

New per-step metrics when token batching is on: **`num_microbatches`** and **`num_padding_microbatches`** (averaged across DP), for both backends.

### Tests

- **CPU unit tests** (`test_token_based_batching_utils.py`): `balanced_binpacking` (incl. oversized-sequence behavior), `TokenBasedBatchIterator`, multi-modal `TensorList` gathering, padding-microbatch shape / `num_padding_microbatches`.
- **GPU tests** (`test_token_based_batching.py`): FSDP/Megatron forward_backward + loss-equivalence vs sample-based (packed and dense, DP=1 and DP=2 with padding microbatches), the new metrics, and `test_megatron_per_minibatch_forward_matches_forward_backward` validating that per-mini-batch forward matches training packing while full-batch forward diverges (DP=2, packed).


skyrl/train/config/config.py
```diff
@@ -660,6 +660,22 @@ class TrainerConfig(BaseConfig):
...
+    """Maximum number of tokens per microbatch. When > 0, microbatches are formed by bin-packing
+    samples based on their token counts (from attention_mask) instead of using a fixed sample count.
+    -1 means disabled (use sample-based micro_train_batch_size_per_gpu / micro_forward_batch_size_per_gpu).
+    Applies to both forward and training micro-batching.
+
+    NOTE: this is a *soft* cap. Sequences are never split across microbatches, so a single sequence
+    longer than ``max_tokens_per_microbatch`` is placed alone in its own microbatch that exceeds the
+    cap (no error, no truncation). The true peak microbatch size is therefore
+    ``max(max_tokens_per_microbatch, longest_sequence_in_batch)``."""
+    recompute_old_logprobs_per_minibatch: bool = True
```
erictang000, line 706: for context on this see: https://wandb.ai/sky-posttraining-uc-berkeley/max_tokens_per_microbatch_megatron?nw=nwusereatang

We can see that when recomputing logprobs using the same mini batch + DP partition as the training step (and hence the same token microbatching strategy, since the token microbatching is deterministic), the clip ratio is noticeable lower, and the minibatching doesn't incur a significant time overhead.

Keeping the option to fall back to just DP sharding the whole batch without mini batching for backwards compatibility, but I think it's better for numerics to have this true by default

Image

erictang000:
all new gpu ci tests passing (`test_token_based_batching.py`)

image

erictang000:
# Megatron Experiments

All experiments run for token based microbatching run with `861e005b48489b41eb8f4cae6b5f1c9a04be63ad` as anchor on main.

## Qwen3-1.7B DAPO
https://api.wandb.ai/links/sky-posttraining-uc-berkeley/6nwz9a69
### Accuracy

Green + dark brown are with token microbatching - green is additionally with fwd using the same minibatching as train. Eval accuracy is slightly lower in both cases but within variance, and other metrics are relatively stable.

image
image

### Speedups
Roughly 2x speedup in train from using train_micro_batch_size=4 vs max_tokens_per_microbatch=8000 (max context length is 10K)
image



## Qwen3-30B-A3B DAPO
https://wandb.ai/sky-posttraining-uc-berkeley/max_tokens_per_microbatch_dapo

### Accuracy
Looks matching with baseline run on same commit

imageimage

Both new baseline + new max_tokens_per_microbatch does worse than previous baseline - looking into see if there has been a regression on MoE megatron perf since the prompt_mean commit in #1719 - seems like there might be a minor one (perhaps due to numerics from #1751)

image

### Speedups
3x speedup from using micro_train_batch_size=2 vs max_tokens_per_microbatch=16000
image

erictang000:
# FSDP Experiments

## Regression after #1679 
Issue detailed in #1767 on training stability

## Qwen3-1.7B DAPO
https://wandb.ai/sky-posttraining-uc-berkeley/max_tokens_per_microbatch_dapo
Seems to roughly match a baseline that also has training instabilty, but does see eval accuracy drop significantly lower.

image

### Speedup
Roughly 2x speedup from micro_train_batch_size=4 vs max_tokens_per_microbatch=20000
image

skyrl/train/config/config.py
erictang000: also see: https://github.com/THUDM/slime/issues/1964#issuecomment-4598051449

SumanthRH (commented):
@erictang000 slowly making my way through the PR. Meanwhile let's add results for a PP > 1 run.

skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
```diff
@@ -838,19 +994,35 @@ def forward_backward(
...
 
         seq_len = micro_buffer[0]["sequences"].shape[1]
-        micro_bsz = micro_buffer[0]["sequences"].shape[0]
+
+        if use_token_batching:
+            # With token-based batching, microbatches may have different batch sizes.
+            # Megatron's forward_backward_func requires uniform micro_batch_size,
+            # so pad all microbatches to the max batch size across microbatches.
+            max_micro_bsz = max(m["sequences"].shape[0] for m in micro_buffer)
+            micro_buffer = [self._pad_microbatch_to_size(m, max_micro_bsz) for m in micro_buffer]
```
SumanthRH, line 1025: WHy is padding to the same microbatch size needed if we're using `remove_microbatch_padding` and packing all samples in a microbatch in the megatron worker? That would make the micro batch size for all microbatches 1 right? 

I'd imagine all we need is padding microbatches to equalize the number of microbatches, and no padding within a microbatch. 

Is this to support `remove_microbatch_padding=False`?

skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
```diff
@@ -881,10 +1053,16 @@ def forward_backward(
...
+        for m_batch, metrics in zip(micro_buffer, metrics_list):
             # Extract loss_fn_outputs before reduce_metrics (it's not a scalar metric)
             if "loss_fn_outputs" in metrics:
                 all_loss_fn_outputs.extend(metrics.pop("loss_fn_outputs"))
+            # Skip fully-padding microbatches: their metrics (clip_ratio=0, policy_entropy=0,
+            # ...) are meaningless and would drag down the mean-reduced metrics. Summed
+            # metrics (e.g. policy_loss) are unaffected since padding contributes 0, but
+            # excluding them here keeps both reductions correct.
+            if m_batch["loss_mask"].sum().item() == 0:
+                continue
```
SumanthRH, line 1068: Great

tests/backends/skyrl_train/gpu/gpu_ci/test_token_based_batching.py
```diff
@@ -0,0 +1,601 @@
...
+async def test_fsdp_token_based_forward_backward(ray_init_fixture, worker_type):
+    """
+    Test that forward_backward with max_tokens_per_microbatch works correctly for FSDP.
+
+    Verifies:
+    1. Token-based batching runs without errors for both policy and critic
+    2. Returns valid metrics with expected keys
+    3. For policy: loss is close to sample-based baseline (both use pre-scaled advantages)
+    """
+    try:
```
SumanthRH: Nit: `try` with `ray.shutdown` in `finally` is not needed with the `ray_init_fixture`

skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
erictang000: resolved offline

for future context this is needed for megatron's pp schedule which requires each microbatch passed in to `forward_backward_func` to have equal [micro_batch_size, max_seq_len], prior to us removing padding/doing sequence packing inside the `forward_backward_func`

skyrl/backends/skyrl_train/workers/worker_utils.py
```diff
@@ -130,3 +120,303 @@ def batch_to_experience(batch: TrainingInputBatch):
...
+
+    def __iter__(self) -> Iterator[TrainingInputBatch]:
+        return iter(self._chunks)
+
+    def reorder_and_combine_batches(self, batches: List[TensorBatch]) -> TensorBatch:
+        """Concatenate output batches. No reordering needed for sample-based splitting."""
+        return TensorBatch.cat(batches)
+
+
+def balanced_binpacking(token_counts: List[int], max_tokens_per_microbatch: int) -> List[List[int]]:
```
SumanthRH: Could we add this as a new `SeqPacker` subclass and use it? 


https://github.com/NovaSky-AI/SkyRL/blob/baf1359cc466b3f7aa17c6c98296cd21c50f6252/skyrl/train/dataset/bin_packing.py#L130

tests/backends/skyrl_train/test_token_based_batching_utils.py
```diff
@@ -0,0 +1,215 @@
...
+        assert padding["sequences"].shape[1] == batch["sequences"].shape[1]
+        assert padding["attention_mask"].shape[1] == batch["attention_mask"].shape[1]
+        # Only a single token is marked valid (full seq_len for shape uniformity, but
+        # cheap to compute in the packed path).
+        assert padding["attention_mask"].sum().item() == padding["attention_mask"].shape[0]
+        assert padding["attention_mask"][:, 0].sum().item() == padding["attention_mask"].shape[0]
+        # Padding rows must not contribute to the loss.
+        assert padding["loss_mask"].sum().item() == 0
+
+    def test_multimodal_tensorlist_microbatching(self):
```
SumanthRH, line 202: have you tested `max_tokens_per_microbatch` with the VLM example script? 


multi-modal support for dynamic batching can be in a separate PR as well, but looks like it's already implemented here

erictang000: hmm yeah this was just flagged by gemini review so added a test for it, but haven't tested with VLM

i think would be good to leave in a separate PR

skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
```diff
@@ -482,6 +486,178 @@ def lora_pre_wrap_hook(model):
...
+                    "sub_seq_lengths": micro.get("sub_seq_lengths"),
+                }
+            )
+
+        if use_token_batching:
+            # Pad microbatches to uniform batch size for Megatron compatibility
+            max_micro_bsz = max(m["sequences"].shape[0] for m in micro_dicts) if micro_dicts else 1
+            for i, m in enumerate(micro_dicts):
+                micro_dicts[i] = self._pad_microbatch_to_size(m, max_micro_bsz)
+            mbs = max_micro_bsz
```
SumanthRH, line 548: Can we move this logic to the token based batch iterator actually? That seems like the right place instead of the worker doing this

tests/backends/skyrl_train/gpu/gpu_ci/test_token_based_batching.py
```diff
@@ -0,0 +1,601 @@
...
+        # Dense attention is bitwise-stable across groupings; packed/THD has
+        # ~1e-4 varlen-kernel fp noise from the shifted cu_seqlens layout.
+        tol = 1e-3 if remove_microbatch_padding else 1e-6
+        print(f"\nMegatron loss equivalence (remove_microbatch_padding={remove_microbatch_padding}, tol={tol}):")
+        for i, (r_baseline, r_token) in enumerate(zip(results_baseline, results_token)):
+            bl = r_baseline.metrics["policy_loss"]
+            tl = r_token.metrics["policy_loss"]
+            print(f"  Rank {i}: baseline={bl:.6f}, token-based={tl:.6f}, diff={abs(bl - tl):.6f}")
+            assert abs(bl - tl) < tol, f"Loss mismatch on rank {i}: {bl} vs {tl} (tol={tol})"
+
```
SumanthRH: Nit: It would be good to assert on other metrics as well here to ensure that padding microbatches haven't affected metrics

SumanthRH: I would add clip ratio because its magnitude is small and it can be sensitive to padding microbatches

skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
erictang000: hmm this is megatron specific though, i think it might be cleaner to have here rather than having two separate token batched iterators for FSDP and megatron

tests/backends/skyrl_train/gpu/gpu_ci/test_token_based_batching.py
erictang000: done

skyrl/backends/skyrl_train/workers/worker_utils.py
erictang000: done!

tests/backends/skyrl_train/gpu/gpu_ci/test_token_based_batching.py
erictang000: fixed

erictang000:
e2e ci fsdp: https://github.com/NovaSky-AI/SkyRL/actions/runs/27451681681

e2e ci megatron: https://github.com/NovaSky-AI/SkyRL/actions/runs/27451674130

SumanthRH (approved):

skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
SumanthRH: Yeah but you could actually just configure the TokenBasedIterator to have a `equalize_microbatch_size` parameter and set it to `True` only for megatron. 

This is okay for now

PR #1486, opened 2026-04-09
[train][multimodal][3/3] Add multi-turn VLM generator 

nithinvc:
## Summary

When reviewing, please diff against #1494 . Generator implementation for #1493 .

Adds `SkyRLVLMGymGenerator`, a VLM-specific generator subclass that handles multi-modal (text + image) observations from VisGym environments. Extends the base generator types to carry vision tensors through the training pipeline.

- Add `pixel_values` and `image_grid_thw` fields to `GeneratorOutput` and `TrajectoryOutput` so vision tensors flow through to the trainer
- Extend `SkyRLGymGenerator.generate()` to collect and forward vision features when trajectories contain them
- Implement `SkyRLVLMGymGenerator` with a "render delta" approach: the conversation (OpenAI-format messages with base64 images) is the source of truth, re-tokenized via `render_chat_completion` at each turn. Generated tokens retain original logprobs; observation tokens are sliced from the re-render and masked out (`loss_mask=0`)
  - While inefficient, this ensure correctness for now. 
- Simplifies the parent by dropping support for batched generation, step-wise trajectories, custom chat templates, and expert indices tracking
- Add E2E GPU test with a 2-turn `ColorSquareEnv` that verifies structural output, multimodal tensors, loss mask correctness, and semantic generation quality

## Test plan

- [x] Existing generator unit tests pass: `uv run pytest tests/train/generators/test_skyrl_gym_generator.py -v`
- [x] GPU E2E test passes (requires local vLLM): `SKYRL_LOCAL_VLLM=1 uv run --isolated --extra dev --extra fsdp pytest tests/backends/skyrl_train/gpu/gpu_ci/test_skyrl_vlm_gym_generator.py -m vllm -v`

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1486" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
+++ b/skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
@@ -68,6 +68,7 @@ import aiohttp
 from skyrl.backends.skyrl_train.inference_engines.base import (
     InferenceEngineInput,
     InferenceEngineOutput,
+    MultiModalFeatures,
 )
 from skyrl.env_vars import (
     SKYRL_GENERATE_CONCURRENCY_PER_ENGINE,
@@ -325,6 +326,7 @@ class RemoteInferenceClient:
             raise ValueError("n > 1 is not supported. Use `config.generator.n_samples_per_prompt` instead.")
 
         session_ids = input_batch.get("session_ids")
+        mm_features = input_batch.get("mm_features")
         get_logprobs = sampling_params.get("logprobs") is not None
 
         # Two semaphores decouple the generate and detokenize stages:
@@ -347,12 +349,14 @@ class RemoteInferenceClient:
                     prompt_token_ids=prompt_token_ids[idx],
                     sampling_params=sampling_params,
                     session_id=session_ids[idx] if session_ids and idx < len(session_ids) else None,
+                    mm_features=mm_features,
                 )
             async with gen_sem:
                 return await self._generate_single(
                     prompt_token_ids=prompt_token_ids[idx],
                     sampling_params=sampling_params,
                     session_id=session_ids[idx] if session_ids and idx < len(session_ids) else None,
+                    mm_features=mm_features,
                 )
 
         async def _throttled_detokenize(token_ids: List[int]) -> str:
@@ -380,6 +384,7 @@ class RemoteInferenceClient:
         prompt_token_ids: List[int],
         sampling_params: Dict[str, Any],
         session_id: Optional[Any],
+        mm_features: Optional[MultiModalFeatures] = None,
     ) -> Dict[str, Any]:
         """
         Generate completion for a single prompt.
@@ -400,11 +405,13 @@ class RemoteInferenceClient:
         # Use LoRA adapter name if one is active, otherwise use base model name
         effective_model = self.active_lora_name if self.active_lora_name else self.model_name
 
-        payload = {
+        payload: dict[str, Any] = {
             "sampling_params": sampling_params,
             "model": effective_model,
             "token_ids": prompt_token_ids,
         }
+        if mm_features:
+            payload["features"] = mm_features
 
         headers = {"Content-Type": "application/json"}
         if session_id:
--- a/skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
+++ b/skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
@@ -349,14 +349,14 @@ class RemoteInferenceClient:
                     prompt_token_ids=prompt_token_ids[idx],
                     sampling_params=sampling_params,
                     session_id=session_ids[idx] if session_ids and idx < len(session_ids) else None,
-                    mm_features=mm_features,
+                    mm_features=mm_features[idx] if mm_features and idx < len(mm_features) else None,
                 )
             async with gen_sem:
                 return await self._generate_single(
                     prompt_token_ids=prompt_token_ids[idx],
                     sampling_params=sampling_params,
                     session_id=session_ids[idx] if session_ids and idx < len(session_ids) else None,
-                    mm_features=mm_features,
+                    mm_features=mm_features[idx] if mm_features and idx < len(mm_features) else None,
                 )
 
         async def _throttled_detokenize(token_ids: List[int]) -> str:
--- a/skyrl/train/generators/base.py
+++ b/skyrl/train/generators/base.py
@@ -2,6 +2,8 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from typing import Any, Dict, List, Literal, Optional, TypedDict, Union
 
+import torch
+
 from skyrl.backends.skyrl_train.inference_engines.base import ConversationType
 
 TrainingPhase = Literal["train", "eval"]
@@ -43,6 +45,8 @@ class GeneratorOutput(TypedDict):
     rollout_expert_indices: Optional[List[List[List[List[int]]]]]  # [batch_size, seq_len, layer_num, topk]
     # Applicable only for step-wise training
     is_last_step: Optional[List[bool]]
+    pixel_values: Optional[List[torch.Tensor]]
+    image_grid_thw: Optional[List[torch.Tensor]]
 
 
 class MetricsOutput(TypedDict):
--- a/skyrl/train/generators/base.py
+++ b/skyrl/train/generators/base.py
@@ -45,6 +45,7 @@ class GeneratorOutput(TypedDict):
     rollout_expert_indices: Optional[List[List[List[List[int]]]]]  # [batch_size, seq_len, layer_num, topk]
     # Applicable only for step-wise training
     is_last_step: Optional[List[bool]]
+    # Applicable only for vision-language models
     pixel_values: Optional[List[torch.Tensor]]
     image_grid_thw: Optional[List[torch.Tensor]]
 
--- a/skyrl/train/generators/skyrl_gym_generator.py
+++ b/skyrl/train/generators/skyrl_gym_generator.py
@@ -12,6 +12,7 @@ from dataclasses import asdict, dataclass
 from typing import Any, Dict, List, Optional, Tuple, Union
 from uuid import uuid4
 
+import torch
 from loguru import logger
 from tqdm.asyncio import tqdm
 
@@ -51,6 +52,8 @@ class TrajectoryOutput:
     rollout_logprobs: Optional[List[float]]
     env_metrics: Dict[str, Any]
     rollout_expert_indices: Optional[List[List[List[int]]]] = None
+    pixel_values: Optional[torch.Tensor] = None
+    image_grid_thw: Optional[torch.Tensor] = None
 
 
 @dataclass
@@ -803,6 +806,14 @@ class SkyRLGymGenerator(GeneratorInterface):
             is_last_step = None
             out_trajectory_ids = None
 
+        has_vision_features = any(getattr(output, "pixel_values", None) is not None for output in all_outputs)
+        pixel_values = (
+            [getattr(output, "pixel_values", None) for output in all_outputs] if has_vision_features else None
+        )
+        image_grid_thw = (
+            [getattr(output, "image_grid_thw", None) for output in all_outputs] if has_vision_features else None
+        )
+
         if sampling_params is not None:
             # sampling params will be a dict in the format of the inference engine backend
             get_logprobs = sampling_params.get("logprobs", None) is not None
@@ -847,6 +858,9 @@ class SkyRLGymGenerator(GeneratorInterface):
             "rollout_expert_indices": rollout_expert_indices,
             "is_last_step": is_last_step,
         }
+        if has_vision_features:
+            generator_output["pixel_values"] = pixel_values
+            generator_output["image_grid_thw"] = image_grid_thw
 
         return generator_output
 
--- a/skyrl/train/generators/skyrl_vlm_generator.py
+++ b/skyrl/train/generators/skyrl_vlm_generator.py
@@ -52,10 +52,7 @@ class SkyRLVLMGymGenerator(SkyRLGymGenerator):
         inference_engine_client: RemoteInferenceClient,
         tokenizer,
     ):
-        # Parent stores as self.inference_engine_client and sets up
-        # generator_cfg, skyrl_gym_cfg, tokenizer, max_turns, env_executor, etc.
         super().__init__(generator_cfg, skyrl_gym_cfg, inference_engine_client, tokenizer)
-        self.inference_client = inference_engine_client
         logger.info("Initialized SkyRLVLMGymGenerator (VLM multi-modal generator)")
 
     def _validate_cfg(self, generator_cfg: GeneratorConfig):
@@ -70,8 +67,8 @@ class SkyRLVLMGymGenerator(SkyRLGymGenerator):
             )
 
     async def _render_conversation(self, conversation: ConversationType) -> RenderedConversation:
-        rendered = await self.inference_client.render_chat_completion(
-            {"json": {"model": self.inference_client.model_name, "messages": conversation}}
+        rendered = await self.inference_engine_client.render_chat_completion(
+            {"json": {"model": self.inference_engine_client.model_name, "messages": conversation}}
         )
         return RenderedConversation(prompt_ids=rendered["token_ids"], features=rendered.get("features", None))
 
@@ -156,7 +153,7 @@ class SkyRLVLMGymGenerator(SkyRLGymGenerator):
                 sampling_params=current_sampling_params,
                 mm_features=latest_features,
             )
-            engine_output = await self.inference_client.generate(engine_input)
+            engine_output = await self.inference_engine_client.generate(engine_input)
 
             gen_text = engine_output["responses"][0]
             gen_ids = engine_output["response_ids"][0]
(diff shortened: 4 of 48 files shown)


skyrl/train/generators/skyrl_vlm_generator.py
```diff
@@ -0,0 +1,232 @@
...
+from skyrl.train.generators.base import GeneratorOutput, TrajectoryID
+from skyrl.train.generators.skyrl_gym_generator import (
+    SkyRLGymGenerator,
+    TrajectoryOutput,
+)
+
+
+class RenderedConversation(TypedDict):
+    prompt_ids: list[int]
+    features: MultiModalFeatures
```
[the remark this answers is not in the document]
nithinvc: done

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -380,6 +383,7 @@ async def _generate_single(
         prompt_token_ids: List[int],
         sampling_params: Dict[str, Any],
         session_id: Optional[Any],
+        mm_features: Optional[Dict[str, Any]] = None,
```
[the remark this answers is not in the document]
nithinvc: done

skyrl/train/generators/skyrl_gym_generator.py
```diff
@@ -803,6 +806,10 @@ async def generate(self, input_batch: GeneratorInput, disable_tqdm: bool = False
             is_last_step = None
             out_trajectory_ids = None
 
+        has_pixel_values = any(output.pixel_values is not None for output in all_outputs)
+        pixel_values = [output.pixel_values for output in all_outputs] if has_pixel_values else None
+        image_grid_thw = [output.image_grid_thw for output in all_outputs] if has_pixel_values else None
```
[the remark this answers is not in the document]
nithinvc: fixed

skyrl/train/generators/skyrl_vlm_generator.py
```diff
@@ -0,0 +1,232 @@
...
+            raise ValueError(
+                "SkyRLVLMGymGenerator requires `use_conversation_multi_turn=True` "
+                "because multi-modal observations must be in separate user messages."
+            )
+
+    async def _render_conversation(self, conversation: ConversationType) -> RenderedConversation:
+        rendered = await self.inference_client.render_chat_completion(
+            {"json": {"model": self.inference_client.model_name, "messages": conversation}}
+        )
+        return RenderedConversation(prompt_ids=rendered["token_ids"], features=rendered.get("features", {}))
```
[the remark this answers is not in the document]
nithinvc: done

skyrl/train/generators/skyrl_vlm_generator.py
```diff
@@ -0,0 +1,232 @@
...
+        per_token_reward: List[float] = [0.0] * len(response_ids)
+        for reward, idx in per_step_rewards:
+            per_token_reward[idx] = float(reward)
+
+        # ── Deserialize vision tensors from the most recent render ────
+        mm_kwargs = decode_mm_kwargs((latest_features or {}).get("kwargs_data"))
+        pixel_values = mm_kwargs.pixel_values
+        image_grid_thw = mm_kwargs.image_grid_thw
+
+        assert pixel_values is not None, "Expected pixel_values in output"
```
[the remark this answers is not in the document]
nithinvc: removed

skyrl/train/generators/skyrl_vlm_generator.py
```diff
@@ -0,0 +1,232 @@
...
+        self,
+        generator_cfg: GeneratorConfig,
+        skyrl_gym_cfg: SkyRLGymConfig,
+        inference_engine_client: RemoteInferenceClient,
+        tokenizer,
+    ):
+        # Parent stores as self.inference_engine_client and sets up
+        # generator_cfg, skyrl_gym_cfg, tokenizer, max_turns, env_executor, etc.
+        super().__init__(generator_cfg, skyrl_gym_cfg, inference_engine_client, tokenizer)
+        self.inference_client = inference_engine_client
```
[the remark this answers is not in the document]
nithinvc: done

skyrl/train/generators/skyrl_vlm_generator.py
```diff
@@ -0,0 +1,232 @@
...
+            # 4. Append assistant message to conversation
+            conversation.append({"role": "assistant", "content": gen_text})
+
+            # 5. Track generated tokens (loss_mask=1)
+            response_ids.extend(gen_ids)
+            loss_mask.extend([1] * len(gen_ids))
+            if rollout_logprobs is not None:
+                rollout_logprobs.extend(gen_logprobs if gen_logprobs else [0.0] * len(gen_ids))
+
+            per_step_rewards.append((step_reward, len(response_ids) - 1))
```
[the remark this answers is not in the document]
nithinvc: done

skyrl/train/generators/skyrl_vlm_generator.py
```diff
@@ -0,0 +1,232 @@
...
+            if len(input_ids) > max_input_length:
+                stop_reason = "length"
+                break
+
+            # 2. Generate
+            engine_input = InferenceEngineInput(
+                prompt_token_ids=[input_ids],
+                session_ids=[session_id],
+                sampling_params=current_sampling_params,
+                mm_features=latest_features,
```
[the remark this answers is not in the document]
nithinvc: done

tests/backends/skyrl_train/gpu/gpu_ci/test_skyrl_vlm_gym_generator.py
```diff
@@ -0,0 +1,280 @@
...
+        )
+
+    def get_metrics(self) -> Dict[str, Any]:
+        return {"answers": self.answers}
+
+
+register(
+    id="color_square_test_env",
+    entry_point="tests.backends.skyrl_train.gpu.gpu_ci.test_skyrl_vlm_gym_generator:ColorSquareEnv",
+)
```
SumanthRH, line 118: Tests should ideally register and deregister envs at cleanup. We didn't do this properly before I think. let's add a `deregister` helper in skyrlgym's registration file for this? You can then run register + deregister in a autouse module-level fixture

skyrl/train/generators/base.py
```diff
@@ -43,6 +45,8 @@ class GeneratorOutput(TypedDict):
     rollout_expert_indices: Optional[List[List[List[List[int]]]]]  # [batch_size, seq_len, layer_num, topk]
     # Applicable only for step-wise training
     is_last_step: Optional[List[bool]]
+    pixel_values: Optional[List[torch.Tensor]]
```
SumanthRH, line 49: nit: Add a comment that these are for vlm - the previous comment makes it look like these are only for step wise

skyrl/train/generators/skyrl_vlm_generator.py
```diff
@@ -0,0 +1,206 @@
...
+                rollout_logprobs.extend(gen_logprobs if gen_logprobs else [0.0] * len(gen_ids))
+
+            if gen_ids:
+                per_step_rewards.append((step_reward, len(response_ids) - 1))
+
+            # 6. If episode continues, defer obs token extraction to next render
+            if not done:
+                conversation.extend(new_obs)
+                pending_obs_offset = len(input_ids) + len(gen_ids)
+
```
SumanthRH, line 179: isn't this book-keeping wrong for some chat templates?

if rendering for an assistant message changes depending on whether this message is the last assistant message or not, then `gen_ids` doesn't match the rendered token ids for `new_obs` in the new conversation. 


Example, with some thinking models like Qwen 3 (text-only), previous assistant messages have their thinking token stripped when you apply chat template. In this case, `gen_ids` will include thinking tokens, but once you add `gen_text` and `new_obs` to the conversation and re-tokenize, the thinking part in `gen_text` is stripped out. So `pending_obs_offset`  will be incorrect.

There are probably multimodal models where this holds true as well. 

We did this for getting obs tokens correctly for text-only: https://github.com/nithinvc/SkyRL/blob/da798404f9c99e4466bf61a56e5d9cec2247aeda/skyrl/train/generators/skyrl_gym_generator.py#L584-L593

tests/backends/skyrl_train/gpu/gpu_ci/test_skyrl_vlm_gym_generator.py
SumanthRH, line 1: It would be good to add a test asserting the exact return value of agent_loop here so that the concatenation logic is tested. The right way would be to have a CPU test with a dummy vllm endpoint. For rendering, we make use of a real HuggingFace renderer and we have a hardcoded expected result on the full token sequence. Ideally this is with a thinking model with a non-standard chat template

SumanthRH (changes requested):
Looking good! I have some questions on the concatenation logic for  `obs_tokens`

SumanthRH:
Q: Have you tested the generate entrypoint as well E2E ? It looks like that will just work given the current generator changes, but would be good to run E2E

https://github.com/NovaSky-AI/SkyRL/blob/66d401a1473a891925e69f6c7092b3f567986340/skyrl/train/entrypoints/main_generate.py#L2

tests/backends/skyrl_train/gpu/gpu_ci/test_skyrl_vlm_gym_generator.py
nithinvc: done, added a deregister function

skyrl/train/generators/base.py
nithinvc: done

skyrl/train/generators/skyrl_vlm_generator.py
nithinvc: Yeah, that makes sense. I got rid of the `pending_obs_offset` logic.  I switched to book-keeping the last render length `prev_render_len`. In the next agent loop iteration, we walk from `prev_render_len` looking for an EOS token. The renderer will add the EOS token to indicate the end of the previous turn. The observation tokens start from there.

Alternatively, we can break up the conversation history to do a `/render` call on individual observation chunks (like the code snippet linked). The downside here is that there will be more generator side book-keeping, especially keeping track of the multi-modal inputs. Each `render` call we would have to deserialize and build a GenerateRequest. This would have two steps:
1. Concatenate the serialized tensor data (deserialize, cat, serialize) 
2. Update the placeholder indexing returned by vllm since, from the endpoint's perspective, the new observation is the only observation.

I figured the `prev_render_len` approach is simpler, but let me know what you think

tests/backends/skyrl_train/gpu/gpu_ci/test_skyrl_vlm_gym_generator.py
nithinvc: Makes sense. I added some CPU tests in a file alongside the existing generator tests `tests/train/generators/test_skyrl_vlm_generator.py`

nithinvc:
> Q: Have you tested the generate entrypoint as well E2E ? It looks like that will just work given the current generator changes, but would be good to run E2E
> 
> https://github.com/NovaSky-AI/SkyRL/blob/66d401a1473a891925e69f6c7092b3f567986340/skyrl/train/entrypoints/main_generate.py#L2

Yes, but it does require the trainer changes in #1498

skyrl/train/generators/skyrl_vlm_generator.py
```diff
@@ -0,0 +1,216 @@
...
+            input_ids = rendered_conversation["prompt_ids"]
+            latest_features = rendered_conversation["features"]
+
+            # 1b. Flush pending obs tokens from the previous turn.
+            # Find the end of the previous turn's assistant message by
+            # scanning for the first EOS token from prev_render_len.
+            if prev_render_len is not None:
+                obs_offset = None
+                for i in range(prev_render_len, len(input_ids)):
+                    if input_ids[i] == self.tokenizer.eos_token_id:
```
SumanthRH: Hmm I guess this works for now because we pass `include_stop_str_in_output=True`: https://github.com/NovaSky-AI/SkyRL/blob/c2d6d4754cb999273122dea80b2870d564a3925c/skyrl/backends/skyrl_train/inference_engines/utils.py#L22

skyrl/train/generators/skyrl_vlm_generator.py
```diff
@@ -0,0 +1,216 @@
...
+                        obs_offset = i + 1
+                        break
+                if obs_offset is None:
+                    # If no EOS token found, assume no thinking tokens
+                    obs_offset = prev_render_len
+                    logger.warning("No EOS token found after prev_render_len; obs offset may be incorrect")
```
SumanthRH: actually instead of having a "sometimes wrong" codepath here, let's just error out loudly?

skyrl/train/generators/skyrl_vlm_generator.py
```diff
@@ -0,0 +1,216 @@
...
+            new_obs = env_step_output["observations"]
+            step_reward: float = env_step_output["reward"]
+            done = env_step_output["done"]
+
+            # 4. Append assistant message to conversation
+            conversation.append({"role": "assistant", "content": gen_text})
+
+            # 5. Track generated tokens (loss_mask=1)
+            response_ids.extend(gen_ids)
+            loss_mask.extend([1] * len(gen_ids))
```
SumanthRH: This is the logic for the models with a standard chat template - otherwise loss mask can't be fully append only. 

Ok let's just make this example to be for standard append only chat templates itself? I believe otherwise we'd need to go for step wise training or have custom chat templates

tests/train/generators/test_skyrl_vlm_generator.py
```diff
@@ -0,0 +1,327 @@
...
+    prompt = [[{"role": "user", "content": "a"}]]
+    input_batch: GeneratorInput = {
+        "prompts": prompt,
+        "env_extras": [{"answer": "4"}],
+        "env_classes": ["cpu_vlm_test_env"],
+    }
+    output: GeneratorOutput = await generator.generate(input_batch)
+
+    response_ids = output["response_ids"][0]
+    loss_masks = output["loss_masks"][0]
```
SumanthRH: Nit: this is `loss_mask` for a single entry, not `loss_masks`

skyrl/train/generators/skyrl_vlm_generator.py
SumanthRH, line 168: This is the logic for the models with a standard chat template - otherwise loss mask can't be fully append only. this is as such okay in this PR - this will lead to some off-policy ness in training for thinking models (we basically retain all the thinking tokens in all the turns - even tho we sampled by ignoring thining tokens for previous turns).

For proper support for thinking models, step wise training would be the way

tests/train/generators/test_skyrl_vlm_generator.py
```diff
@@ -0,0 +1,327 @@
...
+    SkyRLGymConfig,
+)
+from skyrl.train.generators.base import GeneratorInput, GeneratorOutput
+from skyrl.train.generators.skyrl_vlm_generator import SkyRLVLMGymGenerator
+from skyrl_gym.envs import register
+from skyrl_gym.envs.base_text_env import BaseTextEnv, BaseTextEnvStepOutput
+
+MODEL_NAME = "Qwen/Qwen3-0.6B"
+THINKING_PREFIX = "<think>\nmock thinking\n</think>\n\n"
+
```
SumanthRH: Let's just focus the test on non-thinking model actually. non-standard chat templates will need step wise training support

SumanthRH (changes requested):
Ok i realize there is more complexity with thinking models - loss_mask can't be made append only. Let's focus on regular non-thinking models for this PR. 

The best way to add support for thinking models would be to add step-wise training support for the SkyRLVLMGenerator, best done in a follow-up

skyrl/train/generators/skyrl_vlm_generator.py
nithinvc: switched to raising an error

tests/train/generators/test_skyrl_vlm_generator.py
nithinvc: renamed var to `loss_mask` but dict key is still `loss_masks`

skyrl/train/generators/skyrl_vlm_generator.py
nithinvc: I agree, this makes sense to me

skyrl/train/generators/skyrl_vlm_generator.py
nithinvc: yes - but I also think it works since it adds stop strings to intermediate turns. For example, for a generated:

```
<user> </user>
<assistant></EOS> 
```

Even if EOS is excluded during generation, I believe it's added back when we rerender the next turn to indicate the end of the assistant message:
```
<user></user>
<assistant></EOS>
<user></user>
```

nithinvc:
@SumanthRH Updated and removed thinking tests - everything focuses on standard tokenizers now.

skyrl/train/generators/skyrl_vlm_generator.py
SumanthRH: Actually because it stops at stop string, you will get an issue with `response_ids` being incorrect. Because `gen_ids` won't have EOS token

SumanthRH:
```
# stop string -> <stop>
gen_ids as tokens ->  [Hi, <stop>]
obs_tokens ->  [Hello]

response_ids = [Hi, <stop>, Hello]
expected -> [Hi, <stop>, EOS, Hello]
```

SumanthRH (approved):

skyrl/train/generators/skyrl_vlm_generator.py
SumanthRH, line 177: can we make sure to append EOS token to gen_ids incase it stoppped at a stop token @nithinvc ? use the same flag `append_eos_token_after_stop_str_in_multi_turn`

SumanthRH (approved):

PR #1506, opened 2026-04-13
[fix][train] Fix advantage response masking for step wise training 

### LINKED ISSUE #1492: [train] Incorrect advantage assignment for step_wise_trajectores in RayPPOTrainer?
I've been having some stability issues when training with step_wise_trajectories. I see in the RayPPOTrainer we are doing 

```
traj_ids = (
    torch.cat([torch.tensor([False], device=is_last_step.device), is_last_step[:-1]]).int().cumsum(dim=0)
)
num_groups = traj_ids[-1].item() + 1
assert num_groups == len(
    last_step_advantages
), f"number of groups {num_groups} doesn't match the number of trajectories as given by `is_last_step` {len(last_step_advantages)}. The `is_last_step` tensor is likely malformed"
advantages = last_step_advantages[traj_ids]
returns = last_step_returns[traj_ids]
```

[link](https://github.com/NovaSky-AI/SkyRL/blob/main/skyrl/train/trainer.py#L823)

But assigning `advantages = last_step_advantages[traj_ids]` doesn't seem quite right to me since `last_step_advantages` is a `scalar * response_mask` of the last step. This means the previous steps in a trajectory are getting an advantage with a different response mask. This is especially problematic if the last step `response_ids` is short and the prior steps see much longer obs_ids (thus there is virtually no overlap in action tokens due to the left padding). Shouldn't we be getting the advantage of the last step as a scalar and multiplying that by `data["response_mask"]`? 

When making this fix I'm seeing much better stability, but do let me know if I'm looking at this correctly. Also happy to open a small PR if that helps.

SumanthRH:
# What does this PR do?

Fixes #1492

### DIFF
--- a/skyrl/backends/skyrl_train/utils/ppo_utils.py
+++ b/skyrl/backends/skyrl_train/utils/ppo_utils.py
@@ -1064,6 +1064,8 @@ def compute_reinforce_plus_plus_outcome_advantage(
     Args:
         - token_level_rewards: Float[torch.Tensor, "batch_size seqlen"]
         - response_mask: Float[torch.Tensor, "batch_size seqlen"]
+        - return_raw_scores: bool — when True, skips response_mask multiplication
+          so raw [N, seqlen] advantages flow through for step-wise training.
 
     Returns:
         - advantages: Float[torch.Tensor, "batch_size seqlen"]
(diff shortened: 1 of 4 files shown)


SumanthRH:
Closing in favour of #1507

Issue #1501, opened 2026-04-13
[stepwise] AssertionError when trajectories have inconsistent number of steps


zhudotexe:
When training with stepwise mode on, if the generator produces trajectories with an inconsistent number of steps, it is possible for SkyRL to encounter an AssertionError when dispatching steps to the train workers. For example, when training with `train_batch_size=mini_batch_size=64, n=8`, I encountered this error:

```
File ".../python3.12/site-packages/skyrl/backends/skyrl_train/distributed/dispatch.py", line 191, in stage_chunks
    len(data) % mini_batch_size == 0
AssertionError: data batch size must be divisible by mini_batch_size, got 1032 and 512
```

In this case, most trajectories contained two steps, but some contained three, making the data batch size exceed 1024 (i.e., 64 * 8 * 2).

This seems to have been introduced by https://github.com/NovaSky-AI/SkyRL/pull/1376.

SkyRL Version: 706bba5852f8444f599428ae10c1628f006edf9c

CharlieFRuan:
Thanks for pointing it out! It should be fixed after this PR gets merged: https://github.com/NovaSky-AI/SkyRL/pull/1483 We will merge this in asap after more verifications

zhudotexe:
Ah, great to see that's already on the horizon! I'd be happy to help test out the PR since I was thinking I might need to implement something similar. Is there a particular area that could use an additional set of eyes?

CharlieFRuan:
@zhudotexe Merged! Apologies for the delay. https://github.com/NovaSky-AI/SkyRL/pull/1529

Step-wise training at HEAD will work now. You can see the commands I used to produce the search-r1 curves in the PR description.

CharlieFRuan:
Feel free to try it out again and let us know if there are other problems! Will close this issue for now.

PR #1507, opened 2026-04-13
[fix][train][step-wise] Broadcast step-wise advantage with each step's own response_mask

### LINKED ISSUE #1492: [train] Incorrect advantage assignment for step_wise_trajectores in RayPPOTrainer?

CharlieFRuan:
In step-wise training, `advantages = last_step_advantages[traj_ids]` copied the last step's already-masked tensor verbatim to every step of the trajectory. Because each step has its own right-aligned response_mask at its own token positions (prompts and responses vary in length across turns), the scalar advantage landed at the last step's token positions — which almost never overlap with the earlier step's own response positions. Earlier steps silently received advantages = 0 at their loss_mask=1 positions, so roughly (N-1)/N of step-samples are trained incorrectly.

Fix:
  - Call the estimator with an all-ones mask so it returns the scalar advantage at every position (outcome-based estimators produce `scalar * mask`).
  - Broadcast by indexing into `traj_ids`, then multiply by each step's own `response_mask` so values land at the right token positions.
  - Reject `step_wise_trajectories=True` + `gae`/`reinforce++` in `validate_cfg`: temporal estimators produce per-token advantages that the scalar-broadcast model can't represent, and the step-wise path only sees the last step's slice anyway (no cross-step credit).
  - Add regression test `test_calc_advantages_and_returns_step_wise_broadcast` with a realistic right-aligned batch, plus a parametrized `test_validate_cfg_step_wise_estimator_compatibility` covering all 5 built-in estimators.
  - Document the outcome-only constraint in the step-wise tutorial.

Fixes https://github.com/NovaSky-AI/SkyRL/issues/1492.

### Testing

https://wandb.ai/sky-posttraining-uc-berkeley/skyrl-search-padding/reports/PR-1285-search-r1-PAD-left-only-baseline-after-PR-and-step-wise---VmlldzoxNjE3MTA0OQ

Since we will run into https://github.com/NovaSky-AI/SkyRL/issues/1501 at HEAD, we make this hacky hot fix, mainly the `pad_batch()` change: https://github.com/CharlieFRuan/SkyRL/commit/f2783a48006b9b4207345202388818568ba1c0df

```py
    if self.cfg.generator.step_wise_trajectories:
        n_samples = self.cfg.generator.n_samples_per_prompt
        pad_target = max(
            self.cfg.trainer.policy_mini_batch_size * n_samples,
            self.cfg.trainer.critic_mini_batch_size * n_samples,
            dp_size,
        )
    else:
        pad_target = dp_size
    pad_size = math.ceil(training_input.batch_size / pad_target) * pad_target - training_input.batch_size
```

We compare against the red curve from https://github.com/NovaSky-AI/SkyRL/pull/1285

Note that the actual curve should converge slower than this after we merge in https://github.com/NovaSky-AI/SkyRL/pull/1483, in which each mini batch will truly be one optimizer step. Currently we can have multiple optimizer step due to the number of turns can inflate the number of sequences in each mini batch.

image

The reward curve being higher is likely due to:

1) The fix we apply in this PR
2) In the previous PR (red line), we silently strip the amount of data and only train `num_mini_batches = len(data) // mini_batch_size`, which will no longer be an issue with the `pad_target` hacky fix above, or properly fixed after https://github.com/NovaSky-AI/SkyRL/pull/1483

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1507" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/docs/content/docs/tutorials/step-wise-training.mdx
+++ b/docs/content/docs/tutorials/step-wise-training.mdx
@@ -183,6 +183,8 @@ for i, step_output in enumerate(trajectory_steps):
         rewards = [0.0] * len(step_output.response_ids)
 ```
 
+Note that SkyRL currently only supports trajectory-level reward for step-wise training. Therefore, the reward should be placed at the last step's last token, and all non-last-step rewards are ignored. We then use the last step's reward to estimate the advantage and broadcast it to previous turns. Because of this, you should only use outcome-based advantage estimators (`cfg.trainer.algorithm.advantage_estimator` in `grpo`, `rloo`, or `maxrl`); `reinforce++` and `gae` are rejected at config validation time.
+
 ### 4. Ensure Contiguous Ordering
 
 All steps of trajectory A must appear before any steps of trajectory B in the output lists:
--- a/skyrl/train/trainer.py
+++ b/skyrl/train/trainer.py
@@ -805,7 +805,7 @@ class RayPPOTrainer:
             # We use the last step of each trajectory to compute advantages and broadcast them to
             # all steps of that trajectory, so we ignore per-step rewards in step-wise training.
             # We pass an all-ones mask here so the estimator returns the scalar advantage at every
-            # position. The real per-step `response_mask` is re-applied on broadcast below. See issue #1492.
+            # position. The real per-step `response_mask` is re-applied on broadcast below.
             # Shapes:
             #   traj_ids, (batch_size,):         trajectory id per step (cumsum of shifted is_last_step)
             #   last_step_advantages/returns,
--- a/skyrl/train/trainer.py
+++ b/skyrl/train/trainer.py
@@ -798,19 +798,22 @@ class RayPPOTrainer:
         token_level_rewards = data["rewards"]
 
         if self.cfg.generator.step_wise_trajectories:
-            # Step-wise only supports outcome-based estimators (GRPO, RLOO, MAXRL); ensured by `validate_cfg`.
             is_last_step = data["is_last_step"].bool()
             index = np.array(data.metadata["uids"])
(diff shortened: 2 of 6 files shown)


SumanthRH (approved):

SumanthRH:
Thanks! Would be good to get a E2E run!

PR #1508, opened 2026-04-14
[ci] Add CI for supported models for megatron backend, and add supported model page to docs

erictang000:
Add round trip test with tiny model versions for full coverage of all models that we claim first class support for in SkyRL.

This PR adds test coverage for:
1. Qwen3
2. GLM-4.7/4.7-Flash

Also add a docs page listing supported models:
image

### Add more Models
Pending models to add to CI (roughly in order of priority):
1. Qwen3.5 (need to debug some issues with tiny model...)
2. Nemotron 3
3. Deepseek
4. GLM-5
5. Kimi-K2.5
### Add LoRA coverage
We can initialize non-zero LoRA adapters, sync, generate, forward, compare logprobs, sync, verify logprobs.


docs/content/docs/getting-started/supported_models.mdx
```diff
@@ -0,0 +1,29 @@
...
+
+## FSDP
+Any model that is supported via HuggingFace transformers + vLLM is supported via the FSDP backend. The FSDP backend also comes with support for Ulysses-style sequence parallelism,
+making it a viable option for training on dense models with up to 32B parameters and 32K/64K context lengths.
+
+## Megatron
+The [Megatron backend](/docs/examples/megatron) comes with support for 5D parallelism (DP/TP/CP/PP/EP) and allows for efficient scaling of large MoE models (100B+).
+Support for mapping from HuggingFace model definitions to Megatron GPTModels is provided by NVIDIA's [Megatron-Bridge](https://github.com/NVIDIA-NeMo/Megatron-Bridge) library.
+
+The following models are supported in SkyRL's Megatron backend via Megatron-Bridge + vLLM:
```
SumanthRH: There are a lot more models here right? What about all the older models like Qwen 2.5 that have support in Megatron-bridge and vLLM? 

It should be safe to sayt that for whatever model supported by both our Megatron-bridge commit and our vLLM version, we can run them (with some configuration tweaking ex; disabling sample packing)?

pyproject.toml
```diff
@@ -116,6 +116,7 @@ megatron = [
     "flash-attn==2.8.3; sys_platform == 'linux'",
     "flash-linear-attention; sys_platform == 'linux'",
     "causal-conv1d; sys_platform == 'linux'",
+    "mamba-ssm>=2.3.0; sys_platform == 'linux'",
```
SumanthRH, line 119: If this is the base `megatron` extra then we should update the nemotron 3 example:

https://github.com/NovaSky-AI/SkyRL/blob/e6550f2a19ca17007493bf4efab3564321b2cbe7/examples/train/nemotron_3/run_nemotron_3_nano_4b_gsm8k.sh#L31

docs/content/docs/getting-started/supported_models.mdx
erictang000: Hmm yeah qwen2.5 we can definitely add, but models like kimi/gpt-oss/gemma3/minimax, etc... which are supported in both i'm not sure we can be confident enough to say we support fully.

We can add a comment listing that additional models that are supported by megatron-bridge should theoretically be supported, and tell users to open an issue if they aren't.

tests/backends/skyrl_train/gpu/gpu_ci/megatron/test_megatron_models.py
```diff
@@ -0,0 +1,288 @@
...
+        loss_masks=generator_output["loss_masks"],
+    )
+
+
+@pytest.mark.asyncio
+@pytest.mark.megatron_models
+@pytest.mark.parametrize(
+    "tp,pp,cp,ep,etp,inference_tp,num_gpus,model_name,vllm_threshold,megatron_threshold",
+    [
+        pytest.param(2, 1, 1, 2, 1, 2, 4, "eatang/qwen3-moe-tiny-random", 1e-1, 2e-1, id="qwen3-moe_tp2_ep2"),
```
SumanthRH, line 150: how did we get these thresholds btw? I'd imagine that the logprobs drift for tiny models to be much smaller than the larger moe models, but these thresholds seem similar to what you'd see for the large moe models

erictang000: this was kind of by inspection, but i think the reason the values are higher is because the weights are randomly initialized, so the magnitude of the logprobs overall is higher

pyproject.toml
erictang000: fixed

docs/content/docs/getting-started/supported_models.mdx
erictang000: added a comment

PR #1512, opened 2026-04-14
[feat] chunked ipc support for new inference

hao-aaron:
While we wait for https://github.com/vllm-project/vllm/pull/39212 and https://github.com/vllm-project/vllm/pull/37476, support for chunked updates is still unavailable on vllm. In the meantime, this PR adds a new worker extension class for new inference that adds the relevant start/finish weight update api, allowing chunked weight updates for SkyRL IPC. We maintain both the old, one shot weight update api that NCCL uses, as well as exposing a new api that only IPC uses to do chunked weight update.

tested by running test_policy_local_engines_e2e.py and ensuring all tests pass


SumanthRH:
Looking good @hao-aaron ! I'm planning to get https://github.com/NovaSky-AI/SkyRL/pull/1476 in first, which also has the world size fixes for num_engines > 1 with DP. Feel free to leave comments on that PR

SumanthRH (changes requested):
can you unskip the full param test in `test_megatron_worker.py` ?

SumanthRH (approved):

PR #1515, opened 2026-04-15
IPC Chunked fix

hao-aaron:
While we wait for https://github.com/vllm-project/vllm/pull/39212 and https://github.com/vllm-project/vllm/pull/37476, support for chunked updates is still unavailable on vllm. In the meantime, this PR adds a new worker extension class for new inference that adds the relevant start/finish weight update api, allowing chunked weight updates for SkyRL IPC. We maintain both the old, one shot weight update api that NCCL uses, as well as exposing a new api that only IPC uses to do chunked weight update.


SumanthRH (changes requested):
@hao-aaron with this PR, we can enable the colocated megatron weight sync tests IMO. Can we test weight sync with megatron in a colocated setting with the new inference codepath and enable it in CI? 

Currently it is disabled: 

https://github.com/NovaSky-AI/SkyRL/blob/02be377857cf2b5e274617ddbdc771149c002527/tests/backends/skyrl_train/gpu/gpu_ci/test_megatron_worker.py#L124-L127

hao-aaron:
added tests and ensured they are passing

SumanthRH (approved):
LFG

PR #1550, opened 2026-04-21
[tinker] Fix optimizer offload in SkyRL train backend after weight sync

SumanthRH:


SumanthRH:
I ran `math_rl.train` from the tinker cookbook locally and it successfully finished 6 iterations with expected memory savings after `save_weights_for_sampler` calls

PR #1568, opened 2026-04-23
[feat][megatron] Add freeze_moe_router config to skip router parameters

SumanthRH:
# What does this PR do?

Adds `MegatronConfig.freeze_moe_router` (default False). When `True`, the Megatron policy/ref workers call the existing freeze_moe_router helper after model load and before optimizer construction, setting requires_grad=False on router parameters. The Megatron DistributedOptimizer reads requires_grad at construction, so frozen params are excluded from its reduce buckets and param groups.

Also fixed a bug:
- hasattr(x, "bias") returns True for nn.Linear(bias=False) where bias is None, matching Megatron's TopKRouter(add_bias_linear=False) default. Swap to getattr(x, name, None) is not None.

## Test plan

 New `tests/backends/skyrl_train/distributed/test_freeze_moe_router.py` covering default behaviour, testing the `bias=None` case and the wrapped models
 
 TODO:

- [x] E2E run


SumanthRH:
Made the PR ready for review by mistake and the bots are unforgiving

PR #1579, opened 2026-04-28
[feat] Multi-LoRA serving for RemoteInferenceClient

### LINKED ISSUE #1573: Enable multi-lora serving for vLLM in SkyRL for synchronous RL
# Summary

Tinker serves trained adapters to multiple tenants from a shared base model. Today, SkyRL's Tinker API implements multi-tenancy only in the Jax backend. As a part of supporting multi-lora training with the Tinker API with FSDP/Megatron, we need to add support for multi-lora serving for vLLM in SkyRL. 

The default inference codepath consists of three main components - `RemoteInferenceClient` , `VLLMRouter` and `vLLMServerActor`.  Multi-lora serving is already enabled with the `enable_lora` initialization parameter in vLLM. We've further configured `VLLM_ALLOW_RUNTIME_LORA_UPDATING=True` for updating lora weights at runtime - this already allows for dynamically loading/ unloading LoRA adapters. 

To enable multi-lora serving in SkyRL, we need to 

1. Support sampling from different lora adapters 
2. Support unloading a LoRA adapter (needed for multi-tenancy with the Tinker API)

## Rough implementation 

Overall we need to

1. Move away from the current single-lora / single-model bookkeeping in `RemoteInferenceClient`
2. Make `model` (model name) an explicit parameter for all generation / chat completion / sample methods for `RemoteInferenceClient` (this can be the base model or any of the active lora adapters)
3. Add a new `unload_lora_adapter` method for unloading a LoRA adapter

The interfaces will look something like
```
# client will maintain a set of active loras being trained internally
client = RemoteInferenceClient(server_urls=server_urls, enable_lora=True)

# generation - explicit model name
client.generate(model="skyrl-lora-1", input_batch)

# Load lora adapter for `lora_model_name` from disk
# rename the current `update_lora_from_disk` method to be consistent with vLLM
client.load_lora_adapter(lora_model_name, lora_path)

# Unload LoRA adapter 
client.unload_lora_adapter(lora_model_name)
```

Both `load_lora_adapter` and `unload_lora_adapter` will simply fan out requests to all servers. 

## Scope of changes

The changes will mostly be at the `RemoteInferenceClient` layer since LoRA serving related changes are already present in `VLLMServerActor`.

## Test strategy

Since multi-lora training is not yet enabled for FSDP / Megatron, we should write out unit tests for generation and weight syncing with multiple lora adapters and ensure they pass. 
1. Generation: Load 2 different lora adapters from disk, make interleaved generation requests to different adapters, ensure outputs are as expected
2. Weight syncing: This will be inference-only, no trainer initiated `broadcast_to_inference_engines`. Load lora adapters A and B from disk -> generate from both ->  issue a `load_lora_adapter` call for A -> generate again -> ensure generation for A changes, no changes for B

hao-aaron:
## Summary

Adds dynamic multi-LoRA serving to the new inference path: `RemoteInferenceClient` can now load/unload arbitrary LoRA adapters at runtime, route every data-plane call to a specific adapter (or the base model) via an explicit `model` parameter, and serve from multiple adapters concurrently.

## Key design decision: `model` is always explicit

Every data-plane method requires the caller to identify the target model — there is no implicit fallback to a client-side default.

| Method | How `model` is supplied |
|---|---|
| `generate(input_batch, model)` | Required keyword argument |
| `sample(payload)`, `chat_completion(payload)`, `completion(payload)`, `render_chat_completion(payload)` | Required field in the request body; missing/empty raises `ValueError` |

Callers know (or can resolve) which adapter they are addressing — there is no client-side guessing about which LoRA is "active". `RemoteInferenceClient.model_name` is now *only* the base model the vLLM server was started with; it is consumed internally by `tokenize`/`detokenize` (LoRA-agnostic) and is no longer auto-injected into data-plane requests.

## What's new

- **Control plane** on `RemoteInferenceClient`:
  - `load_lora_adapter(name, path, load_inplace=False)` — fans out to all backend servers via `POST /v1/load_lora_adapter`.
  - `unload_lora_adapter(name)` — symmetric `POST /v1/unload_lora_adapter`.
- **Config knobs** plumbed to vLLM:
  - `trainer.policy.model.lora.max_loras` (concurrent adapters per batch)
  - `trainer.policy.model.lora.max_cpu_loras` (CPU LRU cache size)
- **`SKYRL_LORA_ADAPTER_NAME`** promoted to a public module-level constant; FSDP and Megatron (unmerged) workers register their trained adapter under this name.

## Single source of truth for "what does the policy resolve to?"

Added `resolve_policy_model_name(cfg)` in `skyrl/backends/skyrl_train/inference_servers/utils.py`. It returns:

- `SKYRL_LORA_ADAPTER_NAME` when the worker registers a LoRA adapter on the inference engine (FSDP + LoRA, or Megatron + LoRA with `merge_lora=False`).
- `cfg.trainer.policy.model.path` otherwise (including Megatron + LoRA with `merge_lora=True`, where the engine receives merged base weights).

This is called **once at wiring time** and threaded through:

- `skyrl_train_backend._sample_with_remote_client`
- `SkyRLGymGenerator` / `SkyRLVLMGymGenerator` (new required `policy_model_name: str` constructor arg)
- `main_base.get_generator`
- `SkyRLBackend` in `skyrl-agent`

The `is_single_lora` ternary that previously decided `client.model_name` at construction is gone from both `skyrl_train_backend` and `main_base`.

## Breaking change: legacy inference path

The legacy `InferenceEngineClient` (Ray-wrapped) path is **not updated** for multi-LoRA. We assume `_SKYRL_USE_NEW_INFERENCE=1` (the current default) is in use everywhere. The legacy `LoraLoadRequest` smuggling path still works for the single-LoRA legacy flow, but anything calling the legacy client's `generate`/`sample`/etc. without `model=` plumbing will need updating before that path can serve LoRAs.

## Tests

- **Mock tests** (`tests/backends/skyrl_train/inference_servers/test_remote_inference_client.py`):
  - `TestLoRAControlPlane` — fan-out, in-place reload, conflict on `load_inplace=False`, fan-out unload, 404 on unknown unload.
  - `TestExplicitModelRequired` — every body-style method raises `ValueError` when `model` is missing; `generate` raises `TypeError` without the kwarg.
  - All existing + new mock tests pass.
- **Real-GPU multi-LoRA tests** (`tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_multi_lora_serving.py`):
  - `test_multi_lora_interleaved_generation` — load `lora-meow` + `lora-woof`, interleave per-call routing, verify each adapter's signature output.
  - `test_lora_inplace_reload_isolated` — in-place reload of one adapter must not perturb the other.


SumanthRH:
@hao-aaron would recommend updating the legacy inference codepath as well for now

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -320,6 +346,7 @@ async def _post(self, url: str, json: Dict[str, Any], headers: Optional[Dict[str
     async def generate(
         self,
         input_batch: InferenceEngineInput,
+        model: str,
```
SumanthRH: can do this;

1. RemoteInferenceCLient keeps track of whether we're training with LoRA , `use_lora`
2. model default is None: If None, 
-  use base model if `use_lora` is `False`. 
- If not, error loudly -> prevents an accidental call to base model with lora.

skyrl/train/config/ppo_base_config.yaml
SumanthRH, line 1: Revert this @hao-aaron

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_chat_template.py
SumanthRH, line 1: Why change this?

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_vlm_inference_generation.py
SumanthRH, line 1: again, let's not change this

tests/backends/skyrl_train/gpu/gpu_ci/test_lora.py
```diff
@@ -145,5 +153,9 @@ async def test_policy_local_engines_e2e(
         policy.offload_to_cpu()
         await client.wake_up(tags=["kv_cache"])
         await client.reset_prefix_cache()
-        outputs = await run_inference(client, get_test_prompts(MODEL), sampling_params)
+        # When vLLM has the LoRA adapter loaded, address it by name so this
+        # test actually exercises LoRA inference. For megatron+merge_lora the
+        # adapter isn't loaded — merged weights live under the base model.
+        inference_model = SKYRL_LORA_ADAPTER_NAME if needs_vllm_lora else None
```
SumanthRH: COuld reuse the helper `resolve_policy_model_name`

tests/backends/skyrl_train/gpu/gpu_ci/test_pause_and_continue_generation.py
SumanthRH, line 1: don't change this

SumanthRH (approved):
Left a few nits, gtg otherwise

tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_remote_inference_client_chat_template.py
hao-aaron: autofilled by new code, but it doesn't hurt to leave it in

PR #1596, opened 2026-04-30
[feat] Recursive language models (RLMs) environment

sox8502:
## Summary

Adds an RLM (Recursive Language Model) environment to SkyRL: a multi-turn task where the model drives a stateful Python REPL with access to its context, scaffold tools (FINAL/FINAL_VAR/SHOW_VARS), LM-query helpers, and recursive child RLM agents with their own REPLs.

W&B of RLM training run: https://wandb.ai/rehaan14-alphaxiv/rlm/reports/RLM-2b-4b-E2E-Runs--VmlldzoxNjc3OTM5OA?accessToken=mo7hf9stx3iplltm2yj1isezmig1f692vuaw8ubfkd7w8j3cu8gb7flp17468igc

## Key files

- [`skyrl-gym/skyrl_gym/envs/rlm/repl.py`](./skyrl-gym/skyrl_gym/envs/rlm/repl.py) (new) — PersistentREPL: stateful Python sandbox with restricted builtins (blocks `eval`/`exec`/`compile`), `SIGALRM` timeout + thread-based fallback for Ray workers. Injects scaffold names (`FINAL_VAR`, `SHOW_VARS`, `context`) and LM-query helpers (`llm_query`, `rlm_query`, batched variants); reserved names are restored after every `execute()` so the model can't shadow them. `rlm_query` dispatches through `subcall_fn` to spawn a child rollout and `ast.literal_eval`s the result.

- [`skyrl-gym/skyrl_gym/envs/rlm/env.py`](./skyrl-gym/skyrl_gym/envs/rlm/env.py) (new) — BaseRLMEnv: builds the system prompt, boots the REPL with the rollout's `context_text`, and runs the multi-turn loop. Parses the last ```repl block per turn and detects final answers via `FINAL_VAR()` inside exec or text-parsed `FINAL(...)`. Uses an **ephemeral user-prompt mechanism** — each turn injects a "think step-by-step…" message and pops the previous one off `chat_history` so per-turn instructions never accumulate. Three subclass hooks: `_get_reward`, `_get_system_prompt`, `_get_repl_tools`.

- [`examples/train/rlm/rlm_generator.py`](./examples/train/rlm/rlm_generator.py) (new) — RLMGymGenerator, a SkyRLGymGenerator subclass that plugs in via the two new hooks (no upstream reward code touched). Manages a parent/child rollout tree in `self.active_rollouts`; children carry a `_rlm_parent_rid` sentinel and link to parents at registration. Builds `lm_callback` (routes to a frozen OpenRouterInferenceEngine when `frozen_openrouter_model` is set, else the policy engine) and `subcall_fn` (spawns a child `agent_loop` and returns its `final_answer`). On root finalize, optionally inlines child step-outputs into the parent trajectory (`train_child_trajectories`). Requires `step_wise_trajectories=True`.

- [`skyrl/train/generators/skyrl_gym_generator.py`](./skyrl/train/generators/skyrl_gym_generator.py) (modified) — adds two no-op subclass hooks (`_setup_env_extras`, `_post_process_agent_loop_output`) into `agent_loop` so subclasses can extend env construction and post-reward output assembly without forking. Threads `loss_masks` into `get_rollout_metrics` for assistant-token stats and surfaces `env_metrics` in generator output.


SumanthRH (changes requested):
Looking great! I did a quick pass and noticed that you're using the legacy inference codepath, specifically a customized subclass of `RemoteInferenceEngine`  when `frozen_openrouter_model` is `True`.  

The legacy inference code in `inference_engines/` will be removed soon. Could you shift to the new `inference_servers/` codepath (the default)? In your case, the `OpenRouterInferenceEngine` seems to be a lightweight client - perhaps you can just make a custom class of `RemoteInferenceClient` for this? Also, for the sake of the RLM example, this custom inference client code should live in `examples/train/rlm/`, not in `backends/skyrl_train/inference_servers/`.  It would be great to also re-run your E2E RLM run to ensure that rewards are not changed.

SumanthRH (approved):
Sick

PR #1599, opened 2026-04-30
Add FP8 text-to-SQL training script; drop SQLEnv stop-tag assertion

hao-aaron:
- examples/train/text_to_sql/run_skyrl_sql_fp8.sh: colocated GRPO run with vLLM fp8 quantization, aligned with run_skyrl_sql.sh knobs.
- skyrl-gym: remove _validate_action; stop strings are enforced in the generator config instead of failing mid-rollout.

Wandb: https://wandb.ai/sky-posttraining-uc-berkeley/skyrlsql-quantized?nw=nwuserahao

### DIFF
--- a/examples/train/text_to_sql/run_skyrl_sql_fp8.sh
+++ b/examples/train/text_to_sql/run_skyrl_sql_fp8.sh
@@ -33,7 +33,7 @@ uv run --isolated --extra fsdp -m skyrl.train.entrypoints.main_base \
   data.train_data="['$DATA_DIR/train.parquet']" \
   data.val_data="['$DATA_DIR/validation.parquet']" \
   trainer.policy.model.path="Qwen/Qwen2.5-Coder-7B-Instruct" \
-  trainer.epochs=125 \
+  trainer.epochs=30 \
   trainer.placement.colocate_all=true \
   trainer.strategy=fsdp2 \
   trainer.policy.fsdp_config.cpu_offload=false \
@@ -53,9 +53,9 @@ uv run --isolated --extra fsdp -m skyrl.train.entrypoints.main_base \
   trainer.policy.optimizer_config.lr=1.0e-6 \
   trainer.policy_mini_batch_size=256 \
   trainer.algorithm.use_kl_loss=false \
-  trainer.ckpt_interval=-1 \
-  trainer.hf_save_interval=-1 \
-  trainer.dump_data_batch=false \
+  trainer.ckpt_interval=60 \
+  trainer.hf_save_interval=30 \
+  trainer.dump_data_batch=true \
   generator.inference_engine.backend=vllm \
   generator.inference_engine.run_engines_locally=true \
   generator.inference_engine.weight_sync_backend=nccl \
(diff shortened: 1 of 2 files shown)


examples/train/text_to_sql/run_skyrl_sql_fp8.sh
```diff
@@ -0,0 +1,84 @@
...
+
+uv run --isolated --extra fsdp -m skyrl.train.entrypoints.main_base \
+  trainer.algorithm.off_policy_correction.tis_ratio_type=$TIS_TYPE \
+  trainer.algorithm.off_policy_correction.sequence_tis_ratio_clip_high=$TIS_IMP_RATIO_CAP \
+  generator.sampling_params.logprobs=$LOGPROBS \
+  trainer.algorithm.advantage_estimator="grpo" \
+  data.train_data="['$DATA_DIR/train.parquet']" \
+  data.val_data="['$DATA_DIR/validation.parquet']" \
+  trainer.policy.model.path="Qwen/Qwen2.5-Coder-7B-Instruct" \
+  trainer.epochs=125 \
```
SumanthRH: Seems excessive

examples/train/text_to_sql/run_skyrl_sql_fp8.sh
SumanthRH, line 1: can you actually make the configuration as close to the baseline script as possible? `run_skyrl_sql.sh`? Configurations like ckpt interval and hf save interval seem wrong here

skyrl-gym/skyrl_gym/envs/sql/env.py
```diff
@@ -95,18 +95,8 @@ def _is_done(self, action: str) -> bool:
             return True
         return "<solution>" in action and "</solution>" in action
 
-    def _validate_action(self, action: str):
```
CharlieFRuan, line 98: Could you elaborate on "skyrl-gym: remove _validate_action; stop strings are enforced in the generator config instead of failing mid-rollout." please? Specifically what do you mean by "generator config"

Wouldn't this be enforced by the stop string vllm config?

I also see that @SumanthRH  removed this for the search-r1 example in https://github.com/NovaSky-AI/SkyRL/pull/1301

Why would that be the case? Are the output not obeying the stop strings?

CharlieFRuan: Or is it because the last token generated can be `>\n` and hence breaking the validation which requires `>`

SumanthRH: > Or is it because the last token generated can be >\n and hence breaking the validation which requires >

Yes this is the main reason!

SumanthRH: Overall the regex parsing in SQL code will actually allow for stray characters after `<answer/>` so the change should not affect accuracy

SumanthRH (approved):

skyrl-gym/skyrl_gym/envs/sql/env.py
CharlieFRuan: Got it, thanks!

PR #1606, opened 2026-05-01
[bug] Fix LoRA disk reloading and sleep issues

### LINKED ISSUE #1582: _SKYRL_USE_NEW_INFERENCE=1 (default) causes ~15x performance regression vs legacy path
### Summary

I encountered a severe performance regression when using the new HTTP-based inference path (`_SKYRL_USE_NEW_INFERENCE=1`, the default) compared to the legacy Ray-actor-based path (`_SKYRL_USE_NEW_INFERENCE=0`) for colocated FSDP+vLLM training with tensor parallelism, using the Tinker API entrypoint (`skyrl_train` backend). 

I don't know the root cause, and I'm unsure whether others can easily reproduce this.

**Measured:** Qwen3-30B-A3B on 8x H200 GPUs, colocated TP=8:
- **Legacy path:** ~4-5 min/iteration, generation batches complete in 5-10s
- **New path:** ~67 min/iteration, generation ~20x slower, repeated ~10 min CUDA warmups per turn

The only difference between the two runs was the `_SKYRL_USE_NEW_INFERENCE` environment variable. All other config (model, hardware, parallelism, colocation) was identical.

### Environment

- **SkyRL version:** v0.2.0 (`skyrl_train` backend / Tinker API)
- **vLLM version:** 0.19.0
- **Model:** Qwen/Qwen3-30B-A3B (MoE, 30B total, 3B active)
- **Hardware:** 8x NVIDIA H200 (141 GiB each), NVLink
- **Config:** `colocate_all=true`, `tensor_parallel_size=8`, `enforce_eager=false`, `gpu_memory_utilization=0.5`

### Observations

With the new path, the per-iteration breakdown was approximately:

| Phase | Duration | Notes |
|---|---|---|
| FSDP model load (one-off) | ~5 min | `create_model` took ~304s |
| vLLM server start (one-off) | ~4 min | includes model load + graph compile |
| Per-turn CUDA warmup | ~9-10 min | happens 3x per iteration |
| Per-turn generation | ~12-13 min | 120 requests x ~6s/request |
| Training step | ~67s | fast |

With the legacy path, generation batches of 28-34 samples completed in **8-10 seconds**. CUDA graph compilation happened once at startup (46s) and was not repeated on subsequent wake cycles. Sleep took ~10s, wake(weights) ~1.4s, wake(kv_cache) ~1.0s.

### Additional Observations (maybe related, unclear)

**1. GPU under-utilization on one TP rank (new path only)**

During the slow new-path run, `nvidia-smi dmon` showed one TP worker (GPU 1) consistently at **0% SM utilization** while the other 7 GPUs were at 93-100%. This was not a hardware issue: ECC errors were 0, clocks were normal (1980 MHz), temperature and power were within range. Process-level inspection showed the stalled worker had far fewer voluntary context switches (95K vs 646K for rank 0), suggesting it was CPU-spinning rather than doing GPU work. This was not observed during the legacy path run.

**2. Non-colocated run hits FSDP OOM during model loading**

A non-colocated run with the following config hit a `torch.OutOfMemoryError` during FSDP model initialization:

```
colocate_all=false
policy_num_gpus_per_node=2
num_engines=6
gpu_memory_utilization=0.95
```

This is surprising because the full Qwen3-30B-A3B model is ~60 GiB, well below the 141 GiB H200 capacity, and with `colocate_all=false` the FSDP trainer and vLLM engines should be on separate GPUs. This may be a separate issue, but mentioning here in case there is a common cause (e.g. FSDP not sharding properly, or GPU placement not isolating trainer and inference GPUs correctly).

hao-aaron:
Two bugs fixed here:
1. When lora is enabled with merge_lora=False, a lora reload request with `load_inplace=True` causes each step to trigger load from disk causing massive slowdowns. This requires an upstream vllm fix, but I provide a temporary workaround by unloading the lora first then loading with `load_inplace=False`. This could be the issue behind https://github.com/NovaSky-AI/SkyRL/issues/1582
2. New inference code path does not override sleep level 2 to level 1 when using lora, since when we are using unmerged LoRA weights we should not discard the base weights. Discarding base weights with level 2 sleep led to garbage output

Validated by running previously slow lora run and ensuring nonzero reward: https://wandb.ai/sky-posttraining-uc-berkeley/gsm8k_0.5b_lora/runs/f44fhwwl

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1606" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
+++ b/skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
@@ -200,6 +200,12 @@ class RemoteInferenceClient:
     active_lora_name: Optional[str] = None
     """Name of the active LoRA adapter. If set, generation requests use this adapter instead of the base model."""
 
+    uses_lora_weight_sync: bool = False
+    """True when the trainer syncs LoRA adapters (rather than full/merged weights). When True,
+    `sleep()` is forced to level=1: level=2 discards the base model from VRAM with no CPU backup,
+    and LoRA-only broadcasts cannot repopulate it. Must be kept in sync with the same gate vLLM
+    uses for `enable_lora` (see `_uses_lora_weight_sync` in inference_servers/utils.py)."""
+
     tokenizer: Optional[Any] = None
     """Optional HF tokenizer for local tokenize/detokenize (avoids HTTP round-trips)."""
 
@@ -924,6 +930,15 @@ class RemoteInferenceClient:
         Returns:
             Dict mapping server_url to response.
         """
+        # Mirror BaseVLLMInferenceEngine.sleep: when the trainer syncs LoRA adapters
+        # only, force level=1 so the base model survives via CPU backup. level=2
+        # discards weights with no source to restore from on wake_up(["weights"]).
+        if self.uses_lora_weight_sync and level != 1:
+            logger.info(
+                "Forcing sleep level=1 (uses_lora_weight_sync=True); requested level=%d would discard the base model.",
+                level,
+            )
+            level = 1
         params: Dict[str, Any] = {"level": str(level)}
         if tags:
             params["tags"] = tags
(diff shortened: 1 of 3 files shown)


skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -200,6 +200,12 @@ class RemoteInferenceClient:
     active_lora_name: Optional[str] = None
     """Name of the active LoRA adapter. If set, generation requests use this adapter instead of the base model."""
 
+    uses_lora_weight_sync: bool = False
```
SumanthRH, line 203: Why do we need a separate flag here? Can't we use just `active_lora_name is not None` as the condition?

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
```diff
@@ -1085,44 +1100,51 @@ async def update_lora_from_disk(
...
-        After loading, generation requests will automatically use the LoRA adapter
-        by setting the model name to the LoRA adapter name.
+        Implementation: unload then load (with load_inplace=False), rather than a single
+        load with load_inplace=True. Reason: vLLM's /v1/load_lora_adapter caches the
+        LoRARequest object server-side keyed by lora_name (see OpenAIServingModels.lora_requests).
+        If load_inplace=True is set on that cached request, every subsequent generate request
+        causes LRUCacheWorkerLoRAManager.add_adapter to re-read the adapter from disk on every
+        model step (line 299-307 of vllm/lora/worker_manager.py checks `load_inplace` per-step).
+        The cleanest fix is upstream (clear load_inplace before caching); until that lands,
+        unloading first lets us reuse the same lora_name without ever needing load_inplace=True.
```
SumanthRH: Nit: Docstring stylke is not consistent with the rest of the codebase. Can you clean this up and also convert this into a TODO:

> The cleanest fix is upstream (clear load_inplace before caching); until that lands,
        unloading first lets us reuse the same lora_name without ever needing load_inplace=True.

skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
hao-aaron: planning to get rid of active_lora_name in multi-lora, https://github.com/NovaSky-AI/SkyRL/pull/1579/changes#diff-e098add986cd397d2cb2fa04042e96a468524e733f54be101ee04d400e4c4270L200

SumanthRH (approved):

PR #1634, opened 2026-05-08
[tinker] Separate out SkyRLTrainBackend sample requests into EXTERNAL request path

erictang000:
Allows for speedups with multi-lora!


erictang000:
closing in favor of #1638

PR #1638, opened 2026-05-09
[tinker] Forward sample requests directly to backend vLLM (non-colocated)

erictang000:
## Description

When SkyRL-Train runs non-colocated (`colocate_all=false`), vLLM is always-on but its sample capacity is wasted: the Tinker engine subprocess serializes `sample` behind `forward_backward` / `optim_step` in its 100ms tick loop, so a sample submitted at the start of a 30s training step waits 30s. For multi-tenant RL the queueing compounds — every `sample` from any tenant queues behind every other tenant's training step.

This PR hoists sample requests off the engine queue and into the API process's asyncio loop, where they're forwarded directly to the engine-managed vLLM. The colocated and JAX paths are unchanged.

## How it works

We reuse the existing `RequestType.EXTERNAL` plumbing (already in place for fully external vLLM URLs):

```
       ┌─ POST /api/v1/asample ─────────────────────────┐
client ┤  api.py:                                       │
       │    create FutureDB(type=EXTERNAL)              │
       │    asyncio.create_task(client.call_and_store)  │
       │    return future_id immediately                │
       └────────────────────────────────────────────────┘
                            │
                            ▼
       ┌─ SkyRLTrainInferenceForwardingClient ──────────┐
       │  POST {vllm_proxy_url}/v1/completions          │
       │    model=<model_id>  (LoRA registered already) │
       │  on completion: write FutureDB.result_data     │
       └────────────────────────────────────────────────┘

   (parallel)
       ┌─ engine subprocess ────────────────────────────┐
       │  process_pending_requests loop:                │
       │    forward_backward / optim_step /             │
       │    save_weights_for_sampler / ...              │
       │  (NO sample work here)                         │
       └────────────────────────────────────────────────┘
```

The cross-process plumbing was already wired for external vLLM URLs; the gap was that nothing was pointing it at the *engine-managed* vLLM. This PR closes that gap.

## What's in the diff

8 files, +796/-3 LOC.

**New `EngineStateDB` (`skyrl/tinker/db_models.py`)** — singleton row carrying engine→API handoff state (currently just `inference_proxy_url` + `updated_at`). The backend writes it after building vLLM; the forwarding client reads it lazily on first sample.

**`SkyRLTrainBackend._publish_engine_state` (`skyrl/backends/skyrl_train_backend.py`)** — called inside `_create_new_inference_client` after `build_new_inference_client` returns, and on `delete_model` teardown to clear the row. Best-effort: a DB write failure logs and returns rather than corrupting controller state. Wired through a new `set_engine_database_url(...)` setter that the engine calls right after constructing the backend.

**New `SkyRLTrainInferenceForwardingClient` (`skyrl/tinker/extra/skyrl_train_inference_forwarding.py`)** — pair to `ExternalInferenceClient`, with the same EXTERNAL future-write contract but resolves the target URL from `EngineStateDB` instead of from a user-supplied `EngineConfig.external_inference_url`. Uses `model=<model_id>` (the name `save_weights_for_sampler` registered with vLLM via `load_lora_adapter`).

Key design choices:
- **No API-side semaphore.** Backpressure is layered: httpx connection pool → vllm-router → each vLLM server's `max_num_seqs`. Adding a Python semaphore on top would just serialize work above what vLLM already manages.
- **Persistent `httpx.AsyncClient`** with `forwarding_inference_max_connections` defaulting to `None` (unlimited). The only cost of "unlimited" is file descriptors, so operators just raise `ulimit -n` to match peak concurrent samples. Set an int to enforce a per-process cap. Closed via `aclose()` from the API lifespan.
- **Polymorphic `.stop` dispatch** matches `api.py SamplingParams.to_types()` (list[str] → `stop`, list[int] → `stop_token_ids`).
- **Response parsing mirrors the in-process path**: `finish_reason` normalized to Literal["stop", "length"]; missing `token_logprobs` zero-filled; non-JSON responses (e.g. proxy 502 HTML) surfaced with content-type + body excerpt for diagnosis.
- **Retry covers all transport-level errors** (`httpx.RequestError` umbrella — `ConnectError`, `ReadError`, `RemoteProtocolError`, `TimeoutException`, `PoolTimeout`). HTTP 4xx/5xx from vLLM is NOT retried — it's a real upstream signal.

**`api.py` lifespan** — installs the forwarding client when `backend in ("megatron", "fsdp")` AND `trainer.placement.colocate_all=False`. Colocated runs and JAX keep the synchronous engine flow. Calls `aclose()` on shutdown.

**Synchronization invariants preserved:**

- **(I1)** Sample for checkpoint X requires save-X to have completed. Already enforced by the SDK + `validate_checkpoint(...)` at `api.py:1037` before the future is created.
- **(I2)** In-flight sample during a re-broadcast for the same model. `WorkerDispatch.save_weights_for_sampler` brackets the broadcast with `pause_generation` / `resume_generation`; vLLM's KEEP-mode pause freezes in-flight requests in its scheduler.
- **(I3)** Result writes don't conflict with engine writes. EXTERNAL futures are written by the API process; all other futures by the engine. SQLite WAL handles concurrent readers.
- **(I4)** Colocated path is unchanged. Lifespan strictly gates on `colocate_all=false && backend ∈ (megatron, fsdp)`.

## Test plan

- ✅ **`test_engine_state_published`** — after `save_weights_for_sampler`, `EngineStateDB.inference_proxy_url` is populated with the engine-managed vLLM proxy URL.
- ✅ **`test_sample_uses_external_path`** — issued sample creates a `FutureDB` row with `request_type=EXTERNAL` (off the engine queue).
- ✅ **`test_sample_concurrent_with_training_is_fast`** — central parallelism test. While 24 `forward_backward`+`optim_step` calls stream in a background thread, a concurrent sample resolves in **1.3s vs 10s training stream (ratio 0.13×)**. Without async routing the sample would queue behind the entire training stream and the latencies would be comparable.
- ✅ **`test_concurrent_samples_per_adapter`** — 8 concurrent samples across two adapters all resolve via the forwarding client's connection pool and per-adapter `model_id` routing on vLLM.
- ✅ All 6 existing `test_multi_lora_megatron.py` tests continue to pass under the new forwarding path (backwards compatible with the multi-LoRA RL workload).
- ✅ `tests/tinker/test_api.py` single-tenant paths pass.

**End-to-end multi-tenant RL bench** following `docs/tinker/multi_tenancy.mdx#quickstart---two-rl-clients` (Qwen3-0.6B, GSM8K, 4 policy GPUs + 1 vLLM, 3 iters × batch 16 × group 4 × max 128 tokens per client):

| Mode | A wall-clock | B wall-clock | Total wall-clock |
|---|---|---|---|
| Sequential (A then B) | 37.8s | 31.7s | **85s** |
| Concurrent (A ∥ B) | 36.0s | 34.7s | **44s** |

**Speedup ~1.93×**, close to the theoretical 2× ceiling. Per-iter cost grows ~1-2s under contention (vLLM multiplexes both adapters' samples + engine batches both fwd_bwds), but total wall-clock halves because both tenants progress simultaneously instead of waiting their turn.

Server-side instrumentation: **240 sample requests forwarded** via the API-process client; **0 served by the engine** (single startup-time engine sample is pre-bench). `process_batch_requests(forward_backward, n=2)` lines in the server log confirm the engine batches concurrent A+B fwd_bwd calls together.

## Operator config

To enable, set in `--backend-config`:

```json
{
    "trainer.placement.colocate_all": false,
    "trainer.policy.megatron_config.lora_config.merge_lora": false,
    "trainer.policy.model.lora.max_loras": <max concurrent adapters in a batch>,
    "trainer.policy.model.lora.max_cpu_loras": <total adapter capacity>
}
```

For very high fan-out workloads, raise the host's `ulimit -n` to match peak concurrent samples (default httpx pool is unbounded). Operators who want a per-API-process FD cap can set `--forwarding-inference-max-connections <N>`.

See [docs/content/docs/tinker/multi_tenancy.mdx](docs/content/docs/tinker/multi_tenancy.mdx) (added in the multi-lora-rl PR) for the full operator contract.

## Out of scope

- **Colocated mode acceleration.** No parallelism to exploit there — vLLM is asleep during training and the engine's synchronous sample path is what wakes it.
- **Auto-recovery from vLLM eviction.** If `max_cpu_loras` is too low and vLLM evicts an adapter mid-run, the next sample 404s. Re-loading from disk on demand is a follow-up; for now the operator sizes `max_cpu_loras` ≥ expected concurrent adapters.
- **Sample retries beyond one transport-level retry.** The forwarding client refreshes its cached proxy URL and retries once on `httpx.RequestError`. Application-level retry is the SDK's job (already implemented in `tinker/retry_handler`).
- **Parallelizing training requests across model_ids.** Multi-tenant `forward_backward` / `optim_step` still serialize through the engine's main loop.

### DIFF
--- a/skyrl/backends/skyrl_train_backend.py
+++ b/skyrl/backends/skyrl_train_backend.py
@@ -342,12 +342,15 @@ class SkyRLTrainBackend(AbstractBackend):
         """
         self._engine_database_url = database_url
 
-    def _publish_engine_state(self, *, proxy_url: str | None, server_urls: list[str], is_colocated: bool) -> None:
+    def _publish_engine_state(self, *, proxy_url: str | None) -> None:
         """Upsert the singleton row in EngineStateDB.
 
         Idempotent. Safe to call repeatedly. ``proxy_url=None`` signals that
         no vLLM is currently up (post-teardown) — the API process will fall
         back to the engine's synchronous sample path.
+
+        Best-effort: on any DB error we log and return rather than raise.
+        Callers rely on local state being reset regardless of publish outcome.
         """
         if self._engine_database_url is None:
             # Backend not wired to a Tinker engine (unit tests, non-Tinker uses).
@@ -359,24 +362,25 @@ class SkyRLTrainBackend(AbstractBackend):
 
         from skyrl.tinker.db_models import EngineStateDB, enable_sqlite_wal
 
-        db_engine = create_engine(self._engine_database_url, echo=False)
-        enable_sqlite_wal(db_engine)
-        # The API process creates the schema, but in unit tests the backend
-        # may run without an API. Make sure the table exists before writing.
-        SQLModel.metadata.create_all(db_engine, tables=[EngineStateDB.__table__])
         try:
-            with Session(db_engine) as session:
-                row = session.get(EngineStateDB, 1)
-                if row is None:
-                    row = EngineStateDB(singleton_id=1)
-                row.inference_proxy_url = proxy_url
-                row.inference_server_urls = list(server_urls)
-                row.is_colocated = bool(is_colocated)
-                row.updated_at = datetime.now(timezone.utc)
-                session.add(row)
-                session.commit()
-        finally:
-            db_engine.dispose()
+            db_engine = create_engine(self._engine_database_url, echo=False)
+            enable_sqlite_wal(db_engine)
+            # The API process creates the schema, but in unit tests the backend
+            # may run without an API. Make sure the table exists before writing.
+            SQLModel.metadata.create_all(db_engine, tables=[EngineStateDB.__table__])
+            try:
+                with Session(db_engine) as session:
+                    row = session.get(EngineStateDB, 1)
+                    if row is None:
+                        row = EngineStateDB(singleton_id=1)
+                    row.inference_proxy_url = proxy_url
+                    row.updated_at = datetime.now(timezone.utc)
+                    session.add(row)
+                    session.commit()
+            finally:
+                db_engine.dispose()
+        except Exception as e:
+            logger.warning(f"Failed to publish engine state (proxy_url={proxy_url!r}): {e}")
 
     def _create_new_inference_client(self):
         """Create new HTTP-based inference client."""
@@ -396,11 +400,7 @@ class SkyRLTrainBackend(AbstractBackend):
 
         # Publish inference endpoint so the API can forward samples directly
         # (only meaningful in non-colocated mode; the API gates on this).
-        self._publish_engine_state(
-            proxy_url=server_setup.proxy_url,
-            server_urls=list(server_setup.server_urls),
-            is_colocated=bool(is_colocated),
-        )
+        self._publish_engine_state(proxy_url=server_setup.proxy_url)
 
     def _ensure_inference_engines(self):
         """Lazily create inference engines and init weight sync on first sampling-related call."""
@@ -550,9 +550,6 @@ class SkyRLTrainBackend(AbstractBackend):
         ray.shutdown()
         self._model_ids_to_role = {}
         self._model_metadata = {}
-        # Tell the API there is no inference engine to forward to. Next
-        # _create_new_inference_client will repopulate this row.
-        self._publish_engine_state(proxy_url=None, server_urls=[], is_colocated=False)
         self._cfg = None
         self._dispatch = None
         self._inference_engine_client = None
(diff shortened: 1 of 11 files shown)


skyrl/backends/skyrl_train_backend.py
```diff
@@ -325,6 +331,57 @@ def _create_legacy_inference_client(self):
             self._cfg.generator.inference_engine,
         )
 
+    def set_engine_database_url(self, database_url: str) -> None:
```
pcmoritz: Let's keep the database out of the backends, and even better would be to not write the `proxy_url` into the database (since it is ephemeral), how about we just make it an (optional) attribute of the backend?

If it really needs to be in the database, I think the better way would be to re-use the DB connection in `engine.py`.

It is better to keep the database out of the backend to avoid propagating it everywhere in the future.

erictang000: hmm good point. i think because the proxy URL doesn't exist until `_create_new_inference_client` is run in the backend it's tricky to configure as an attribute up front

but I can move it up to reuse the engine.py db connection so it's cleaner

PR #1663, opened 2026-05-13
[feat] Add support for `loss_fn` with `forward` and enable evaluation support in `SFTTrainer`

SumanthRH:
# What does this PR do?

## Summary

Adds Tinker-compatible `forward(loss_fn=...)` to SkyRL's FSDP and Megatron backends, wires eval-dataset support into `SFTTrainer` for periodic eval-loss logging during training, and refactors the worker output contract to a single `WorkerOutput` dataclass mirroring Tinker's `ForwardBackwardOutput`.

### `forward(loss_fn=...)` on policy workers

`WorkerDispatch.forward` now accepts `loss_fn` and `loss_fn_config` kwargs. When set, the worker runs the model under `torch.no_grad()` (no backward), computes the requested loss, and returns per-sample `loss_fn_outputs` alongside scalar metrics. When unset, the existing logprobs-only path is preserved.

- `PolicyWorkerBase.forward` dispatches on `loss_fn`; new `_forward_micro_with_loss` mirrors the SFT branch of `_forward_backward_micro` but skips `strategy.backward` and KL/entropy.
- `MegatronModelWrapper.forward_backward_mini_batch` gains a `forward_only` flag so Megatron's pipeline schedule machinery is reused for the no-grad path.
- DP all-reduce on metrics matches the existing training-path semantics.

### Eval dataset support in `SFTTrainer`

New `SFTConfig` fields enable periodic eval during training:
- `eval_dataset_name: Optional[str] = None` — HF dataset name/path. `None` disables eval.
- `eval_dataset_split: str = "validation"`
- `eval_interval: int = 0` — run eval every N training steps; 0 disables.

Eval iterates the eval dataset in chunks of `micro_train_batch_size_per_gpu * dp_size`, calls `dispatch.forward(loss_fn="cross_entropy", ...)`, and aggregates a token-weighted mean loss across batches. The result is logged to wandb as `eval/eval_loss` alongside `train/loss` in a single `tracker.log(commit=True)` call per training step, preserving wandb's step ordering. A final eval pass at end of training is also logged.

### `WorkerOutput` dataclass

`forward` and `forward_backward` now both return a `WorkerOutput` dataclass mirroring Tinker's `ForwardBackwardOutput`:

```python
@dataclass
class WorkerOutput:
    loss_fn_output_type: str = "scalar"
    loss_fn_outputs: List[Dict[str, Any]] = field(default_factory=list)
    metrics: Dict[str, Any] = field(default_factory=dict)
```

- Single `WorkerOutput.cat(actor_infos, shards)` classmethod replaces the previous dual concat code paths. List-extend for `loss_fn_outputs`, take-rank-0 for already-DP-reduced `metrics`.
- Freestanding `loss_fn_outputs_to_tensor(outputs, key, ...)` helper for callers that need a padded batch-dim tensor.
- `skyrl_train_backend.py`'s HTTP serving path simplifies: workers now emit per-sample lists directly, so the manual per-row tensor slicing is removed.

### Worker signature cleanup

- `Worker.forward` is abstract (`*args, **kwargs`, raises `NotImplementedError`). Each subclass (`PolicyWorkerBase`, `RefWorkerBase`, `CriticWorkerBase`) declares its own concrete signature — no LSP violation.
- All worker `forward` methods return `Dict[str, Any]` (no more `Union[TrainingOutputBatch, Dict]`).
- The shared `_inference_forward` / `_megatron_inference_forward` helpers are removed from the base classes; each subclass's `forward` inlines the logprob-only path with the appropriate output key (`logprobs` / `values`). Some duplication, but base classes shouldn't carry helpers used by only some subclasses.

### Config and helper polish

- `SFTTrainer.collate_batch.batch_size` is now a required positional arg.
- `eval_batch_size` removed from `SFTConfig` — eval uses `micro_train_batch_size_per_gpu * dp_size` per dispatch. (Side effect: `eval_loss` is now the true per-non-pad-token NLL, independent of batch size; previous values were scaled by `batch_size / (micro_batch_size * dp_size)`.)

### Example script

`examples/train/sft/run_sft_megatron_tulu3_eval.sh` — end-to-end SFT run on the tulu-3-sft-mixture dataset with the Megatron backend and eval enabled.

## Test plan

- [x] `tests/backends/skyrl_train/gpu/gpu_ci/test_training_step.py::test_sft_forward_with_cross_entropy[fsdp2|megatron]` — both PASS (81s + 126s)
- [x] `tests/backends/skyrl_train/gpu/gpu_ci/test_training_step.py::test_sft_forward_backward_with_cross_entropy[fsdp2|megatron]` — both PASS (75s + 119s)
- [x] `tests/train/test_sft_tokenization.py` + `tests/train/generators/test_utils.py` — 79 passed
- [x] End-to-end tulu Megatron run (Qwen2.5-0.5B-Instruct, 200 steps, eval every 50 steps) — eval_loss decreases monotonically: 2.1463 → 2.1419 → 2.1407 → 2.1400 ([wandb run 8t8opq7z](https://wandb.ai/sky-posttraining-uc-berkeley/skyrl-sft-eval/runs/8t8opq7z))
- [x] ruff clean on `skyrl/train/` + `skyrl/backends/skyrl_train/workers/`


SumanthRH:
Closing since this PR was split into #1667  and #1668

PR #1666, opened 2026-05-14
[multi-tenant-lora] Expose per-LoRA pause state over HTTP

erictang000:
## Summary

Exposes the per-LoRA pause state from `RemoteInferenceClient.sample_with_retry` over HTTP so out-of-process callers (the Tinker `SkyRLTrainInferenceForwardingClient` from #1638) can participate in the same pause/resume cycle as in-process callers.

This is the first of two PRs that finish the fully-async multi-tenant Tinker path. Follow-up PR will use these endpoints to add `sample_with_retry`-equivalent abort recovery to the Tinker forwarding client and add Tinker-side tests that mirror `test_pause_lora.py`.

## Background

Today `RemoteInferenceClient.sample_with_retry` (added in #1657) uses an in-process `_lora_pause_events: dict[str, asyncio.Event]` to gate retries during a per-LoRA pause window. The Tinker API process runs in a separate Python process and cannot observe that dict, so an in-flight Tinker asample for a paused LoRA today:

1. Sees `finish_reason="abort"` in the vLLM response but maps it to `stop_reason="length"` (silent corruption — fix lands in the follow-up PR), and
2. A fresh Tinker asample for the same LoRA submitted during the pause window races with `load_lora_adapter` and can observe torn adapter weights.

## Changes

`skyrl/backends/skyrl_train/inference_servers/vllm_server_actor.py`:
- `app.state.paused_loras: dict[str, asyncio.Event]` initialised in `_add_custom_endpoints`. Mirrors the in-process gate.
- Existing `POST /skyrl/v1/abort_lora_requests` (#1657) now also clears the per-LoRA event before the `engine.abort()` fan-out.
- New `POST /skyrl/v1/resume_lora_requests {lora_name}`: sets the event.
- New `POST /skyrl/v1/wait_lora_unpaused {lora_name, timeout_s?}`: long-polls; returns `{paused: false}` once the event is set or `{paused: true}` after timeout so the caller can loop.
- New pure-ASGI submission-gate middleware: pending `/v1/completions` and `/v1/chat/completions` for a currently-paused LoRA block until resume. Closes the race in (2) above. Pure ASGI (not `@app.middleware("http")`) because `BaseHTTPMiddleware` buffers streaming responses and would regress vLLM's SSE completions path.

`skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py`:
- `resume_generation(lora_name=X)` now also POSTs `/skyrl/v1/resume_lora_requests` to all `server_urls` so the server-side flag matches the in-process gate. The local set still happens first, so in-process retries don't wait on the HTTP round-trip; brief inconsistency between the two gates just causes one extra retry-loop iteration.

All new endpoints carry the same TRANSIENT marker as the existing abort endpoint: delete the stack when vLLM ships native per-LoRA pause.

## Test plan

- [x] `ruff` + `black` clean
- [ ] `tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_pause_lora.py` (existing pause/resume suite) passes — every test exercises `pause_generation` and `resume_generation`, so my changes are validated indirectly. New endpoints (`/skyrl/v1/wait_lora_unpaused`, submission gate) are exercised by the follow-up PR's Tinker-side test.

### DIFF
--- a/skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
+++ b/skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
@@ -1161,11 +1161,27 @@ class RemoteInferenceClient:
         )
 
     async def resume_generation(self, lora_name: Optional[str] = None) -> Dict[str, Any]:
-        """Resume generation. ``lora_name`` must match the prior pause call."""
+        """Resume generation. ``lora_name`` must match the prior pause call.
+
+        For the per-LoRA branch, releases both the local in-process gate (so
+        in-process ``sample_with_retry`` callers can resubmit immediately) and
+        the server-side gate via ``/skyrl/v1/resume_lora_requests`` (so the
+        submission gate on each worker stops blocking fresh /v1/completions
+        and out-of-process callers polling ``/skyrl/v1/wait_lora_unpaused``
+        unblock). The local set happens first so in-process retries don't
+        wait on the HTTP round-trip; brief inconsistency between the two
+        gates is fine — the retry loop just spins one extra iteration.
+
+        TRANSIENT: the ``lora_name`` branch (and the HTTP call) should be
+        deleted when vLLM ships native per-LoRA pause.
+        """
         if lora_name is None:
             return await self.resume()
         self._get_lora_pause_event(lora_name).set()
-        return {"status": "ok"}
+        return await self._call_all_servers(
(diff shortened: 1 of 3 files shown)


erictang000:
closed since we can just use load_lora_adapter without pausing....

PR #1679, opened 2026-05-16
[ci] H100 CI

hao-aaron:
# H100 CI tests for large MoE models

Introduces an opt-in H100 CI lane that exercises two ~30B-class MoE models end-to-end:

- **Qwen/Qwen3.5-35B-A3B** — FSDP e2e (`test_policy_local_engines_e2e`), both colocated and non-colocated with vLLM. Megatron logprob roundtrip (`test_megatron_models`).
- **nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16** — Megatron logprob roundtrip (`test_megatron_models`), re-enabled after the original 8-GPU param was skipped.

Tests are gated by `pytest.mark.h100` (auto-skipped unless `-m h100` is passed). The new GitHub workflow submits them as an Anyscale staging job on the `llm-team-h100-4x:1` compute config.

## Test infrastructure

- Register the `h100` marker in `tests/backends/skyrl_train/gpu/conftest.py` and auto-skip those tests unless `-m h100` is explicitly passed.
- Add Qwen3.5-35B-A3B parameterizations to `test_policy_local_engines_e2e` (colocated + non-colocated) and `test_megatron_models` (TP=4 EP=4).
- Re-enable the Nemotron-3-Nano param on 4 GPUs (TP=1 EP=4) with the `h100` marker.

## Core fixes uncovered along the way

- **FSDP2 meta-init reordering** (`fsdp_strategy.py`): swap the module to meta *before* `apply_fsdp2`, so sharded DTensors allocate directly at their final shard size instead of materializing the full model on rank 0 first. Snapshot/restore non-persistent buffers (e.g. `RotaryEmbedding.inv_freq`) around the meta swap so they survive the broadcast.
- **Replace manual broadcast loop with `set_model_state_dict`** (`fsdp_utils.py`): the ~80-line per-parameter broadcast + `distribute_tensor` loop is now PyTorch's `set_model_state_dict(..., StateDictOptions(full_state_dict=True, broadcast_from_rank0=True))`.
- **Remove the FSDP1-era offload+load dance** (`fsdp_utils.py`): the trick (`offload_fsdp2_model_to_cpu` → `empty_cache` → `load_fsdp2_model_to_gpu`) was meant to clear reserved-but-unallocated PyTorch memory. For FSDP2 it's a no-op (`model.to("cpu")` doesn't move FSDPParam-managed storage during init), then the reload allocates a second copy — doubling memory. Removed; `set_model_state_dict` already leaves us with exactly the shard on GPU.
- **NCCL weight-sync receive must be wrapped in `set_current_vllm_config`** (`new_inference_worker_wrap.py`, `remote_inference_client.py`, `broadcast_strategy.py`): MoE models (FlashInfer CUTLASS kernel) read `get_current_vllm_config()` during `load_weights`, which is only set around `init_device`/`load_model`. Added a new `update_weights_nccl` worker method that wraps `weight_transfer_engine.receive_weights` in `set_current_vllm_config`, and route the broadcast sender through `start_weight_update` + `update_weights_nccl` + `finish_weight_update` instead of vLLM's native `/update_weights` endpoint. Same pattern CUDA IPC already used. Tracked against upstream [vllm-project/vllm#42577](https://github.com/vllm-project/vllm/pull/42577).
- **Rename `update_weights_chunk` → `update_weights_ipc`** across `new_inference_worker_wrap.py`, `remote_inference_client.py`, and `cuda_ipc_strategy.py` so the IPC and NCCL paths have parallel names.
- **Introduce new `inference_only_init` config flag** : disables optimizer state and allows bf16 trainer to save memory for tests. 
- **Large-MoE Megatron config** (`test_megatron_models.py`): for Qwen3.5-35B and Nemotron-3-Nano, set `use_precision_aware_optimizer=True` + `optimizer_cpu_offload=True` + `optimizer_offload_fraction=1.0`. Megatron eagerly materializes the fp32 AdamW state on GPU at init (unlike PyTorch's lazy AdamW), so the optimizer state alone OOMs without offload.
- **Lower vLLM `gpu_memory_utilization` to 0.5** for these large MoE models in the Megatron test (`_engine_overrides_for_model`) so the policy shard + vLLM pool both fit on each H100.
- **`sleep_level=2` in `test_megatron_models`**: the test explicitly syncs weights, so full sleep (matching the `InferenceEngineState.create` default and the FSDP e2e test) is the right level. Previously was hardcoded to 1.


skyrl/backends/skyrl_train/workers/fsdp/fsdp_worker.py
```diff
@@ -159,7 +159,6 @@ def init_model(self, model_path, num_training_steps: int = None):
         wrapped_model = HFModelWrapper(
             model_path,
             use_flash_attention_2=self.cfg.flash_attn,
-            bf16=False,
             lora_rank=self.cfg.policy.model.lora.rank,
             lora_alpha=self.cfg.policy.model.lora.alpha,
             lora_dropout=self.cfg.policy.model.lora.dropout,
```
SumanthRH, line 168: This change - using`bf16=True` - is going to initialize the model weights in BF16, and will lead to pure BF16 training.

You want to initialize the model weights in FP32 and let FSDP handle casting to BF16 during forward pass. 

https://docs.pytorch.org/docs/2.12/fsdp.html#torch.distributed.fsdp.MixedPrecision

"Outside forward and backward, the sharded parameters are kept in full precision (e.g. for the optimizer step), and for model checkpointing, the parameters are always saved in full precision."

tests/backends/skyrl_train/gpu/gpu_ci/megatron/test_megatron_models.py
```diff
@@ -72,6 +72,20 @@ def get_test_actor_config(model_name) -> SkyRLTrainConfig:
         # sample packing not yet supported for GDN
         # https://github.com/NVIDIA/Megatron-LM/pull/2644
         cfg.trainer.use_sample_packing = False
+    # Large MoE models: AdamW's fp32 optimizer state (~6x model) is the
+    # dominant per-rank GPU cost during Megatron init.
+    # use_precision_aware_optimizer keeps it in bf16 (halves it), and
+    # optimizer_cpu_offload moves it off GPU entirely. Without these the
+    # optimizer state alone OOMs the H100 on 4 GPUs.
```
SumanthRH: It would be good to just initialize policy without any optimizer state

We might need a fix in SkyRL FSDP code for this, which is the better way to do this than using optimizer offload. 

We can explicitly pass `optimizer_config` as `None` to the FSDPStrategy : 

https://github.com/NovaSky-AI/SkyRL/blob/33f18badd5cbd0d5da7594d79dcadf399281b0b2/skyrl/backends/skyrl_train/distributed/fsdp_strategy.py#L231-L234

Currently this assertion willl prevent us from having `optimizer` as `None` for PolicyWorker, 

https://github.com/NovaSky-AI/SkyRL/blob/33f18badd5cbd0d5da7594d79dcadf399281b0b2/skyrl/backends/skyrl_train/workers/fsdp/fsdp_worker.py#L189-L191

 but we can just modify it to say that it should be non-null only if `optimizer_config` is non-null

SumanthRH (changes requested):
Can you break up this PR into 

1. A minimal PR to initialize H100 CI with Qwen 3.5 35B 
2. A PR to add support for Nemotron 30B

Also, it is hard to see which of the changes in FSDP worker initialization were essential and also which model they apply to. When you break up the PR, please also add only the relevant changes needed.

This PR also switches to initialize the model weights directly in BF16. While this saves memory, it introduces a correctness bug because training will no longer be in mixed precision but instead in pure BF16.

SumanthRH (commented):
- FSDP2 meta-init reordering : Makes sense. Does the basic policy weight sync test pass after this change (to ensure model init is correct)?
- Replace manual broadcast loop with set_model_state_dict: Is this needed?
- NCCL weight-sync receive must be wrapped in set_current_vllm_config : Makes sense
- Rename update_weights_chunk → update_weights_ipc across new_inference_worker_wrap.py, remote_inference_client.py, and cuda_ipc_strategy.py : Makes sense
- Drop stray bf16=False in policy FSDP init : Don't do this
- Large-MoE Megatron config : Instead of this we shoudl really just support creating a policy model without optimizer state
- Lower vLLM gpu_memory_utilization to 0.5 : Makes sense
- sleep_level=2 in test_megatron_models : Makes sense

hao-aaron:
> Replace manual broadcast loop with set_model_state_dict: Is this needed?

`set_model_state_dict` releases intermediate staging tensors as it goes and leaves exactly the shard on GPU. This also lets us drop the verl-style offload+reload that was meant to reclaim allocator memory — diagnostics on Qwen3.5-35B-A3B showed the offload is a no-op for FSDP2 (the shard storage is owned by FSDPParamGroup, not by anything Module.to("cpu") walks) while the reload doubled GPU memory to 65 GiB.

> Drop stray bf16=False in policy FSDP init : Don't do this

Added a test only option for this to prevent oom, default is bf16=False

> Large-MoE Megatron config : Instead of this we shoudl really just support creating a policy model without optimizer state

Added a skip_optimizer_init field in trainer cfg

skyrl/train/config/config.py
```diff
@@ -230,6 +230,15 @@ class PolicyConfig(BaseConfig):
...
     Loads only the language model backbone using AutoModelForCausalLM."""
+    bf16_init: bool = False
+    """When True, initialize policy weights in bf16 instead of fp32. Defaults to fp32
+    init so FSDP mixed precision keeps fp32 master weights; only flip this on for
+    memory-constrained tests where fp32 init would OOM (e.g. large MoE on 4xH100)."""
+    skip_optimizer_init: bool = False
+    """When True, skip optimizer and LR-scheduler construction in the policy worker.
+    Use for inference-only flows (forward + weight sync, no train_step) — primarily
+    Megatron, where DistributedOptimizer eagerly materializes fp32 master + AdamW
+    state on GPU and OOMs large MoE models on memory-constrained nodes."""
```
SumanthRH: Can you make this into a single attribute `inference_only_init` ? Otherwise it's pretty confusing why `bf16_init` is provided as an option in the policy config - it is not going to be mixed precision training.

tests/backends/skyrl_train/gpu/gpu_ci/conftest.py
```diff
@@ -37,6 +37,33 @@ def _build_ray_env_vars():
...
+        try:
+            _ver = subprocess.run(["strings", _p], capture_output=True, text=True, timeout=10).stdout
+        except (subprocess.TimeoutExpired, OSError):
+            continue
+        if "NCCL version 2.29" in _ver and "cuda13" in _ver:
+            env_vars["VLLM_NCCL_SO_PATH"] = _p
+            existing_preload = os.environ.get("LD_PRELOAD", "")
+            env_vars["LD_PRELOAD"] = f"{_p}:{existing_preload}" if existing_preload else _p
+            log_once(f"Pinning VLLM_NCCL_SO_PATH and LD_PRELOAD={_p} (cu13-matched NCCL)")
+            break
```
SumanthRH: umm this seems pretty unexpected given backwards compatibility guarantees.... are you sure this fix is needed? We've used the same H100s with the same SkyRL docker image for training btw without any issues

skyrl/backends/skyrl_train/distributed/fsdp_utils.py
```diff
@@ -129,100 +129,56 @@ def _sync_non_persistent_buffers(model: torch.nn.Module, loaded_sd: dict):
...
+
+    # NOTE: removed the offload_fsdp2_model_to_cpu + load_fsdp2_model_to_gpu
+    # dance from verl PR #1667. That trick was meant to release reserved-but-
+    # unallocated cache memory, but for FSDP2 the offload is a no-op
+    # (`model.to("cpu")` doesn't move FSDPParam-managed storage) and the reload
+    # then *allocates a second copy*. Diagnostic on Qwen3.5-35B-A3B:
+    #   after set_model_state_dict: 32.71 GiB
+    #   after offload+empty_cache:  32.71 GiB (unchanged)
+    #   after load:                 65.42 GiB (doubled)
+    # set_model_state_dict already leaves us with exactly the shard on GPU;
```
SumanthRH: this comment is not needed. Details best left in the PR

ci/gpu_ci_run_h100.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+set -xeuo pipefail
+
+export CI=true
+
+# Prepare datasets used in tests (Megatron test uses gsm8k env_class).
+uv run examples/train/gsm8k/gsm8k_dataset.py --output_dir $HOME/data/gsm8k
+
+# Run FSDP h100 tests. Use --extra fsdp so the fsdp-only tests can collect.
```
SumanthRH: this comment is incorrect. `--extra` doesn't determine which tests are selected. 


```suggestion
# Run FSDP h100 tests. 
```

ci/gpu_ci_run_h100.sh
```diff
@@ -0,0 +1,15 @@
...
+uv run --directory . --isolated --extra dev --extra fsdp pytest -s -vvv -m h100 \
+    tests/backends/skyrl_train/gpu/gpu_ci/test_policy_local_engines_e2e.py
+
+# Run Megatron h100 tests. Use --extra megatron so the megatron-only tests can collect.
```
SumanthRH:
```suggestion
# Run Megatron h100 tests. 
```

skyrl/backends/skyrl_train/distributed/fsdp_strategy.py
```diff
@@ -219,6 +219,33 @@ def _fsdp_init_model(self, model, is_train=True, is_wrapped=False):
         }
         module = model.model if is_wrapped else model
         full_state = module.state_dict()
+
+        # Move the entire module to meta before apply_fsdp2 so the sharded
```
SumanthRH, line 222: The changes look good here. It would be good to run `tests/backends/skyrl_train/gpu/gpu_ci/test_meta_init.py` after these changes to ensure meta init is correct. We added this test recently #1725

tests/backends/skyrl_train/gpu/gpu_ci/conftest.py
hao-aaron: this might be something wonky on my setup that needed it, can take it out

skyrl/backends/skyrl_train/distributed/fsdp_strategy.py
hao-aaron: ran and ensured it is passing

SumanthRH (approved):

PR #1690, opened 2026-05-19
feat: add max_training_steps config for CI smoke tests

dinhxuanvu:
## Summary

Adds a `max_training_steps` config field that caps training after N steps regardless of epochs or dataset size. This enables fast CI smoke tests that validate full training pipelines end-to-end without running to completion.

- Supported in all trainer paths: sync RL, async RL, and SFT
- Validated as a positive integer when set (ValueError on <= 0)
- SFT: caps `num_steps` after resolution from `num_epochs` and propagates to the LR scheduler
- RL: caps `total_training_steps` at initialization and adds an early exit after step completion
- Async RL: cancels in-flight generation tasks on early exit to avoid resource leaks

## Motivation

Running full training configs in nightly CI is prohibitively expensive. `max_training_steps=2` exercises the core pipeline — data loading, model init, forward/backward, and logging — in minutes rather than hours/days, validating that the full stack initializes and trains without errors. Unlike `dummy_run_max_steps` (SFT-only, uses fabricated data), this uses real data and works across RL and SFT configs.

## Test plan

- [x] Unit tests covering config construction, validation, capping logic, early exit, and post-loop cleanup
- [x] 38 tests (33 pass, 5 skip on macOS due to `transformers` unavailability)
- [x] Pre-commit checks pass (ruff, black)
- [ ] GPU integration test with a real config + `max_training_steps=2` (requires cluster)


skyrl/train/trainer.py
```diff
@@ -353,8 +356,20 @@ async def train(self):
 
                 self.global_step += 1
 
+                if (
+                    self.cfg.trainer.max_training_steps is not None
+                    and self.global_step > self.cfg.trainer.max_training_steps
+                ):
+                    logger.info(f"Reached max_training_steps={self.cfg.trainer.max_training_steps}, stopping early.")
+                    self.global_step = self.cfg.trainer.max_training_steps
```
SumanthRH: this logic is not needed after merging from `main`

We decrement the global step by 1 before a final checkpoint save.

https://github.com/NovaSky-AI/SkyRL/blob/b3cd1ac7f171fd3c0369d26b61bbca7b9f5cd2ff/skyrl/train/trainer.py#L464-L467

tests/train/test_max_training_steps.py
SumanthRH, line 1: I would avoid such extensive mocking

I believe you should be able to get away with the FakeActorGroup class without extensive mocking for workers or transformers? `transformers` is  a dependency in `skyrl-train`

skyrl/train/fully_async_trainer.py
```diff
@@ -457,6 +460,18 @@ async def train(self):
                 self.all_timings = {}
                 self.global_step += 1
 
+                if (
+                    self.cfg.trainer.max_training_steps is not None
+                    and self.global_step > self.cfg.trainer.max_training_steps
+                ):
+                    logger.info(f"Reached max_training_steps={self.cfg.trainer.max_training_steps}, stopping early.")
+                    self.global_step = self.cfg.trainer.max_training_steps
```
SumanthRH: Same comment: Please merge from `main` and we can avoid this

dinhxuanvu:
@SumanthRH Thanks for the review. I rebased and addressed your feedback. PTAL.

tests/train/test_max_training_steps.py
```diff
@@ -0,0 +1,341 @@
...
+            total = min(total, max_steps)
+        assert total == expected
+
+
+# ---------------------------------------------------------------------------
+# RL Trainer: early exit condition (shared by sync and async)
+# ---------------------------------------------------------------------------
+
+
+class TestRLTrainerEarlyExit:
```
SumanthRH: I would actually not have these tests - they are simply validating the code in the test rather than an actual test with the RL trainer.

tests/train/test_max_training_steps.py
```diff
@@ -0,0 +1,341 @@
...
+            "caps_explicit_num_steps",
+            "no_cap_when_none",
+            "no_effect_larger",
+            "epoch_resolution_no_cap",
+            "single_step",
+        ],
+    )
+    def test_capping(self, kwargs, expected):
+        assert self._resolve_num_steps(**kwargs) == expected
+
```
SumanthRH: this is again just testing the code in the test rather than the actual resolution in SFTTrainer. We can place the num_steps logic in an internal method in SFTTrainer and add this test against it, otherwise we shouldn't have this test

SumanthRH:
left a couple nits @dinhxuanvu , should be gtg with these changes!

tests/train/test_max_training_steps.py
dinhxuanvu: Removed both TestRLTrainerStepsCapping and TestRLTrainerEarlyExit. Done.

tests/train/test_max_training_steps.py
dinhxuanvu: Took your suggestion and reworked the test. Thanks.

dinhxuanvu:
> left a couple nits @dinhxuanvu , should be gtg with these changes!

Appreciated the feedback. Addressed. PTAL.

SumanthRH (approved):

SumanthRH:
LFG

PR #1691, opened 2026-05-19
fix: prevent process hang when vLLM engine dies during async generation

dinhxuanvu:
## Summary

- Fix process hang caused by `raise RuntimeError()` inside `asyncio.create_task()` being silently swallowed
- Use `os._exit(1)` instead of `sys.exit(1)` to force-terminate from async task context
- Release capacity slot on `CancelledError` to prevent assertion failures in `validate_state_at_epoch_end`
- Reset `slot_acquired` after `on_rollout_accepted()` to avoid stale state across loop iterations

## Problem

`_run_generate_for_a_group_loop` runs inside `asyncio.create_task()`. When vLLM's `EngineDeadError` occurs during generation, `slot_acquired` is True, so the handler executes:

```python
raise RuntimeError("Generation workers should only run into error when they finish running.")
```

This exception is **silently swallowed** by asyncio (task exceptions only surface if you `await` the task). The `sys.exit(1)` on the next line is never reached because `raise` exits the except block.

Even if `sys.exit(1)` were reached, it raises `SystemExit` which is also captured by the asyncio task machinery and does not terminate the process.

The training process hangs indefinitely with all GPUs idle.

## Fix

1. Remove the `raise RuntimeError()` guards that prevented process termination
2. Use `os._exit(1)` which bypasses asyncio and forces immediate termination
3. On `CancelledError`, call `on_rollout_rejected()` to release the capacity slot
4. Reset `slot_acquired = False` after `on_rollout_accepted()` for correct state tracking

## Impact

Without this fix, a single `EngineDeadError` causes all training GPUs to sit idle indefinitely. Observed in production: 128 H100 GPUs were idle for 5+ days due to this bug.

## Test plan

- [x] Verified the diff is minimal (6 insertions, 10 deletions) and only touches exception handling logic
- [x] Existing tests pass (no behavioral change in the happy path)
- [ ] Integration test: kill a vLLM engine during generation and confirm the process exits instead of hanging

### DIFF
--- a/skyrl/train/fully_async_trainer.py
+++ b/skyrl/train/fully_async_trainer.py
@@ -595,8 +595,8 @@ class FullyAsyncRayPPOTrainer(RayPPOTrainer):
                 slot_acquired = False
         except asyncio.CancelledError:
             if "slot_acquired" in locals() and slot_acquired:
-                logger.warning("Generation worker cancelled while slot was acquired. Releasing slot.")
-                await self._staleness_manager.on_rollout_rejected()
+                logger.error("Generation worker cancelled mid-flight (slot acquired). Exiting.")
+                os._exit(1)
             return
         except Exception as e:
             logger.error(f"Generator worker errored out with exception: {e}")
--- a/skyrl/train/fully_async_trainer.py
+++ b/skyrl/train/fully_async_trainer.py
@@ -14,7 +14,6 @@ High-level notes:
 import asyncio
 import inspect
 import os
-import sys
 import traceback
 from dataclasses import dataclass
 from typing import Iterable, List, Set, Tuple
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
@SumanthRH FYI

skyrl/train/fully_async_trainer.py
```diff
@@ -596,19 +592,24 @@ async def _run_generate_for_a_group_loop(self, generation_output_group_buffer: a
...
                 await self._staleness_manager.on_rollout_accepted()
+                slot_acquired = False
         except asyncio.CancelledError:
-            # If a slot was acquired but we exit early, release running count
-            try:
-                if "slot_acquired" in locals() and slot_acquired:
-                    raise RuntimeError("Generation workers should only be cancelled when they finish running.")
-            finally:
-                return
+            if "slot_acquired" in locals() and slot_acquired:
```
SumanthRH, line 597: THis is a good fix! However, the current async trainer has some more assumptions around failures - I believe currently we expect generations for all prompts to go through without failures and thus there is likely to be an error with this fix: https://github.com/NovaSky-AI/SkyRL/blob/ccc181e27c04b9f02fe1f7d30483aad96902d7a5/skyrl/train/fully_async_trainer.py#L148-L151

skyrl/train/fully_async_trainer.py
```diff
@@ -596,19 +592,24 @@ async def _run_generate_for_a_group_loop(self, generation_output_group_buffer: a
...
+                logger.warning("Generation worker cancelled while slot was acquired. Releasing slot.")
+                await self._staleness_manager.on_rollout_rejected()
+            return
         except Exception as e:
             logger.error(f"Generator worker errored out with exception: {e}")
             logger.error(f"Traceback: \n{traceback.format_exc()}")
-            if "slot_acquired" in locals() and slot_acquired:
-                raise RuntimeError("Generation workers should only run into error when they finish running.")
-            sys.exit(1)
+            os._exit(1)
```
SumanthRH, line 604: Curious if child processes like vLLM actors spawned from this process were killed with `os._exit`? Or did you have to do manual cleanup

skyrl/train/fully_async_trainer.py
dinhxuanvu: Removed `on_rollout_rejected()` from the `CancelledError` handler and it would break the `submitted == accepted` invariant. If `slot_acquired` is true during cancellation, then exit.
Let me know what you think.

skyrl/train/fully_async_trainer.py
dinhxuanvu: `os._exit(1)` would kill the driver on the head node so Ray GCS would terminate all of workers. This should take care of the cleanup. I don't think manual cleanup is needed here.

SumanthRH: Hmm actually the trainer code doesn't run on the driver, it runs in a task `skyrl_entrypoint` at the moment, which is why I was curious

SumanthRH (approved):

PR #1698, opened 2026-05-26
feat(tracking): plumb wandb tags through Tracking init

dinhxuanvu:
## Why

`Tracking` doesn't currently pass `tags` to `wandb.init`, so SkyRL-driven runs are always untagged. Tags are W&B's primary primitive for filtering, grouping, and organizing runs in the UI; not having them means callers can't distinguish, e.g., "baseline vs. ablation," "SFT vs. RL," or sibling runs in a sweep.

This also closes a small surprise gap: the lower-level `WandbTracker` in `skyrl/utils/log.py` already plumbs arbitrary kwargs to `wandb.init`, so power users discovered they could set tags through that path but not through the supported `Tracking` adapter.

## What

- `Tracking.__init__` accepts a new explicit `tags: Optional[List[str]] = None` parameter.
- The wandb branch forwards it to `wandb.init(tags=tags)`.
- New `trainer.tags` config field (default `None`).
- `get_tracker` in `main_base.py` forwards `cfg.trainer.tags`.
- The SFT path mirrors this: `SFTConfig.tags` bridges to `cfg.trainer.tags`, and `SFTTrainer._init_tracker` forwards it.
- Tags are wandb-specific; other backends are unchanged.

## Why an explicit param

`Tracking` is an adapter over five backends (wandb, mlflow, swanlab, tensorboard, console). A `**kwargs` passthrough would couple the adapter to wandb's API surface. An explicit, typed parameter keeps the contract clear and leaves a clean place to add matching primitives (mlflow tags, etc.) without breaking changes.

## Test plan

- [x] Unit test: `Tracking(..., tags=["a","b"])` results in `wandb.init(..., tags=["a","b"])`.
- [x] Unit test: default `tags=None` is forwarded when not passed.
- [x] Format/lint clean.
- [x] No behavior change when `tags` is not passed.

## Compatibility

Strictly additive. `wandb.init(tags=None)` is a no-op; the SDK still honors `WANDB_TAGS` env var.

### DIFF
--- a/docs/content/docs/configuration/config.mdx
+++ b/docs/content/docs/configuration/config.mdx
@@ -119,6 +119,7 @@ log_path: "/tmp/skyrl-logs"
 dump_data_batch: false
 dump_eval_results: true
 log_example_interval: 1
+tags: null
 
 ```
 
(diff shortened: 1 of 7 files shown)


dinhxuanvu:
Please ignore the Gemini's review as there were some unrelated commits which were pushed accidentally and Gemini reviewed polluted code which was incorrect. These were removed.

SumanthRH (approved):

SumanthRH:
Can you resolve merge conflicts  @dinhxuanvu ?

dinhxuanvu:
@SumanthRH Sorry for the delay. The PR is now rebased.

PR #1700, opened 2026-05-26
setup-slurm: default key path to ~/.ssh/runpod_key

dzorlu:
## Summary
- Default for `RUNPOD_KEY` now matches the script's comment and error message (`~/.ssh/runpod_key`) instead of `~/.ssh/id_ed25519`.
- Fixes "Permission denied (publickey,password)" for users who saved their RunPod key at the documented path. With `IdentitiesOnly yes` written into `~/.slurm/config`, SSH refuses to fall back to other keys, so the wrong default silently breaks the verify step.
- `RUNPOD_SSH_KEY` env var override is unchanged.

## Test plan
- [x] `bash scripts/setup-slurm.sh` succeeds when `~/.ssh/runpod_key` exists and no env var is set
- [x] `RUNPOD_SSH_KEY=/path/to/other_key bash scripts/setup-slurm.sh` still respects the override


dzorlu:
Wrong repo — opening against fleet-ai/SkyRL-Fleet instead.

PR #1722, opened 2026-05-28
[fix] handle "uni" processor case on new inference code path

erictang000:
### Problem

  Running two colocated TP=1 jobs on the same node (e.g. two instances of
  `run_dapo_qwen3_1.7b_aime.sh`) caused the second run's vLLM to start on `cuda:0` —
  already occupied by the first run's training process — instead of its own
  placement-group GPU:

```
  ValueError: Free memory on device cuda:0 (17.41/178.35 GiB) on startup is less
  than desired GPU memory utilization (0.8, 142.68 GiB).
```
  The new inference server path only supported the `ray`/`mp` executor backends. For
  single-GPU engines it used `ray`, scheduling the actor with `num_gpus=0` and relying
  on vLLM's Ray worker placement + `VLLM_RAY_BUNDLE_INDICES`. With
  `RAY_EXPERIMENTAL_NOSET_*_VISIBLE_DEVICES` set (SkyRL's default) and multiple
  placement groups on one node, that single-GPU worker placement does not reliably land
  on the bundle's physical GPU.

  ### Fix

  Port the proven `uni` (in-process) backend from the legacy path
  (`ray_wrapped_inference_engine.py`) into the new server path. For `TP × PP == 1`,
  vLLM now runs in-process inside the actor, which is scheduled directly onto its
  placement-group bundle and holds the GPU — so it always lands on the correct physical
  GPU regardless of how many jobs share the node.

  - **`server_group.py`**: resolve the backend to `uni` when `num_gpus_per_server == 1`;
    the actor itself requests the GPU (`0.2` colocated / `1.0` non-colocated) via the new
    `_actor_num_gpus()`.
  - **`vllm_server_actor.py`**: add the `uni` branch — when NOSET is set, pin
    `CUDA_VISIBLE_DEVICES = ray.get_gpu_ids()[0]` and skip the Ray-only env vars.

  The change is **resource-neutral** (same GPU fraction as today's
  `VLLM_RAY_PER_WORKER_GPUS`, just requested on the actor instead of a spawned worker)
  and leaves the TP>1 `ray`/`mp` path untouched.

  ### Testing

  - [x] `ruff` / `black` clean; both modules import.
  - [x] Logic test confirms backend resolution and GPU-fraction selection across
    TP=1/TP=2 × colocated/non-colocated.
  - [ ] TP=2 GPU CI regression (`test_inference_server_group.py`)
  - [ ] Two-concurrent-jobs repro on a node with ≥8 GPUs

  That's the raw GitHub-flavored markdown — copy the block above into the PR description. Want me to open the PR with gh using this?

### DIFF
--- a/skyrl/backends/skyrl_train/inference_servers/server_group.py
+++ b/skyrl/backends/skyrl_train/inference_servers/server_group.py
@@ -25,6 +25,10 @@ logger = logging.getLogger(__name__)
 # about the GPU ID to pack actors appropriately on different nodes.
 # Thus we use a fractional CPU allocation for colocated actors.
 COLOCATED_ACTOR_CPU_FRACTION = 0.2
+# For the "uni" backend (single-GPU engines), the actor itself is the vLLM worker
+# and holds the GPU. Colocated engines share each GPU bundle with a training worker
+# (fractional); non-colocated engines own the whole GPU.
+COLOCATED_ACTOR_GPU_FRACTION = 0.2
 
 
 class ServerGroup:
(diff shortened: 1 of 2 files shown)


erictang000:
accidental ai slop, the real issue was not running `ray start` so the placement groups overlapped

PR #1729, opened 2026-05-29
fleet-common: scope DATA_ROOT by sky cluster on shared /workspace

dzorlu:
## Summary

Concurrent training launches with the same `MODALITY` on the RunPod slurm cluster were stomping on each other's prepared parquet files. Both `fleet-common-setup.sh` and `fleet-common-run.sh` wrote/read `/workspace/data/fleet/{modality}/train.parquet` without any per-cluster namespacing, so two concurrent browser_use runs (or two concurrent computer_use, etc.) silently corrupted each other's training data.

**Concrete incident**: a browser_use run with `DATA_VERSION=sre-tasks` (105 rows) wrote sre-tasks data. A second concurrent browser_use run with `DATA_VERSION=v6` (10K rows) overwrote the same file. The first cluster's training loaded the clobbered v6 data and trained on the wrong dataset for 1h+ before being caught.

## Fix

Derive the sky cluster name from the script's absolute path (sky places scripts under `.sky_clusters/<name>/sky_workdir/`) and include it in `DATA_ROOT` when running on shared `/workspace`:

```
DATA_ROOT=/workspace/clusters/<sky-cluster-name>
```

Both `setup` and `run` scripts use identical resolution so they agree on every node. Falls back to plain `/workspace` if the path doesn't contain `.sky_clusters/` (e.g. when invoked outside a sky-managed env, or single-cluster RunPod direct).

## Trade-off

Each cluster re-downloads its dataset (no shared cache across clusters). Data is tiny (sre-tasks ~150 KB, v6 ~3 MB), so re-download is trivial vs. silent data corruption.

## Test plan

- [x] Bash syntax: `bash -n` on both scripts
- [x] Regex extraction: verified `sed` resolves cluster name from sky workdir paths and returns empty for non-sky paths
- [ ] Smoke: launch two concurrent same-modality runs and verify they each have their own `/workspace/clusters/<name>/data/fleet/<modality>/train.parquet`


dzorlu:
Wrong remote, closing. Opening on fleet-ai/skyrl-fleet instead.

PR #1732, opened 2026-05-29
[tinker] Fix checkpointing for multi-node runs

pcmoritz:
Before this fix, there would be an error like
```
  2026-05-29 20:09:06,716 - ERROR - skyrl: Error saving checkpoint for model model_5b6cb10d, checkpoint <...>: [Errno 2] No such file
  or directory: '/tmp/tmp6kkqkzrg/model'
```


erictang000 (approved):
makes sense to me!

PR #1662, opened 2026-05-13
vllm metrics docs

hao-aaron:
add docs for vllm metrics. Also, ran /examples/train/megatron/run_megatron_dapo_qwen3.5_35b_a3b.sh to ensure that mp backend also logs metrics correctly:
https://wandb.ai/sky-posttraining-uc-berkeley/qwen3_5_dapo/runs/mhq1q8aa?nw=nwuserahao

PR #1736, opened 2026-06-01
[docs] Fix prod build: use `text` instead of unknown `promql` code fence

hao-aaron:
The vllm-metrics doc (#1662) uses a ```promql code fence, but Shiki ships no `promql` grammar, so `next build` aborts with a ShikiError and the Vercel docs deploy fails. Switch the fence to ```text so it renders unhighlighted instead of breaking the build.


SumanthRH (approved):
LGTM, deploying to Vercel to confirm

PR #1743, opened 2026-06-02
Avoid dense packed logits in Megatron logprobs

jinghanyao1-hub:
# PR README: Memory-safe Megatron packed logprobs

## Problem

With Megatron context parallelism and packed microbatches enabled, the old wrapper path
postprocessed packed logits back into dense batch form before computing token logprobs:

```text
packed CP logits [1, T_local, V_tp]
  -> postprocess_packed_seqs(...)
  -> dense logits [B, S, V_tp]
  -> selected token logprobs [B, S-1]
```

That dense `[B, S, V_tp]` materialization causes a large transient memory spike.

## 8k memory comparison

At the same 8k setup on `Qwen/Qwen3-4B-Instruct-2507`, the new packed-logprob path removes the
dense-logit spike:

| path | profile | peak allocated | peak reserved |
| --- | --- | ---: | ---: |
| old dense packed-logit path | 8k CUDA memory snapshot | 41.31 GiB | 47.06 GiB |
| new packed-logprob path | 8k CUDA memory snapshots, max across ranks | 29.97 GiB | 30.03 GiB |

This is an 11.34 GiB allocated-memory reduction and a 17.03 GiB reserved-memory reduction at 8k.
Both numbers were reconstructed from PyTorch CUDA memory snapshots of the policy forward/backward
region.

## Test setup

Benchmarks used:

```text
model: Qwen/Qwen3-4B-Instruct-2507
node config: 1 node, 8 GPUs
Megatron parallelism: TP=1, PP=1, CP=8
policy/ref placement: 8 GPUs per node, colocate_all=true
CP communication: p2p and a2a
micro train/forward batch size per GPU: 1
train batch size: 1
response length: 1024
```

## Change

The new wrapper path keeps logits in packed/context-parallel form and computes only the selected
token logprobs:

```text
packed CP logits [1, T_local, V_tp]
  + packed target ids [1, T_padded]
  -> from_parallel_logits_to_logprobs_packed_sequences(...)
  -> packed scalar logprobs [B, S-1]
  -> action logprobs [B, num_actions]
```

Concretely, when `remove_microbatch_padding` creates `packed_seq_params`, the wrapper now:

1. Builds matching packed target token IDs.
2. Calls `from_parallel_logits_to_logprobs_packed_sequences(...)`.
3. All-gathers only scalar logprobs across CP ranks.
4. Skips `postprocess_packed_seqs(...)` for packed logits.

The normal non-packed logprob path is unchanged.

## 128k result

After removing the 8k dense-logit memory spike, the patched path completed the final CP benchmark
sweep through 128k total length:

At 128k total tokens (`max_total_len=131072`, prompt `130048`, response `1024`):

```text
p2p peak allocated: 53.53 GiB
p2p peak reserved:  57.26 GiB
a2a peak allocated: 53.53 GiB
a2a peak reserved:  56.68 GiB
```

So the patch reduces 8k peak memory substantially and enables clean 128k packed Megatron runs,
without gathering dense `[B, S, V]` logits.

### DIFF
--- a/skyrl/backends/skyrl_train/workers/megatron/megatron_model_wrapper.py
+++ b/skyrl/backends/skyrl_train/workers/megatron/megatron_model_wrapper.py
@@ -12,13 +12,13 @@ from omegaconf import OmegaConf
 from skyrl.backends.skyrl_train.distributed.megatron.megatron_utils import (
     get_model_config,
     make_batch_generator,
-    postprocess_packed_seqs,
     preprocess_packed_seqs,
     recover_left_padding,
     remove_left_padding,
 )
 from skyrl.backends.skyrl_train.distributed.megatron.model_utils import (
     from_parallel_logits_to_logprobs,
+    from_parallel_logits_to_logprobs_packed_sequences,
     vocab_parallel_entropy,
 )
 from skyrl.backends.skyrl_train.utils.ppo_utils import (
@@ -31,6 +31,23 @@ from skyrl.backends.skyrl_train.utils.replay_utils import (
 )
 from skyrl.backends.skyrl_train.utils.torch_utils import masked_mean
 from skyrl.train.config import TrainerConfig
+
+
+def _build_packed_targets(
+    sequences: torch.Tensor,
+    attention_mask: torch.Tensor,
+    packed_seq_params,
+) -> torch.Tensor:
+    """Pack full target token IDs without context-parallel sharding."""
+    attention_mask = attention_mask.to(device=sequences.device, dtype=torch.bool)
+    cu_padded = packed_seq_params.cu_seqlens_q_padded.to(device=sequences.device, dtype=torch.long)
+    total_padded_tokens = cu_padded[-1]
+
+    targets = torch.zeros((total_padded_tokens,), dtype=sequences.dtype, device=sequences.device)
+    token_offsets = attention_mask.to(torch.long).cumsum(dim=1) - 1
+    packed_indices = cu_padded[:-1].unsqueeze(1) + token_offsets
+    targets[packed_indices[attention_mask]] = sequences[attention_mask]
+    return targets.unsqueeze(0)
 
 
 class MegatronModelWrapper:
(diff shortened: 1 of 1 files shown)


skyrl/backends/skyrl_train/workers/megatron/megatron_model_wrapper.py
```diff
@@ -33,6 +33,24 @@
...
+    cu_padded_cpu = packed_seq_params.cu_seqlens_q_padded.tolist()
+    seq_lens_cpu = attention_mask.sum(dim=1, dtype=torch.int32).cpu().tolist()
+    total_padded_tokens = cu_padded_cpu[-1]
+
+    targets = torch.zeros(total_padded_tokens, dtype=sequences.dtype, device=sequences.device)
+    for i, seq_len in enumerate(seq_lens_cpu):
+        start = cu_padded_cpu[i]
+        if seq_len > 0:
+            targets[start : start + seq_len] = sequences[i, attention_mask[i]]
+    return targets.unsqueeze(0)
```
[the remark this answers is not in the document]
erictang000: this seems resolved

skyrl/backends/skyrl_train/workers/megatron/megatron_model_wrapper.py
```diff
@@ -12,13 +12,13 @@
 from skyrl.backends.skyrl_train.distributed.megatron.megatron_utils import (
```
erictang000, line 12:
```bash
uv run --isolated --extra dev --extra megatron -- pytest -s tests/backends/skyrl_train/gpu/gpu_ci/megatron/test_megatron_worker.py::test_megatron_forward -k cp_2_policy_seq_packing
```

These tests currently fail with the new implementation with error:
```bash
        if ep == 1:
            # Loose bound to absorb bf16 reduction-order noise under combined
            # parallelism (e.g. tp=2 + cp=2); avg_diff stays in line (~0.06).
>           assert max_diff < 5e-1, f"Max diff {max_diff} is too large"
E           AssertionError: Max diff 34.635921478271484 is too large
E           assert tensor(34.6359) < 0.5
```

it seems like there may be a correctness issue with the new implementation, can you look into this? We should make sure that both `test_megatron_forward` and `test_megatron_train` tests that exercise the cp + seq packing paths are still passing

Ideally we can get a run with https://github.com/NovaSky-AI/SkyRL/blob/main/examples/train/megatron/run_megatron.sh working and showing that it matches the baseline reward with CP passing as well

erictang000: we should validate the above for both ring + a2a, and add an explicit test for a2a in `test_megatron_worker.py`

skyrl/backends/skyrl_train/workers/megatron/megatron_model_wrapper.py
```diff
@@ -148,16 +183,7 @@ def forward_step(batch_iter, model):
                 packed_seq_params=packed_seq_params,
             )
 
-            if self.remove_microbatch_padding:
-                outputs = postprocess_packed_seqs(
```
erictang000, line 152: if this is no longer needed after the above change to use `from_parallel_logits_to_logprobs_packed_sequences`, then we should remove the `postprocess_packed_seqs` function where it is defined since it will be dead code.

but we should double check the correctness before removal. also would be good to check sanity check that CP + PP still work together

erictang000 (commented):
seeing some failing correctness tests on this PR that we should resolve + a comment on the removal of entropy calculation + removing calls to `postprocess_packed_seqs`

skyrl/backends/skyrl_train/workers/megatron/megatron_model_wrapper.py
```diff
@@ -332,15 +374,16 @@ def loss_func(logits, data):
...
+            if loss_config.use_entropy_loss:
                 action_logits = logits[:, -num_actions - 1 : -1, :]
                 entropy_BS = vocab_parallel_entropy(action_logits)
                 entropy = masked_mean(entropy_BS, loss_mask)
-
-            if loss_config.use_entropy_loss:
                 entropy_loss_term = entropy * loss_config.entropy_loss_coef
             else:
-                entropy_loss_term = torch.tensor(0.0)
+                entropy = torch.tensor(0.0, device=logits.device)
```
erictang000, line 385: hmm computing entropy here is still pretty important, since it's a pretty important metric for monitoring training stability, even if it's not used for the loss - does this operation cause another memory spike and that's why it's disabled?

erictang000: actually it seems like we could keep the entropy calculation here using `_VocabParallelEntropy` as long as we modify that logic that relies on

```python
  action_logits = logits[:, -num_actions - 1 : -1, :]   # assumes [batch, seq, vocab], CP-gathered, unpacked
```

since `_VocabParallelEntropy` only relies on tp gathering and entropy gets computed per token.

We can still using `_VocabParallelEntropy` on the packed CP sharded logits and then do a reduction (accounting for the number of tokens on each CP rank) to get the mean entropy

jinghanyao1-hub:
Close this PR as the rebase and more modifications are included in a new PR #1751. With the proposed feature also adapting to SFT (per the current upstream SFT implementation).

PR #1746, opened 2026-06-03
Make prompt dataloader `num_workers`/`persistent_workers` configurable

### LINKED ISSUE #1744: Configuration for SkyRL to store training data in-memory, instead of reloading every epoch
My compute cluster's NFS filesystem can get overwhelmed. I occasionally see training jobs unnecessarily die when reloading the dataset from the filesystem.

With SkyRL v0.2.0's dataloader, it tears down and re-creates its worker processes every epoch, leading to training code re-reading in the dataset from the filesystem every epoch. This issue is exacerbated when using train batch size = dataset size, as the data gets reloaded from the filesystem every step. My request is to allow SkyRL to just load the training and val datasets _once_ into memory (they are <2 KiB `.parquet`) and use across epochs.

In SkyRL v0.2.0:
- The `dataloader.num_workers` is 8 by default: https://github.com/NovaSky-AI/SkyRL/blob/skyrl-v0.2.0/skyrl/train/utils/trainer_utils.py#L779
- Setting `num_workers` to 0 disables multiprocessing (uses [`_StatefulSingleProcessDataLoaderIter`](https://github.com/meta-pytorch/data/blob/v0.11.0/torchdata/stateful_dataloader/stateful_dataloader.py#L383-L384)), which has the benefit of using an in-memory dataset: https://github.com/NovaSky-AI/SkyRL/blob/skyrl-v0.2.0/skyrl/train/dataset/dataset.py#L43

The request is to:

- Allow configuration of `dataloader.num_workers` (not having it always be auto-derived), so one can specify `num_workers=0`
- Creating some user-facing `in_memory` or `persistent_workers` configuration

jamesbraza:
Adds a `data.dataloader` config section exposing `num_workers` and `persistent_workers`(previously hardcoded) with backward-compatible defaults.

Closes https://github.com/NovaSky-AI/SkyRL/issues/1744.

### DIFF
--- a/skyrl/train/config/config.py
+++ b/skyrl/train/config/config.py
@@ -55,7 +55,7 @@ class DataLoaderConfig(BaseConfig):
         metadata={
             "help": (
                 "Keep DataLoader workers alive across epochs instead of respawning them at "
-                "every epoch boundary. Setting this requires num_workers above 0."
+                "every epoch boundary. Setting this requires `num_workers > 0`"
             )
         },
     )
--- a/skyrl/train/config/config.py
+++ b/skyrl/train/config/config.py
@@ -39,9 +39,37 @@ class BaseConfig(ABC):
 
 
 @dataclass
+class DataLoaderConfig(BaseConfig):
+    num_workers: Optional[int] = field(
+        default=None,
+        metadata={
+            "help": (
+                "Prompt DataLoader worker processes. Default of None auto-derives the value "
+                "(0 with the inference HTTP endpoint, else 8). Set 0 for in-process loading "
+                "that never respawns workers at epoch boundaries."
+            )
+        },
+    )
+    persistent_workers: bool = field(
+        default=False,
+        metadata={
+            "help": (
+                "Keep DataLoader workers alive across epochs instead of respawning them at "
+                "every epoch boundary. Setting this requires num_workers above 0."
+            )
+        },
+    )
+
+    def __post_init__(self) -> None:
+        if self.num_workers is not None and self.num_workers < 0:
+            raise ValueError(f"data.dataloader.num_workers must be None or >= 0, got {self.num_workers}.")
+
+
+@dataclass
 class DataConfig(BaseConfig):
     train_data: List[str] = field(default_factory=lambda: [os.path.expanduser("~/data/gsm8k/train.parquet")])
     val_data: List[str] = field(default_factory=lambda: [os.path.expanduser("~/data/gsm8k/validation.parquet")])
+    dataloader: DataLoaderConfig = field(default_factory=DataLoaderConfig)
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


skyrl/train/utils/trainer_utils.py
```diff
@@ -770,19 +770,22 @@ def build_dataloader(
...
+
     dataloader = StatefulDataLoader(
         dataset,
         batch_size=batch_size if not is_fully_async else 1,
         shuffle=True if is_train else False,
         collate_fn=dataset.collate_fn,
-        # TODO(Charlie): debug why inference http endpoint is slow when num_workers is 8
-        num_workers=0 if cfg.generator.inference_engine.enable_http_endpoint else 8,
+        num_workers=num_workers,
+        persistent_workers=cfg.data.dataloader.persistent_workers,
```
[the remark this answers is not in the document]
jamesbraza: The eval dataloader isn't iterated once and discarded — `RayPPOTrainer` builds it once in `__init__` (`self.eval_dataloader`) and re-iterates it on every evaluation (`eval_before_train`, then every `eval_interval` steps). So `persistent_workers=True` keeps its workers alive between evaluations instead of respawning them each time — exactly the respawn this knob avoids (each respawn re-imports the venv, which can hang on a slow/NFS filesystem); the train dataloader benefits the same way across epochs. Forcing it off for `is_train=False` would also silently override the user's explicit opt-in, so I've kept it a passthrough and added a comment noting why it isn't branched on `is_train`.

skyrl/train/config/config.py
```diff
@@ -38,10 +38,38 @@ def from_dict_config(cls, cfg: DictConfig) -> "BaseConfig":
...
+                "that never respawns workers at epoch boundaries."
+            )
+        },
+    )
+    persistent_workers: bool = field(
+        default=False,
+        metadata={
+            "help": (
+                "Keep DataLoader workers alive across epochs instead of respawning them at "
+                "every epoch boundary. Setting this requires num_workers above 0."
```
SumanthRH:
```suggestion
                "every epoch boundary. Setting this requires `num_workers > 0`"
```

skyrl/train/utils/trainer_utils.py
```diff
@@ -770,19 +770,24 @@ def build_dataloader(
     seeded_generator = torch.Generator()
     seeded_generator.manual_seed(cfg.trainer.seed)
 
+    num_workers = cfg.data.dataloader.num_workers
+    assert num_workers is not None, "num_workers is resolved to an int in SkyRLTrainConfig.__post_init__"
```
SumanthRH:
```suggestion
    assert num_workers is not None, "dataloader `num_workers` should be non-null"
```

SumanthRH (approved):
Left some minor feedback, LGTM otherwise

PR #1760, opened 2026-06-08
[ci] add commit names to gpu ci jobs to avoid collisions with `anyscale job wait`

erictang000:
## Make Anyscale CI job names unique per run

**Problem:** All CI workflows submitted Anyscale jobs with a hardcoded `name`, then called `anyscale job wait --name <name>`. When two runs of the same workflow overlap, `job wait` resolves to the *most recently created* job of that name (per Anyscale's API), so a passing run and a failing run can report each other's status — marking both GitHub checks failed.

**Fix:** Each workflow now derives a unique job name `<base>-<short-sha>-<run_id>` and passes it to both `submit --name` and `wait --name`. The short SHA keeps names identifiable in the Anyscale console; `github.run_id` guarantees uniqueness even across concurrent runs of the same commit. Applied to all 13 submit/wait workflows; no `ci/*.yaml` configs needed changing since `--name` overrides the file's `name:`.

### DIFF
--- a/.github/workflows/gpu_ci_h100.yaml
+++ b/.github/workflows/gpu_ci_h100.yaml
@@ -65,5 +65,7 @@ jobs:
           ANYSCALE_CLI_TOKEN: ${{ secrets.ANYSCALE_CLI_TOKEN_STAGING }}
           ANYSCALE_HOST: https://console.anyscale-staging.com
         run: |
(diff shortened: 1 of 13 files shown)


erictang000:
example of this working here: https://github.com/NovaSky-AI/SkyRL/actions/runs/27165213480/job/80190489963

PR #1784, opened 2026-06-13
[logging] Flush pending step metrics to the tracker when training crashes

### LINKED ISSUE #1687: Recorded generation metrics unnecessarily lost when `RayPPOTrainer.fwd_logprobs_values_reward` raises
With SkyRL v0.2.0, I had a training run with:
1. A successful generation period taking 42-min. It completed all rollouts and collected valuable information (e.g. reward signal)
2. Hit an OOM (`torch.OutOfMemoryError`) during `RayPPOTrainer.fwd_logprobs_values_reward`'s reference model forward pass: https://github.com/NovaSky-AI/SkyRL/blob/skyrl-v0.2.0/skyrl/train/trainer.py#L1006

Due to the OOM, MLFlow never received metrics, even though many were already present in `RayPPOTrainer.all_metrics`:
- Reward metrics: `reward/avg_raw_reward`, `reward/mean_positive_reward`, etc.
- Timing metrics: `timing/generate`, `timing/convert_to_training_input`, etc.
- Many metrics in `generator_output["rollout_metrics"]`

This request is for SkyRL to better log so we don't have total information loss. Basically, can SkyRL add a `finally` branch that calls `tracker.log` when there is an unhandled exception, so metrics get preserved as the system goes down:
1. Put this in a `try`: https://github.com/NovaSky-AI/SkyRL/blob/skyrl-v0.2.0/skyrl/train/trainer.py#L215
2. Put `tracker.log` in a `finally`, to handle both normal exit and/or teardown: https://github.com/NovaSky-AI/SkyRL/blob/skyrl-v0.2.0/skyrl/train/trainer.py#L338

discobot:
Fixes #1687.

`train()` only logs `all_metrics`/`all_timings` once a step fully completes, so a mid-step crash drops everything already recorded for that step.

#1706 added the crash path that surfaces the exception itself to the tracker (`BasePPOExp.run` → `Tracking.log_exception`); this PR extends that same path to first flush the pending step metrics. A new `RayPPOTrainer.flush_pending_metrics()` logs the same `timing/`-prefixed payload shape as the per-step log at the current `global_step`. The flush must run before `log_exception`, because `log_exception` finishes the wandb run to push the traceback table. It never raises (a tracker failure must not mask the original exception) and clears the accumulators, so it is idempotent.

Keeping the flush on the `run()` except path rather than a `try/finally` inside `train()` keeps crash handling in the one place #1706 created, and covers `FullyAsyncRayPPOTrainer` (which inherits the method) for free.

## Test plan
- New CPU unit tests in `tests/train/test_trainer.py`: payload shape/step/clearing + idempotence, tracker failures swallowed, and `BasePPOExp.run` ordering (flush before `log_exception`, exception still re-raised).

### DIFF
--- a/skyrl/train/entrypoints/main_base.py
+++ b/skyrl/train/entrypoints/main_base.py
@@ -410,7 +410,7 @@ class BasePPOExp:
             if self.trainer is not None and self.trainer.tracker is not None:
                 # Flush metrics already recorded for the in-flight step (e.g.
                 # reward/timing metrics from a completed generation phase)
-                # before log_exception finishes the wandb run. See #1687.
+                # before log_exception finishes the wandb run.
                 self.trainer.flush_pending_metrics()
                 self.trainer.tracker.log_exception(e, step=self.trainer.global_step)
             else:
--- a/skyrl/train/entrypoints/main_base.py
+++ b/skyrl/train/entrypoints/main_base.py
@@ -408,6 +408,10 @@ class BasePPOExp:
             # worker logs; route them through the tracker so wandb users see
             # them as an `error/tracebacks` table row.
             if self.trainer is not None and self.trainer.tracker is not None:
+                # Flush metrics already recorded for the in-flight step (e.g.
+                # reward/timing metrics from a completed generation phase)
+                # before log_exception finishes the wandb run. See #1687.
+                self.trainer.flush_pending_metrics()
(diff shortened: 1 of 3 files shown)


erictang000 (approved):
trimmed some comments but lgtm thanks!

PR #1636, opened 2026-05-08
[feat] Add prometheus logging to new inference, add vllm metrics to wandb 

hao-aaron:
Connected prometheus logging on new inference, metrics can be viewed on anyscale serve llm panel. 

A couple vllm metrics were chosen to be logged per step on wandb in a separate vllm panel:
```
_GAUGE_NUM_RUNNING = "ray_vllm_num_requests_running"
_GAUGE_NUM_WAITING = "ray_vllm_num_requests_waiting"
_GAUGE_KV_CACHE_USAGE = "ray_vllm_kv_cache_usage_perc"
_COUNTER_PREFIX_QUERIES = "ray_vllm_prefix_cache_queries_total"
_COUNTER_PREFIX_HITS = "ray_vllm_prefix_cache_hits_total"
_COUNTER_PROMPT_TOKENS = "ray_vllm_prompt_tokens_total"
_COUNTER_GENERATION_TOKENS = "ray_vllm_generation_tokens_total"
_HIST_TTFT_SUM = "ray_vllm_time_to_first_token_seconds_sum"
_HIST_TTFT_COUNT = "ray_vllm_time_to_first_token_seconds_count"
_HIST_ITL_SUM = "ray_vllm_inter_token_latency_seconds_sum"
_HIST_ITL_COUNT = "ray_vllm_inter_token_latency_seconds_count"
```

Example run on async CI: https://wandb.ai/sky-posttraining-uc-berkeley/gsm8k_fully_async_ci/runs/jvzuyj0d?nw=nwuserahao

grafana: https://grafana-cld-hxkifz7xa22mwicp21nzkds1lw.anyscale-cloud.dev/d/5187f19337e14051361f870b0b3527e2/serve-llm-dashboard-2-51-1-5187f193?var-ClusterId=ses_d714lyt2s4glseublds6snxnkm&var-ProjectId=prj_4b6c498rypyq6g7yhk6vzgjevt&var-WorkspaceId=expwrk_i8nllm3il9kbezfzsvl26y6hi8&orgId=1&from=1778280294687&to=1778282094687

PR #1794, opened 2026-06-16
[Fix] vLLM Metrics Scrapper throughput calculation

zanderjiang:
For more context checkout: 
[#1636 comment 1](https://github.com/NovaSky-AI/SkyRL/pull/1636/changes#r3411074771)
[#1636 comment 2](https://github.com/NovaSky-AI/SkyRL/pull/1636/changes#r3411079924)

The current vLLM Metrics Scrapper calculates rollout throughput with the full step size time, and doesn't account for variances caused during eval (different sampling params, eval batch sizes). This PR addresses those issues by updating the timing logic and separating metrics for training and eval rollout (for synchronous RL, in async the rollouts are combined)


SumanthRH (commented):
Quick test: did you confirm that the reported throughput on wandb is close to the throughputs reported by vllm for that time window from prometheus server / from the vllm logs?

zanderjiang:
> Quick test: did you confirm that the reported throughput on wandb is close to the throughputs reported by vllm for that time window from prometheus server / from the vllm logs?

Yes, I checked the reported throughput against a stand alone vllm server on a similar workload, the throughput matches.

skyrl/train/utils/vllm_metrics_scraper.py
```diff
@@ -212,32 +213,95 @@ async def sample(self) -> Dict[str, float]:
...
+            out = self._window_metrics(None, snapshot, None, "vllm/")  # gauges only
 
         self._prev_aggregated = snapshot
         self._prev_timestamp = now
         return out
 
+    async def mark_pre_eval(self) -> None:
+        """Snapshot the counters right before the eval rollout for :meth:`sample_split`."""
+        self._mid_snapshot = await self._read_snapshot()
+
```
SumanthRH, line 262: I don't like this approach of marking eval boundary. It seems brittle relying on the ordering of `generate` and `eval` in the caller.

How about pushing the timing bookkeeping into the vllm metrics scraper? It can be something like this:


```python
class VLLMMetricsScraper:
    def __init__(
            self,
            urls: Optional[List[str]] = None,
            request_timeout_s: float = 2.0,
        ):
        ...

    def start(self, label):
        self._prev_snapshot = ...
        self._cur_time = time.monotonic()
        self._label = label
        self.window_time_s = 0
        ...
    
    def pause(self):
        if self.paused:
            raise ValueError("`pause` called without `resume`")
        cur_time = time.monotonic()
        self.window_time_s += cur_time - self._cur_time
        self.paused = True
    
    def resume(self):
        self._cur_time = time.monotonic()
        self.paused = False
    
    def stop(self) -> dict:
        new_snapshot = ...
        end_time = time.monotonic()
        self.window_time_s += end_time - self._cur_time
        ...
        metrics = self._window_metrics(self._prev_snapshot, new_snapshot, self.window_time_s, self._label)
        self._label = None
        self._cur_time = None
        return metrics # metrics nested as `{label}/generation_throughput_tok_s`



# usage
metrics_scraper = VLLMMetricsScraper(...)
# generate call
metrics_scraper.start("vllm/train")
with Timer("generate", self.all_timings):
    ...
train_vllm_metrics = metrics_scraper.stop()

# eval
metrics_scraper.start("vllm/eval")
metrics_scraper.pause()
for _, prompts in enumerate(eval_dataloader):
        pbar.update(1)
        generator_input, uids = prepare_generator_input(
            prompts,
            cfg.generator.eval_n_samples_per_prompt,
            get_sampling_params_for_backend(cfg.generator.inference_engine.backend, sampling_params),
            cfg.environment.env_class,
            "eval",
            global_step,
        )
        metrics_scraper.resume()
        generator_output: GeneratorOutput = await generator.generate(generator_input)
        metrics_scraper.pause()
eval_vllm_metrics = metrics_scraper.stop()
```

That way all the trainer / evaluation code can be clean

SumanthRH (changes requested):

skyrl/train/utils/vllm_metrics_scraper.py
zanderjiang: updated the metrics scrapper, now we keep track of train and eval metrics via labels

SumanthRH (approved):
lfg

PR #1798, opened 2026-06-16
[async] add flag to toggle whether to clear kv cache after weight sync for fully async training, update default max_staleness_steps to 0

erictang000:
Previously, we always reset prefix cache prior to weight sync, even for fully async training:

https://github.com/NovaSky-AI/SkyRL/blob/649b16fb58ddcba6e2288368ae1e7f24b0fed211/skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py#L1274

Adds a new flag `trainer.fully_async.clear_kv_cache_on_weight_sync` which defaults to `False`, meaning the default behavior during fully async training is to keep stale KV cache.

Also updates the default for `trainer.fully_async.max_staleness_steps` from `4` -> `0`, to be able to determine whether or not fully async training is enabled. All existing fully async scripts already manually set max_staleness_steps, and users should manually set this to a non zero value for all fully async scripts anyway.

### DIFF
--- a/docs/content/docs/configuration/config.mdx
+++ b/docs/content/docs/configuration/config.mdx
@@ -603,11 +603,13 @@ def ppo_policy_loss(
 fully_async:
   max_staleness_steps: 4
   num_parallel_generation_workers: 768
+  clear_kv_cache_on_weight_sync: true
 
 ```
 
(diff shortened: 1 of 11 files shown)


erictang000:
verified similar performance with not clearing kv cache

https://api.wandb.ai/links/sky-posttraining-uc-berkeley/p8c50yl8

image

CharlieFRuan:
I think max_staleness = 0 with fully async might still be used for debugging. Treating > 0 as a flag for fully async is a bit implicit to me.

Could we set `clear_kv_cache_on_weight_sync` default to True instead (which applies to sync RL), and add it as a flag to be false for the example fully async scripts?

PR #1799, opened 2026-06-17
[feat] Add a serve entrypoint to serve inference endpoints

SumanthRH:
# What does this PR do?

Adds a `serve` entrypoint to spin up inference servers with SkyRL.

### DIFF
--- a/skyrl/train/entrypoints/serve.py
+++ b/skyrl/train/entrypoints/serve.py
@@ -76,6 +76,9 @@ class InferenceOnlyEntrypoint(BasePPOExp):
         try:
             while True:
                 time.sleep(3600)
+        # TODO: Currently this is run inside a ray task so a KeyboardInterrupt on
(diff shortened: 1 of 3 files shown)


skyrl/train/entrypoints/serve.py
```diff
@@ -0,0 +1,149 @@
...
+        )
+
+        try:
+            while True:
+                time.sleep(3600)
+        # TODO: Currently this is run inside a ray task so a KeyboardInterrupt on
+        # driver never reaches here - Ray sends a SIGTERM. We should propagate the
+        # interrupt for better shutdown
+        except KeyboardInterrupt:
+            logger.info("Received interrupt, shutting down inference servers...")
```
SumanthRH, line 83: This is a non issue, but if someone runs this code on it's own outside of the entrypoint task it is helpful.

Also the AI bots don't let you go unless you have exception handling

PR #1801, opened 2026-06-17
[algorithm] make zero-variance filter generic (tolerance + loss-mask aware + metric)

erictang000:
## Summary

Make the trainer's generic `zero_variance_filter` a strict superset of bespoke per-generator zero-variance filters so those can be removed and the logic lives in one place.

- **Loss-mask aware:** within-group variance is now computed over *live* trajectories only (`sum(loss_mask) > 0`). Trajectories masked upstream (e.g. failed / timed-out rollouts) no longer make a genuine zero-variance group look like it has reward spread.
- **Tolerance:** new `algorithm.zero_variance_filter_tol` (default `0.0`, exact / unchanged) so near-equal float rewards from LLM-judge rubrics (e.g. `0.6667`) count as zero-variance. Set `~1e-6` for float rewards.
- **Metric:** emits `reward/num_zero_variance_filtered` via `postprocess_generator_output`, so it flows for both the sync and fully-async trainers.

A group is filtered only when it has more than one live trajectory with reward spread within `tol`; singletons (and groups whose other trajectories were masked upstream) are always kept. `tol=0.0` + no `loss_masks` reproduces the previous `np.std > 0` behavior exactly.

This is phase 1 of generalizing zero-variance filtering; a follow-up adds active drop+refill (`sample_full_batch`) for the fully-async trainer.

## Test plan

- New unit tests in `tests/train/test_trainer_utils.py`: tolerance, loss-mask awareness, single-live-trajectory; existing three still pass (backward compatible).
- `tests/train/test_generator_postprocess.py` passes.
- ruff + black clean.

### DIFF
--- a/skyrl/train/config/config.py
+++ b/skyrl/train/config/config.py
@@ -449,6 +449,9 @@ class AlgorithmConfig(BaseConfig):
     grpo_norm_by_std: bool = True
     zero_variance_filter: bool = False
     """Loss-mask prompts with zero-variance rewards. Only applicable when rewards are response-level."""
+    zero_variance_filter_tol: float = 0.0
+    """Two rewards within this absolute tolerance count as equal when detecting zero-variance groups.
+    Only used when ``zero_variance_filter=True``. 0.0 = exact; set ~1e-6 for float (LLM-judge) rewards."""
     lambd: float = 1.0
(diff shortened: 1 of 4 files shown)


erictang000:
closed in favor of #1802

PR #1802, opened 2026-06-17
[fully_async] add sample_full_batch: drop zero-variance groups and refill the mini-batch

erictang000:
## Summary

Adds `trainer.fully_async.sample_full_batch` (requires `trainer.algorithm.zero_variance_filter=True`) — the async-native equivalent of DAPO `dynamic_sampling.type="filter"`. When enabled, the fully-async consumer **drops** zero-variance groups at drain time and keeps pulling generated groups until the mini-batch is full of non-zero-variance groups, blocking on generation as needed. With `sample_full_batch=False` (default), behavior is unchanged (zero-variance groups are loss-masked, occupying a slot).

## What it does and the invariants it preserves

- **Drain-loop filtering** (`_should_keep_group` reuses the generic `zero_variance_filter`): drops a group with no live-reward variance, refills from the buffer.
- **Staleness capacity (deadlock fix):** `_AsyncStalenessManager.on_rollout_filtered()` reclassifies a dropped group from `accepted` to `filtered`, restoring producer capacity. Without it, dropped groups keep `accepted` climbing against a fixed staleness ceiling and starve the producers. Epoch-end invariant becomes `submitted == accepted + filtered`.
- **Consumed-UID accounting / resume:** `_AsyncDataloader` splits trained vs filtered UIDs. The do-not-redraw set is their union; both are persisted to `fully_async_state.pt`, so dropped (and end-of-epoch discarded) prompts are **skipped — not regenerated — on resume**, while only trained UIDs count toward completed steps.
- **Dynamic epoch length:** dropping consumes extra prompts per trained step, so an epoch can exhaust before `num_steps_per_epoch`. A generators-done event + exhaustion-aware drain (`_drain_next_group`) lets the consumer detect exhaustion instead of blocking forever; the partial mini-batch is discarded (marked consumed) and the epoch ends. Per-epoch step accounting now uses an explicit counter and the epoch is persisted, since `global_step` no longer aligns with epoch boundaries under early-ending epochs.

## Test plan

New CPU unit tests in `tests/train/test_fully_async_trainer.py`:
- staleness manager: `on_rollout_filtered` restores producer capacity (deadlock regression) + epoch-end accounting with filtered groups;
- dataloader: trained-vs-filtered tracking, skipping filtered UIDs, resume restore, backward-compat load;
- `_drain_next_group`: buffered-items-then-exhaustion, drain-before-exhaustion, blocks-until-item;
- `_should_keep_group`: zero-variance drop, variance keep, singleton keep, masked-trajectory handling, tolerance, token-level rejection.

All pass; ruff + black clean. (Docs build not run — `npm` unavailable in the dev box; MDX change is prose-only.) The full async loop is exercised by the existing GPU e2e scripts, which I can't run on CPU.


erictang000:
summary of metrics with `sample_full_batch`. in below figures, red is a fully async baseline, and blue is with `zero_variance_filtering=True` and `sample_full_batch=True`.

Adds async/num_groups_dropped per step
image


## Reward
Reward is computed over both filtered and kept samples to keep things comparable to non `sample_full_batch` runs.

image


## Loss
Since loss is only relevant for kept samples, loss metrics are only computed on kept samples

image

## Async Metrics
staleness is counted only on kept samples, since the staleness of dropped samples is not relevant for training

image

## Generation
generation metrics are computed both for kept and dropped samples to keep things comparable to non `sample_full_batch` runs.

image


## Policy
Only computed on kept samples 

## Trainer
an epoch is a full pass over the dataset regardless of whether samples are trained or dropped - so epochs go up faster now

image

erictang000:
claude validation of checkpointing resumption behavior: https://gist.github.com/erictang000/0c14ab992a25c50e98698b2594872100

skyrl/train/config/config.py
```diff
@@ -449,6 +449,9 @@ class AlgorithmConfig(BaseConfig):
     grpo_norm_by_std: bool = True
     zero_variance_filter: bool = False
     """Loss-mask prompts with zero-variance rewards. Only applicable when rewards are response-level."""
+    zero_variance_filter_tol: float = 0.0
```
CharlieFRuan: I think we could just hard code to `1e-6` instead of introducing a new cnofig; or make the default 1e-6 perhaps. It might be quite easy to miss

CharlieFRuan (approved):
Thank you! We could make it less verbose later.

skyrl/train/fully_async_trainer.py
```diff
@@ -392,33 +471,60 @@ async def train(self):
...
+                            for _ in cur_generation_group_mini_batch:
+                                await self._staleness_manager.on_rollout_filtered()
+                            await self.async_train_dataloader.mark_filtered_uids(
+                                [g.uid for g in cur_generation_group_mini_batch]
+                            )
+                            logger.warning(
+                                f"sample_full_batch: epoch {epoch} exhausted with a partial mini-batch of "
+                                f"{len(cur_generation_group_mini_batch)} group(s); discarding and ending the epoch."
+                            )
+                        break
```
CharlieFRuan, line 510: This will skip checkpoint saving at the end of an epoch, which we might want?

erictang000: good point, fixed!

skyrl/train/config/config.py
erictang000: updated default to 1e-6

PR #1805, opened 2026-06-18
[megatron] Accept dtype-string optimizer_config_kwargs (coerce exp_avg_dtype etc. to torch.dtype)

dyurk-lila:
## Reviewers: Where to Look

The behavioral core of this PR is small and lives in **one new file**:

- **`skyrl/backends/skyrl_train/distributed/megatron/optimizer_dtype.py`** — read this closely. `coerce_optimizer_dtype_kwargs()` does the actual `str -> torch.dtype` coercion, per-field legal-set validation, and `ValueError` raising. Scrutinize the two tables it depends on:
  - `_DTYPE_NAME_TO_TORCH` — the accepted names/aliases and, notably, the non-obvious `fp8 -> torch.uint8` mapping (TE convention, *not* from Megatron's `dtype_map`).
  - `_LEGAL_FIELD_DTYPES` — which fields are legal-set-enforced (`main_params_dtype` = fp32/fp16; `exp_avg_dtype`/`exp_avg_sq_dtype` = +bf16/fp8) and which are deliberately *not* (`params_dtype`, `main_grads_dtype` — coerced but left for Megatron to validate). The correctness argument here rests on which kwargs actually reach TE FusedAdam at the pinned megatron-core rev; that claim is the thing most worth checking.
- **`skyrl/backends/skyrl_train/distributed/megatron/optimizer.py`** (~line 51) — the single wiring change: `init_megatron_optim_config` now routes `optimizer_config_kwargs` through the coercion helper instead of a raw `.update(...)`. Worth confirming this is the sole optimizer-construction choke point (it serves both SFT and the RL policy).

Needs **less scrutiny** (mechanical / non-load-bearing):

- **Docs** (`docs/content/docs/configuration/config.mdx`, `docs/content/docs/examples/megatron.mdx`) — descriptive only; just check the name/alias table matches `_DTYPE_NAME_TO_TORCH`.
- **Tests** (`tests/backends/skyrl_train/distributed/test_optimizer_dtype_coercion.py`) — the CPU class runs torch-only (no megatron-core); the end-to-end class is megatron-gated. Verify coverage rather than line-by-line logic.

## Summary

Megatron's precision-aware `OptimizerConfig` types its `*_dtype` fields (`exp_avg_dtype`, `exp_avg_sq_dtype`, `main_params_dtype`, `params_dtype`) as real `torch.dtype`, but `optimizer_config_kwargs` is forwarded verbatim from YAML/Hydra, which delivers plain strings (e.g. `"bf16"`). Such a string would reach TransformerEngine FusedAdam and crash. This adds a central `str -> torch.dtype` coercion at the single optimizer-construction choke point so low-precision optimizer state can be configured from YAML.

## What changed

- New torch-only module `skyrl/backends/skyrl_train/distributed/megatron/optimizer_dtype.py` holding `coerce_optimizer_dtype_kwargs(dict) -> dict` plus two mapping tables:
  - `_DTYPE_NAME_TO_TORCH`: canonical dtype-name -> `torch.dtype`. The short forms `fp32`/`bf16`/`fp16`/`fp8` follow Megatron-LM's own `dtype_map`; common alias spellings (`bfloat16`, `float16`/`half`, `float32`/`float`, `float8`/`uint8`) are also accepted. `fp8 -> torch.uint8` is a TransformerEngine convention (TE represents FP8 optimizer state as uint8) and is added here, not sourced from Megatron's `dtype_map`.
  - `_LEGAL_FIELD_DTYPES`: per-field legal sets for the fields actually forwarded to TE FusedAdam — `main_params_dtype` (master weights) restricted to `{fp32, fp16}`; `exp_avg_dtype`/`exp_avg_sq_dtype` allow `{fp32, fp16, bf16, fp8}`. Illegal values raise a clear `ValueError` before reaching FusedAdam.
- `optimizer.py`: `init_megatron_optim_config` now calls `coerce_optimizer_dtype_kwargs(optimizer_config_kwargs)` in place of the raw `.update(...)`. The coercion sits at the single shared construction point (sole caller serves SFT and the RL policy).
- Docs: documented the new string-dtype support for `optimizer_config_kwargs` in `docs/content/docs/configuration/config.mdx` (next to the existing `use_precision_aware_optimizer` callout, with the full name/alias table) and in `docs/content/docs/examples/megatron.mdx` (concise note cross-linking to the table) — accepted names/aliases, the per-field legal sets, and the `fp8 -> uint8` convention — noting these short forms differ from the full `bfloat16`/`float16`/`float32` spellings accepted by `str_to_torch_dtype` elsewhere.

The helper is deliberately kept free of any `megatron.core` import so it can be unit-tested on the CPU CI lane (torch only). Coercion lives at the optimizer-construction choke point rather than `MegatronConfig.__post_init__`, which would replace the YAML strings with `torch.dtype` objects in the dataclass and break the serializable config path (`asdict`/`yaml.dump`).

Byte-identical to current behavior unless a `*_dtype` key is explicitly set in `optimizer_config_kwargs`. With no `*_dtype` overrides the coercion is a pure pass-through copy, so `OptimizerConfig` keeps its existing fp32 defaults for `exp_avg_dtype`/`exp_avg_sq_dtype`/`main_params_dtype` and the hardcoded `params_dtype=torch.bfloat16` seed is unchanged. The default optimizer kwargs contain no `*_dtype` keys, so the default path is unchanged. The only intentional behavior change: a `*_dtype` string that previously would have reached FusedAdam and crashed now becomes the correct `torch.dtype` (enabling low-precision optimizer state), and an illegal `main_params_dtype` now fails fast with a clear message instead of a cryptic TE error. Values already `torch.dtype` and non-dtype kwargs pass through untouched; non-string/non-dtype `*_dtype` values (e.g. `None`) pass through so Megatron's own validation surfaces them. `main_grads_dtype` is coerced str->dtype but intentionally has no legal-set row: at the pinned megatron-core rev it is not forwarded to TE FusedAdam, so there is no TE-backed legal set to enforce; it is left for `OptimizerConfig.__post_init__` to validate (mirroring how `params_dtype` is handled).

## Generality & follow-ups

- Covers all Megatron optimizer construction reachable via `init_megatron_optim_config` (SFT and RL policy). The critic worker does not construct an optimizer through this path.
- The FSDP optimizer-state path is intentionally out of scope — it is a separate code path with its own mixed-precision config; no change made there.
- A separate `str -> torch.dtype` helper already exists (`str_to_torch_dtype` / `PrecisionType.to_dtype`), but neither knows the `fp8 -> uint8` mapping nor does per-field legal-set validation, both of which the precision-aware optimizer-state feature requires; consolidating the canonical name table is a possible follow-up.

## Test plan

`tests/backends/skyrl_train/distributed/test_optimizer_dtype_coercion.py`:
- `TestCoerceOptimizerDtypeKwargs` (CPU, no skip-guard — runs on the CPU lane): parametrized name->dtype coercion for all aliases, `fp8->uint8`, case/whitespace insensitivity, `torch.dtype` pass-through, `main_params_dtype` accepts fp32/fp16 and rejects bf16/fp8, `params_dtype` coercion, `main_grads_dtype` coerced-but-not-field-validated, unrecognized-name `ValueError`, unrelated kwargs untouched, `None` pass-through, input not mutated.
- `TestInitMegatronOptimConfigDtypeCoercion` (megatron-gated via `_has_megatron` skip-guard, no GPU): end-to-end that string kwargs reach a real `OptimizerConfig` with coerced dtypes; `params_dtype` string override replaces the seeded default; default (no override) keeps fp32 defaults; and `use_precision_aware_optimizer=False` + non-fp32 state fast-fails with megatron's own `AssertionError`.

Run:

```bash
uv run --isolated --extra megatron --extra dev pytest \
    tests/backends/skyrl_train/distributed/test_optimizer_dtype_coercion.py -v
```

The CPU class runs on the CPU lane (torch only, megatron-core not required); the megatron-gated class runs wherever megatron-core is installed. No GPU is required by either.

### DIFF
--- a/tests/backends/skyrl_train/distributed/test_optimizer_dtype_coercion.py
+++ b/tests/backends/skyrl_train/distributed/test_optimizer_dtype_coercion.py
@@ -1,0 +1,207 @@
+"""CPU tests for central Megatron optimizer-state dtype coercion.
+
+Verifies that ``*_dtype`` string values forwarded through
+``MegatronConfig.optimizer_config_kwargs`` (e.g. "bf16" from YAML/Hydra) are
+coerced to real ``torch.dtype`` before reaching Megatron's precision-aware
+``OptimizerConfig`` / TransformerEngine FusedAdam, using Megatron-LM's own
+canonical dtype mapping; that illegal values for ``main_params_dtype`` (master
+weights, fp32/fp16 only) are rejected; and that unrelated kwargs pass through
+untouched.
+
+``TestCoerceOptimizerDtypeKwargs`` exercises the pure-Python (torch-only)
+coercion helper, which lives in a module that does NOT import megatron-core, so
+it runs unconditionally on the cheap CPU CI lane (which installs torch but not
+megatron-core). ``TestInitMegatronOptimConfigDtypeCoercion`` builds a real
+``OptimizerConfig`` and is therefore skipped when megatron-core is not installed
+(mirroring ``test_megatron_correctness.py``). No CUDA is required by either.
+
+uv run --isolated --extra megatron --extra dev pytest \
+    tests/backends/skyrl_train/distributed/test_optimizer_dtype_coercion.py -v
+"""
+
+import sys
+
+import pytest
+import torch
+
+from skyrl.backends.skyrl_train.distributed.megatron.optimizer_dtype import (
+    coerce_optimizer_dtype_kwargs,
+)
+
+_has_megatron = "megatron" in sys.modules or __import__("importlib").util.find_spec("megatron") is not None
+
+
+class TestCoerceOptimizerDtypeKwargs:
+    """``coerce_optimizer_dtype_kwargs`` maps dtype-name strings to torch.dtype.
+
+    Runs unconditionally: the helper module is torch-only (no megatron-core).
+    """
+
+    def _coerce(self, kwargs: dict) -> dict:
+        return coerce_optimizer_dtype_kwargs(kwargs)
+
+    @pytest.mark.parametrize(
+        "name,expected",
+        [
+            ("bf16", torch.bfloat16),
+            ("bfloat16", torch.bfloat16),
+            ("fp16", torch.float16),
+            ("float16", torch.float16),
+            ("half", torch.float16),
+            ("fp32", torch.float32),
+            ("float32", torch.float32),
+            ("float", torch.float32),
+            ("fp8", torch.uint8),
+            ("float8", torch.uint8),
+            ("uint8", torch.uint8),
+        ],
+    )
+    def test_string_names_coerce_to_torch_dtype(self, name, expected):
+        """Each canonical/alias dtype name maps to the right torch.dtype."""
+        # exp_avg_dtype legally accepts fp32/fp16/bf16/fp8, so it can exercise all names.
+        out = self._coerce({"exp_avg_dtype": name})
+        assert out["exp_avg_dtype"] == expected
+        assert isinstance(out["exp_avg_dtype"], torch.dtype)
+
+    def test_fp8_maps_to_uint8(self):
+        """TE represents fp8 optimizer state as uint8."""
+        out = self._coerce({"exp_avg_sq_dtype": "fp8"})
+        assert out["exp_avg_sq_dtype"] is torch.uint8
+
+    def test_case_and_whitespace_insensitive(self):
+        out = self._coerce({"exp_avg_dtype": "  BF16 "})
+        assert out["exp_avg_dtype"] is torch.bfloat16
+
+    def test_already_torch_dtype_passes_through(self):
+        """A value already a torch.dtype is preserved as-is."""
+        out = self._coerce({"exp_avg_dtype": torch.bfloat16})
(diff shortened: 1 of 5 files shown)


SumanthRH (approved):

tests/backends/skyrl_train/distributed/test_optimizer_dtype_coercion.py
SumanthRH, line 1: This test requires megatron but it is placed in `tests/backends/skyrl_train/distributed` and it will not run on our CPU CI workflows. 

We only run with fsdp backend right now:

https://github.com/NovaSky-AI/SkyRL/blob/af36dc041c8d32a00460d742aad0c1d6bcaacd88/.github/workflows/cpu_skyrl_train.yaml#L72-L75


Could you use mocking to ensure that the megatron library is not required here?

dyurk-lila: Addressed in 6f50dbe0. The test now re-imports the torch-only helper with `megatron` explicitly unavailable in `sys.modules`, so any Megatron dependency fails during fixture setup. The focused non-Megatron CPU test passes: `4 passed`.

PR #1765, opened 2026-06-09
[megatron] Triton backend for fused LM-head log-prob

dyurk-lila:
## Reviewers: Where to Look

> **Rebase note (re-scoped onto #1841):** [#1841](https://github.com/NovaSky-AI/SkyRL/pull/1841) landed the torch fused LM-head log-prob path before this PR. This PR now sits on top of #1841 and only adds the remaining Triton backend pieces. It does **not** implement entropy-with-gradient for fused LM-head; upstream #1841 keeps fused entropy metric-only/no-grad, and this PR preserves that contract.

Please review closely:

1. **Triton backend selector + dispatch** - `skyrl/backends/skyrl_train/distributed/megatron/model_utils.py`, `_fused_lm_head_logprob_apply(...)`. This chooses between upstream `FusedLinearChunkedDistributedLogprob` and the vendored Triton adapter. Both backends now share the same logprob-only apply contract and return TP-combined `[B, S]` log-probs.

2. **Vendored Triton kernel adapter** - `skyrl/backends/skyrl_train/distributed/megatron/fused_linear_logprob_triton.py`, `FusedLinearLogprobTriton`. Review the SkyRL adapter around the vendored verl/NVIDIA kernel: arbitrary shard-layout target re-keying, fully-OOV forward masking/backward pass-through, hidden-to-weight dtype casting, and import-safe CUDA/Triton availability checks.

3. **Backend config wiring** - `skyrl/train/config/config.py` and `skyrl/backends/skyrl_train/workers/megatron/megatron_model_wrapper.py`. This adds `trainer.fused_lm_head_logprob_backend: "torch" | "triton"` as a backend selector behind the existing `trainer.fused_lm_head_logprob` on/off flag.

4. **Triton GPU tests** - `tests/backends/skyrl_train/gpu/gpu_ci/megatron/test_fused_linear_logprob_triton.py`. These drive `FusedLinearLogprobTriton.apply` directly against the stock materialized-logits path for logprob forward, grad-hidden, and grad-weight across TP1/TP2, chunked/single-chunk, fp32/bf16, and OOV targets.

Needs less scrutiny: the vendored Triton kernel internals and the H100 benchmark harness. The top-level `NOTICE` and standalone `triton` extra were removed after review; attribution remains in the vendored file header and the Megatron/FSDP dependency graph already resolves Triton through the pinned torch stack.

## Summary

[...]

PR #1806, opened 2026-06-18
[megatron] Stream ChunkedDistributedLogprob.backward into a preallocated buffer (lower peak memory)

dyurk-lila:
## Reviewers: Where to Look

The entire behavioral change is one method:

- **`skyrl/backends/skyrl_train/distributed/megatron/model_utils.py` — `ChunkedDistributedLogprob.backward` (approx. lines 204-281).** This is the sole production change and the only place needing close review. The scrutiny should focus on the correctness of the streaming rewrite:
  - The preallocated `torch.empty(...)` fp32 buffer (line ~232) and, critically, that the chunk loop tiles `[0, seq_size)` with **no gap or overlap** so every slice of the uninitialized buffer is written exactly once (the per-chunk write `grad_input[:, chunk_start:chunk_end, :] = chunk_grad_input`, line ~279).
  - That the per-chunk math is unchanged and the rename `grad_input` -> `chunk_grad_input` (lines ~271-277) is byte-identical to the old list+`torch.cat` form.

Everything else needs **less** scrutiny:

- The two new test files are the correctness gates, not logic to audit line-by-line: `tests/backends/skyrl_train/distributed/test_chunked_logprob_backward_streaming.py` (CPU `world_size=1` byte-identity via `torch.equal`) and `tests/backends/skyrl_train/gpu/gpu_ci/megatron/test_chunked_logprob_backward_tp.py` (TP>1 GPU parity + no-overlap tiling check).
- `tests/backends/skyrl_train/gpu/gpu_ci/megatron/test_chunked_logprob_backward.py` is a one-line import-path fix (`get_free_port` moved), mechanical.
- `DistributedLogprob`, `forward()`, and the vendored/fused-LCE path are untouched.

---

## Summary

Refactor `ChunkedDistributedLogprob.backward` (the vocab-parallel chunked-logprob autograd function used by the Megatron worker for SFT cross-entropy and RL policy/ref losses) to stream each chunk's gradient into a single preallocated fp32 buffer instead of appending to a Python list and concatenating with `torch.cat` at the end. This lowers peak activation memory on the chunked backward path with **no change to numerics**.

## What changed

`skyrl/backends/skyrl_train/distributed/megatron/model_utils.py`, `ChunkedDistributedLogprob.backward`:

- Removed `all_grad_input = []` and the final `grad_input = torch.cat(all_grad_input, dim=1)`.
- Preallocate once before the loop:
  `grad_input = torch.empty((batch_size, seq_size, partition_vocab_size), dtype=torch.float32, device=vocab_parallel_logits.device)`.
- Renamed the per-chunk grad to `chunk_grad_input` and, after the `scatter_add_`, write it into its sequence slice: `grad_input[:, chunk_start:chunk_end, :] = chunk_grad_input`.

The change is **unconditional** (no flag) because it is numerically byte-identical. It only engages on the chunked dispatch path (i.e. when `chunk_size < seq_len_local`). The non-chunked `DistributedLogprob`, `forward()`, and the vendored Triton fused-LCE path are untouched. `forward()` retains the list+`cat` form intentionally: its accumulator holds per-chunk `[batch_size, chunk_len]` log-prob tensors (tiny), not the `[batch_size, chunk_len, V//TP]` fp32 grads that make the backward `cat` expensive, so streaming it would not meaningfully lower peak.

The old list-then-`cat` form kept every per-chunk `[B, chunk_len, V//TP]` fp32 grad alive **and** allocated the full concatenated output at the cat moment, so peak was ~2x the full `[B, seq, V//TP]` fp32 grad. Streaming drops peak to full-buffer + one live chunk = ~`(1 + 1/num_chunks)`x of the full grad. The win scales with chunk count; it is **not** a flat halving.

Byte-identical by construction:

- The per-chunk math is unchanged: same `_compute_distributed_log_softmax` on the same fp32 slice, same `.exp()`, same `neg_`/`mul_`/`scatter_add_` formulation. `chunk_grad_input` is a fresh fp32 tensor, so the slice write is a same-dtype copy with no cast.
- The chunks tile `[0, seq_size)` exactly and contiguously: chunk `i` covers `[i*chunk_size, min(seq_size, (i+1)*chunk_size))`, consecutive chunks meet with no gap/overlap, and `num_chunks = ceil(seq_size/chunk_size)`. `torch.cat(dim=1)` placed chunk `i`'s columns at those same `[chunk_start:chunk_end]` offsets, so values land at identical positions and each slice is written exactly once.
- The new buffer is contiguous fp32 of shape `[B, seq_size, V//TP]` — identical shape/dtype/contiguity to the previous `torch.cat` output — and is fully overwritten by the full-coverage tiling, so no uninitialized `torch.empty` memory survives. Full-coverage tiling is a load-bearing invariant for the `torch.empty` buffer (a partial write would leak garbage silently); the prime-length coverage test below is the regression guard.

A deliberate choice of a **separate fp32 buffer** (rather than writing in place into the autograd-saved `vocab_parallel_logits`): the separate buffer keeps the gradient in fp32 regardless of the saved logits' dtype, does not mutate an autograd-saved tensor (avoiding version-counter / double-backward hazards), and trades only ~`(1/num_chunks)`x extra memory for that safety.

## Test plan

> Written to SkyRL CI conventions; `python -m py_compile` passes and lint is clean (`ruff`: all checks passed; `black --line-length 120`: unchanged) on all three files.

- `tests/backends/skyrl_train/distributed/test_chunked_logprob_backward_streaming.py` (CPU lane): stubs `megatron.core.parallel_state` into `sys.modules` via a module-scoped autouse save/restore fixture (mirroring `test_preprocess_packed_seqs_cp.py`), uses a gloo `world_size=1` TP group (every `all_reduce` is the identity), and asserts `torch.equal` between the chunked-streamed grad and the single-shot `DistributedLogprob` grad across `chunk_size in {1,3,7,16,32,64}` x with/without OOV targets, plus edge cases (seq_len=1, all-in/all-out mask, a prime/ragged tiny-vocab config). This is the **byte-identity gate for the storage refactor**: the log-softmax + scatter-add backward math is purely per-position (reductions only over the vocab dim), so chunk boundaries cannot change any value, and the world_size=1 `torch.equal` fully validates the device/dtype-agnostic slice-write. The prime-length (`seq_len=17`, `chunk_size=5`) coverage test additionally pins that every sequence slice of the preallocated buffer is written (an unwritten `torch.empty` slice cannot coincidentally match the reference). Collected by the SkyRL-Train-CPU lane (`pytest tests/backends/skyrl_train/ --ignore=.../gpu`).

  Run with:
  ```
  uv run --isolated --extra dev -- pytest -s \
    tests/backends/skyrl_train/distributed/test_chunked_logprob_backward_streaming.py
  ```

- `tests/backends/skyrl_train/gpu/gpu_ci/megatron/test_chunked_logprob_backward_tp.py` (`@pytest.mark.megatron`): spawns TP NCCL ranks (parametrized `tp_size in {2, 4}`, guarded by a `device_count()` skip), runs `mpu.initialize_model_parallel`, shards the vocab across ranks, runs fwd+bwd on each rank's slice, and asserts (a) the rank vocab slices tile `[0, vocab)` exactly once via `torch.equal` on a coverage counter, and (b) each rank's local streamed grad matches the full-vocab single-process autograd reference columns **within fp32 tolerance** via `torch.testing.assert_close(atol=1e-5, rtol=1e-4)` — a tolerance is correct here because the cross-rank all-reduce reorders the fp32 reduction vs. the single-tensor reference (matching the existing GPU test's grad tolerance), while the no-overlap tiling check uses exact equality. Spawned ranks set the conftest-mandated NCCL env (`NCCL_CUMEM_ENABLE=0`, etc.) before `init_process_group`, since `mp.spawn` children do not inherit the runtime env set by the CI conftest.

  Run with (>=2 free GPUs; the `tp_size=4` case is skipped unless 4 are present):
  ```
  uv run --isolated --extra dev --extra megatron -- \
    pytest -s tests/backends/skyrl_train/gpu/gpu_ci/megatron/test_chunked_logprob_backward_tp.py
  ```

## Scope & follow-ups

- **Backends covered:** Megatron only, by nature — `ChunkedDistributedLogprob` is the vocab-parallel (TP/CP) chunked-logprob used on the Megatron worker, reached through the single `forward_backward_mini_batch` train-step chokepoint, so all Megatron-backed training pathways (SFT, sync/async/fully-async RL, full-context) inherit the optimization with no per-pathway edit. FSDP has no vocab-parallel chunked-logprob equivalent, so it is intentionally out of scope.
- **Scope:** Limited to the single `backward` method; `DistributedLogprob`, `forward()`, and the fused-LCE path are unchanged.
- **Deferred:** A direct streaming-vs-old-`cat` comparison at TP>1 was intentionally not added, since it would require shipping the pre-refactor `cat`-based backward as dead reference code. The byte-identity gate for the refactor itself lives on the CPU `world_size=1` lane (`torch.equal`); the GPU lane validates the distributed math against the reference (with tolerance) plus exact no-overlap tiling.

## Related PRs (merge ordering)

- **#1765** (fused LM-head log-prob + entropy) edits the same `ChunkedDistributedLogprob.backward` and **keeps** the `all_grad_input = []` / `torch.cat` form. It refactors the per-chunk chosen-token scatter-add into a shared `_add_chosen_token_grad` helper — touching the exact lines this PR renames to `chunk_grad_input` and streams — so the two **will textually collide on this hunk**. #1765 does **not** subsume this PR's peak-memory win (it preserves the list+`cat`), and #1765 is otherwise complementary in intent. Whichever lands first forces a rebase of the other. **Recommended ordering:** land #1765 first, then rebase this streaming change on top of it (the preallocated-buffer write replaces the `append`+`cat` #1765 keeps); if the two are reviewed together they can be folded into one change.
- **#1543** (WIP) independently removes the same `torch.cat` 2x-peak, but does so **in place** into the autograd-saved logits with no extra buffer. This PR deliberately does **not** take that approach: the in-place variant downcasts the fp32 grad to the saved logits' dtype (breaking byte-identity) and mutates an autograd-saved tensor (the version-counter / double-backward hazard). This PR's separate fp32 buffer preserves numeric fidelity and autograd-safety at the cost of ~`(1/num_chunks)`x extra memory. #1543 is also built on pre-merge code and currently conflicts with main, so it is not a viable supersession.


SumanthRH (approved):
PR looks good to me. Can you resolve merge conflicts?

PR #1809, opened 2026-06-18
[train] Async batch collation (double-buffering) for the SFT trainer

dyurk-lila:
## Summary

Overlap the next SFT `StatefulDataLoader` batch with the current GPU step. A single background thread advances the dataloader and performs collation while the current batch runs forward/backward.

This rebased version integrates with the stateful dataloader and custom sampler support added in #1842:

- `AsyncBatchCollator` provides a one-worker, one-future buffer with strict step matching.
- `SFTTrainer.train()` consumes the buffered batch, handles dataloader exhaustion and epoch transitions, and always joins the worker on exit.
- Checkpoints save the dataloader state after the current batch, not the already-collated next batch, so resume does not skip data.
- `SFTConfig.async_batch_collation` defaults to `True`; set it to `False` for the serial path.

## Validation

- Async and serial batches match across epoch boundaries for default and packed collators, including partial tail batches.
- Checkpoint-resume coverage verifies that a step-N checkpoint resumes at the buffered step-N+1 batch.
- `AsyncBatchCollator` unit tests cover step mismatch, single-slot enforcement, worker errors, draining, and shutdown.
- Focused local suite: `54 passed` across `test_async_batch_collation.py` and `test_sft_dataloader.py`.
- Ruff 0.11.9 and Black 24.10.0 pass on all touched files.


SumanthRH (approved):
LGTM!


Tests look sufficient to capture basic correctness and ckpt-resume behaviour. To be sure, I also did a manual run and tested the PR E2E in the following way:

Script used: `run_sft_fsdp.sh`: https://github.com/NovaSky-AI/SkyRL/blob/fd79ceec9f66b0562fb15b7e70295652026eefd6/examples/train/sft/run_sft_fsdp.sh

1. Serial baseline: Run the above script for 6 steps with `async_batch_collation` `False`
2. Async: Run the script with `async_batch_collation=True`
3. Async, Ckpt + Resume: Run training for 4 steps, ckpt and then resume from step 4 ckpt. 


Run 3 matches Run 1 and Run 2 in overall loss values and the `input_ids` used in step 5 in Run 3 matches step 5 in Run 2.

SumanthRH:
Can you merge from main @dyurk-lila ?

PR #1813, opened 2026-06-19
[feat] Skip policy fwd logprobs for rollout-based losses (rollout_is, dppo) + worker train/rollout logprob diff metric

erictang000:
## What

Skips the policy forward pass that computes the "old" logprobs when the configured loss optimizes against the **rollout** logprobs rather than the old policy logprobs (`policy_loss_type="rollout_is"` or `"dppo"`). In that setting the forward-pass logprobs are not consumed by the training step, so the extra full forward over the batch is wasted work.

When the forward pass is skipped, the trainer's existing `policy/rollout_train_logprobs_abs_diff_*` metric (rollout vs. a single full-batch forward) is unavailable. To keep monitoring logprob drift, the workers emit a new per-mini-batch metric computed from the logprobs the loss **actually optimizes against**.

## Changes

- **Skip the policy forward pass** in `RayPPOTrainer.fwd_logprobs_values_reward` when `_skip_policy_forward(...)` holds (also inherited by the fully-async trainer):
  - `policy_loss_type` is in `LOSSES_WITHOUT_OLD_LOGPROBS` (`rollout_is`, `dppo`),
  - rollout logprobs are present (these losses fall back to the old logprobs without them),
  - the KL reward penalty is disabled (`use_kl_in_reward=False`), and
  - no off-policy correction is active.
  - `action_log_probs` is set to `None` and threaded through (already supported as an optional field); the forward-pass logprob-diff metric is guarded on its presence.

- **New worker-side metric** `policy/minibatch_rollout_logprobs_abs_diff_{mean,std,max,min}`:
  - Absolute diff between the train-step logprobs and the rollout logprobs over response tokens, computed per micro-batch in `compute_minibatch_rollout_logprob_diff_metrics` (in `worker_utils.py`), shared by **both** the FSDP worker and the Megatron model wrapper.
  - Named distinctly from the existing `rollout_train_logprobs_abs_diff`: this one is between the **train-step** logprobs and rollout, and is influenced by mini-batching drift when `train_batch_size > policy_mini_batch_size`.
  - **Aggregation:** `_mean`/`_max`/`_min` reduce correctly (mean/max/min) across micro-batches, DP ranks, and mini-batches via the existing suffix-aware `reduce_metrics`/`all_reduce_metrics`. The std is reconstructed from the first and second moments (`_mean`, `_sq_mean`) — which *do* mean-reduce correctly — in `finalize_minibatch_rollout_logprob_diff_std` (`trainer_utils.py`), since a std cannot itself be mean-reduced across those axes.
  - Padding micro-batches don't pollute the metric: FSDP padding micro-batches carry no rollout logprobs (helper returns `{}`), and the Megatron aggregation loop already skips fully-padding micro-batches.

- **Shared helper** `off_policy_correction_enabled(...)` extracted in `off_policy_correction_utils.py` (and reused by `compute_off_policy_correction`'s early return) so the skip decision and the correction itself agree on exactly which paths read the old logprobs.

- Docs: new "Logprob diff metrics" subsection in `docs/algorithms/off_policy_correction.mdx` describing both metrics and the skip behavior.

## `action_log_probs=None` safety

Audited every consumer of the forward-pass logprobs on the skip path: the only code that dereferences them (`apply_reward_kl_penalty`, off-policy correction) is gated off by the same conditions that trigger the skip, and all batch transforms (slice/chunk/`pad_training_input_batch`/`to_device`/serialize) explicitly handle `None`. The load-bearing invariant — `rollout_is`/`dppo` never dereference the old logprobs when off-policy correction is disabled — is covered by a unit test.

## Tests

`tests/train/algorithms/test_skip_fwd_logprobs.py`:
- `off_policy_correction_enabled` across every correction path.
- `rollout_is`/`dppo` produce a finite loss with `old_log_probs=None` (skip-path invariant).
- The worker metric matches a manual masked computation; absent when rollout logprobs are missing or all tokens are masked.
- End-to-end reduction: the std derived from reduced moments equals the exact pooled population std under equal-size micro-batches/mini-batches.

## E2E validation (Megatron, Qwen3-0.6B, GSM8K, 4×B200)

Ran `examples/train/megatron/run_megatron.sh` with `generator.sampling_params.logprobs=1`:

- **Without `rollout_is`** (forward pass runs, both metrics emitted): new vs. old mean `0.0148 / 0.0150` (step 1), `0.0152 / 0.0155` (step 2); **std identical at `0.0336`**, max/min identical. Near-identical magnitude, as expected with a single optimizer step per batch.
- **With `rollout_is`** (forward pass skipped): new metric present and stable (`0.0149`, `0.0153`), same magnitude; old `rollout_train_logprobs_abs_diff` **fully absent** — confirming the policy forward was skipped while rollout logprobs were still present. Training proceeded with no errors.

## Notes

- The metric means are unweighted per micro-batch (consistent with existing mean-reduced metrics such as `policy_entropy`); with unequal micro-batch token counts this is a close approximation rather than a token-exact mean. Per-mini-batch drift is expected and intended to be visible when `train_batch_size > policy_mini_batch_size`.

### DIFF
--- a/skyrl/backends/skyrl_train/workers/worker_utils.py
+++ b/skyrl/backends/skyrl_train/workers/worker_utils.py
@@ -1,5 +1,5 @@
 import math
-from typing import Dict, Iterator, List
+from typing import Dict, Iterator, List, Optional
 
 import torch
 import torch.distributed as dist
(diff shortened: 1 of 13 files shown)


erictang000:
Verified that when using non rollout logprobs, the rollout/train logprob diff almost exactly matches the rollout/fwd logprob diff, with differences due to different numbers of tokens per DP worker.

image

skyrl/backends/skyrl_train/workers/worker_utils.py
```diff
@@ -1,14 +1,53 @@
...
+    if rollout_logprobs is None:
+        return {}
+    abs_diff = (action_log_probs - rollout_logprobs).abs()
+    masked_abs_diff = abs_diff if loss_mask is None else abs_diff * loss_mask
+    return {
+        MINIBATCH_ROLLOUT_LOGPROB_DIFF_MEAN_KEY: masked_mean(abs_diff, loss_mask).item(),
+        MINIBATCH_ROLLOUT_LOGPROB_DIFF_SQ_MEAN_KEY: masked_mean(abs_diff.square(), loss_mask).item(),
+        MINIBATCH_ROLLOUT_LOGPROB_DIFF_MAX_KEY: masked_abs_diff.max().item(),
+        MINIBATCH_ROLLOUT_LOGPROB_DIFF_MIN_KEY: masked_abs_diff.min().item(),
+    }
```
SumanthRH, line 49: Can you explain why we do a masked mean here directly? 

I would imagine the right thing to do is to treat this as separate sum loss metrics:

(1) the rollout logprobs diff  (and squared diff) across all DP workers 
(2) the loss mask nnz  

and then do a final mean with these after `reduce_metrics` call

There can be cases where some DP workers only get failed trajectories or get heavily imbalanced `loss_mask` `nnz` values. In these cases, the values from some DP workers can contribute more to the final statistics than if you used the sum-based approach.

SumanthRH (changes requested):
Looking good! My only comment is that I don't think we should do the per-worker reduction - do sum based to avoid imbalanced all-reduce across DP workers.

SumanthRH (approved):
Discussed offline. The issue with per-worker reduction is also applicable for other metrics like entropy. 

We will track this issue separately regarding metrics reduction. 


This PR is overall correct and would be great to have as-is as well.

PR #1819, opened 2026-06-20
fix(tracking): finalize MLflow runs on SIGTERM instead of leaving them RUNNING

### LINKED ISSUE #1516: MLflow runs stuck in `RUNNING` after `SIGTERM`
With https://github.com/NovaSky-AI/SkyRL/tree/c8bdc6320da93c7647a2aede61ef95a7b9ae3c42, when a training process is terminated by a signal (e.g. SLURM `scancel` sending `SIGTERM`), the MLflow run is left in `status="RUNNING"` with `end_time=null` on the tracking server.

The run never gets a terminal status because `Tracking.finish` (which calls `mlflow.end_run()`) only fires from `Tracking.__del__`. Python does not guarantee `__del__` runs on signal-driven termination.

You can see many runs in my MLflow server are not considered done, despite being `scancel`'d:

screenshot of MLflow Runs

A possible fix is to:

1. Add a `status` arg to `Tracking.finish`
2. Add a `signal.signal(SIGTERM, …)` handler that calls `self.finish(status="KILLED")`

EazyReal:
## What

When a training process is killed by a signal (e.g. SLURM `scancel` → SIGTERM), its MLflow run is left in `status="RUNNING"` with `end_time=null` forever. `Tracking.finish()` (which calls `mlflow.end_run()`) is only invoked from `Tracking.__del__`, and Python does not guarantee `__del__` runs on signal-driven termination.

## Fix

Implements the approach suggested in the issue:
- `Tracking.finish(status="FINISHED")` and the MLflow adapter forward `status` to `end_run`.
- Register an `atexit` hook and a `SIGTERM` handler that finalize the run (`status="KILLED"` on signal). The handler chains to the previous disposition so normal process teardown is preserved, and is skipped off the main thread (e.g. Ray workers) where `signal.signal` raises.
- `finish()` is idempotent, so the `signal` / `atexit` / `__del__` paths end the run exactly once.

CPU tests cover default/forwarded status, idempotency, and the SIGTERM path.

Fixes #1516

### DIFF
--- a/skyrl/train/utils/tracking.py
+++ b/skyrl/train/utils/tracking.py
@@ -15,9 +15,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import atexit
 import dataclasses
 import pprint
+import signal
 import traceback
+import weakref
 from enum import Enum
 from functools import partial
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


EazyReal:
Good catch on the reference cycle. Fixed: the atexit hook and SIGTERM handler now reference `self` through a `weakref` (so the instance stays collectable and `__del__` keeps working), and `finish()` unregisters the atexit hook and restores the previous SIGTERM handler. Added a test asserting the instance is GC'd after `finish()`.

EazyReal:
@SumanthRH @erictang000 Could I get a quick review on this tracking finalization fix? It closes MLflow runs on SIGTERM so interrupted jobs do not stay RUNNING forever.

PR #1821, opened 2026-06-21
fix(generators): compute batched rollout metrics from truncated responses

EazyReal:
## What

In `SkyRLGymGenerator.generate_batched`, each engine response is truncated to `max_tokens` into `truncated_responses`, and `loss_masks` are built from those truncated lengths. The `GeneratorOutput` returns `truncated_responses` as `response_ids`. But the rollout metrics were computed from the **untruncated** `responses`:

```python
rollout_metrics = get_rollout_metrics(responses, rewards, env_metrics, env_classes, loss_masks)
```

So `generate/{min,max,avg,std}_num_tokens` described sequences longer than the `response_ids` actually returned and trained, and disagreed with the assistant-token metrics derived from the truncated `loss_masks` passed into the same call. `generate/max_num_tokens` could exceed `max_generate_length`, violating the invariant added in #764.

## Fix

Pass `truncated_responses` instead of `responses` — a one-token change that matches the async multi-turn path (`generate`), which feeds `get_rollout_metrics` the same `responses` it returns.

## Test

`tests/train/generators/test_skyrl_gym_generator.py::test_generate_batched_metrics_use_truncated_responses` sets `max_generate_length=2` below the 4-token mock output and asserts `generate/max_num_tokens == 2`, matching the truncated `response_ids` and `loss_masks`. Fails on `main` (`4 == 2`), passes after the fix. All 24 tests in the file pass; black + ruff clean.

### DIFF
--- a/skyrl/train/generators/skyrl_gym_generator.py
+++ b/skyrl/train/generators/skyrl_gym_generator.py
@@ -776,7 +776,7 @@ class SkyRLGymGenerator(GeneratorInterface):
             # Close the environment
             await self._run_in_executor_if_available(env.close)
 
-        rollout_metrics = get_rollout_metrics(responses, rewards, env_metrics, env_classes, loss_masks)
+        rollout_metrics = get_rollout_metrics(truncated_responses, rewards, env_metrics, env_classes, loss_masks)
 
         if self.generator_cfg.apply_overlong_filtering:
             # set loss mask to 0 if the stop reason is not "stop"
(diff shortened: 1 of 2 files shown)


EazyReal:
@SumanthRH @erictang000 Could I get a quick review on this rollout metrics fix? It computes batched generator metrics from the truncated responses that are actually returned/trained.

erictang000 (approved):

PR #1824, opened 2026-06-23
Add `reset_running_requests` to `reset_prefix_cache` call for async RL

SumanthRH:
# What does this PR do?

Adds `reset_running_requests` to `reset_prefix_cache` for async RL.

This unifies the behaviour of `reset_prefix_cache` for sync and async RL

Currently, we call `reset_prefix_cache` without any arguments [here](https://github.com/NovaSky-AI/SkyRL/blob/ffb3f95b7192e2682947b32dcf2efd41fb3f6a73/skyrl/backends/skyrl_train/workers/fsdp/fsdp_worker.py#L284-L285). In vLLM, this is finally passed to the scheduler in  the engine core process of each GPU worker:

https://github.com/vllm-project/vllm/blob/635c38338afe132f9555ebf4a7c8ac7dfb05b2a0/vllm/v1/core/sched/scheduler.py#L2154-L2190

if there are running requests, nothing is done, as can be seen in the KV Cache Block pool implementation [here](
https://github.com/vllm-project/vllm/blob/b529bfd6c51b252cee5741062ef140a34b43aad3/vllm/v1/core/block_pool.py#L470-L493). Thus, even if `clear_kv_cache_on_weight_sync` is set , stale KV cache is not cleared after a weight update with async RL right now, except at epoch boundaries. 

This PR adds `reset_running_requests` to clear kv cache for running requests. 

I did a quick test on the `fully_async_gsm8k.sh` script. 

With clear cache (purple), TTFT shoots up when there are running requests as expected.

image

image

### DIFF
--- a/skyrl/train/config/config.py
+++ b/skyrl/train/config/config.py
@@ -503,10 +503,11 @@ class FullyAsyncConfig(BaseConfig):
     Dropped groups are marked consumed (not regenerated on resume), so the per-epoch step count becomes
     an upper bound: if the epoch's prompts run out mid mini-batch, the partial batch is discarded and
     the epoch ends."""
-    clear_kv_cache_on_weight_sync: bool = True
-    """Whether or not to clear the KV cache on weight sync. Defaults to True, matching synchronous RL.
-    Set to False for fully async training to reuse KV cache from stale policies during generation
-    (avoids recomputation at the cost of using slightly stale KV cache)."""
+    clear_kv_cache_on_weight_sync: bool = False
+    """Whether or not to clear the KV cache on weight sync. Defaults to False.
+    If False, we reuse KV cache from stale policies during generation
+    (avoids recomputation at the cost of using slightly stale KV cache).
+    """
 
     # --- Trainer simulation (no real trainer components) ---
     simulate_training: bool = False
--- a/skyrl/train/config/config.py
+++ b/skyrl/train/config/config.py
@@ -491,6 +491,8 @@ class FullyAsyncConfig(BaseConfig):
     """Knobs for fully async training.
     See https://docs.skyrl.ai/docs/tutorials/fully_async#step-2-config-knobs-to-tune-for-fully-async-training."""
 
+    enabled: bool = False
+    """Indicates whether fully async training is enabled"""
     max_staleness_steps: int = 4
     """Maximum off-policy steps allowed. If a trajectory group is scheduled at step *i* and trained at step *j*,
     then ``j - i <= max_staleness_steps``. Larger values increase throughput but also off-policy-ness."""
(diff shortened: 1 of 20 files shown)


skyrl/train/config/config.py
```diff
@@ -503,10 +503,11 @@ class FullyAsyncConfig(BaseConfig):
...
     the epoch ends."""
-    clear_kv_cache_on_weight_sync: bool = True
-    """Whether or not to clear the KV cache on weight sync. Defaults to True, matching synchronous RL.
-    Set to False for fully async training to reuse KV cache from stale policies during generation
-    (avoids recomputation at the cost of using slightly stale KV cache)."""
+    clear_kv_cache_on_weight_sync: bool = False
+    """Whether or not to clear the KV cache on weight sync. Defaults to False.
+    If False, we reuse KV cache from stale policies during generation
+    (avoids recomputation at the cost of using slightly stale KV cache).
+    """
```
SumanthRH, line 512: `False` by default. We will run more ablations on throughput vs accuracy for `clear_kv_cache_on_weight_sync=True` soon.

skyrl/train/config/config.py
```diff
@@ -503,10 +503,11 @@ class FullyAsyncConfig(BaseConfig):
     Dropped groups are marked consumed (not regenerated on resume), so the per-epoch step count becomes
     an upper bound: if the epoch's prompts run out mid mini-batch, the partial batch is discarded and
     the epoch ends."""
-    clear_kv_cache_on_weight_sync: bool = True
-    """Whether or not to clear the KV cache on weight sync. Defaults to True, matching synchronous RL.
-    Set to False for fully async training to reuse KV cache from stale policies during generation
-    (avoids recomputation at the cost of using slightly stale KV cache)."""
+    clear_kv_cache_on_weight_sync: bool = False
```
erictang000, line 508: ok unfortunately, currently, this default cannot be false, since we currently don't have a conditional in the workers on whether or not we are doing fully async training or just sync

https://github.com/NovaSky-AI/SkyRL/blob/165653d85ce90c0ac36efb944c4d5416a68901e3/skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py#L1281

so changing this default to false would make it so that kv cache isn't cleared for sync training as well. We should improve how this is gated but per charlie's comment on my pr: https://github.com/NovaSky-AI/SkyRL/pull/1798#issuecomment-4724710732 we didn't want to gate on max_staleness > 0 in case the async trainer is used with no staleness for testing.

SumanthRH: Crazy. let me fix

SumanthRH: Fixed now! Thanks !


I am explictly avoiding the alternative of `clear_cache=True` for `pause_generation` because that is a deprecated param in vLLM - we will remove it soon

PR #1836, opened 2026-06-26
[generator] Add per global_step cache salt for each trajectory to invalidate super stale KV

erictang000:
# Overview
Adds `generator.use_cache_salt` config - defaults to `True`.

vLLM hashes prefix-cache blocks by token content alone, so two requests with an identical token prefix share KV blocks even across a weight update -- where the cached KV is now stale. 

When `generator.use_cache_salt=True` (the default), a per-policy-version string (``cache_salt``) is mixed into the block hash so trajectories that begin at the same policy version share cache (throughput) while requests across versions do not. The salt is keyed on the inference engine's weight version (the number of weight
syncs broadcast to the engines), captured at the start of each ``generate`` batch (for a multi-turn trajectory, we still get prefix hits across turns, since the salt stays consistent throughout the lifetime of the trajectory).

This matters for fully-async RL, where the prefix cache is deliberately not fully cleared on every weight sync. It is effectively a no-op for synchronous training (which already resets the prefix cache on each weight sync) and whenever prefix caching is disabled on the inference engine, so it is safe to leave on by default


skyrl/backends/skyrl_train/inference_engines/vllm/vllm_engine.py
SumanthRH, line 1: let's just not add this feature for old inference, gonna be removed soon: #1835

erictang000:
image

logprob diff is reduced with cache salt on

PR #1837, opened 2026-06-26
feat: Arctic RL training backend integration

sfc-gh-kganesan:
## Summary

Adds an `arctic_rl` integration under `integrations/arctic_rl/` that lets SkyRL recipes train against Snowflake's Arctic RL stack ([Arctic-Platform](https://github.com/snowflakedb/Arctic-Platform) + [Arctic-Inference](https://github.com/snowflakedb/ArcticInference)). It mirrors the existing `harbor` / `dapo` integration shape — each integration ships its own entrypoint, config, generator, trainer, BIRD-SQL environment, and example launchers.

**Core SkyRL changes are intentionally minimal** (1 file, +15 / -1 line):

- `skyrl/train/entrypoints/main_base.py`: one ~15-line dispatch hook that inspects `trainer.override_entrypoint=<module>` on argv BEFORE strict config parsing and forwards to the integration's `main()`. This lets integrations add config fields without subclassing `SkyRLTrainConfig` and without forking the recipe scripts. UX is just:

  ```
  uv run -m skyrl.train.entrypoints.main_base \
      trainer.override_entrypoint=integrations.arctic_rl.entrypoint \
      <... any existing SkyRL recipe flags ...>
  ```

- `pyproject.toml`: adds an additive `[arctic-rl]` extra (`skyrl[arctic-rl]`) pulling in `arctic-platform`, `arctic-inference[vllm]`, and `liger-kernel`. **No existing pins touched.**

## Integration contents (`integrations/arctic_rl/`)

- `entrypoint.py` — boots an `ArcticRLClient` against arctic-platform and forwards weight-sync, rollouts, and metrics
- `config.py` — `ArcticRLTrainerConfig` exposing the colocate / ZoRRo / FCA / spec-dec / CUDA-IPC-weight-sync knobs
- `trainer.py`, `generator.py` — thin adapters from SkyRL's PPO loop to Arctic-Platform's RL client
- `envs/bird*.py` — vendored BIRD-SQL environment so the recipe is self-contained (no private branch dep)
- `examples/` — smoke / 8B / 32B BIRD GRPO recipes + a GSM8K sanity check. Env vars in launchers are trimmed to only what's recipe-required.

## Diff scope

- **18 files / +3871 / -1**
- 16 net-new files all under `integrations/arctic_rl/`
- 2 surgical edits to core SkyRL (`main_base.py`, `pyproject.toml`)

## Test plan

- [x] Smoke GSM8K + BIRD recipes converge on single-node 4×H200
- [x] 4-node 32×H200 BIRD GRPO with Qwen3-8B / Qwen3-32B (Arctic-Inference: FCA + speculative-decoding + CUDA-IPC weight sync)
- [x] `uv run --extra arctic-rl -m skyrl.train.entrypoints.main_base trainer.override_entrypoint=...` boots without touching any of SkyRL's typed config
- [ ] Final perf table (2× rollout speedup vs SkyRL FSDP-native baseline) — last-mile timing capture in progress

Happy to iterate on review feedback.

### DIFF
--- a/integrations/arctic_rl/README.md
+++ b/integrations/arctic_rl/README.md
@@ -116,19 +116,6 @@ uv pip install flash-attn-3 --index-url https://download.pytorch.org/whl/cu128
 
 Skip and set `trainer.arctic_rl.attn_implementation=flash_attention_2` if you don't want FA3.
 
-## What you get vs. SkyRL FSDP2
-
-For the deep dive on where the numbers come from — split attention, Forest Cascade, Arctic Speculator — read the [ZoRRo blog post](https://www.snowflake.com/en/blog/engineering/zorro-enterprise-rl-training/). The headline:
-
-| Optimization | Speedup on Arctic-Text2SQL-R2 (Qwen3-32B, 32 × H200) |
-| --- | --- |
-| ZoRRo training (split attention) | 6× actor update, 5× log-prob |
-| Forest Cascade + Arctic spec-dec | 1.7× rollout generation |
-| End-to-end RL iteration | **3.5×** (5 days → 1.5 days) |
-| Max context window | **3.2×** (20 K → 64 K) |
-
-This integration exposes the same stack to SkyRL users.
-
 ## File layout
 
 ```
--- a/integrations/arctic_rl/README.md
+++ b/integrations/arctic_rl/README.md
@@ -36,9 +36,12 @@ uv run -m skyrl.train.entrypoints.main_base \
     trainer.override_entrypoint=integrations.arctic_rl.entrypoint \
     trainer.arctic_rl.use_zorro=true \
     trainer.arctic_rl.use_arctic_inference=true \
+    trainer.arctic_rl.speculative_model=<hf-id-or-path> \
     trainer.arctic_rl.zero_stage=2 \
     <... your existing recipe overrides ...>
 ```
+
+The integration translates each typed `arctic_rl.*` knob into the matching arctic-platform / vLLM payload — no raw `vllm_config` blocks required for FCA, speculative decoding, or the multi-replica FlashInfer workaround.
 
 `trainer.override_entrypoint` tells `main_base` to dispatch into the Arctic entrypoint after Hydra parsing. No `PYTHONPATH` setup, no edits to `main_base.py`.
 
@@ -47,7 +50,9 @@ uv run -m skyrl.train.entrypoints.main_base \
 | Knob | Default | Purpose |
 | --- | --- | --- |
 | `use_zorro` | `false` | Enable ZoRRo split-attention + prompt dedup in the trainer. |
-| `use_arctic_inference` | `false` | Enable Forest Cascade Attention + Arctic spec-dec in the rollout. |
+| `use_arctic_inference` | `false` | Enable Forest Cascade Attention in the rollout (auto-injects the multi-replica `fuse_allreduce_rms` workaround when needed). |
+| `speculative_model` | `None` | HF id / local path of an Arctic draft head. Set this to turn on Arctic speculative decoding — no raw `vllm_config` needed. |
+| `num_speculative_tokens` | `3` | Draft tokens per target-model step (only when `speculative_model` is set). |
 | `use_liger` | `false` | Liger fused kernels in the policy fwd/bwd. |
 | `zero_stage` | `0` | DeepSpeed ZeRO (use `2` for ~1B, `3` for ≥7B; required for bf16 + fp32 grad). |
 | `offload_optimizer` | `false` | CPU offload optimizer state (`zero_stage≥2`). |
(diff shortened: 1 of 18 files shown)


SumanthRH (changes requested):
Can you look into the uv resolution failures? 

CI workflows are failing because of the pyproject.toml changes in your PR: https://github.com/NovaSky-AI/SkyRL/actions/runs/28268995579/job/83762333405?pr=1837


It looks like arctic-inference[vllm] only supports vllm <=0.18, but we use vllm 0.23.0 in SkyRL. Can you look into that? Is that a hard conflict?

 I would recommend using dependency overrides as a fallback: You can remove `arctic-rl` as a native extra in the `pyproject.toml` and then instruct users to install the dependencies `--with` overrides:

```bash
uv run --with arctic-platform --with arctic-inference[vllm] --with liger-kernel ....
```

this will of course require more changes from the user, but can be a workaround while we address the conflicts in a follow-up PR

SumanthRH (commented):
Did you confirm that the Arctic RL script runs with the new dependency changes? 

You're syncing some custom dependencies with `uv pip install` into the current venv.

However, the uv + ray integration will create a new venv for the workers based on the `pyproject.toml` and the `uv run` command

"Under the hood, this integration uses Ray’s [runtime environment](https://docs.ray.io/en/latest/ray-core/handling-dependencies.html) feature, which allows the environment to be specified in code. We define a new low level runtime environment plugin called [py_executable](https://docs.ray.io/en/latest/ray-core/handling-dependencies.html#using-uv-for-package-management). It allows you to specify the Python executable (including arguments) that Ray workers will be started in. In the case of uv, the py_executable is set to uv run with the same parameters that were used to run the driver. Also, the working_dir runtime environment is used to propagate the working directory of the driver (including the pyproject.toml) to the workers. This allows uv to set up the right dependencies and environment for the workers to run in."

https://www.anyscale.com/blog/uv-ray-pain-free-python-dependencies-in-clusters

So custom dependencies will not be synced AFAIK, unless you use `uv run --with ...`

sfc-gh-kganesan:
Done — exactly what you suggested. Latest commit (b6e010e) drops the `_setup_env.sh` workaround and switches every launcher in `integrations/arctic_rl/examples/` to the same `uv run --isolated --extra ... --with ...` shape that `examples/train/flash_rl`, `examples/train_integrations/harbor`, and `examples/train/mini_swe_agent` already use:

```bash
uv run --isolated --extra skyrl-train \
    --with arctic-platform \
    --with 'arctic-inference[vllm]' \
    --with liger-kernel \
    --with 'transformers==4.57.6' \
    --with "flash-attn@<torch2.10 cu128 wheel>" \
    -- python -m skyrl.train.entrypoints.main_base ...
```

A few notes on what changed and why:

- `pyproject.toml` is now untouched by this PR (no `arctic-rl` extra, no source overrides, no conflict-table edits) — the launchers are the only diff.
- `--isolated` is the bit that lets vLLM 0.18 coexist with the fsdp/megatron backends' vLLM 0.23 + transformers 5.x pins. The project lock isn't applied for the launcher's env, and `--with` overrides win over `tool.uv.sources`.
- The `--with` chain matters: `arctic-platform` + `arctic-inference[vllm]` pulls vLLM 0.18 → transformers <5 and torch 2.10. So I also pin `transformers==4.57.6` (exact pin, not `<5`, because Ray's worker-spawn shell parses `<5` as redirect-from-fd-5 — same reason flash_rl uses `transformers==4.53.3`) and swap the torch-2.11 flash-attn wheel in `tool.uv.sources` for the torch-2.10 ABI wheel.
- Ray workers automatically get the same env via the uv+Ray `py_executable` plumbing you linked — visible in the launch log as `(raylet) Installed 139 packages in 376ms` lines as workers come up.

Verified E2E on GSM8K (Qwen3-0.6B, 4 GPUs) with this branch on a clean machine:

```
ArcticRLClient ready — training_job=1, sample_job=2
Arctic RL ZoRRO path enabled
ArcticPPOTrainer: build_models → training routed to Arctic RL server
arctic_inference.server ReplicaPool reset prefix cache (1 workers)
Weight sync complete in 6.211s          (first sync)
Weight sync complete in 0.016s          (steady state)
Step 2: loss=0.0423  grad_norm=0.231  pass@4=0.5195  tokens/s/gpu=12451
```

The user-facing UX is now just `bash integrations/arctic_rl/examples/run_gsm8k_grpo_4gpu.sh` — identical to any other SkyRL recipe, no manual install steps.

I also looked at the Gemini comments (the `try/finally` ones in `bird_reward.py`/`preprocess_bird.py`, the `arctic_inference_config` translation, and the logprob list/numpy/tensor check in `trainer.py`). They're all reasonable; happy to push them as a follow-up if you'd prefer to keep this PR scoped to the dependency-resolution fix.

SumanthRH (approved):

PR #1838, opened 2026-06-26
[megatron] add opt-in async distributed checkpoint save

dinhxuanvu:
## Summary

The Megatron `torch_dist` checkpoint save is fully parallel across ranks but **synchronous** — training blocks for the entire shard-write duration on every `ckpt_interval`. For large models checkpointing frequently, that stall is significant.

This adds an opt-in `MegatronConfig.async_dist_ckpt_save` (default `False`, so existing behavior is unchanged). When enabled, `dist_checkpointing.save` stages each rank's shards to host memory and writes them to disk in a background process, letting training resume immediately. This replaces the long-standing `TODO(tgriggs): Support configurable async saves` in `megatron_strategy.py`.

## Changes

- `MegatronConfig.async_dist_ckpt_save` knob (off by default).
- `MegatronStrategy.save_checkpoint`: passes `async_sharded_save` through and schedules the returned request on the existing persistent `AsyncCallsQueue` instead of asserting it is `None`. Blocks on the previous async save before issuing a new one.
- `MegatronStrategy.finalize_pending_saves()`: drains in-flight writes; called at the start of `load_checkpoint` and exposed up through `Worker` → `WorkerDispatch` → end-of-training in the sync, fully-async, and SFT trainers.

## Correctness

- On-disk format is identical to the synchronous save.
- The pending write is finalized before the next save, before any reload, and at end of training.
- Async falls back to synchronous for cloud destinations, where `local_work_dir` uploads on context exit and would otherwise race a partial checkpoint. Only active for local/shared filesystems.

## Test plan

- [ ] GPU run with `async_dist_ckpt_save=true`: verify checkpoints are byte-identical to sync saves and reload correctly.
- [ ] Confirm training resumes before the disk write completes (timing/save_checkpoint drops).
- [ ] Verify final checkpoint at end of training is fully written before exit.

### DIFF
--- a/skyrl/backends/skyrl_train/distributed/megatron/megatron_strategy.py
+++ b/skyrl/backends/skyrl_train/distributed/megatron/megatron_strategy.py
@@ -316,6 +316,7 @@ class MegatronStrategy(DistributedStrategy):
                 checkpoint_dir=work_dir,
                 sharded_strategy=save_strategy,
                 async_sharded_save=async_save,
+                async_strategy=self.megatron_config.async_dist_ckpt_strategy,
                 validate_access_integrity=True,
             )
             if async_save:
--- a/skyrl/backends/skyrl_train/distributed/megatron/megatron_strategy.py
+++ b/skyrl/backends/skyrl_train/distributed/megatron/megatron_strategy.py
@@ -299,16 +299,30 @@ class MegatronStrategy(DistributedStrategy):
             save_strategy, mpu.get_data_parallel_group(with_context_parallel=True)
         )
 
+        async_save = self.megatron_config.async_dist_ckpt_save
+        if async_save and io.is_cloud_path(ckpt_dir):
+            # local_work_dir uploads on context exit, which would race the still-running
+            # background write. Async is only safe writing in-place to a local/shared FS.
+            self.print(f"async_dist_ckpt_save unsupported for cloud path {ckpt_dir}; saving synchronously")
+            async_save = False
+        if async_save:
+            # The queue holds one in-flight write; the prior checkpoint dir must be
+            # fully written before this save reuses it.
+            ckpt_base.async_calls.maybe_finalize_async_calls(blocking=True)
+
         with io.local_work_dir(ckpt_dir) as work_dir:
-            # TODO(tgriggs): Support configurable async saves.
             async_save_request = dist_checkpointing.save(
                 sharded_state_dict=sharded_state_dict,
                 checkpoint_dir=work_dir,
                 sharded_strategy=save_strategy,
-                async_sharded_save=False,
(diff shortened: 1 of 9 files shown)


skyrl/backends/skyrl_train/distributed/megatron/megatron_strategy.py
SumanthRH, line 1: How does this interact with the Tinker codepath? 

With Tinker, we call `save_checkpoint` and then immediately archive the saved checkpoint directory:

https://github.com/NovaSky-AI/SkyRL/blob/a07f44ed008db65d56fe952171902404df99b8ed/skyrl/backends/skyrl_train_backend.py#L1200-L1204

With async save, the ckpt would be incomplete. 

We should call `finalize_pending_saves` here before archiving

skyrl/backends/skyrl_train/distributed/megatron/megatron_strategy.py
```diff
@@ -299,16 +299,30 @@ def save_checkpoint(
...
+            self.print(f"async_dist_ckpt_save unsupported for cloud path {ckpt_dir}; saving synchronously")
+            async_save = False
+        if async_save:
+            # The queue holds one in-flight write; the prior checkpoint dir must be
+            # fully written before this save reuses it.
+            ckpt_base.async_calls.maybe_finalize_async_calls(blocking=True)
+
         with io.local_work_dir(ckpt_dir) as work_dir:
-            # TODO(tgriggs): Support configurable async saves.
             async_save_request = dist_checkpointing.save(
```
SumanthRH, line 314: If I run with the default `async_strategy` argument for `.save`, I get the error:

```bash
ModuleNotFoundError: A compatible nvidia-resiliency-ext installation is required for
async_strategy="nvrx". Please install it or set async_strategy to "mcore".
```

https://github.com/NVIDIA/Megatron-LM/blob/f258d4fa82de72705f846fb713bbe3b4180f38e8/megatron/core/dist_checkpointing/serialization.py#L342

Perhaps we can add a configuration for `async_strategy` in the config file? It can default to `mcore` maybe? I haven't looked into this too closely, but it looks like `mcore`  strategy should be fine and we don't need the `nvidia-resilient-ext` package

SumanthRH (commented):
Thanks for the PR! Left a few comments.

Can we add a test to `tests/backends/skyrl_train/gpu/gpu_ci/test_save_load_checkpoint.py` to test the `async_dist_ckpt_save` codepath ? Let's test with the `mcore` `async_strategy`

dinhxuanvu:
@SumanthRH Thanks for the feedback and suggestion. Addressed. PTAL. Thanks.

SumanthRH (approved):

Issue #1839, opened 2026-06-26
[RFC] Stateful dataloader and custom samplers for `SFTTrainer`


SumanthRH:
# Summary

The current `SFTTrainer` in SkyRL is pretty barebones: We load the dataset and store tokenized sequences as a list in memory, and sample directly from this list. 

This RFC proposes a simple refactor for the SFT trainer to 
1. Switch to  `torchdata.stateful_dataloader.StatefulDataLoader` for train and eval data loading, matching the RL trainer's dataloader and checkpointing pattern.
2. Support customization for dataloading through stateful samplers that implement `__iter__`. In addition, `state_dict`, and `load_state_dict` can be implemented for easier resume.

## Goals

- Replace manual list shuffling and slicing in `SFTTrainer.train()` with
  `StatefulDataLoader`, with ckpt and resume support
- Keep existing SFT tokenization and collation behavior unchanged.
- Support stateful custom samplers for ordered data, data mixing, and
  curriculum learning.

## Non-Goals

- Changing the tokenization or data collation implementation

## Proposed Trainer modifications

Add these fields to `SFTTrainer`:

```python
self.train_dataloader = None
self.eval_dataloader = None
```

Add builder methods:

```python
def build_train_dataloader(self, tokenized: list[dict]) -> StatefulDataLoader:
    ...

def build_eval_dataloader(self, eval_tokenized: list[dict]) -> StatefulDataLoader:
    ...

def build_train_sampler(self, tokenized: list[dict]) -> torch.utils.data.Sampler[int]:
    ...
```

The default train dataloader should be:

```python
def collate_sft_examples(examples, collator, batch_size):
    return collator(examples, batch_size=batch_size)


collate_fn = functools.partial(
    collate_sft_examples,
    collator=self.collator,
    batch_size=self.sft_cfg.batch_size,
)

seeded_generator = torch.Generator()
seeded_generator.manual_seed(self.sft_cfg.seed)

self.train_dataloader = StatefulDataLoader(
    tokenized,
    batch_size=self.sft_cfg.batch_size,
    sampler=self.build_train_sampler(tokenized),
    collate_fn=collate_fn,
    drop_last=True,
    generator=seeded_generator,
    num_workers=self.sft_cfg.dataloader_num_workers,
    persistent_workers=self.sft_cfg.dataloader_persistent_workers,
    multiprocessing_context=(
        "spawn" if self.sft_cfg.dataloader_num_workers > 0 else None
    ),
)
```

[...]

PR #1842, opened 2026-06-26
[SFT] Stateful dataloader and custom samplers for SFTTrainer

avigyabb:
# Stateful dataloader and custom samplers for `SFTTrainer`

Addresses #1839.

## Summary

`SFTTrainer` previously stored tokenized sequences as an in-memory list and sampled by manually shuffling and slicing that list inside `train()`, with epoch shuffles replayed by hand on resume. This PR replaces that machinery with a `torchdata.stateful_dataloader.StatefulDataLoader` for both train and eval, matching the RL trainer's dataloader and checkpointing pattern, and adds a pluggable sampler interface so dataloading can be customized for ordered data, data mixing, and curriculum learning.

Tokenization and collation are unchanged (per the RFC non-goals); only how examples are ordered, batched, and checkpointed changes.

## What changed

**Config (`skyrl/train/config/sft_config.py`)** — new dataloader/sampler fields plus validation:

- `dataloader_num_workers: int = 0`
- `dataloader_persistent_workers: bool = False`
- `sampler: str = "random"` (`"random"`, `"sequential"`, or `"custom"`)
- `sampler_class_path: Optional[str] = None` (required when `sampler="custom"`)
- `sampler_kwargs: dict = {}`

`validate_sft_cfg` now rejects an unknown `sampler`, a `custom` sampler with no `sampler_class_path`, and a negative `dataloader_num_workers`. The existing `num_workers` (tokenization) field is left untouched.

**Core samplers module (`skyrl/train/dataset/samplers.py`, new)**:

- `StatefulSequentialSampler` — yields `0..len-1` in order, tracks an internal cursor, and exposes `state_dict`/`load_state_dict` so a `StatefulDataLoader` can resume mid-epoch from the exact next index. Backs `sampler="sequential"`.
- `import_sampler_class(class_path)` — resolves a `module.path.ClassName` string for the `sampler="custom"` path, with a clear error on a bare (non-dotted) name.

**Trainer (`skyrl/train/sft_trainer.py`)**:

- `build_train_sampler`, `build_train_dataloader`, and `build_eval_dataloader` replace the inline list shuffling/slicing. `train_dataloader`/`eval_dataloader` are built in `train()` once data is tokenized.
- `sampler="random"` returns `None` and uses the dataloader's built-in `shuffle=True` (statefully checkpointed), since an explicit sampler and `shuffle` are mutually exclusive.
- A module-level `collate_sft_examples` (not a closure/lambda) keeps the collate function picklable under the `spawn` start method when `dataloader_num_workers > 0`. It takes a `pad_to_batch_size` flag that, on the train path, pads a final short batch up to `batch_size` rows (see the no-dropped-examples point below).
- The train loader uses `drop_last=False`: rather than dropping the trailing partial batch, the final short batch is padded up to `batch_size` in `collate_sft_examples` (padded rows are masked out of the loss via `pad_training_input_batch`), so every example in an epoch is trained on while each dispatch still sees a full, evenly shardable batch. Packed batches (`use_sequence_packing=true`) are never row-padded — their rows are FFD bins already rounded to a multiple of `dp_size` — so padding is applied only on the un-packed path (`pad_to_batch_size = not use_sequence_packing`).
- Throughput accounting excludes the padded tail rows: the token count uses `batch.metadata["pad_size"]` to drop the masked copies, so reported tokens/sec reflect only real tokens.
- Checkpointing: `save_checkpoint` writes the dataloader `state_dict` to `data.pt`; `load_checkpoint` restores it (mirroring the RL trainer's `data.pt`). A missing `data.pt` (older checkpoints) logs a warning and starts from the beginning of the sampling order rather than failing.
- Training loop: a single iterator is advanced on `global_step` instead of looping epoch-by-epoch. Epoch boundaries are detected lazily via iterator exhaustion, and the iterator is only re-created at those boundaries. This preserves state for custom samplers that span the whole run in one pass (per the RFC "Other considerations").
- `steps_per_epoch` is derived from `len(train_dataloader)`. With `drop_last=False` this is `ceil(len(sampler) / batch_size)` — the padded trailing batch counts as a step.
- An empty-dataloader guard raises an explicit error only when the sampler yields no indices at all (a truly empty dataset, or a custom sampler with `num_samples=0`). With `drop_last=False` a sub-`batch_size` dataset now produces a single padded batch instead of zero, so the guard no longer fires on small-but-nonempty data.
- `run_eval()` no longer takes an `eval_tokenized` argument; it iterates `eval_dataloader` (sequential, `drop_last=False`) and pads the final short chunk as before.

**Examples (`examples/train/sft/`)** — reference custom samplers, intentionally not part of core:

- `curriculum_sampler.py` (`CurriculumLearningSampler`) — walks difficulty-ordered subsets, progressively unlocking harder data.
- `data_mixing_sampler.py` (`DataMixingSampler`) — mixes concatenated sources by per-source weights.

**Docs (`examples/train/sft/README.md`)** — new config-table rows and a "Samplers and the stateful dataloader" section covering the three strategies, resume semantics, and the curriculum example.

## Resume semantics

- The in-progress epoch resumes bit-exactly (the common case) for all samplers.
- For the built-in `random` sampler, epochs *after* the resumed one are reshuffled into a valid but not byte-identical order, because the generator advances differently after a partially replayed epoch. This matches the RL trainer's behavior.
- Custom samplers that cover the whole run in a single pass (e.g. `CurriculumLearningSampler` with `num_samples = num_steps * batch_size`) resume bit-exactly across the entire schedule, since the iterator is never re-created.

## Behavior changes and compatibility

- Default behavior is unchanged in spirit: `sampler="random"` still reshuffles each epoch using `seed`.
- No examples are dropped within an epoch. The train loader uses `drop_last=False` and pads the trailing partial batch (padded rows masked out of the loss) instead of either dropping it or wrapping across the epoch boundary mid-batch as the old slicing did. `steps_per_epoch` is therefore `ceil(len(sampler) / batch_size)`.
- `run_eval()` is now argument-free; any direct caller must read from `eval_dataloader` instead of passing `eval_tokenized`.

## Test plan

New CPU tests in `tests/train/test_sft_dataloader.py` cover the RFC test plan and the new code paths:

- `StatefulSequentialSampler`: in-order yield, length, reset after exhaustion, and `state_dict`/`load_state_dict` resuming at the next sample.
- `import_sampler_class`: resolves a dotted path, resolves a test-local class, and rejects a bare name.
- `build_train_sampler`: `random` returns `None`, `sequential` and `custom` return the expected sampler, `custom` requires a class path, and an unknown sampler raises.
- Random dataloader: deterministic for the same seed, differs for a different seed, keeps the padded partial tail batch (41 items at `batch_size=4` gives 11 batches, not 10), mid-epoch resume, and that `random` uses `shuffle` rather than an explicit sampler.
- Custom-sampler checkpointing: sampler state is included in the dataloader checkpoint and resumes mid-epoch.
- Empty-dataloader guard: a sub-`batch_size` dataset and a custom sampler with `num_samples < batch_size` each yield a single padded batch, while only a truly empty dataset yields zero batches (the guarded condition).
- Tail-batch padding: a final short batch is padded up to `batch_size` with the padded rows masked out of the loss (real rows keep a nonzero `loss_mask`), and full batches are left unaffected.
- Example samplers: curriculum (length, lengths-sum validation, progressive unlocking, determinism, and `state_dict` resume) and data mixing (length, weight/length alignment, determinism, and that weighting biases the mix).

GPU/e2e tests are unchanged.

## Usage

Sequential, in-order training:

```bash
python -m skyrl.train.main_sft sampler=sequential
```

Custom curriculum sampler (directory on `PYTHONPATH` so the dotted path resolves):

```bash
PYTHONPATH=examples/train/sft python -m skyrl.train.main_sft \
    sampler=custom \
    sampler_class_path=curriculum_sampler.CurriculumLearningSampler \
    'sampler_kwargs={lengths: [2000, 2000, 2000], num_samples: 800, seed: 42}'
```

Weighted data mixing:

```bash
PYTHONPATH=examples/train/sft python -m skyrl.train.main_sft \
    sampler=custom \
    sampler_class_path=data_mixing_sampler.DataMixingSampler \
    'sampler_kwargs={lengths: [8000, 2000], weights: [0.5, 0.5], num_samples: 800, seed: 42}'
```


examples/train/sft/curriculum_sampler.py
```diff
@@ -0,0 +1,120 @@
...
+
+This is an *example* of a user-supplied stateful sampler plugged into
+``SFTTrainer`` via the ``sampler=custom`` config path -- it is intentionally
+NOT part of the core library. Point ``sampler_class_path`` at this class (with
+the directory on ``PYTHONPATH``) to use it::
+
+    PYTHONPATH=examples/train/sft python -m skyrl.train.main_sft \
+        sampler=custom \
+        sampler_class_path=curriculum_sampler.CurriculumLearningSampler \
+        'sampler_kwargs={lengths: [2000, 2000, 2000], num_samples: 800, seed: 42}'
```
SumanthRH: Have you actually run this E2E? 


I don't think this works because the `sampler_class_path` import is run inside `SFTTrainer`:

https://github.com/NovaSky-AI/SkyRL/pull/1842/changes#diff-93c21553d2f8ddd1b7a0f6d2f9c1bb941ec97ae858ad3a5d46c1de1fea15f767R1161

which calls 
https://github.com/NovaSky-AI/SkyRL/pull/1842/changes#diff-
6b1f5f06404d568195eaf1a15901f768eb456a82965a382080c41cd4e86cc2ebR29


and `SFTTrainer` is run inside a Ray task: 

https://github.com/NovaSky-AI/SkyRL/blob/7473a5a5a8455eeacdca3ca2ccd2ead9e71ab921/skyrl/train/main_sft.py#L27-L41

The `PYTHONPATH` env var is set only on the driver so the ray task will not inherit this

can you actually just use the full path ? AFAIK even without `__init__.py` the import should work 

```bash
➜  SkyRL git:(sumanthrh/fix-readme) ✗ mkdir -p test/test2
➜  SkyRL git:(sumanthrh/fix-readme) ✗ touch test/test2/module.py
➜  SkyRL git:(sumanthrh/fix-readme) ✗ uv run python -c "import test.test2.module; print('Success')"
warning: The `extra-build-dependencies` option is experimental and may change without warning. Pass `--preview-features extra-build-dependencies` to disable this warning.
Success
```

examples/train/sft/curriculum_sampler.py
```diff
@@ -0,0 +1,120 @@
...
+        self.position = 0
+
+    def __len__(self) -> int:
+        return len(self._plan)
+
+    def state_dict(self) -> dict:
+        return {"position": self.position}
+
+    def load_state_dict(self, state: dict) -> None:
+        self.position = state["position"]
```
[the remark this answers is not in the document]
SumanthRH: @avigyabb

skyrl/train/sft_trainer.py
```diff
@@ -1222,50 +1379,47 @@ def run_eval(self, eval_tokenized: list) -> tuple[dict, int]:
...
                 "by setting eval_dataset_name=None."
             )
 
-        # One micro-batch per DP rank per dispatch call — keeps memory usage bounded
-        # and removes the need for a separate `eval_batch_size` knob.
-        dp_size = self.dispatch.dp_size("policy")
-        eval_chunk_size = self.sft_cfg.micro_train_batch_size_per_gpu * dp_size
+        # The dataloader yields one chunk per DP rank's micro-batch; the final
+        # (possibly short) chunk is padded below up to the full chunk size.
+        eval_chunk_size = self.eval_dataloader.batch_size
```
SumanthRH, line 1519: Why was this change made? 

this is a subtle but important change - the `eval_chunk_size` dictates the evaluation batch size. We will send the batch over to the workers in a `.forward` call. 

Previously this was set to `micro_train_batch_size_per_gpu * dp_size` - > this is a size known to fit for a forward pass on the workers. 

Swiching to `batch_size` can lead to OOMs during the evaluation forward pass 

Can you revert this change? You can still iterate over the dataloader in batches of `batch_size`, but each `.forward` call can only use a chunk of size `micro_train_batch_size_per_gpu * dp_size`

examples/train/sft/README.md
```diff
@@ -121,6 +126,61 @@ bash examples/train/sft/run_sft_megatron_tulu3_50k.sh \
...
+Each source's weight is divided across its examples, so the source-level mixing
+proportion matches `weights` regardless of how many examples each source has.
+Order the dataset by source and give `lengths`/`weights` per source:
+
+```bash
+PYTHONPATH=examples/train/sft python -m skyrl.train.main_sft \
+    sampler=custom \
+    sampler_class_path=data_mixing_sampler.DataMixingSampler \
+    'sampler_kwargs={lengths: [8000, 2000], weights: [0.5, 0.5], num_samples: 800, seed: 42}' \
+    num_steps=200 batch_size=4
```
SumanthRH: can you actually change this to be overrides on top of the base sft example script? This command as is is too barebones and will break.  

https://github.com/NovaSky-AI/SkyRL/blob/main/examples/train/sft/run_sft_megatron.sh

```bash
bash examples/train/sft/run_sft_megatron.sh \
     sampler=custom \ 
     sampler_class_path=data_mixing_sampler.DataMixingSampler \
     ...
```

examples/train/sft/README.md
```diff
@@ -121,6 +126,61 @@ bash examples/train/sft/run_sft_megatron_tulu3_50k.sh \
...
+across epoch boundaries and makes resume bit-exact across the entire run).
+Put the file's directory on `PYTHONPATH` so the dotted path resolves:
+
+```bash
+PYTHONPATH=examples/train/sft python -m skyrl.train.main_sft \
+    sampler=custom \
+    sampler_class_path=curriculum_sampler.CurriculumLearningSampler \
+    'sampler_kwargs={lengths: [2000, 2000, 2000], num_samples: 800, seed: 42}' \
+    num_steps=200 batch_size=4
+```
```
SumanthRH, line 189: Similar comment as https://github.com/NovaSky-AI/SkyRL/pull/1842/changes#r3541503061

SumanthRH (changes requested):
Needs some small fixes, please take a look!

skyrl/train/sft_trainer.py
avigyabb: Actually, the batch_size attribute on the eval dataloader is different from the normal training batch_size. In build_eval_dataloader() in sft_trainer.py we set batch_size = self.sft_cfg.micro_train_batch_size_per_gpu * dp_size, so the eval_chunk_size in run_eval() is the same value as before- just read from the dataloader instead of recomputed.

examples/train/sft/curriculum_sampler.py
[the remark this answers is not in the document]
avigyabb: My understanding was that iter() only happens on train start and on epoch boundary. Since no other code path after the loop calls iter() this flag would never be used.

examples/train/sft/curriculum_sampler.py
avigyabb: Yes sorry, you're right, will go ahead and update that!

skyrl/train/sft_trainer.py
SumanthRH: You're right, I missed the assignment, can you copy over my comment: 

"# One micro-batch per DP rank per dispatch call — keeps memory usage bounded
        # and removes the need for a separate `eval_batch_size` knob."
        
  to `build_eval_dataloader`  where you set `eval_chunk_size` ?

avigyabb: Done!

SumanthRH (approved):

PR #1414, opened 2026-03-30
[megatron] Fix redundant downloading of shards across workers/nodes for megatron dist checkpointing

erictang000:
## Summary

- **Per-rank cloud checkpoint loading for Megatron**: When loading a distributed checkpoint from S3/GCS, each rank now downloads only its own shard file(s) into a shared node-local temp directory, rather than every rank downloading the entire checkpoint. Local rank 0 creates the directory and downloads common metadata files; after a barrier, each rank downloads its shard in parallel. This reduces total download volume per node from `N * full_checkpoint` to `1 * full_checkpoint`.
- **New `io.download_file()` utility**: Adds a single-file cloud download helper (with S3 retry support) to complement the existing `download_directory()`.
- **New `test_save_load_checkpoint_cloud` test**: End-to-end test that saves a Megatron checkpoint to S3, loads it back via per-rank downloading, and verifies logit-exact reproducibility.

## Test plan

- [x] `test_save_load_checkpoint_cloud` -- saves checkpoint to S3, loads via per-rank shard download, verifies identical logits after reload + re-training
- [x] Existing `test_save_load_checkpoint` parametrizations (local filesystem) -- should be unaffected since the new cloud path is only triggered for `is_cloud_path()` URLs

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1414" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

Issue #800, opened 2025-12-22
[skyrl-train] s3 checkpointing + multi-node results in redundant downloads from s3


erictang000:
Currently, with s3 checkpointing, saving works as expected (each rank saves to the same s3 checkpoint path), but loading causes each worker to download the entire model checkpoint from s3

https://github.com/NovaSky-AI/SkyRL/blob/3ab8eaad5118f592f9d80a3d91829906c981c082/skyrl-train/skyrl_train/distributed/megatron/megatron_strategy.py#L227

Update this logic to just download the relevant ranks on each node (so local rank 0 on each node should handle the s3 download)

PR #1844, opened 2026-06-27
[fsdp] Fix redundant downloading of shards across ranks for FSDP checkpointing

kurtislin:
## Summary

When using FSDP with cloud checkpointing, each rank downloads the full
checkpoint on load even though it only needs its own shard. This PR makes
each rank fetch only its own shard files, cutting per-node download from
`N * full_checkpoint` to `1 * full_checkpoint` (N = world_size).

- Adds `io.local_read_files()`: downloads a given list of files from a cloud
  directory into a temp dir (returns the dir directly for local paths).
  Built on the `download_file()` helper added in #1414.
- `FSDPStrategy.load_checkpoint` now uses `local_read_files` to fetch only
  this rank's `model` / `optim` / `extra_state` shard files, instead of
  `local_read_dir` (which downloads the whole directory).
- Adds `tests/backends/skyrl_train/utils/test_local_read_files.py`.

Follows the approach in #1414 (the Megatron counterpart of this change).
Part of #800.

## Test plan

- [x] `test_local_read_files.py` — asserts only the requested files are
      downloaded for cloud paths, and no download for local paths.

### DIFF
--- a/tests/backends/skyrl_train/utils/test_local_read_files.py
+++ b/tests/backends/skyrl_train/utils/test_local_read_files.py
@@ -1,0 +1,41 @@
+from unittest.mock import patch
+
+from skyrl.backends.skyrl_train.utils.io.io import local_read_files
+
+
+@patch("skyrl.backends.skyrl_train.utils.io.io.download_file")
+@patch("skyrl.backends.skyrl_train.utils.io.io.exists")
+@patch("skyrl.backends.skyrl_train.utils.io.io.is_cloud_path")
+def test_cloud_downloads_only_requested_files(mock_is_cloud_path, mock_exists, mock_download_file):
+    """
+    Test that when a cloud path is provided, only the requested files are downloaded to a temporary directory.
+    """
+    mock_is_cloud_path.return_value = True
+    mock_exists.return_value = True
+
+    requested = [
+        "model_world_size_8_rank_3.pt",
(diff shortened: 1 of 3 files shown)


erictang000 (approved):
looks good to me, thanks!

tests/backends/skyrl_train/utils/test_local_read_files.py
```diff
@@ -0,0 +1,41 @@
...
+
+    downloaded = [call.args[0] for call in mock_download_file.call_args_list]
+    assert len(downloaded) == len(requested)
+    for name in requested:
+        assert any(name in src for src in downloaded)
+
+
+@patch("skyrl.backends.skyrl_train.utils.io.io.download_file")
+@patch("skyrl.backends.skyrl_train.utils.io.io.is_cloud_path")
+def test_local_path_does_not_download(mock_is_cloud_path, mock_download_file):
```
erictang000: =========================== short test summary info ============================
FAILED tests/backends/skyrl_train/utils/test_local_read_files.py::test_local_path_does_not_download - FileNotFoundError: Path does not exist: /some/local/dir
= 1 failed, 878 passed, 6 skipped, 1 deselected, 83 warnings in 303.70s (0:05:03) =

erictang000 (changes requested):
can you fix this cpu test failure?

erictang000 (approved):

PR #1855, opened 2026-07-01
[chore] remove mentions of flashrl from the repo and point to vllm quantization support instead

### LINKED ISSUE #1658: [cleanup] remove flashrl mentions from the repo
since we now use: https://docs.vllm.ai/en/latest/training/layerwise/, flashrl extras and docs: https://docs.skyrl.ai/docs/examples/flash_rl should be removed.

In it's place, we should add docs for quantized training directly with vllm.

erictang000:
closes #1658


SumanthRH:
@erictang000 flashrl cleanup is being done as a part of #1835

SumanthRH (commented):
Can we make this PR just about a new "Training with Quantized Rollouts" doc?

PR #1862, opened 2026-07-02
[megatron] 1/n towards Kimi K2.6: fake QAT training for Qwen3.6-35B-INT4

casper-hansen:
INT4 serving now adds essentially nothing over BF16 serving. The gap with fake-quant on equals the same-weights floor. The example for Megatron uses https://huggingface.co/casperhansen/Qwen3.6-35B-A3B-INT4-RTN which includes the code to use vLLM project's llm-compressor to produce these weights.

The difference between the Qwen3.6 INT4 weights and Kimi is a slight nuance in the SCALE_DIV and Q_MIN variable which is intended to be forward compatible when Kimi K2.6 is supported at a later time.

## Results

STE removes ~55% of the uncorrected train/infer gap, halves the spread, and cuts the tail hardest.

| arm | mean (per step) | avg mean | avg std | worst max |
|---|---|---|---|---|
| OFF (no fake-quant) | 0.0259 – 0.0315 | **0.0279** | 0.086 | 4.99 |
| STE (bit-exact) | 0.0118 – 0.0141 | **0.0126** | 0.040 | 2.24 |

## Can we get closer to zero?

The remaining 0.0118 is the cross-engine numerics floor, and we're sitting exactly on it. The cleanest evidence is STE step 1: at that point the LoRA adapters are still zero and the expert grid is bit-exact, so trainer and vLLM score bit-identical weights and the gap is 0.0118.

## Inspiration

This was inspired by the work done in miles https://github.com/radixark/miles/ on the Megatron side.


casper-hansen:
@erictang000 @SumanthRH could you please review this? Before I move forward to larger models, I would like to land this core feature for quantized models like Kimi K2.6

examples/train/megatron/run_megatron_dapo_qwen3.6_35b_a3b_lora_int4_qat.sh
```diff
@@ -0,0 +1,172 @@
...
+  generator.eval_sampling_params.temperature=$TEMPERATURE \
+  generator.eval_sampling_params.max_generate_length=$MAX_RESPONSE_LENGTH \
+  trainer.algorithm.use_kl_loss=$USE_KL_LOSS \
+  trainer.algorithm.clip_ratio_c=$CLIP_RATIO_C \
+  trainer.policy.model.path="$MODEL_NAME" \
+  trainer.policy.model.fake_int4_qat.enabled=$FAKE_QUANT_ENABLED \
+  trainer.policy.model.fake_int4_qat.group_size=32 \
+  trainer.policy.model.fake_int4_qat.scale_divisor=7.5 \
+  trainer.policy.model.fake_int4_qat.bf16_base_path="$FAKE_QUANT_BF16_PATH" \
+  trainer.policy.megatron_config.lora_config.merge_lora=false \
```
erictang000, line 121: if i'm understanding correctly, it seems like using lora + merge_lora=false is a requirement for `fake_int4_qat.enabled=True`, since expert_hf_weights for weight sync from Megatron-Bridge always ships bf16 weights.

This makes sense to me as a first step, but can you add an assertion in the `ModelConfig` to flag the current limitation?

examples/train/megatron/run_megatron_dapo_qwen3.6_35b_a3b_lora_int4_qat.sh
```diff
@@ -0,0 +1,172 @@
...
+# isolates the effect of fake-quant + TIS:
+#   QAT_MODE=off : fake-quant OFF + TIS OFF  -> uncorrected BF16(train)/INT4(infer) mismatch
+#   QAT_MODE=on  : fake-quant ON  + TIS ON   -> corrected (on-policy)
+#
+#   QAT_MODE=off bash examples/train/megatron/run_megatron_dapo_qwen3.6_35b_a3b_lora.sh
+#   QAT_MODE=on  bash examples/train/megatron/run_megatron_dapo_qwen3.6_35b_a3b_lora.sh
+
+# INT4 actor served by vLLM; BF16 masters loaded by the trainer (Megatron-Bridge
+# can't load compressed-tensors, so it reads BF16 from FAKE_QUANT_BF16_PATH).
+MODEL_NAME="${MODEL_NAME:-/data/qwen36-int4/Qwen3.6-35B-A3B-INT4-RTN}"
```
erictang000: can you use the huggingface paths here instead of relative local paths? or add a download command above

skyrl/train/config/config.py
```diff
@@ -100,10 +100,45 @@ class SkyRLLoraConfig(BaseConfig):
...
+    ``casperhansen/Qwen3.6-35B-A3B-INT4-RTN``); ``7.0`` = Kimi K2-Thinking / K2.6 /
+    Miles (``[-7, 7]``). Set ``q_min`` consistently."""
+    q_min: float = -8.0
+    """Lower clamp of the INT4 code range: ``-8`` for llm-compressor RTN
+    (``scale_divisor=7.5``), ``-7`` for Kimi/Miles (``scale_divisor=7.0``, whose
+    QAT never emits ``-8``)."""
+    bf16_base_path: Optional[str] = None
+    """Megatron-Bridge cannot load a compressed-tensors INT4 checkpoint, so when
+    ``model.path`` points at the INT4 model the trainer loads its BF16 master
+    weights from this path instead (the Miles ``--ref-load`` pattern). The INT4
```
erictang000: can we remove this reference to the miles flag

skyrl/train/config/config.py
```diff
@@ -100,10 +100,45 @@ class SkyRLLoraConfig(BaseConfig):
     >= ``max_loras`` if explicitly set."""
 
 
+@dataclass
+class FakeInt4QatConfig(BaseConfig):
+    """Fake-INT4 quantization-aware training for MoE experts (Megatron only).
```
erictang000, line 105: can we add an assertion either in a post init or in https://github.com/NovaSky-AI/SkyRL/blob/abd553a125bb9284f1333ccba475f5fc3e512e18/skyrl/train/utils/utils.py#L221 that if this is enabled, we are using megatron?

erictang000 (commented):
@casper-hansen thanks for the PR, this is great! just a few minor comments

examples/train/megatron/run_megatron_dapo_qwen3.6_35b_a3b_lora_int4_qat.sh
casper-hansen: Your understanding is correct. This is intended for merge_lora=False since you can't really merge a LoRA adapter into quantized weights.

examples/train/megatron/run_megatron_dapo_qwen3.6_35b_a3b_lora_int4_qat.sh
casper-hansen: Done

skyrl/train/config/config.py
casper-hansen: Done

skyrl/train/config/config.py
casper-hansen: Done

erictang000 (approved):
Great PR, thanks @casper-hansen!

Validated the script on B200 and everything looks as expected:

Image

Image

Image

PR #1872, opened 2026-07-07
[deps] bump ray to 2.56.0, add TE wheel to pyproject.toml

erictang000:
### Summary
  Upgrades the pinned Ray version from `2.51.1` → `2.56.0` across Docker images, CI configs, docs, and the lockfile, and moves the Megatron backend's Transformer Engine install from a build-from-source step to prebuilt wheels
  (`2.10.0` → `2.11.0`).

  ### Changes

  **Ray 2.56.0 bump**
  - Updated base images in `docker/Dockerfile` and `docker/Dockerfile.megatron` to `anyscale/ray:2.56.0-slim-py312-cu128`.
  - Bumped `image_uri`/`ray_version` in all `ci/anyscale_*.yaml` CI configs to `2.56.0`.
  - Updated `override-dependencies` in `skyrl-agent/pyproject.toml` and refreshed `uv.lock`.
  - Updated the Modal integration example and docs (`installation.mdx`, `.claude/docs/ci.md`, READMEs) to reference the new `2.56.0` image tags and version.
  - Adapted to the Ray API change: `PlacementGroupSchedulingStrategy` is now imported from `ray.util.scheduling_strategies` instead of `ray.util.placement_group` (in `worker.py` and `utils/utils.py`).

  **Transformer Engine 2.11.0 via prebuilt wheels**
  - Removed the in-Dockerfile `pip wheel` / build step for `transformer-engine-torch` from `Dockerfile.megatron`; bumped `torch` `2.10.0` → `2.11.0` in the base layer.
  - Declare `transformer-engine-torch==2.11.0` directly in the `megatron` extra so the `[tool.uv.sources]` URL (a prebuilt torch-2.11 / cu128 / cp312 wheel) is honored — uv ignores sources for transitive-only deps.
  - Added `transformer-engine-cu12==2.11.0` explicitly: the meta `transformer-engine` package hardcodes `transformer-engine-cu13` as its core dep (overridden to `never`), so the matching cu12 core lib must be pulled in
  directly — otherwise the meta package's sanity check fails on this CUDA 12.8 image.


SumanthRH (commented):
update image URIs in the docs as well?

docker/Dockerfile.megatron
```diff
@@ -45,10 +45,4 @@ RUN uv pip install  --system --no-cache-dir \
     wheel
 
 ENV CMAKE_ARGS=-DPython_EXECUTABLE=/home/ray/anaconda3/bin/python3
-
-RUN sudo mkdir /opt/wheels && sudo chown ray: /opt/wheels && \
```
erictang000, line 49: this was not working previously:

https://console.anyscale.com/cld_hxkifz7xa22mwicp21nzkds1lw/prj_4b6c498rypyq6g7yhk6vzgjevt/jobs/prodjob_mp24u22lj9zg7vsmm6qyl4a9vb?job-logs-section-tabs=application_logs&job-tab=overview

the cleaner solution is just to directly save the wheel source in the pyproject.toml for TE-torch.

SumanthRH: Hmm this looks like an issue with the path in our docker builds;

```bash
Installed 344 packages in 967ms
  × Failed to download `transformer-engine-torch==2.11.0`
  ├─▶ Failed to read from the distribution cache
  ╰─▶ failed to query metadata of file
      `/tmp/ray/opt/wheels/transformer_engine_torch-2.11.0-cp312-cp312-linux_x86_64.whl`:
      No such file or directory (os error 2)
  help: `transformer-engine-torch` (v2.11.0) was included because
        `skyrl[megatron]` depends on `transformer-engine[pytorch]` (v2.11.0)
        which depends on `transformer-engine-torch`
```

Interesting that the path used is `/tmp/ray/opt/wheels` rather than `/opt/wheels`.

SumanthRH (approved):

SumanthRH (commented):
Can you also update `examples/train/thunder_agent/scripts/r2egym_32b/setup_env.sh`: 

https://github.com/NovaSky-AI/SkyRL/blob/1ab51f1b965c0a3a6fa7dfd4e965758525bfd89b/examples/train/thunder_agent/scripts/r2egym_32b/setup_env.sh

PR #1876, opened 2026-07-09
[tinker] expose dppo through tinker api

casper-hansen:
Small change to expose the new DPPO through the Tinker API.


mojtaba-komeili (approved):
It looks like we only need to match the request args for the running SkyRL service. The core of changes LGTM if it is passing the smoke test runs.

casper-hansen:
@erictang000 would love a review on this small change

erictang000 (approved):
lgtm, thanks!

PR #1880, opened 2026-07-10
[tinker] Report rollout/train logprob gap in forward_backward metrics

### LINKED ISSUE #1870: forward_backward never reports the rollout/train logprob gap (rollout_logprobs is never populated, metrics get dropped)
When you drive RL training through the tx Tinker API (`skyrl.tinker.api`), there is currently no way to see the train/inference logprob mismatch — the `policy/rollout_train_logprobs_abs_diff_*` family that skyrl-train logs natively. `forward_backward` responses only carry a small whitelist (`total_loss:sum`, `pg_loss:sum`, ...), so a Tinker client that wants this signal has to pull the per-token `logprobs` out of `loss_fn_outputs` and diff them against its own datums client-side. At long context that means walking hundreds of thousands of floats in Python every step, just to reproduce a number the workers already compute.

The pieces all exist, they're just not connected on the tx path:

1. `SkyRLTrainBackend._to_training_batch` maps each datum's sampling logprobs to `action_log_probs` but never fills in `rollout_logprobs`. On the Tinker path the datum logprobs *are* the rollout engine's logprobs, so mirroring them is the natural mapping. Because the field is missing, `compute_minibatch_rollout_logprob_diff_metrics(...)` in the workers early-returns `{}`. As a side effect, `trainer.algorithm.off_policy_correction` (TIS etc.) is also silently a no-op over the Tinker API, since the correction keys off `rollout_logprobs` too.
2. Even if the workers did emit the metrics, `SkyRLTrainBackend._extract_metrics` drops everything that isn't in its fixed whitelist, so they'd never reach `ForwardBackwardOutput.metrics`.

Two implementation details worth knowing for anyone picking this up:

- std can't be mean-reduced across micro-batches / DP ranks, which is why the workers emit first/second moments. The backend should reconstruct `std = sqrt(E[x²] − E[x]²)`, same as `finalize_minibatch_rollout_logprob_diff_std` does for the native trainer.
- the Tinker SDK's client-side chunk combiner (`combine_fwd_bwd_output_results`) does `name, reduction = key.split(":")` on every metric key, so any new key must carry a `:mean` / `:max` / `:min` suffix or the client errors out when a request gets split into chunks.

Here's the patch I've been testing against current main. With it, `forward_backward` responses report `policy/rollout_train_logprobs_abs_diff_{mean,std,max,min}`; I checked the values against direct masked `|logp_train − logp_rollout|` stats, and the suffixed keys reduce correctly through the SDK's chunk combiner (mean weighted by datum count, max/min exact).

```diff
diff --git a/skyrl/backends/skyrl_train_backend.py b/skyrl/backends/skyrl_train_backend.py
index 016ff2ec..75b37163 100644
--- a/skyrl/backends/skyrl_train_backend.py
+++ b/skyrl/backends/skyrl_train_backend.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import io
+import math
 import os
 import tarfile
 import tempfile
@@ -23,6 +24,12 @@ from skyrl.backends.skyrl_train.training_batch import (
 )
 from skyrl.backends.skyrl_train.workers.worker import PPORayActorGroup
 from skyrl.backends.skyrl_train.workers.worker_dispatch import WorkerDispatch
+from skyrl.backends.skyrl_train.workers.worker_utils import (
+    MINIBATCH_ROLLOUT_LOGPROB_DIFF_MAX_KEY,
+    MINIBATCH_ROLLOUT_LOGPROB_DIFF_MEAN_KEY,
+    MINIBATCH_ROLLOUT_LOGPROB_DIFF_MIN_KEY,
+    MINIBATCH_ROLLOUT_LOGPROB_DIFF_SQ_MEAN_KEY,
+)
 from skyrl.env_vars import SKYRL_RAY_PG_TIMEOUT_IN_S
 from skyrl.tinker import types
 from skyrl.train.config import SkyRLTrainConfig, get_config_as_yaml_str
@@ -591,6 +598,14 @@ class SkyRLTrainBackend(AbstractBackend):
         has_returns = any(len(r) > 0 for r in prepared_batch.all_returns)
         if has_logprobs:
             batch_dict["action_log_probs"] = torch.tensor(action_log_probs_list, dtype=torch.float32)
+            # Tinker datums carry the *sampling* (rollout-engine) logprobs.
+            # Mirror them into `rollout_logprobs` so the policy workers emit
+            # the train-vs-rollout logprob-gap metrics
+            # (`minibatch_rollout_logprobs_abs_diff_*`), surfaced by
+            # `_extract_metrics` below.  Loss behaviour only changes when
+            # `trainer.algorithm.off_policy_correction` is explicitly
+            # configured (rollout logprobs are its intended input).
+            batch_dict["rollout_logprobs"] = torch.tensor(action_log_probs_list, dtype=torch.float32)
         if has_advantages:
             batch_dict["advantages"] = torch.tensor(advantages_list, dtype=torch.float32)
         if role == "critic":
@@ -682,6 +697,33 @@ class SkyRLTrainBackend(AbstractBackend):
         if "critic_lr" in data:
             metrics["critic_lr:last"] = float(data["critic_lr"])
 
+        # Train-vs-rollout logprob gap, computed by the policy workers per
+        # micro-batch (masked |logp_train - logp_rollout| over action tokens)
+        # and reduced across micro-batches / DP ranks.  Reconstruct the std
+        # from the reduced first/second moments (std itself cannot be
+        # mean-reduced; same as finalize_minibatch_rollout_logprob_diff_std)
+        # and surface the family under skyrl-train's familiar
+        # `policy/rollout_train_logprobs_abs_diff_*` names.  The `:mean` /
+        # `:max` / `:min` suffixes drive the Tinker SDK's cross-chunk
+        # reduction when a request is split into multiple chunks.
+        if MINIBATCH_ROLLOUT_LOGPROB_DIFF_MEAN_KEY in data:
+            mean = float(data[MINIBATCH_ROLLOUT_LOGPROB_DIFF_MEAN_KEY])
+            metrics["policy/rollout_train_logprobs_abs_diff_mean:mean"] = mean
+            if MINIBATCH_ROLLOUT_LOGPROB_DIFF_SQ_MEAN_KEY in data:
+                sq_mean = float(data[MINIBATCH_ROLLOUT_LOGPROB_DIFF_SQ_MEAN_KEY])
+                # max(0, ...) guards tiny negatives from float round-off.
+                metrics["policy/rollout_train_logprobs_abs_diff_std:mean"] = math.sqrt(
+                    max(0.0, sq_mean - mean**2)
+                )
+            if MINIBATCH_ROLLOUT_LOGPROB_DIFF_MAX_KEY in data:
+                metrics["policy/rollout_train_logprobs_abs_diff_max:max"] = float(
+                    data[MINIBATCH_ROLLOUT_LOGPROB_DIFF_MAX_KEY]
+                )
+            if MINIBATCH_ROLLOUT_LOGPROB_DIFF_MIN_KEY in data:
+                metrics["policy/rollout_train_logprobs_abs_diff_min:min"] = float(
+                    data[MINIBATCH_ROLLOUT_LOGPROB_DIFF_MIN_KEY]
+                )
+
         return metrics
 
     def _sleep_inference_engines(self):
```

Populating `rollout_logprobs` only changes loss behavior when `off_policy_correction` is explicitly configured — and since that config currently does nothing over the Tinker API, that half is arguably a bug fix on its own. If you'd rather be conservative, the batch-dict line could sit behind a backend config flag. Happy to turn this into a PR if the approach looks right.

casper-hansen:
Mirror Tinker datum sampling logprobs into `rollout_logprobs` so policy workers emit the logprob-gap metrics, and surface them in ForwardBackwardOutput as `policy/rollout_train_logprobs_abs_diff_*` with SDK-reducible suffixes (std reconstructed from first/second moments).

Fixes #1870

### DIFF
--- a/skyrl/backends/skyrl_train_backend.py
+++ b/skyrl/backends/skyrl_train_backend.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import io
+import math
 import os
 import tarfile
 import tempfile
@@ -23,6 +24,12 @@ from skyrl.backends.skyrl_train.training_batch import (
 )
 from skyrl.backends.skyrl_train.workers.worker import PPORayActorGroup
 from skyrl.backends.skyrl_train.workers.worker_dispatch import WorkerDispatch
+from skyrl.backends.skyrl_train.workers.worker_utils import (
+    MINIBATCH_ROLLOUT_LOGPROB_DIFF_MAX_KEY,
+    MINIBATCH_ROLLOUT_LOGPROB_DIFF_MEAN_KEY,
+    MINIBATCH_ROLLOUT_LOGPROB_DIFF_MIN_KEY,
+    MINIBATCH_ROLLOUT_LOGPROB_DIFF_SQ_MEAN_KEY,
+)
 from skyrl.env_vars import SKYRL_RAY_PG_TIMEOUT_IN_S
 from skyrl.tinker import types
 from skyrl.train.config import SkyRLTrainConfig, get_config_as_yaml_str
@@ -591,6 +598,14 @@ class SkyRLTrainBackend(AbstractBackend):
         has_returns = any(len(r) > 0 for r in prepared_batch.all_returns)
         if has_logprobs:
             batch_dict["action_log_probs"] = torch.tensor(action_log_probs_list, dtype=torch.float32)
+            # Tinker datums carry the *sampling* (rollout-engine) logprobs.
+            # Mirror them into `rollout_logprobs` so the policy workers emit
+            # the train-vs-rollout logprob-gap metrics
+            # (`minibatch_rollout_logprobs_abs_diff_*`), surfaced by
+            # `_extract_metrics` below.  Loss behaviour only changes when
+            # `trainer.algorithm.off_policy_correction` is explicitly
+            # configured (rollout logprobs are its intended input).
+            batch_dict["rollout_logprobs"] = torch.tensor(action_log_probs_list, dtype=torch.float32)
         if has_advantages:
             batch_dict["advantages"] = torch.tensor(advantages_list, dtype=torch.float32)
         if role == "critic":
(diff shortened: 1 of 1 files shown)


casper-hansen:
@erictang000 tagging you for review

erictang000 (approved):
lgtm, thanks!

PR #1883, opened 2026-07-10
[feat][SFT] Support multiple datasets for training and evaluation in SFTTrainer

### LINKED ISSUE #1875: [RFC] Support for training and evaluation on multiple datasets in `SFTTrainer`
## Summary

Add native `SFTTrainer` support for training on a weighted mixture of multiple
datasets and evaluating on multiple named datasets. This should be a thin layer on top of the existing `StatefulDataLoader` and still allow for custom samplers on top. Evaluation metrics should be logged per-evaluation dataset.

The intended user experience is:

```bash
bash examples/train/sft/run_sft_megatron.sh \
  train_datasets="['allenai/tulu-3-sft-mixture','yahma/alpaca-cleaned','openchat/openchat_sharegpt4_dataset']" \
  train_dataset_splits="['train[:50000]','train[:10000]','train[:10000]']" \
  train_dataset_weights="[0.8,0.1,0.1]" \
  eval_datasets="['allenai/tulu-3-sft-mixture','yahma/alpaca-cleaned']" \
  eval_dataset_splits="['train[-500:]','train[200:700]']" \
  eval_dataset_names="['tulu3','alpaca']"
```

## Motivation

Often, researchers are experimenting with multiple datasets for SFT. For training, it is common to iterate on different data mixtures for SFT. Further, evaluation is typically performed on multiple datasets, often not only on the datasets present in the training mixture but additional datasets to evaluate model capabilities on other axes, track any catastrophic forgetting, etc.

This RFC details a simple extension to the `SFTTrainer` to support multiple datasets for training and evaluation in SkyRL.

## Goals

- Support multiple training datasets with source-level sampling weights.
- Keep weighted sampling independent of source dataset size.
- Support multiple eval datasets with metrics logged separately per dataset.
- Support custom samplers with multiple training datasets: For example, users can change the dataset mixing weights after training progresses. 
- All datasets should have the same relevant columns for training - `messages_key`, `tools_key` and `system_key`, but can have additional unique columns beyond these.

## Non-goals

- Streaming dataset loading. This RFC keeps the current in-memory tokenized dataset model.
- Dataset-specific preprocessing or tokenization settings.
- Native dynamic mixing schedules. Users can express these through `sampler=custom`.
- Mixing datasets with different modalities or different schemas. All datasets should support the same `messages_key`, `tools_key` and `system_key`. Further, they should all be text-only or with images but not both.

## Proposed config

Add list-based train and eval dataset fields:

```python
class SFTConfig:
    # Training data
    train_datasets: Optional[List[str]] = None
    train_dataset_splits: Optional[List[str]] = None
    train_dataset_weights: Optional[List[float]] = None

    # Evaluation data
    eval_datasets: Optional[List[str]] = None
    eval_dataset_splits: Optional[List[str]] = None
	# optional, shorthand names used only for logging
    eval_dataset_names: Optional[List[str]] = None
```

1. Compatibility:

- `dataset_name` and `dataset_split` are deprecated in favour of `train_datasets` and `train_dataset_splits`. This is similarly true for `eval_datasets` and `eval_dataset_name`.
- SkyRL will still support user-provided overrides with the old fields, but internally translate them to the new list format. Add a deprecation warning in such cases. 

2. What does `train_dataset_weights` mean?
- Ratios of samples from each dataset within a batch (approximately), unrelated to the lengths of the dataset.
- Default ratio is `1/N` where `N` is the number of training datasets -> equal mixing from all the datasets.

## Training design

- Load and tokenize each training dataset independently using the existing
`_load_and_tokenize(dataset_name, dataset_split)` path. This preserves current
tokenization behavior and cache keys. After tokenization, concatenate the lists. Keep track of the lengths of each dataset.
- When there is a single dataset specified, keep the behaviour unchanged: Simply use basic shuffling or user provided sampler for dataloading.
- With multiple datasets: load into a single dataloader as before, but with a [`DataMixingSampler`](https://github.com/NovaSky-AI/SkyRL/blob/1ab51f1b965c0a3a6fa7dfd4e965758525bfd89b/examples/train/sft/data_mixing_sampler.py#L37) by default. Configure the lengths according to the individual dataset sizes and weights according to user specified `train_dataset_weights`.
	- Allow users to provide a custom sampler override. This can choose to respect the ratios or not
	- Ensure that the weighted mixing in `DataMixingSampler` resamples a _fresh_ list of indices every epoch. Currently, sampler's indices list `_plan` is instantiated once and re-used. Update the state of the sampler to include the `rng` state for resumption.
- Checkpointing and resume has no additional considerations: We built on top of the existing dataloading save/resume logic.

## Evaluation design

- Support evaluation on multiple datasets separately. Build one evaluation dataloader per evaluation dataset.
- Run evaluation separately and log metrics individually, grouped under `eval/{eval_dataset_name}`. This nesting will be used even if only one dataset is provided so that users can easily compare WandB metrics across runs with and without dataset mixing.

## Implementation Plan
1. Add support for the new configuration + validation for translating deprecated params
2. Add support for multiple training datasets in the dataset preprocessing stage
3. Move the `DataMixingSampler` into a native sampler in SkyRL and use that in the case with multiple datasets
4. Add support for multiple evaluation datasets as specified above
5. Update documentation with an example for training with multiple datasets and evaluation on multiple datasets. Add an example for user-provided custom sampler on top. 

## Test Plan

The tests can be CPU only tests:
1. Test with user-specified `train_dataset_weights` and `train_datasets`: ensure that representation in a batch is close to the weights specified. 
2. Test with default `train_dataset_weights` and user-specified `train_datasets`: equal mixing
3. Test with user-specified `train_dataset_weights` and `train_datasets` with varying dataset lengths: Mixing ratios are independent of length. Representation in a batch should be roughly unchanged.

avigyabb:
Implements RFC #1875: native `SFTTrainer` support for training on a weighted mixture of multiple datasets and evaluating on multiple named datasets, as a thin layer on top of the `StatefulDataLoader`/sampler infrastructure from #1842.

### Config (`skyrl/train/config/sft_config.py`)
- New list-based fields: `train_datasets` / `train_dataset_splits` / `train_dataset_weights` and `eval_datasets` / `eval_dataset_splits` / `eval_dataset_names`.
- `dataset_name`/`dataset_split` and `eval_dataset_name`/`eval_dataset_split` are **deprecated**: still accepted, translated to the list form with a `DeprecationWarning` (both CLI and programmatic paths), and rejected when combined with the new fields.
- Validation: list lengths must match, weights strictly positive (default `1/N` equal mixing), eval names unique; explicit `train_dataset_weights` only valid with `sampler=random` (custom samplers take ratios via `sampler_kwargs`).

### DataMixingSampler promoted to core (`skyrl/train/dataset/samplers.py`)
- Moved from `examples/train/sft/data_mixing_sampler.py` (which now re-exports it, so existing `sampler_class_path` configs keep working).
- Fixes vs the example, per the RFC: draws a **fresh weighted plan every epoch** (the example materialized `_plan` once and replayed it), and `state_dict` now includes the generator state, so mid-epoch and epoch-boundary checkpoint resumes reproduce the uninterrupted sample stream exactly. Old position-only states load best-effort with a warning.

### Trainer (`skyrl/train/sft_trainer.py`)
- Each training dataset is tokenized independently through the existing `_load_and_tokenize` path (per-`(name, split)` cache keys preserved), then concatenated with per-dataset lengths tracked.
- Single dataset: behavior completely unchanged (`sampler=random` still uses the dataloader's built-in shuffle). Multiple datasets: `sampler=random` selects `DataMixingSampler` configured with the tokenized lengths and `train_dataset_weights` — per-batch source ratios, independent of dataset sizes.
- Custom samplers receive the per-dataset `lengths` via an injected kwarg (user-supplied `sampler_kwargs` win).
- Eval: one dataloader per eval dataset, evaluated separately, logged under `eval/{name}/loss` — nested even with a single eval dataset so runs with and without mixing chart the same keys.
- Checkpoint/resume reuses the existing `data.pt` dataloader state; no new mechanism.

### :warning: Breaking change
The eval metric key `eval/eval_loss` is now `eval/{name}/loss` (intentional per the RFC). Dashboards/callbacks keying on the old name need updating.

### Tests (CPU-only, per the RFC test plan)
- Batch representation tracks user-specified weights (`[0.8, 0.2]` → ~80/20), default weights give equal mixing, and ratios are independent of dataset sizes (20-vs-180 sources still ~50/50).
- Core sampler: fresh plan per epoch, deterministic multi-epoch stream per seed, exact resume mid-epoch and across epoch boundaries (sampler-level and through `StatefulDataLoader`), old-format state compatibility.
- Config normalization/deprecation/validation coverage in `test_sft_config.py`; callback test updated to the nested eval keys.

### Docs & examples
- `examples/train/sft/README.md`: new *Multi-dataset training and evaluation* section, config reference updates, breaking-change note; removed the single-dataset limitation.
- New `examples/train/sft/run_sft_megatron_multi_dataset.sh`; existing SFT run scripts (and the GPU e2e script) migrated to the list fields so layering `train_datasets=[...]` overrides on them doesn't conflict with the deprecated fields they used to set.

## Testing

`tests/train/test_sft_dataloader.py`, `test_sft_config.py`, `test_sft_callbacks.py`: **106 passed**. Broader CPU suite (`tests/train/` + `tests/backends/skyrl_train/`, GPU dirs excluded): 965 passed; the 3 remaining failures are missing `vllm`/torchvision optional deps in the local env, unrelated to this change.

Closes #1875


avigyabb:
Ran the GPU e2e test (`tests/train/gpu_e2e_test/sft_tulu3_megatron.sh`) on 4x L40S against this branch — **all CI assertions passed** ✅

- 100/100 steps completed; `train/loss` trended down (2.80 @ step 20 → 2.19 @ step 100), no NaN/inf.
- Final eval logged under the new nested key: `eval/allenai_tulu-3-sft-mixture/loss = 1.9843` (62 batches) — exercising the default sanitized `eval_dataset_names` and the `eval/{name}/loss` scheme end-to-end.
- The run also exercises the migrated script path: the e2e layers `"train_dataset_splits=['train[:2000]']"` over the base script's list-based fields.
- wandb run: https://wandb.ai/sky-posttraining-uc-berkeley/skyrl_sft_ci/runs/rii88p8v

The first e2e attempt caught a real bug, fixed in fe512549: OmegaConf/YAML can't parse HF slice syntax unquoted inside a list override (`train_dataset_splits=[train[:2000]]` → ParserError on the nested brackets). All scripts/README examples now quote each element, with a CPU regression test encoding the syntax.

examples/train/sft/data_mixing_sampler.py
SumanthRH, line 1: Just remove it?

examples/train/sft/run_sft_megatron_multi_dataset.sh
```diff
@@ -0,0 +1,53 @@
...
+uv run --isolated --extra megatron \
+    python -m skyrl.train.main_sft \
+    strategy=megatron \
+    model.path=Qwen/Qwen2.5-0.5B-Instruct \
+    "train_datasets=['allenai/tulu-3-sft-mixture','yahma/alpaca-cleaned']" \
+    "train_dataset_splits=['train[:800]','train[:200]']" \
+    "train_dataset_weights=[0.8,0.2]" \
+    "eval_datasets=['allenai/tulu-3-sft-mixture','yahma/alpaca-cleaned']" \
+    "eval_dataset_splits=['train[-100:]','train[200:300]']" \
+    "eval_dataset_names=[tulu3,alpaca]" \
```
SumanthRH: For my understanding: `OmegaConf` will fail to parse the below override? 


```bash
    train_datasets="['allenai/tulu-3-sft-mixture','yahma/alpaca-cleaned']" \
```

tests/train/test_sft_dataloader.py
```diff
@@ -1,17 +1,25 @@
...
+    tracks the weights
+  - default weights: equal mixing across datasets
+  - mixing ratios are independent of the individual dataset sizes
+  - DataMixingSampler draws a fresh plan every epoch and resumes exactly
+    (mid-epoch and across epoch boundaries) via its RNG state
+
+The custom-sampler path is exercised with small test-local samplers (curriculum
+learning remains a user-supplied example under
+``examples/train/sft/curriculum_sampler.py``, loaded by file path here to keep
+it covered).
```
SumanthRH: Can you please make sure that Claude doesn't add such "in the process" comments? 

Comments should describe what the code is doing, not what the goal was or what learnings-on-the-journey the agent stumbled into.

See examples here: https://github.com/NovaSky-AI/SkyRL/blob/main/.claude/docs/contributing.md#comments

SumanthRH (commented):
Looking good! Left a few minor comments

tests/train/test_sft_dataloader.py
avigyabb: Yep, sorry!

examples/train/sft/run_sft_megatron_multi_dataset.sh
avigyabb: Running this in the terminal, I believe OmegaConf parses as expected?

```
uv run --isolated --no-project --with omegaconf -- python -c "from omegaconf import OmegaConf; print(OmegaConf.from_cli([\"train_datasets=['allenai/tulu-3-sft-mixture','yahma/alpaca-cleaned']\"]).train_datasets)"
```

SumanthRH: Can you change all the scripts to use this format then:

```bash
    train_datasets="['allenai/tulu-3-sft-mixture','yahma/alpaca-cleaned']" \
```

it's cleaner

SumanthRH (approved):

PR #1884, opened 2026-07-11
[megatron] Enable grouped outer expert LoRA targeting for higher accuracy

casper-hansen:
Add `megatron_config.lora_config.experts_shared_outer_loras`: fc1 (gate_up) lora_A and fc2 (down) lora_B are shared across all experts while the inner matrices are trained per expert, matching Megatron-Bridge's `LoRA(experts_shared_outer_loras=True)`.

For merge_lora=false sync, vLLM has no shared-expert-LoRA contract, so the adapter export stacks the per-expert emissions of packed-HF models back into 3D and expands the shared side to all experts in
`_convert_moe_experts_lora_to_vllm` (per-expert-HF models get indexed-key replication instead). Also patches Megatron-Bridge's `PackedPerExpertLinear.sharded_state_dict`, which omits the required `pg_collection` kwarg and breaks dist-checkpoint saving.

Note that we bump megatron bridge to [`fbf2d3c07f8813b86886b3102a1a9ce0532431c4`](https://github.com/NVIDIA-NeMo/Megatron-Bridge/tree/fbf2d3c07f8813b86886b3102a1a9ce0532431c4) which is a newer commit that adds this feature.

## Held-out GSM8K accuracy (1319 problems, greedy)

I ran a few examples to demonstrate that it works. We observe below that the grouped expert approach improves faster.

| | step 0 | step 8 | step 16 | RL gain (0→16) |
|---|---|---|---|---|
| OFF seed 42 | 79.61 | 82.03 | 83.70 | +4.09 |
| OFF seed 7 | 79.76 | 82.87 | 84.08 | +4.32 |
| ON seed 42 | 79.61 | 82.87 | **85.06** | **+5.45** |
| ON seed 7 | 79.68 | 83.47 | **84.69** | **+5.01** |

## Checkpoint size

One caveat is that this seems to produce much larger checkpoints.

Measured from the real Qwen3.6-35B-A3B artifacts (rank 32, TP=4/EP=8, 40 layers) for ON, with OFF computed by the same per-tensor accounting.

| | shared-outer ON (measured) | OFF, default (computed) | ratio |
|---|---|---|---|
| Distinct trained adapter params | 521 M (1.04 GB bf16) | 70.5 M (0.14 GB bf16) | 7.4× |
| Per-rank `adapter_*.pt` in checkpoint | 142.8 MB | ~20.8 MB | 6.9× |
| All 8 rank files | 1.14 GB | ~0.17 GB | 6.9× |
| Optimizer state (`.distcp` shards, fp32 + Adam) | 6.7 GB measured | ~0.9 GB | ~7.4× |
| **Total policy checkpoint dir** | **7.4 GB measured** | **~1.0–1.1 GB** | **~7×** |


erictang000:
@casper-hansen pulled out the dependency bump into: https://github.com/NovaSky-AI/SkyRL/pull/1888, will come back to this after that is merged

PR #1886, opened 2026-07-13
[bugfix] Fused LM Head Compatibility with HybridModel

dyurk-lila:
## Problem

The fused LM-head path introduced in #1841 passes `output_processor` and `output_processor_context` to the Megatron model. This works for `GPTModel`, whose `forward` method exposes that hook, but Nemotron-3 uses Megatron's `HybridModel`, which does not.

As a result, enabling `trainer.fused_lm_head_logprob` for a Nemotron-3 policy or reference model fails on the first fused forward pass, before training begins.

## Reproduction

Prepare the GSM8K data as described in [`examples/train/nemotron_3/README.md`](https://github.com/NovaSky-AI/SkyRL/tree/main/examples/train/nemotron_3), then run the existing 8-GPU Nemotron-3 example with fused LM-head log-probs enabled:

```bash
bash examples/train/nemotron_3/run_nemotron_3_nano_4b_gsm8k.sh \
  trainer.fused_lm_head_logprob=true
```

On `main` before this fix, the failure terminates with:

```text
Traceback (most recent call last):
  File ".../skyrl/backends/skyrl_train/workers/megatron/megatron_model_wrapper.py", line 803, in forward_step
    outputs = model(
  File ".../torch/nn/modules/module.py", line 1775, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File ".../torch/nn/modules/module.py", line 1786, in _call_impl
    return forward_call(*args, **kwargs)
TypeError: HybridModel.forward() got an unexpected keyword argument 'output_processor'
```

Ray and launcher frames above this traceback tail vary by environment; the `HybridModel.forward` error is the underlying failure.

## Fix

This PR adds a compatibility helper for fused LM-head forwards:

- Models with a native `output_processor` parameter retain the existing `GPTModel` path.
- On the final pipeline stage, `HybridModel` temporarily disables `post_process` to return decoder hidden states, restores it even if forward fails, and invokes the same fused output processor with the model's output layer.
- Intermediate pipeline stages continue returning hidden states unchanged.
- The skipped `ColumnParallelLinear` behavior is preserved for sequence-parallel gathers and non-sequence-parallel input-gradient reduction.
- Fine-grained activation offloading still marks output-layer parameters as non-offloadable before `post_process` changes.

MXFP8 output projection, MuP output scaling, and HybridModel MTP remain explicit unsupported cases because bypassing their normal output paths would change semantics.

## Why this fixes the bug

The fused loss needs decoder hidden states and the LM-head weight, not materialized vocabulary logits. `GPTModel` already exposes those through its output-processor hook. The compatibility path obtains the same inputs from `HybridModel` without passing unsupported kwargs or monkey-patching Megatron, then reuses the existing fused processor. The default non-fused path is unchanged.

## Validation

- `uv run --isolated --extra dev --extra ray pytest -q tests/backends/skyrl_train/distributed/test_fused_lm_head.py` — 9 passed
- targeted pre-commit hooks on all changed files — Ruff, Black, and secret scanning passed

---

> [!NOTE]
> **Medium Risk**
> Touches Megatron forward/post_process and tensor-parallel gradient paths for fused log-probs; incorrect handling could break Nemotron training or TP correctness, but scope is gated behind fused_lm_head_logprob and covered by new tests.
> 
> **Overview**
> Fixes **`trainer.fused_lm_head_logprob`** for Nemotron-style Megatron **`HybridModel`**, which does not accept `output_processor` on `forward` (unlike `GPTModel`).
> 
> Fused LM-head logic moves into **`fused_lm_head.py`**: **`call_model_with_fused_lm_head`** routes native models through the existing hook and, on the last pipeline stage, temporarily turns off **`post_process`** so **`HybridModel`** returns decoder hidden states, then runs the shared **`fused_lm_head_output_processor`** (LM-head weight in context, no full logits). Intermediate PP stages skip the processor; **`post_process`** is restored in a **`finally`** block. The processor now mirrors skipped **`ColumnParallelLinear`** behavior (sequence-parallel gather with **`tp_group`**, **`copy_to_tensor_model_parallel_region`** when **`allreduce_dgrad`**).
> 
> **`MegatronModelWrapper`** inference and training forwards call the helper instead of passing kwargs directly. MXFP8 output projection, MuP, and HybridModel MTP stay explicit **`NotImplementedError`**s. Unit tests cover Hybrid/GPT paths, PP staging, unsupported configs, and TP grad plumbing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 888b7360cd8d2774219d9c359c66b985d20a4d5e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/skyrl/backends/skyrl_train/distributed/megatron/fused_lm_head.py
+++ b/skyrl/backends/skyrl_train/distributed/megatron/fused_lm_head.py
@@ -54,6 +54,7 @@ def call_model_with_fused_lm_head(
         raise NotImplementedError("fused_lm_head_logprob does not support MXFP8 output projection")
     if unwrapped_model.config.use_mup:
         raise NotImplementedError("fused_lm_head_logprob does not support MuP logit scaling")
+    # Standard GPTModel objects directly take `output_processor`, just call the model like normal
     if "output_processor" in inspect.signature(unwrapped_model.forward).parameters:
         return model(
             *args,
@@ -69,10 +70,12 @@ def call_model_with_fused_lm_head(
     if not unwrapped_model.post_process:
         return model(*args, **kwargs)
 
+    # Manually mark output layer as not offloadable; this is normally done by Megatron post-processing, but we disable that
     if unwrapped_model.config.fine_grained_activation_offloading:
         for parameter in unwrapped_model.output_layer.parameters():
             off_interface.mark_not_offloadable(parameter)
 
+    # Disable post-processing since we manually handle the LM head output downstream
     unwrapped_model.post_process = False
     try:
         hidden_states = model(*args, **kwargs)
--- a/skyrl/backends/skyrl_train/distributed/megatron/fused_lm_head.py
+++ b/skyrl/backends/skyrl_train/distributed/megatron/fused_lm_head.py
@@ -1,0 +1,92 @@
+import inspect
+from collections.abc import Callable
+from typing import Any
+
+from torch import Tensor
+
+
+def fused_lm_head_output_processor(
+    *,
+    hidden_states: Tensor,
+    output_layer: Any,
+    output_weight: Tensor | None = None,
+    context: dict[str, Any] | None = None,
+    **_: Any,
+) -> Tensor:
+    """Return hidden states and retain the LM-head weight for fused log-probs."""
+    if context is not None:
+        context["lm_head_weight"] = output_weight if output_weight is not None else output_layer.weight
+
+    if output_layer.sequence_parallel:
+        from megatron.core.tensor_parallel import gather_from_sequence_parallel_region
+
+        hidden_states = gather_from_sequence_parallel_region(
+            hidden_states,
+            tensor_parallel_output_grad=True,
+            group=output_layer.tp_group,
+        )
+    elif output_layer.allreduce_dgrad:
+        from megatron.core.tensor_parallel import copy_to_tensor_model_parallel_region
+
+        hidden_states = copy_to_tensor_model_parallel_region(hidden_states, group=output_layer.tp_group)
+
+    return hidden_states.transpose(0, 1).contiguous()
+
+
+def call_model_with_fused_lm_head(
+    model: Any,
+    *args: Any,
+    output_processor: Callable[..., Any],
+    output_processor_context: dict[str, Any],
+    **kwargs: Any,
+) -> Any:
+    """Run GPT-style or HybridModel forwards without materializing logits."""
+    # Megatron is optional outside the Megatron backend, so keep these imports lazy.
+    from megatron.core.fp8_utils import is_mxfp8_output_proj_active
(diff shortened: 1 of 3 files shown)


skyrl/backends/skyrl_train/distributed/megatron/fused_lm_head.py
```diff
@@ -0,0 +1,95 @@
...
+
+        hidden_states = gather_from_sequence_parallel_region(
+            hidden_states,
+            tensor_parallel_output_grad=True,
+            group=output_layer.tp_group,
+        )
+    elif output_layer.allreduce_dgrad:
+        from megatron.core.tensor_parallel import copy_to_tensor_model_parallel_region
+
+        hidden_states = copy_to_tensor_model_parallel_region(hidden_states, group=output_layer.tp_group)
```
avigyabb, line 34: I think that when TP>1 and `sequence_parallel=False` we train on unreduced gradients.

My understanding is that in [megatron's code](https://github.com/NVIDIA/Megatron-LM/blob/ebfa13852f3d6bdb61c8571d68ccab13cb239089/megatron/core/tensor_parallel/mappings.py#L31) it assumes the `input_` is contiguous. In the backward pass since we transpose (`hidden_states.transpose(0, 1).contiguous()`) we pass in a transposed (non-contiguous) input into `copy_to_tensor_model_parallel_region` (this is not a problem during the forward pass). As a result, the megatron function returns the original tensor and not the reduced one when running `torch.distributed.all_reduce(input_.contiguous(), group=group)`. I think if we transpose before the collective it should fix?
```
hidden_states = hidden_states.transpose(0, 1).contiguous()
if output_layer.allreduce_dgrad:
    hidden_states = copy_to_tensor_model_parallel_region(hidden_states, group=output_layer.tp_group)
return hidden_states
```
Btw, I think the [latest commit to this megatron path](https://github.com/NVIDIA/Megatron-LM/commit/60f338a9cd328e71b272c42d6723ebeabb51fef3) might fix this LOL, but I don't think the latest changes are reflected in SkyRL's pinned version.

avigyabb: Ah, actually I think we hard code sequence_parallel=True if TP>1: https://github.com/NovaSky-AI/SkyRL/blob/23a210970cb86fc6fb247fd93f31103e3f2923e5/skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py#L471

dyurk-lila: Got it, made a fix so this doesn't error in the future if we stop enforcing `sequence_parallel`

erictang000 (approved):
sorry for the delay, this lgtm!

PR #1890, opened 2026-07-13
Add bounded-memory training and colocation controls

YJHMITWEB:
# Bounded-Memory Megatron Training and Colocation

## Summary

This PR adds controls that bound temporary training memory and make colocated Megatron offload behavior complete.

## Changes

| Change | Explanation |
| --- | --- |
| CPU-resident policy microbatches | Keeps the DP-sharded policy batch on CPU and transfers one nested microbatch to CUDA immediately before its forward step. |
| Bounded vocabulary entropy | Chunks Megatron vocabulary entropy by an explicit size or an automatically calculated memory budget, and skips fully masked chunks. |

CPU-resident policy microbatches is opt-in; automatic vocabulary-entropy chunk sizing, which defaults to a 512 MiB temporary-memory budget.

## Memory Validation

| Variant | Memory Spike | Outcome |
| --- | ---: | --- |
| Earlier configuration | 9.3 GiB | OOM easily as the context/vocab size gets larger |
| Bounded-memory configuration | 0.5 GiB | All memory spikes are under control |

### DIFF
--- a/skyrl/backends/skyrl_train/distributed/megatron/model_utils.py
+++ b/skyrl/backends/skyrl_train/distributed/megatron/model_utils.py
@@ -1410,7 +1410,7 @@ def vocab_parallel_entropy_weighted_sum(
     """Compute ``sum(entropy * weights)`` with bounded temporary memory."""
     resolved_chunk_size = _resolve_vocab_entropy_chunk_size(vocab_parallel_logits, chunk_size, chunk_memory_mb)
     if resolved_chunk_size is None:
-        entropy_tokens = _VocabParallelEntropy.apply(vocab_parallel_logits).squeeze(0)
+        entropy_tokens = _VocabParallelEntropy.apply(vocab_parallel_logits)
         return (entropy_tokens * weights).sum()
 
     # Keep an autograd edge even when every chunk is masked out.
@@ -1421,6 +1421,6 @@ def vocab_parallel_entropy_weighted_sum(
         weight_chunk = weights[start:end]
         if torch.count_nonzero(weight_chunk).item() == 0:
             continue
-        entropy_chunk = _VocabParallelEntropy.apply(vocab_parallel_logits[..., start:end, :]).squeeze(0)
+        entropy_chunk = _VocabParallelEntropy.apply(vocab_parallel_logits[..., start:end, :])
         local_entropy_sum = local_entropy_sum + (entropy_chunk * weight_chunk).sum()
     return local_entropy_sum
--- a/skyrl/backends/skyrl_train/distributed/megatron/model_utils.py
+++ b/skyrl/backends/skyrl_train/distributed/megatron/model_utils.py
@@ -17,6 +17,7 @@
 # limitations under the License.
 
 import warnings
+from math import prod
 from typing import Any, Optional
 
 import megatron.core.parallel_state as mpu
@@ -1367,7 +1368,10 @@ def _resolve_vocab_entropy_chunk_size(
         return chunk_size if chunk_size < seq_len else None
 
     budget_bytes = int(chunk_memory_mb) * 1024 * 1024
-    bytes_per_token = int(vocab_parallel_logits.shape[-1]) * vocab_parallel_logits.element_size() * peak_factor
+    leading_elements = prod(vocab_parallel_logits.shape[:-2])
+    bytes_per_token = (
+        leading_elements * int(vocab_parallel_logits.shape[-1]) * vocab_parallel_logits.element_size() * peak_factor
+    )
     if bytes_per_token <= 0:
         return None
 
--- a/skyrl/backends/skyrl_train/distributed/megatron/model_utils.py
+++ b/skyrl/backends/skyrl_train/distributed/megatron/model_utils.py
@@ -1001,6 +1001,8 @@ def vocab_parallel_entropy_packed_sequences(
     loss_mask: Optional[torch.Tensor],
     cp_group: Optional[torch.distributed.ProcessGroup],
     sub_seq_lengths: Optional[list[list[int]]] = None,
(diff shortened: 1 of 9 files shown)


skyrl/backends/skyrl_train/workers/worker.py
```diff
@@ -243,9 +245,86 @@ def init_model(self, *args, **kwargs):
...
+                gpu_uuid = raw_gpu_uuid.decode("ascii")
+            else:
+                gpu_uuid = str(raw_gpu_uuid) if raw_gpu_uuid else None
+            pynvml.nvmlInit()
+            try:
+                handle = (
+                    pynvml.nvmlDeviceGetHandleByUUID(gpu_uuid)
+                    if gpu_uuid
+                    else pynvml.nvmlDeviceGetHandleByIndex(device)
+                )
```
jinghanyao1-hub: fixed

skyrl/backends/skyrl_train/distributed/megatron/model_utils.py
```diff
@@ -1342,4 +1389,38 @@ def vocab_parallel_entropy(vocab_parallel_logits: torch.Tensor) -> torch.Tensor:
...
+
+    # Keep an autograd edge even when every chunk is masked out.
+    local_entropy_sum = vocab_parallel_logits[..., :0, :].sum()
+    seq_len = int(vocab_parallel_logits.shape[-2])
+    for start in range(0, seq_len, resolved_chunk_size):
+        end = min(start + resolved_chunk_size, seq_len)
+        weight_chunk = weights[start:end]
+        if torch.count_nonzero(weight_chunk).item() == 0:
+            continue
+        entropy_chunk = _VocabParallelEntropy.apply(vocab_parallel_logits[..., start:end, :]).squeeze(0)
```
jinghanyao1-hub: fixed

skyrl/backends/skyrl_train/workers/megatron/megatron_model_wrapper.py
```diff
@@ -325,6 +345,8 @@ def collection_func(logits, data):
 
         def forward_step(batch_iter, model):
             batch = next(batch_iter)
+            if self.cpu_resident_microbatches:
```
erictang000: can we get rid of this config flag and just make this the default behavior? (i.e. only put microbatches onto device when needed?) this seems reasonable to me, overhead probably low anyway

skyrl/backends/skyrl_train/workers/worker.py
```diff
@@ -567,6 +568,9 @@ def __init__(
         self.colocate_all = colocate_all
         self.sequence_parallel_size = sequence_parallel_size
         self.record_memory = record_memory
+        self._pg = pg
+        self._num_gpus_per_actor = num_gpus_per_actor
```
erictang000: i think these don't get used anywhere else? can remove the above 2 lines?

skyrl/backends/skyrl_train/distributed/megatron/model_utils.py
```diff
@@ -1342,4 +1393,38 @@ def vocab_parallel_entropy(vocab_parallel_logits: torch.Tensor) -> torch.Tensor:
...
+
+    entropy_chunks = []
+    seq_len = int(vocab_parallel_logits.shape[-2])
+    for start in range(0, seq_len, resolved_chunk_size):
+        end = min(start + resolved_chunk_size, seq_len)
+        entropy_chunks.append(_VocabParallelEntropy.apply(vocab_parallel_logits[..., start:end, :]))
+    return torch.cat(entropy_chunks, dim=-1)
+
+
+def vocab_parallel_entropy_weighted_sum(
```
erictang000, line 1408: just to confirm these changes are stable, can you run a basic gsm8k example before and after this PR, and link it here? Just want to check there is no significant speed regression from chunking entropy calculation, and that the entropy metric is exactly matching before

erictang000 (commented):
looks good, just some minor comments

skyrl/backends/skyrl_train/distributed/megatron/model_utils.py
jinghanyao1-hub: We have a unit test on this part: `tests/backends/skyrl_train/distributed/test_vocab_entropy_chunking.py`, this verifies that the chunked CE is numerically identical. For the overhead, the following table shows:

| `chunk_memory_mb` | tokens/chunk | #chunks | exact | op time | vs unchunked | 
| ---: | ---: | ---: | :---: | ---: | ---: | 
| off (unchunked) | — | — | — | 6.3 ms | 1.0× | 
| 4096 | 4096 | 2 | ✓ | 6.4 ms | 1.0× | 
| 2048 | 2048 | 4 | ✓ | 7.5 ms | 1.2× | 
| 1024 | 1024 | 8 | ✓ | 14.9 ms | 2.4× | 
| 512 *(current default)* | 512 | 16 | ✓ | 29.7 ms | 4.7× | 
| 256 | 256 | 32 | ✓ | 59.6 ms | 9.4× | 
| 128 | 128 | 64 | ✓ | 118.6 ms | 18.8× | 

The slowdown is expected, but since the overall step time here is at most hundreds of ms, the impact of chunking is almost negligible.

skyrl/backends/skyrl_train/workers/worker.py
jinghanyao1-hub: fixed

skyrl/backends/skyrl_train/workers/megatron/megatron_model_wrapper.py
jinghanyao1-hub: fixed

erictang000 (approved):
nice, looking good thanks!

PR #1898, opened 2026-07-14
FP8 RL Training and Rollout Weight Synchronization

YJHMITWEB:
This PR adds end-to-end FP8 training and rollout support. It covers persistent Megatron FP8 parameters, serialized blockwise weight synchronization, mixed-dtype transport, and FP8 vLLM execution. 

## Changes

| Change | Explanation |
| --- | --- |
| Serialized blockwise FP8 sync | `fp8_weight_sync_mode=serialized_blockwise` converts eligible weights to E4M3 and sends matching FP32 `weight_scale_inv` tensors in vLLM checkpoint format. |
| FP8 vLLM execution | Builds the rollout model with `quantization=fp8`, `load_format=dummy`, and a matching 128 x 128 block quantization configuration; the first full sync replaces the dummy weights. |
| Qwen3.5 mapping | Handles dense attention and MLP linears, GDN projections, shared experts, and routed MoE experts while keeping incompatible GDN and vision projections in BF16. |
| Mixed-dtype transport | Preserves mixed-dtype logical chunks over NCCL and partitions CUDA IPC buffers by their actual tensor dtype. |
| Persistent FP8 parameters | Keeps eligible Megatron primary parameters in E4M3, reducing the required HBM in Megatron. |
| Exact optimizer initialization | Initializes FP32 masters and hybrid CPU copies from unquantized checkpoint shards before the first optimizer step. |
| FP8 parameter gather | Exposes `fp8_param_gather`, which refreshes persistent FP8 parameters from updated FP32 masters. |
| FP8-safe packed sequences | Uses one TP/CP-aware alignment rule in collation, packed-sequence preprocessing, and MoE replay. |

## Scale Support

Without persistent `fp8_param`, serialized FP8 synchronization supports both FP32 and power-of-two block scales. Hopper supports both modes and defaults to FP32 scales; the current Blackwell path requires power-of-two scales.

Persistent `fp8_param` is supported only with FP32 scales. The main reason is that at a `1e-6` learning rate, 99.97-99.999% of sampled FP32 master values changed per update, no power-of-two block scale changed. The configuration therefore rejects power-of-two scales with `fp8_param`. Because Blackwell currently requires power-of-two scales, persistent FP8 parameters are not enabled there.

## Configuration

`<role>` denotes `policy` or `ref`. All FP8 features are disabled by default, so existing BF16 configurations retain their current behavior.

| Args| Default | Validated FP8 value | Behavior |
| --- | ---: | ---: | --- |
| `generator.inference_engine.fp8_weight_sync_mode` | `null` | `serialized_blockwise` | Sends rollout weights as E4M3 tensors with matching FP32 block scales. |
| `generator.inference_engine.engine_init_kwargs.quantization` | unset | `fp8` | Enables FP8 vLLM execution. |
| `trainer.<role>.megatron_config.transformer_config_kwargs.fp8` | unset | `e4m3` | Enables Transformer Engine FP8 compute for that Megatron model. |
| `trainer.<role>.megatron_config.transformer_config_kwargs.fp8_recipe` | unset | `blockwise` | Selects the Transformer Engine blockwise FP8 recipe. |
| `trainer.<role>.megatron_config.transformer_config_kwargs.fp8_param` | unset (`false`) | `true` | Stores Megatron primary parameters persistently in E4M3. |
| `trainer.policy.megatron_config.ddp_config.fp8_param_gather` | `false` | `true` with policy `fp8_param` | Refreshes policy FP8 parameters from updated FP32 optimizer masters after each step. |

| Environment variable | Runtime default | Validated value (Hopper) | Behavior |
| --- | ---: | ---: | --- |
| `NVTE_FP8_BLOCK_SCALING_FP32_SCALES` | `1` when FP8 is active | `1` | `1` uses FP32 block scales; `0` uses power-of-two scales. Persistent `fp8_param` requires `1`. |
| `NVTE_FP8_BLOCK_AMAX_EPSILON` | `0` | `1e-4` | Applies the same non-negative block-amax floor to Transformer Engine and serialized rollout quantization. |
| `VLLM_USE_DEEP_GEMM_E8M0` | `0` with serialized FP32 scales | `0` | Prevents vLLM from converting FP32 scales to E8M0 power-of-two scales. |

`serialized_blockwise` requires Megatron and full-weight synchronization; adapter-only Megatron LoRA is rejected. On Blackwell, use power-of-two scales and keep `fp8_param` disabled.

### vLLM Rollout Throughput

On `Qwen/Qwen3.5-9B-Base`, NVIDIA H100 80GB, TP1, vLLM 0.23.0, using DAPO recipe dataset.

| Concurrent requests | BF16 output tok/s | FP8 output tok/s | Gain |
| ---: | ---: | ---: | ---: |
| 1 | 140.5 | 198.6 | +41.4% |
| 16 | 1,940.3 | 2,599.4 | +34.0% |
| 32 | 3,418.9 | 4,361.0 | +27.6% |
| 128 | 7,897.7 | 8,989.2 | +13.8% |
| 256 | 9,223.3 | 10,791.0 | +17.0% |

This gain comes from FP8 vLLM execution and does not require persistent FP8 Megatron parameters.

### End-to-End RL Step Timing

The end-to-end comparison uses `Qwen/Qwen3.5-9B-Base` on eight NVIDIA H100 80GB GPUs with Megatron TP2, eight TP1 vLLM engines, and a training batch size of 32.

| Phase | BF16 mean (s) | FP8 mean (s) | FP8 time change | Speedup |
| --- | ---: | ---: | ---: | ---: |
| Rollout generation | 96.51 | 72.25 | -25.1% | 1.34x |
| Log-probability and reward forward passes | 109.87 | 106.31 | -3.2% | 1.03x |
| Policy training | 46.50 | 50.47 | +8.5% | 0.92x |
| Policy-to-vLLM weight sync | 10.03 | 10.26 | +2.3% | 0.98x |
| **End-to-end step** | **268.90** | **245.49** | **-8.7%** | **1.10x** |

### Persistent FP8 Policy HBM

| Megatron primary storage | HBM/GPU | Change at highest sample |
| --- | ---: | ---: |
| BF16 | 59,676-63,638 MiB | Baseline |
| Persistent E4M3 | 56,850-57,810 MiB | -5,828 MiB (-5.69 GiB, -9.2%) |

## Quality Checks
The following WANDB runs log the metrics. All FP8 runs aligns well with the BF16 baseline.
Qwen3.5-4B runs: https://wandb.ai/sky-posttraining-uc-berkeley/qwen35_4b_dapo_mainclean_fp8align_fastbatch_bf16_fp8_20260630?nw=nwuserjinghanyao1

Qwen3.5-9B runs: https://wandb.ai/sky-posttraining-uc-berkeley/qwen35_9b_h100_tp2_bf16_fp8_prready_20260709?nw=nwuserjinghanyao1

Qwen3.5-35B-A3B runs: https://wandb.ai/sky-posttraining-uc-berkeley/qwen35_35b_a3b_fp8_amaxeps_20260705?nw=nwuserjinghanyao1

## Reproduce key configs
H100:
```
export NVTE_FP8_BLOCK_SCALING_FP32_SCALES=1
export NVTE_FP8_BLOCK_AMAX_EPSILON=1e-4
export VLLM_USE_DEEP_GEMM_E8M0=0

trainer.algorithm.kl_loss_coef=0.00 trainer.algorithm.use_kl_in_reward=false \
trainer.policy.model.path=Qwen/Qwen3.5-9B-trainer.placement.colocate_all=true \
trainer.strategy=megatron \
trainer.placement.policy_num_nodes=1 \
trainer.placement.policy_num_gpus_per_node=8 \
trainer.placement.ref_num_gpus_per_node=8 \
trainer.policy.megatron_config.tensor_model_parallel_size=2 \
trainer.policy.megatron_config.pipeline_model_parallel_size=1 \
trainer.policy.megatron_config.context_parallel_size=1 \
trainer.policy.megatron_config.expert_model_parallel_size=1 \
trainer.policy.megatron_config.expert_tensor_parallel_size=1 \
trainer.ref.megatron_config.tensor_model_parallel_size=2 \
trainer.ref.megatron_config.pipeline_model_parallel_size=1 \
trainer.ref.megatron_config.context_parallel_size=1 \
trainer.ref.megatron_config.expert_model_parallel_size=1 \
trainer.ref.megatron_config.expert_tensor_parallel_size=1 \
trainer.epochs=20 trainer.max_training_steps=400 \
trainer.policy.megatron_config.ddp_config.overlap_grad_reduce=false \
trainer.ref.megatron_config.ddp_config.overlap_grad_reduce=false \
trainer.policy.megatron_config.transformer_config_kwargs.fp8_param=true \
trainer.ref.megatron_config.transformer_config_kwargs.fp8_param=true \
trainer.policy.megatron_config.ddp_config.fp8_param_gather=true \
trainer.eval_batch_size=512 \
trainer.train_batch_size=32 \
trainer.policy_mini_batch_size=32 \
trainer.micro_forward_batch_size_per_gpu=1 \
trainer.micro_train_batch_size_per_gpu=1 \
trainer.max_prompt_length=2048 \
trainer.policy.optimizer_config.lr=1e-6 \
trainer.policy.megatron_config.transformer_config_kwargs.fp8=e4m3 \
trainer.ref.megatron_config.transformer_config_kwargs.fp8=e4m3 \
trainer.policy.megatron_config.transformer_config_kwargs.fp8_recipe=blockwise \
trainer.ref.megatron_config.transformer_config_kwargs.fp8_recipe=blockwise \
trainer.policy.megatron_config.transformer_config_kwargs.fp8_amax_compute_algo=most_recent \
trainer.ref.megatron_config.transformer_config_kwargs.fp8_amax_compute_algo=most_recent \
trainer.policy.megatron_config.transformer_config_kwargs.tp_only_amax_red=false \
trainer.ref.megatron_config.transformer_config_kwargs.tp_only_amax_red=false \
trainer.micro_train_batch_size_per_gpu=2 \
trainer.micro_forward_batch_size_per_gpu=2 
generator.sampling_params.max_generate_length=8192 \
generator.n_samples_per_prompt=8 \
generator.eval_n_samples_per_prompt=16 \
generator.inference_engine.max_num_seqs=32 \
generator.eval_sampling_params.max_generate_length=8192 trainer.algorithm.use_kl_loss=false \
generator.inference_engine.num_engines=8 \
generator.inference_engine.tensor_parallel_size=1 \
generator.inference_engine.data_parallel_size=1 \
generator.inference_engine.expert_parallel_size=1 \
generator.inference_engine.engine_init_kwargs.quantization=fp8 \
generator.inference_engine.fp8_weight_sync_mode=serialized_blockwise \
generator.inference_engine.engine_init_kwargs.kv_cache_dtype=fp8_e4m3 \
```

B200 (only show the difference):
```
export FLA_TILELANG=0                          # B200/sm_100: TileLang GDN kernels crash; use triton
export NVTE_FP8_BLOCK_SCALING_FP32_SCALES=0    # B200: pow-2 (MXFP8) block scales — mandatory
export NVTE_FP8_BLOCK_AMAX_EPSILON=1e-4 

trainer.policy.model.path=Qwen/Qwen3.5-35B-A3B \
trainer.policy.megatron_config.tensor_model_parallel_size=1 \
trainer.policy.megatron_config.expert_model_parallel_size=8 \
trainer.policy.megatron_config.expert_tensor_parallel_size=1 \
trainer.ref.megatron_config.tensor_model_parallel_size=1 \
trainer.ref.megatron_config.expert_model_parallel_size=8 \
trainer.policy.megatron_config.transformer_config_kwargs.fp8_param=false \
trainer.ref.megatron_config.transformer_config_kwargs.fp8_param=false\
```

---

> [!NOTE]
> **High Risk**
> Changes weight sync, vLLM loading, FP8 packing alignment, and optimizer initialization on the critical train↔rollout path; misconfiguration is partly guarded by validation but numerical or MoE-loader mismatches could still break rollouts silently without tests on your exact stack.
> 
> **Overview**
> Adds **end-to-end FP8** for Megatron RL: TE FP8 GEMMs, optional **persistent `fp8_param`**, and **`fp8_weight_sync_mode=blockwise`** so rollouts receive trainer FP8 payloads and block scales instead of re-quantizing BF16.
> 
> **Serialized FP8 weight sync** introduces a `weight_sync/fp8` stack (blockwise quantize, vLLM checkpoint wire format, per-model specs starting with Qwen3.5). The Megatron extractor expands weights into mixed fp8/fp32/bf16 chunks with **`derives_metadata_from_chunks`**; **NCCL** sends one update per chunk, **CUDA IPC** splits packs by dtype, and **delta / sharded_rdt** are rejected at validation and send time. vLLM startup is auto-configured (`quantization=fp8`, `load_format=dummy`, quantization_config + ignored layers); **`NewInferenceWorkerWrap`** loads batched MoE FP8 tensors via FusedMoE.
> 
> **Training correctness**: `fp8_recipe` (blockwise vs **mxfp8**) drives TP/CP/FP8 **sequence alignment** in packing, collators, and workers; **`fp8_recipe=auto`** resolves on workers (Hopper blockwise, Blackwell mxfp8) with validation for mxfp8 + no `fp8_param`. **FP8 param** uses deferred bridge load, exact FP32 masters from unquantized shards, and requires **`fp8_param_gather`**. Driver validation and Ray runtime env pin **`NVTE_FP8_BLOCK_SCALING_FP32_SCALES`** / **`VLLM_USE_DEEP_GEMM_E8M0`** to match the scale contract.
> 
> Docs and **`examples/train/fp8/`** scripts cover Hopper blockwise (with optional fp8_param) and Blackwell MXFP8 DAPO runs for Qwen3.5.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c127af6887eb7056c045aa876c4eaf76fe7efce6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


erictang000 (commented):
Reviewed as a distributed-systems pass (correctness / concurrency / distributed-sync / numerical-stability / memory / perf). FP8 quantization numerics are sound (fuzz-verified), the lossy-master-seed window is correctly closed, and the BF16 path is preserved. Four issues below: one crash-on-valid-input (High) and three latent distributed hazards (Medium) that surface on specific hardware/config combos. Line-anchored comments follow.

Also worth confirming against the pinned Megatron/Bridge (couldn't verify here): (1) `fp8_param.py:152` calls `_copy_model_params_to_main_params(state_dict=...)` but stock Megatron's method takes no args and copies from live (quantized) params — assumes a patched Bridge version; (2) the hybrid path re-seeds `param_to_fp32_param`/CPU copies but not `optimizer.state[param]['master_param']`, which `megatron_strategy.py`'s own patch treats as authoritative; (3) the 3D fused-MoE `weight_loader` fast path in `new_inference_worker_wrap.py` assumes vLLM accepts a batched 3D tensor in one call.

skyrl/train/dataset/collators.py
```diff
@@ -174,14 +178,18 @@ def __call__(self, examples: list, batch_size: int) -> TrainingInputBatch:
         # same number of micro-batches. Forcing the global bin count to a
         # multiple of ``dp_size`` makes the per-DP-rank bin count (and thus
         # ``num_microbatches``) identical across ranks.
+        # Pack aligned footprints so per-sequence padding cannot exceed the row
+        # budget. Allow at least one alignment unit per bin.
+        packing_lengths = [_round_up(length, align_size) for length in seq_lengths]
+        packing_capacity = max(bin_capacity, align_size)
```
erictang000: **[High confidence] Collation crashes for valid inputs when `max_tokens_per_microbatch % align_size != 0`.**

`packing_capacity = max(bin_capacity, align_size)` is not rounded up to an `align_size` multiple, but the items fed to `packer.pack` are now aligned footprints `round_up(len, align_size)`. `FirstFitDecreasing._validate_sequence_lengths` (bin_packing.py:66-67) raises `ValueError` for any item `> packing_capacity`. So a sequence with `align_size < len ≤ bin_capacity` rounds up past capacity and aborts the whole collate. Pre-PR the packer got raw lengths, so this never happened.

Reproduces on **BF16** too (not FP8-only): `tp=2, cp=3, fp8=False` → `align_size=12`; `max_tokens_per_microbatch=8192` (not a multiple of 12); a length-8190 seq → `round_up(8190,12)=8196 > 8192` → crash. FP8 example: `tp=4, cp=1` → `align=512`, budget 768, len 600 → 1024 > 768. Masked when the budget is a multiple of `align_size` (why tests miss it).

**Fix:** `packing_capacity = max(_round_up(bin_capacity, align_size), align_size)`. Since `round_up(L, a) ≤ round_up(bin_capacity, a)` for all `L ≤ bin_capacity`, this restores the invariant; cost is ≤ `align_size-1` pad tokens per row, which per-sequence alignment already requires.

skyrl/backends/skyrl_train/workers/megatron/quantization/fp8_param.py
```diff
@@ -0,0 +1,148 @@
...
+    optimizers = getattr(optimizer, "chained_optimizers", None)
+    if optimizers is None:
+        optimizers = [optimizer]
+
+    initialized = 0
+    with torch.no_grad():
+        for megatron_optimizer in optimizers:
+            if _uses_hybrid_device_optimizer(megatron_optimizer):
+                copied = _copy_model_shards_to_main_params(megatron_optimizer, state_dict)
+                if copied == 0:
```
erictang000, line 135: **[Medium] Divergent `RuntimeError` on `copied == 0` can hang training via NCCL timeout.**

This check runs **per chained sub-optimizer** (`optimizers = optimizer.chained_optimizers or [optimizer]`), and `copied` counts main-param shards **local to this rank**. A chained sub-optimizer with zero local shards on a given rank is a legitimate outcome of DistributedOptimizer DP sharding — documented in this repo at `mtp/grad_clip.py:56` ("May legitimately be EMPTY on this rank...").

Scenario: FP8 + CPU-offload (HybridDeviceOptimizer) + a `ChainedOptimizer` whose member covers fewer params than the DP world size (e.g. MoE dense/expert split under large DP). On the empty-shard rank(s) this raises while other ranks proceed to the next collective → mismatch → hang. The non-hybrid branch has no such raise.

**Fix:** all-reduce `copied` across the DP group and raise only if the *global* sum is 0, or downgrade to an aggregated once-per-run warning.

skyrl/backends/skyrl_train/weight_sync/cuda_ipc_strategy.py
```diff
@@ -185,77 +190,91 @@ async def _send_chunks_vllm_native(
...
-                await self._inference_client.update_weights_ipc(chunk_update_info)
-
-            # Keep packed_tensor alive past the barrier so the receiver's
-            # rebuilt view has valid backing storage while it copies into
-            # the model. Post-barrier drops the local ref safely.
-            torch.distributed.barrier()
-            torch.cuda.ipc_collect()
-            torch.cuda.synchronize()
+        for logical_chunk in chunks:
+            for chunk in iter_single_dtype_chunks(logical_chunk):
```
erictang000, line 200: **[Medium] Per-chunk IPC collective count is now dtype-content-dependent → deadlock if ranks' dtype composition diverges.**

`iter_single_dtype_chunks` yields one sub-chunk per distinct dtype (first-seen order), and `_send_single_dtype_chunk_vllm_native` issues an unconditional `all_gather_object` + 2 `barrier` per sub-chunk. So the per-logical-chunk collective count now equals the number of distinct dtypes in that chunk **on that rank**. Pre-PR this count was purely structural (bucket layout, identical across ranks); it is now content-dependent. If any rank's logical chunk has a different dtype set/order (e.g. weight+scale on one rank vs weight-only on another), ranks call `all_gather_object` a different number of times and hang.

This relies on `extract_weights` (out of scope) producing perfectly rank-symmetric dtype composition; the transport takes it on faith. The intra-sub-chunk `if any(tensor.dtype != dtype): raise` guard is dead code and does not protect against cross-rank divergence.

**Fix:** before the loop, `all_gather_object` each rank's per-logical-chunk sub-chunk dtype list and assert equality (fail fast instead of hanging), or drive the partition from a rank-0 broadcast. At minimum document the invariant.

skyrl/train/utils/utils.py
```diff
@@ -805,6 +836,54 @@ def prepare_runtime_environment(cfg: SkyRLTrainConfig) -> dict[str, str]:
...
+
+        if scale_mode == "0" and (policy_fp8_param or ref_fp8_param):
+            raise ValueError(
+                "Persistent fp8_param requires FP32 block scales. Blackwell only supports "
+                "power-of-2 block scales, so use fp8_param=false on Blackwell."
+            )
+
+        if fp8_contract_enabled:
+            fp8_env_defaults["NVTE_FP8_BLOCK_SCALING_FP32_SCALES"] = scale_mode
+        if serialized_fp8 and scale_mode == "1":
```
erictang000, line 982: **[Medium] Power-of-two (Blackwell) serialized path neither defaults nor validates `VLLM_USE_DEEP_GEMM_E8M0` — asymmetric with the FP32 path → silent numeric mismatch.**

This `scale_mode == "1"` (FP32) branch both rejects a conflicting `VLLM_USE_DEEP_GEMM_E8M0 != "0"` and pins the default to `"0"`, so TE and vLLM agree. The `scale_mode == "0"` power-of-two branch (the documented Blackwell case) does neither — the var is only forwarded if already present in the driver env.

Scenario: Blackwell run sets `NVTE_FP8_BLOCK_SCALING_FP32_SCALES=0` + `serialized_blockwise`, `fp8_param=false` (passes all config checks) but forgets `VLLM_USE_DEEP_GEMM_E8M0=1`. TE emits E8M0 power-of-two scales; vLLM interprets them without E8M0 → mis-dequantized weights → silently degraded rollouts, no error. The symmetric FP32 mistake *is* caught.

**Fix:** handle `scale_mode == "0"` symmetrically — default `VLLM_USE_DEEP_GEMM_E8M0=1` and reject a conflicting `!= "1"`.

erictang000:
@jinghanyao1-hub ai code review above, you can ignore for now, still taking a deeper pass, will resolve the above if i find it's not reasonable

skyrl/train/dataset/collators.py
erictang000: max_tokens_per_microbatch is not compatible with this packing anyways.

erictang000:
high level comment on file structure:

can we reorganize to create a new folder `workers/megatron/quantization/` that contains the `_fp8_block_amax_epsilon_patch.py` (maybe rename to just `amax_epsilon_patch.py`, and `fp8_param.py`? We can also move `fake_int4_qat.py` into that folder. So something like:

```
workers/megatron/quantization/
    __init__.py
    amax_epsilon_patch.py    
    fp8_param.py      
    fake_int4_qat.py
```

and then can we split the `serialized_fp8.py` file into a separate folder/components inside `weight_sync`? The structure should be something like this

```
weight_sync/fp8/
    __init__.py            
    quantize.py           # blockwise_cast_to_fp8, batched_blockwise_cast_to_fp8, scale helpers
    vllm_format.py        # SERIALIZED_BLOCKWISE_FP8, wire prefix, scale_name_for_weight,
                          #   get_serialized_fp8_quantization_config, iter_serialized_fp8_tensors
    models/qwen35.py      # _QWEN35_* tables, is_qwen35_config, ignored layers, MoE spec
```

erictang000:
another high level comment:

can we add a gpu ci test in the style of `test_megatron_models` but for full fp8? This will be added to H100 CI.

Essentially this test: https://github.com/NovaSky-AI/SkyRL/blob/da778a52de5ba5a6153054bcb186097a0e7b96bd/tests/backends/skyrl_train/gpu/gpu_ci/megatron/test_megatron_models.py#L271 but with fp8 quantization on both vllm and megatron.

This will help us maintain full fp8 as a tracked feature across transformer engine/megatron/vllm updates. We want to test vllm/megatron init, weight sync, and logprob diff. We can start with qwen3.5-0.8b and qwen3.5-35B-a3B tests with blockwise quantization only.

skyrl/train/utils/utils.py
erictang000: @YJHMITWEB this seems correct, shouldn't export VLLM_USE_DEEP_GEMM_E8M0 be 1 for blackwell? or is there some reason that it doesn't work w/ blockwise + power of 2?

not seeing `VLLM_USE_DEEP_GEMM_E8M0=1`  in the config that you're using as well so curious about this

skyrl/train/dataset/collators.py
```diff
@@ -132,20 +133,23 @@ def __call__(self, examples: list, batch_size: int) -> TrainingInputBatch:
...
         #   - Sequence Parallelism (auto-on when tp>1) shards along the seq
         #     dim, so each segment must be divisible by ``tp_size``.
         #   - Context Parallelism splits each segment into ``2*cp_size`` equal
         #     load-balanced causal chunks, so each segment must be divisible by
         #     ``2*cp_size``.
-        #   - When FP8 is enabled, Transformer Engine GEMMs require each CP
-        #     rank's local token slab to be 16-aligned; globally this means
-        #     ``16*cp_size``.
+        #   - FP8 requires 16-token local slabs for TP=1. With sequence
+        #     parallelism, TE quantizes all-gather inputs in 128-token blocks,
```
erictang000: this 128 number is specific to blockwise quantization right? can we clarify the wording to reflect that

```suggestion
        #     parallelism, TE with fp8_recipe=blockwise quantizes all-gather inputs in 128-token blocks,
```

skyrl/train/config/config.py
```diff
@@ -744,6 +745,9 @@ class InferenceEngineConfig(BaseConfig):
 
     model_dtype: str = "bfloat16"
     """Should match the dtype used by the inference engine."""
+    fp8_weight_sync_mode: Optional[str] = None
```
erictang000, line 1147: can we update this config in https://github.com/NovaSky-AI/SkyRL/blob/main/docs/content/docs/configuration/config.mdx as well

skyrl/backends/skyrl_train/workers/megatron/quantization/amax_epsilon_patch.py
```diff
@@ -0,0 +1,76 @@
...
+
+from __future__ import annotations
+
+import os
+from dataclasses import replace
+from math import isfinite
+
+from loguru import logger
+
+_ENV_VAR = "NVTE_FP8_BLOCK_AMAX_EPSILON"
```
erictang000: quick question for my understanding: so this is a custom environment variable that we introduced? I can't find it anywhere inside TransformerEngine or elsewhere in OSS

And this is needed because with power of 2 scaling (which is needed on blackwell) + blockwise is not well supported in TE, and doesn't handle CP/TP packing where we can get a block of all 0s 

And the reason existing recipes on TransformerEngine using Float8BlockScaling are able to run without issue is because they all use hopper? And everyone is primarily is using `MXFP8BlockScaling` on blackwell where there is no amax_epsilon concept?

skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
```diff
@@ -439,7 +486,11 @@ def init_configs(
                 "(native GDN thd packing path; vision tower dropped)"
             )
 
-        provider = bridge.to_megatron_provider()
+        # Defer persistent-FP8 checkpoint import until the bridge can expose
```
erictang000, line 535: so the problem here is that:

if fp8 param is enabled, then megatron bridge will not store the unquantized weights?

looking through megatron bridge i'm seeing that on `to_megatron_provider` if `load_weights=true`, then we call `load_weights_hf_to_megatron`, which has the following logic to store the unquantized state dict in the bridge: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/47254ca622d35b8063fcc780b070588836ddf20a/src/megatron/bridge/models/conversion/model_bridge.py#L1072

is there a reason we need to defer it from here to manually calling `self.bridge.load_hf_weights(self.actor_module)`? might be missing something, pls correct if my understanding is wrong? Otherwise i think the `_release_fp8_param_unquantized_state` and `initialize_fp8_param_optimizer_masters` makes sense to me since we do need to pass that state dict to initialize the optimizer state.

skyrl/backends/skyrl_train/weight_sync/serialized_fp8.py
```diff
@@ -0,0 +1,385 @@
...
+
+def iter_serialized_fp8_tensors(
+    name: str,
+    tensor: torch.Tensor,
+    target_dtype: torch.dtype,
+    config: SerializedFp8Config,
+) -> Iterator[tuple[str, torch.Tensor]]:
+    """Yield vLLM checkpoint tensors for one Megatron-exported weight."""
+
+    if batched_moe_expert_spec(name) is not None:
```
erictang000: qq: if we have fp8 param on, are the quantization operations here no-ops? or are we unconditionally re-quantizing? 

if so should we handle fp8 param case separately? (potentially using the interface from this pr:  https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1994 )

skyrl/backends/skyrl_train/weight_sync/serialized_fp8.py
```diff
@@ -0,0 +1,385 @@
+"""Convert Megatron-exported tensors to vLLM's blockwise FP8 checkpoint format."""
```
erictang000: one high level thought: there is this PR in megatron-bridge: https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/1994 that was merged more than 3 months ago that introduces a new function: `build_export_fp8_tasks` - this is designed for the case where we have fp8 params, and allows the user to ship fp8 params + scale metadata directly to vLLM

ideally we could push down the logic in this file for exporting an iterable list of fp8 params to Megatron-Bridge as well and just add support for the case where we have bf16 params but we want to export them to fp8.

skyrl/backends/skyrl_train/weight_sync/serialized_fp8.py
```diff
@@ -0,0 +1,385 @@
...
+    try:
+        epsilon = float(raw_value)
+    except ValueError as exc:
+        raise ValueError(f"NVTE_FP8_BLOCK_AMAX_EPSILON must be a float, got {raw_value!r}") from exc
+    if not isfinite(epsilon) or epsilon < 0:
+        raise ValueError(f"NVTE_FP8_BLOCK_AMAX_EPSILON must be finite and non-negative, got {epsilon}")
+    return epsilon
+
+
+_QWEN35_FP8_WEIGHT_SUFFIXES = (
```
erictang000: mentioned this in the previous comment about re-organizing files, but these mappings should definitely be in a separate file if we have them at all, and the references to fp8 weight suffixes, up/down suffixes, and unquantized suffixes should be made generic behind some abstraction.

Ideally which layers are quantized could be configured, but we can leave that as a follow up of this PR. we should at least design the interface that defines the quantization spec rather than hard coding.

Overall some abstraction like this:

```python
@dataclass(frozen=True)
class ModelFp8Spec:
    matches: Callable[[Any], bool]              # hf_config -> bool (replaces is_qwen35_config)
    should_quantize: Callable[[str, Sequence[int]], bool]   # name/shape -> quantize?
    ignored_layers: Callable[[Any], list[str]]  # hf_config -> vLLM ignored modules (residual shape/vLLM denylist, e.g. vision proj)
    moe_expert_spec: Callable[[str], MoeExpertSpec | None]  # gate/up split + expert batching

@dataclass(frozen=True)
class MoeProjection:
    hf_name: str          # "gate_proj" | "up_proj" | "down_proj"  (what the sender emits on the wire)
    vllm_param: str       # ".experts.w13_weight" | ".experts.w2_weight"  (vLLM fused target, weight only)
    shard_id: str         # "w1" | "w3" | "w2"
    split: Optional[tuple[int, int]] = None   # (chunk_index, num_chunks) along split_dim; None = whole tensor

@dataclass(frozen=True)
class MoeExpertSpec:
    source_suffix: str    # ".mlp.experts.gate_up_proj" | ".mlp.experts.down_proj"  (Bridge-exported batched name)
    split_dim: int        # dim to carve projections from the fused [E, ...] tensor (1 = per-expert output rows)
    projections: tuple[MoeProjection, ...]

QWEN35_MOE_EXPERT_SPECS = (
    MoeExpertSpec(
        source_suffix=".mlp.experts.gate_up_proj",
        split_dim=1,
        projections=(
            MoeProjection("gate_proj", ".experts.w13_weight", "w1", split=(0, 2)),  # first half
            MoeProjection("up_proj",   ".experts.w13_weight", "w3", split=(1, 2)),  # second half
        ),
    ),
    MoeExpertSpec(
        source_suffix=".mlp.experts.down_proj",
        split_dim=1,
        projections=(
            MoeProjection("down_proj", ".experts.w2_weight", "w2"),  # whole tensor
        ),
    ),
)

FP8_MODEL_SPECS: list[ModelFp8Spec] = [QWEN35_FP8_SPEC, ...]

def resolve_fp8_spec(hf_config) -> ModelFp8Spec | None:
    return next((s for s in FP8_MODEL_SPECS if s.matches(hf_config)), None)
```

can you help propose some more modular per model abstraction along these lines? Also seeing if any of mappings from megatron-bridge like here: https://github.com/NVIDIA-NeMo/Megatron-Bridge/blob/b3a5514340af701340e34b9ec506d92688d2a1b9/src/megatron/bridge/models/qwen/qwen35_bridge.py#L187 can be reused would go a long way in making this more extensible/maintainable in the future

skyrl/backends/skyrl_train/inference_servers/utils.py
```diff
@@ -11,6 +11,13 @@
     SKYRL_LORA_ADAPTER_NAME,
 )
 from skyrl.backends.skyrl_train.weight_sync import get_transfer_strategy
+from skyrl.backends.skyrl_train.weight_sync.serialized_fp8 import (
+    SERIALIZED_BLOCKWISE_FP8,
+    get_qwen35_fp8_ignored_layers,
```
erictang000: similarly here it would be great to remove any mention of specific models and use some abstraction for layers to ignore - we shouldn't hard code things for qwen3.5 in here

erictang000 (commented):
left some comments:
1. let's design a more extensible/maintainable solution for model specific quantization mappings/logic and try to keep instances of using a model name to a single config/abstraction
2. i'm a little confused by the fp8 param init logic but i might be missing something
3. ideally we can find areas where we could upstream logic to megatron-bridge rather than maintaining it ourselves for things like quantizing weights to fp8 for transfer
4. gpu ci tests for both regular full fp8 and fp8 params for hopper (+ both moe and dense tests for full fp8, so 3 total gpu ci tests)


otherwise the weight sync changes for cuda ipc/nccl look pretty clean, looking forward to seeing the PR with updates!

skyrl/train/utils/utils.py
jinghanyao1-hub: Yes, and the reason it was absent from our configs is that vLLM's own default is already `1`, so Blackwell ran correctly without us setting it. Made this explicit in the new code.

skyrl/train/dataset/collators.py
jinghanyao1-hub: Good catch, fixed!

skyrl/train/config/config.py
jinghanyao1-hub: Done!

skyrl/backends/skyrl_train/workers/megatron/quantization/amax_epsilon_patch.py
jinghanyao1-hub: It was purely ours. At worker startup, we read the env var and rewrite the defaults of TE's Float8BlockScaling recipe so its three QParams (fwd input, fwd weight, bwd grad) carry amax_epsilon = 1e-4 instead of TE's default 0.0.

And yes, MoE experts that receive zero tokens (plus packing padding) emit all-zero 1×128 weight-grad blocks;  pow-2 scales `amax == 0` degenerates to a huge `scale_inv` and the FP32 grad-norm reduction overflows to `inf`.

I verified that if we use `MXFP8` (which use 1x32 block instead of 128) as the default on Blackwell, we no longer need to set this env var, the verified run is on [wandb](https://wandb.ai/sky-posttraining-uc-berkeley/qwen35_35b_a3b_fp8_amaxeps_20260705?nw=nwuserjinghanyao1 ) run name: `qwen35_35b_a3b_mxfp8_tp1ep8etp1_vtp1x8_tb32_steps400_skyrl-b200-cpct-eun1b-1_20260727_205922`

I have updated the default: blockwise + FP32 scaling on Hopper, mxfp8 on SM100. No more env var is need here.

skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
jinghanyao1-hub: With `fp8_param` on, weights live in FP8, but the optimizer needs FP32 master copies built from the original checkpoint values (building them from FP8 would bake the rounding error in forever). Bridge saves those originals as a side copy during loading. 
The one-liner loads the model perfectly but loses the side copy. Bridge internally creates a brand-new helper object every time it's touched, the side copy gets stored on a temporary helper, and nothing copies it back.

skyrl/backends/skyrl_train/weight_sync/serialized_fp8.py
jinghanyao1-hub: done!

skyrl/backends/skyrl_train/weight_sync/serialized_fp8.py
jinghanyao1-hub: Done!

skyrl/backends/skyrl_train/inference_servers/utils.py
jinghanyao1-hub: Done!

skyrl/backends/skyrl_train/weight_sync/serialized_fp8.py
jinghanyao1-hub: I tried the `export_weight_dtype='fp8'` that hands over the fp8 weight directly, got some size mismatch error on Qwen3.5 packed tensor:
```
File ".../megatron/bridge/models/conversion/param_mapping.py", line 3566, in split_gdn_linear_weights
    in_proj = in_proj.reshape(tp_size, -1, feature_dim)
RuntimeError: shape '[1, -1, 1024]' is invalid for input of size 520
```
Megatron bridge tries to quantize them at once, which is wrong because this is a packed tensor. What we do is to split it first, then perform the quantize on each. Note, bridge always creates the bf16 weights internally when we call `export_weight`.

skyrl/backends/skyrl_train/distributed/megatron/packing_utils.py
```diff
@@ -1,27 +1,83 @@
...
+    return resolved
+
+
+def _fp8_token_align(tp_size: int, cp_size: int, fp8_recipe: Any) -> int:
+    # Blockwise FP8 quantizes sequence-parallel all-gather inputs in 1x128
+    # tiles (a 128*tp*cp global segment when tp>1) and requires 16-token
+    # local slabs at TP=1. MXFP8 quantizes in 1x32 tiles, so the TP=1 slab
+    # grows to 32 (TE asserts dims % 32 == 0); the tp>1 segment of
+    # 128*tp*cp is already a multiple of 32 and stays unchanged.
+    if tp_size > 1:
```
erictang000, line 17: does this mean we're unnecessarily aligning to 128 * tp * cp tokens for mxfp8? 

Could we modify this to align to 32 * tp * cp unconditionally for mxfp8 and then have the 128 * tp * cp / 16 * tp * cp just be for blockwise?

skyrl/train/config/config.py
```diff
@@ -1100,6 +1113,14 @@ class InferenceEngineConfig(BaseConfig):
     """Should match the dtype used by the inference engine.
     Also used during full-weight sync, where policy weights are cast to this dtype before being sent
     to the inference engine. The LoRA-adapter sync path exports fp32 instead."""
+    fp8_weight_sync_mode: Optional[str] = None
```
erictang000, line 1147: question here: with this pr if i do `fp8_recipe=auto` and it resolves to mxfp8 on blackwell, i guess the `fp8_weight_sync_mode` has to be `serialized_blockwise`, but in that case does this send fp8 weights/scales per 1x32 block and send the e8m0 scaling factors?

skyrl/train/config/config.py
```diff
@@ -499,7 +503,16 @@ class MegatronConfig(BaseConfig):
     """Pass-through kwargs for Megatron's ``TransformerConfig``:
     https://github.com/NVIDIA/Megatron-LM/blob/core_r0.13.0/megatron/core/transformer/transformer_config.py
     Also the place to put HuggingFace config overrides (e.g. ``rope_parameters``) for the Megatron
-    backend, where FSDP would use ``model_config_kwargs``."""
+    backend, where FSDP would use ``model_config_kwargs``.
+
+    FP8 training is configured here: ``fp8`` (TransformerEngine compute format, e.g. ``"e4m3"`` or
```
erictang000: can we make `fp8_param`, `fp8_recipe`, `fp8` and `fp8_amax_compute_algo` top level MegatronConfig members? the transformer_config_kwargs can still be used to overwrite them (since the kwargs are applied on top of our params), but i think it's worth making them directly visible to the user

skyrl/backends/skyrl_train/distributed/megatron/packing_utils.py
```diff
@@ -1,27 +1,83 @@
...
     """Return whether a Megatron/TE fp8 config value enables FP8 execution."""
     if isinstance(fp8, str):
-        return fp8.strip().lower() not in {"", "0", "false", "none", "null", "no"}
+        return fp8.strip().lower() not in {"", "0", "false", "none", "null", "no", "off"}
     return bool(fp8)
 
 
-def get_packed_seq_align_size(tp_size: int, cp_size: int, fp8_enabled: bool = False) -> int:
-    """Return global per-subsequence padding needed for TP/CP layout."""
+def is_mxfp8_recipe(fp8_recipe: Any) -> bool:
```
erictang000: nit: i know that they are used for the packing logic here, but could we move these more generic util functions to a new file `distributed/megatron/quantization_utils.py`? like `is_fp8_enabled`, `is_mxfp8_recipe`, `is_blackwell_or_newer()`, `resolve_auto_fp8_recipe` are not necessarily specific to packing, and get imported elsewhere in the weight sync/worker code.

skyrl/train/config/config.py
```diff
@@ -137,7 +137,7 @@ class FakeInt4QatConfig(BaseConfig):
     BF16 masters, enabling this fake-quantizes the frozen expert GEMMs onto the
```
erictang000, line 142: could you add the scripts from `Reproduce key configs` from the pr description as separate example files in the top level `examples/train/`? Maybe under a new folder `examples/train/fp8`

One for blockwise fp8 on hopper, one for blockwise fp8 + fp8 param on hopper, and one for mxfp8 (or blockwise fp8?) on blackwell. Also for both MoE and dense.

would be very helpful for users to reference

erictang000: oh ideally colocated + non colocated as well

skyrl/backends/skyrl_train/workers/megatron/quantization/amax_epsilon_patch.py
erictang000: let's delete this file/patch if we can always use mxfp8 for blackwell then to reduce bloat then? we likely should just support the most happy path which is mxfp8 for blackwell rather than maintaining blockwise for it as well

skyrl/backends/skyrl_train/weight_sync/serialized_fp8.py
erictang000: got it makes sense

tests/backends/skyrl_train/gpu/gpu_ci/megatron/test_megatron_models.py
```diff
@@ -279,18 +308,77 @@ async def construct_training_input_from_generator_output(generator_output, token
...
+            2,
+            2,
+            "Qwen/Qwen3.5-0.8B",
+            1e-1,
+            5e-2,
+            "fp8_param",
+            id="qwen3.5-0.8b-dense_tp2_fp8_param",
+            marks=pytest.mark.h100,
+        ),
+        # vLLM must run TP=1 for this model under blockwise FP8: at TP=4 a
```
erictang000: oh interesting, does tp=2 work? also can we do tp=4 if we do mxfp8 on blackwell?

just wondering

tests/backends/skyrl_train/weight_sync/fp8/test_serialized_fp8.py
```diff
@@ -0,0 +1,310 @@
+from types import SimpleNamespace
```
erictang000, line 1: can we make the test here match the file structure? so this goes inside weight_sync/fp8/test_serialized_fp8.py

tests/backends/skyrl_train/workers/megatron/quantization/test_fp8_param.py
```diff
@@ -0,0 +1,171 @@
+import pytest
```
erictang000, line 1: similarly can we put this in tests/backends/skyrl_train/workers/megatron/quantization to match file structure

tests/backends/skyrl_train/workers/megatron/quantization/test_fake_int4_qat.py
```diff
@@ -13,7 +13,7 @@
 import pytest
```
erictang000, line 13: same here

skyrl/backends/skyrl_train/weight_sync/fp8/vllm_format.py
```diff
@@ -0,0 +1,147 @@
...
+        "activation_scheme": "dynamic",
+        "weight_block_size": [block_m, block_n],
+    }
+    if ignored_layers:
+        qconfig["ignored_layers"] = list(ignored_layers)
+    return qconfig
+
+
+def should_use_serialized_fp8(mode: str | None) -> bool:
+    return mode == SERIALIZED_BLOCKWISE_FP8
```
erictang000: nit: can we remove this level of indirection and just replace all instances of this function with `mode == SERIALIZED_BLOCKWISE_FP8` for readability

erictang000 (commented):
thanks for the changes, pr is looking good! i think we're almost there, some more minor comments/questions

tests/backends/skyrl_train/gpu/gpu_ci/megatron/test_megatron_models.py
erictang000: this 4304 number is for the vision part - see if we can do tp > 1 if we do `language_model_only=true` on the vllm side

https://huggingface.co/Qwen/Qwen3.6-35B-A3B/blob/main/config.json

skyrl/train/config/config.py
erictang000: answer discussed offline: currently we've done mxfp8 trainer, then 128x128 blockwise quant on bf16 params to send to vllm which does blockwise quant and that's somehow working decently still

todo: implement mxfp8 serialized weight sync method

jinghanyao1-hub:
> thanks for the changes, pr is looking good! i think we're almost there, some more minor comments/questions

Thanks for all the feedbacks. 

- `_fp8_token_align()` now dispatches on the recipe: `32*tp*cp` at any TP for mxfp8.

- `fp8`, `fp8_recipe`, `fp8_param`, `fp8_amax_compute_algo` are now `MegatronConfig` members. They fold into `transformer_config_kwargs` via `setdefault`.

- `is_fp8_enabled`, `is_mxfp8_recipe`, `is_blackwell_or_newer`, `resolve_auto_fp8_recipe` moved out of `packing_utils`. 

- `examples/train/fp8/`: Hopper blockwise, Hopper blockwise + fp8_param, and Blackwell MXFP8, each for dense 9B and MoE 35B-A3B, with a colocated/non-colocated toggle.

- amax-epsilon patch deleted.

- TP=2 / TP=4 in vllm work now by ignoring every vision block linear.

skyrl/backends/skyrl_train/distributed/megatron/quantization_utils.py
```diff
@@ -0,0 +1,57 @@
...
+            # amax==0 grad blocks whose degenerate scales overflow the grad
+            # norm. The native recipe has no such mode.
+            logger.warning(
+                "fp8_recipe={} is emulated on SM100+; prefer fp8_recipe='mxfp8' (or 'auto').",
+                recipe,
+            )
+        return recipe
+    resolved = "mxfp8" if is_blackwell_or_newer() else "blockwise"
+    transformer_config_kwargs["fp8_recipe"] = resolved
+    return resolved
```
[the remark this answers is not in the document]
erictang000: hmm @YJHMITWEB i think this is a good point

skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
```diff
@@ -1486,10 +1602,15 @@ async def broadcast_to_inference_engines(
...
+                weight_iterator = self.weight_extractor.extract_weights(generator_dtype)
+                derive_metadata_from_chunks = self.weight_extractor.serialized_fp8_config is not None
+                weight_metadata = (
+                    None if derive_metadata_from_chunks else self.weight_extractor.get_weight_metadata(generator_dtype)
+                )
                 await self._weight_transfer_sender.send_chunks(
-                    self.weight_extractor.extract_weights(generator_dtype),
+                    weight_iterator,
                     weight_metadata=weight_metadata,
+                    derive_metadata_from_chunks=derive_metadata_from_chunks,
```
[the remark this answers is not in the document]
erictang000: this seems right too

skyrl/backends/skyrl_train/distributed/megatron/quantization_utils.py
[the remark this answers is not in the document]
erictang000: let's maybe just make it explicitly either "blockwise" or "mxfp8" wdyt

examples/train/fp8/run_fp8_blackwell_mxfp8_qwen35_35b_a3b.sh
```diff
@@ -0,0 +1,122 @@
...
+  trainer.policy.megatron_config.pipeline_model_parallel_size=$MEGATRON_PP \
+  trainer.policy.megatron_config.context_parallel_size=$MEGATRON_CP \
+  trainer.policy.megatron_config.expert_model_parallel_size=$MEGATRON_EP \
+  trainer.policy.megatron_config.expert_tensor_parallel_size=$MEGATRON_ETP \
+  trainer.ref.megatron_config.tensor_model_parallel_size=$MEGATRON_TP \
+  trainer.ref.megatron_config.pipeline_model_parallel_size=$MEGATRON_PP \
+  trainer.ref.megatron_config.context_parallel_size=$MEGATRON_CP \
+  trainer.ref.megatron_config.expert_model_parallel_size=$MEGATRON_EP \
+  trainer.ref.megatron_config.expert_tensor_parallel_size=$MEGATRON_ETP \
+  trainer.policy.megatron_config.fp8=e4m3 \
```
erictang000: nit: can we move all FP8 related params up to the top with aliases? (similar to how we set MEGATRON_TP, MEGATRON_PP, etc...) for better readability for these example scripts?

jinghanyao1-hub: This is fixed.

erictang000:
h100 ci: https://console.anyscale.com/cld_hxkifz7xa22mwicp21nzkds1lw/prj_4b6c498rypyq6g7yhk6vzgjevt/jobs/prodjob_2uyc79satm1xe69hmn1cltsb5r?job-tab=overview&job-logs-section-tabs=application_logs


was all passing but ran into

```bash
FAILED tests/backends/skyrl_train/gpu/gpu_ci/megatron/test_router_replay.py::test_forward_backward[tp2_pp2_ep2] - SystemExit: 1
==== 1 failed, 7 passed, 4 deselected, 2075 warnings in 4324.38s (1:12:04) =====
```
timeout

erictang000:
h100 ci: https://github.com/NovaSky-AI/SkyRL/actions/runs/33453865623
megatron gpu ci: https://github.com/NovaSky-AI/SkyRL/actions/runs/33453869556
megatron models gpu ci: https://github.com/NovaSky-AI/SkyRL/actions/runs/33453865429/job/99689523705

erictang000:
https://wandb.ai/sky-posttraining-uc-berkeley/skyrl_fp8

verified 100 steps of DAPO still matching on scripts
image
image

Issue #1846, opened 2026-06-29
SkyRL Q3 2026 Roadmap


erictang000:
### Note to Community
This roadmap is a living document, and we welcome community input and feedback. We will update this roadmap to link to specific Issues and PRs for each sub-task.

# Overview
SkyRL's primary focuses in Q3 2026 will continue to be pushing performance for async RL on large scale MoE models, and improving support for SkyRL's native Tinker API compatible training engine.

## Tinker Engine
- [x] [P0] Improve sqlite DB path support for higher concurrency 
- [x] [P0] Avoid inference engine initialization for the SFT code path https://github.com/novasky-ai/skyrl/pull/1878
- [ ] [P1] Expose metrics for observability on the Tinker Engine (i.e. concurrent requests, number of tenants, etc...)
- [ ] [P1] Provide support for external postgres deployments as an alternative for sqlite for high concurrency/prod deployments
- [ ] [P1] Support storing only K-most recent checkpoints to save disk space.
- [ ] [P1] Router replay (R3) support for the tinker engine
- [ ] [P2] Support additional Tinker LoraConfig params: https://github.com/NovaSky-AI/SkyRL/issues/1632
- [ ] [P2] Support ref model in SkyRLTrainBackend

old tracker: https://github.com/NovaSky-AI/SkyRL/issues/1380 

### Megatron
- [ ] [P0] Add a DAPO recipe with MoE + megatron to CI: https://github.com/NovaSky-AI/SkyRL/issues/1322 @erictang000 
- [ ] [P1] Enable async checkpointing
- [ ] [P1] Low-precision expert quantization (like in [link](https://fireworks.ai/blog/scaling-optimizing-frontier-model-training)) 
- [ ] [P2] Megatron dynamic context parallel support (https://github.com/NovaSky-AI/SkyRL/issues/1019)
- [ ] [P2] Support megatron native FSDP

old tracker: https://github.com/NovaSky-AI/SkyRL/issues/1392

### Quantization
See full quantization (FP8, MXFP8, NVFP4, INT4-QAT) tracker here: https://github.com/NovaSky-AI/SkyRL/issues/1943
- [ ] [P0] Enable Full FP8 Training @jinghanyao1-hub 

### Router Replay
- [x] [P0] Add router replay test to CI: https://github.com/NovaSky-AI/SkyRL/pull/1770
- [ ] [P0] Validate multi-node inference for R3 with new vLLM implementation + evaluate data transfer with HTTP
[...]

PR #1902, opened 2026-07-15
[feat] delta weight sync

kailash109:
Addresses delta weight sync in https://github.com/NovaSky-AI/SkyRL/issues/1846

## Overview

Added a third weight-sync transport (`weight_sync_backend = "disk"`) for publishing deltas during weight sync, design inspired from [slime delta weight sync](https://github.com/THUDM/slime/blob/main/docs/en/advanced/delta-weight-sync.md). Instead of broadcasting full weights via nccl every step, trainer publishes only zstd-compressed bytes that chnaged since the last sync to a shared filesystem, and inference hosts patch a local checkpoint copy in place and reload. . Because vllm has no disk-reload endpoint like sglang update_weights_from_disk, I implement a custom DiskWeightTransferEngine inherited from vllm's WeightTransferEngine which calls model.load_weights from the disk checkpoint 

## Key Changes

- Sender (weight_sync/disk_strategy.py): Trainer-side half -- rank 0 keeps an in-memory CPU copy of last weights inference engine received (ie. base snapshot). ON the first sync, this snapshot is loaded from the original HF checkpoint (same as inference engines). Each sync, for all weight tensors coming out of the weight_extractor, 1) compare bytes against snapshot to produce a delta (xor of buffers, or overwrite depending on option), compress the delta with zstd level 1, and update snapshot to new bytes, so next sync diffs against what engines will hold. 

- Codec (weight_sync/delta_codec.py): Handles encode/apply functions for xor and overwrite deltas, patches safetensors checkpoint file in place using mmap, and includes a per-tensor checksum computed by sender and reverified by the receiver after appying the delta. 

- Receiver (inference_servers/disk_transfer_engine.py): Inference-side half, implemented as plugin for vllm weight transfer system -- add "disk" to registry of weight transfer engines. On a weight update, single worker downloads new delta versions from shared filesystem and patches the local copy in place, verifying each tensor's checksum and load weights via model.load_weights
- Misc: added new update_weights_disk collective RPC on NewInferenceWorkerWrap and RemoteInferenceClient

## Validation

I use modal for all my experiments, treating a Modal volume as the shared file system for delta syncs. 

`soak_disk_weight_sync_modal.py --steps 20 --no-prune`: observe mean delta ~91.3 MB, mean compression 10.82x with real gradient steps. 

For a full training run, I tested gsm8k on qwen-1.7b with disaggregated inference (2xh200 for training, 2xh200 for inference). 

logprob diff (consistently around 0.01): 
image

reward: 
image

To reproduce, run `modal run examples/train/gsm8k/modal_disk_wsync_train.py` (you can choose which model to run there, all on gsm8k)


SumanthRH:
Hi @kailash109 , Thanks for the PR. 

Co-incidentally, I've been working on supporting delta weight updates in SkyRL. I just put up the RFC: https://github.com/novasky-ai/skyrl/issues/1903 and I plan to fast-follow with the initial implementation. 

If you're interesting, please leave feedback on the [doc](https://docs.google.com/document/d/1tpgWVQlkX3S56Zgy_XywAZJhF515p-iX7zPwLCdryxk/edit?tab=t.zb10tyhmkd8h#heading=h.1td26easin67) 

Would love to work on Modal support as a follow-up once initial support is merged!

kailash109:
Hi @SumanthRH thanks for the response! And yes that sounds good, will look over the doc and leave my comments. 

On another note, I'm working on some features over the Tinker engine (in particular routing replay for moe training on the tinker API + looking into enabling higher concurrency through the Tinker path) that seem to align with your roadmap goals, if I could tackle some of them through this work let me know!

PR #1905, opened 2026-07-16
Add Documentation for MTP RL

zanderjiang:
This PR adds documentation for #1832


docs/content/docs/examples/spec_decode.mdx
```diff
@@ -0,0 +1,155 @@
...
+MTP training requires the `megatron` backend and the `vllm` inference engine.
+</Callout>
+
+In this example we train a model's native **Multi-Token Prediction (MTP)** head with a decoupled draft loss, and reuse that head for **vLLM speculative decoding** to speed up rollout. Because the draft loss is autograd-decoupled, it trains only the MTP head and never perturbs the policy, leaving the RL dynamics unchanged.
+
+Source code: `examples/train/spec_decode/`.
+
+## Overview
+
+During RL training, rollouts (generation) are typically the throughput bottleneck. Speculative decoding improves throughput (at lower batch size) with a small, cheap **drafter** that proposes the next `k` tokens, which the policy then verifies in a single forward pass.
```
erictang000, line 19: can you elaborate a little on why this is only true for lower batch size, and give ranges for what batch sizes are considered low (or what empirically you had to make batch sizes/context length to see mtp speedups) 

also would be helpful for high level guidance on when mtp is helpful, like if we are compute bound w/ high concurrency then it isn't too helpful, but if we are memory bound due to long context kv cache reads or weight reads at low concurrency then it helps (pls correct me if my rough understanding is wrong)

erictang000: i guess another question i have is if turning on mtp can ever slow us down significantly at high concurrency, like is this a safe default to just turn on or is it tricker to pick when to use it

SumanthRH (approved):

PR #1907, opened 2026-07-16
[fix] Render multimodal sample prompts under the base model (#1860)

### LINKED ISSUE #1860: Multimodal Tinker sampling with LoRA 404s: _render_for_sample posts the adapter name to /v1/chat/completions/render
Multimodal RL training via the Tinker backend is broken for any LoRA run: every rollout whose prompt contains an image fails at sampling. We hit this on Qwen3.5-4B-VL, but it is not model-specific — it affects any VLM + LoRA on the new inference path (`_SKYRL_USE_NEW_INFERENCE=1`).

The failure surfaces (through `RemoteInferenceClient.raise_for_status`) as:

```
ClientResponseError: 404, message='The model `model_6fa65be7` does not exist.',
url='http://.../v1/chat/completions/render'
```

### Root cause

For prompts containing image chunks, `RemoteInferenceClient.sample()` calls `_render_for_sample()`, which POSTs the resolved `model` (a loaded LoRA adapter name / Tinker `model_id`) to vLLM's `/v1/chat/completions/render`:

```python
# skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py  (_render_for_sample)
render_payload = {"json": {"model": model, "messages": [{"role": "user", "content": content_parts}]}}
```

vLLM's render endpoint does **not** resolve LoRA adapter names — even loaded ones — so it returns 404. The generate endpoints (`/inference/v1/generate`, `/v1/chat/completions`) *do* accept the adapter name. Since rendering (chat-template application + tokenization + multimodal preprocessing) is adapter-independent, the render call should use the base model.

Text-only prompts are unaffected because `_render_for_sample` flattens their tokens locally and never calls `/render`; the bug only bites when a prompt actually contains an image.

### Minimal repro

vLLM 0.20.2, launched with `VLLM_ALLOW_RUNTIME_LORA_UPDATING=1`:

```bash
vllm serve Qwen/Qwen2.5-0.5B --enable-lora --max-lora-rank 16 &   # served under "Qwen/Qwen2.5-0.5B"
# save any tiny PEFT LoRA adapter for that base to /tmp/adapter, then:

curl -s localhost:8000/v1/load_lora_adapter \
  -d '{"lora_name":"adapter1","lora_path":"/tmp/adapter"}'
# -> 200  "Success: LoRA adapter 'adapter1' added successfully."

curl -s localhost:8000/v1/chat/completions/render \
  -d '{"model":"adapter1","messages":[{"role":"user","content":"hi"}]}'
# -> 404  {"error":{"message":"The model `adapter1` does not exist.","type":"NotFoundError","code":404}}

curl -s localhost:8000/v1/chat/completions \
  -d '{"model":"adapter1","messages":[{"role":"user","content":"hi"}],"max_tokens":1}'
# -> 200
```

i.e. the render endpoint rejects a loaded LoRA adapter that the generate endpoint accepts. (We reproduced the identical signature both directly against vLLM and through `RemoteInferenceClient.render_chat_completion`, and confirmed it on the true multimodal path with a real VLM + image.)

### Suggested fix

Render with the base model in `_render_for_sample`:

```diff
         render_payload: Dict[str, Any] = {
             "json": {
-                "model": model,
+                "model": self.model_name,  # render is adapter-independent; /render can't resolve LoRA names
                 "messages": [{"role": "user", "content": content_parts}],
             }
         }
```

`self.model_name` is the base model the server was started with (`inference_servers/setup.py` sets it to `cfg.trainer.policy.model.path`), and the non-empty value satisfies `_resolve_model` under `uses_lora_weight_sync=True`. Generation continues to target the adapter on `/inference/v1/generate`.

One thing worth verifying at merge: that the spliced placeholder tokens + `mm_features` are identical whether the request is rendered under the base model vs the adapter (they should be, since the multimodal processor is shared) — a small parity assertion would lock this down.

### Environment

- SkyRL `fd7459bdb677a6c651c3dc837d72ab7cf57f03a8` (bug also present on `main`)
- vLLM `0.20.2`
- New inference path (`_SKYRL_USE_NEW_INFERENCE=1`)

sunnyadn:
# What does this PR do?

Fixes #1860.

`_render_for_sample` posts the resolved model to `/v1/chat/completions/render`. Under `uses_lora_weight_sync=True` the resolved model is the LoRA adapter name, which the render endpoint does not resolve, so every image-bearing Tinker sample request 404s. Following @benrhodes26's diagnosis in #1860, the render request now always targets the base model (`self.model_name`): rendering (chat-template application, tokenization, multimodal preprocessing) is adapter-independent. Generation still targets the adapter on `/inference/v1/generate`.

The `model` parameter of `_render_for_sample` becomes unused and is removed.

## Tests

New `test_sample_with_image_renders_under_base_model`: image chunks + `uses_lora_weight_sync=True` + explicit adapter model, asserts the render sub-request targets the base model while generate targets the adapter. On main it fails with `AssertionError: assert 'lora-adapter' == 'base-model'`; with this fix the full file passes (57 passed, `pytest tests/backends/skyrl_train/inference_servers/test_remote_inference_client.py -m "not vllm"`).

Also verified live against vLLM 0.23.0 (the version pinned in `pyproject.toml`) on an L40S, serving `Qwen/Qwen2.5-VL-3B-Instruct` with a LoRA adapter via `--enable-lora --lora-modules`:

- `POST /v1/chat/completions/render` with the adapter name returns 404 (`The model 'lora-test' does not exist`), even though `/v1/models` lists the adapter.
- The same request with the base model name returns 200 with prompt token ids and image placeholder features.
- `POST /v1/chat/completions` with an image under the adapter name returns 200, confirming generation resolves adapters while render does not.

On the parity point raised in #1860 (base-model render output matching what the adapter would produce): in vLLM 0.23.0 this holds by construction. The render module (`vllm/entrypoints/serve/render/`) has no LoRA code path, and its model check goes through `OpenAIModelRegistry`, documented as having "no engine client and no LoRA support". Adapters also cannot alter tokenization in this version (`LoRAConfig` has no extra-vocab field), so there is no adapter-dependent render output to diverge from.

### DIFF
--- a/skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
+++ b/skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
@@ -524,7 +524,6 @@ class RemoteInferenceClient(InferenceEngineInterface):
         self,
         prompt: Dict[str, Any],
         session_id: Optional[str],
-        model: str,
     ) -> Tuple[List[int], Optional[MultiModalFeatures]]:
         """Build token_ids and optional multi-modal features from a Tinker prompt.
 
@@ -532,6 +531,11 @@ class RemoteInferenceClient(InferenceEngineInterface):
         When image chunks are present, calls /v1/chat/completions/render to
         process images, then splices the resulting placeholder tokens into the
         pre-tokenized text stream and adjusts placeholder offsets.
+
+        The render request always targets the base model (``self.model_name``):
+        rendering (chat-template application, tokenization, and multimodal
+        preprocessing) is adapter-independent, and vLLM's render endpoint does
+        not resolve LoRA adapter names, so passing one returns 404 (#1860).
 
         Returns:
             (token_ids, features) where features is None for text-only prompts.
@@ -558,7 +562,7 @@ class RemoteInferenceClient(InferenceEngineInterface):
 
         render_payload: Dict[str, Any] = {
             "json": {
-                "model": model,
+                "model": self.model_name,
                 "messages": [{"role": "user", "content": content_parts}],
             }
         }
@@ -645,7 +649,7 @@ class RemoteInferenceClient(InferenceEngineInterface):
 
         # Render prompt: flatten text tokens and, if images are present,
         # call the render endpoint to get placeholder tokens + features.
-        token_ids, mm_features = await self._render_for_sample(prompt, session_id, model=model)
+        token_ids, mm_features = await self._render_for_sample(prompt, session_id)
 
         # Map Tinker SamplingParams → vLLM format
         sampling_params: Dict[str, Any] = {
@@ -776,8 +780,11 @@ class RemoteInferenceClient(InferenceEngineInterface):
             request_payload: Dict with {"json": <request-body>}.
                 The request body should be OpenAI-compatible chat completion
                 request. ``model`` is optional and resolved via
-                ``_resolve_model``. session_id can be included in json for
-                consistent routing.
+                ``_resolve_model``; note that vLLM's render endpoint only
+                resolves base model names, so under LoRA pass the base model
+                explicitly (adapter names return 404, #1860; see
+                ``_render_for_sample``). session_id can be included in json
+                for consistent routing.
 
         Returns:
             Rendered chat completion response (template-applied prompt and token IDs).
--- a/tests/backends/skyrl_train/inference_servers/test_remote_inference_client.py
+++ b/tests/backends/skyrl_train/inference_servers/test_remote_inference_client.py
@@ -1024,6 +1024,10 @@ class TestExplicitModelRequired:
       ``client.model_name``.
     - If no model is supplied and LoRA **is** in use, the call raises
       ``ValueError`` so requests don't silently target the base model.
+
+    One exception: ``sample`` threads the resolved model to the generate
+    endpoint, but its internal render sub-request for image prompts always
+    targets the base model (see ``_render_for_sample``).
     """
 
     @pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


sunnyadn:
The close/reopen pairs on the timeline were CI retriggers during the July 16 GitHub incident; no code changed between them. Current status: all test jobs pass, and the two remaining red checks (skyrl_gpu_tests, Vercel) fail because fork PRs cannot access ANYSCALE_CLI_TOKEN and the deploy secrets.

bvolpato (approved):
LGTM! Thanks for the focused regression coverage.

sunnyadn:
@erictang000 flagging this since it may never have reached you. It is the fix for #1860, the one you offered to review on the issue, and I picked it up after the reporter went quiet.

Before asking for a review I re-ran the repro against vLLM 0.26.0, which is what main pins now, and the 404 is gone. With a real VLM and an image in the prompt, rendering under a loaded adapter name returns 200, and the token ids and mm_placeholders come back byte identical to a base-model render. An unknown model name still 404s, so the endpoint does validate, it just resolves adapters now. On 0.23.0, the pin when I opened this, the same request did 404. The byte-identical result also settles the parity question the issue asked to confirm at merge, with a measurement rather than an assertion.

The upstream change is vllm-project/vllm#44285, first released in 0.25.0. It splits ServingRender into a renderer and an entrypoint, and the entrypoint now inherits the shared `_check_model`, which already knew about `lora_requests`. That PR never mentions LoRA, so nothing linked it back here, which is probably why #1860 is still open.

So this PR is no longer a bug fix. Happy to close it and #1860, or to keep it as hardening if you would rather not depend on the adapter being registered at render time. Your call.

PR #1910, opened 2026-07-16
perf(r3): cast expert indices to int32 for pp local layers only

dyurk-lila:
> **Note on the diff:** This PR is part of a routed-expert-replay / sampler-support series and builds on the PRs below. GitHub can't show the intermediate branches here, so the diff is cumulative on top of `main` — the changes *new to this PR* sit on top of:
>
> - #1909 — perf(r3) packed-array transport
>
> Reviewing in PR order (lowest number first) shows each incremental change cleanly.

## Problem

Routed-expert IDs are stored in a compact integer dtype, but the Megatron worker
previously expanded the entire global-layer tensor to `int32` before replay setup.
Under pipeline parallelism, each rank consumes only its local router layers, so that
eager expansion allocates a large `int32` temporary on device for layers the rank
never uses.

## Implementation

- Keep routed-expert IDs in their compact dtype through the ordinary whole-batch
  device movement (drop the eager `.to(torch.int32)` in the Megatron worker forward
  paths).
- Determine the current pipeline-parallel stage's router layers and `index_select`
  only those layers before metadata alignment, via a new
  `_get_local_router_layer_indices` helper (patched `layer_number` when available,
  offset-based fallback otherwise).
- Perform the single contiguous `int32` conversion inside `_split_replay_indices`,
  immediately before installing local replay data, so only the bounded PP-local
  slice is ever materialized as `int32`.
- Validate the 4D replay-indices shape up front in `setup_per_microbatch_replay_forward`.

This leaves `Experience.to_device()`, batch generation, and normal batch ownership
unchanged. Part of a small series of routed-expert replay improvements.

## Testing

```
uv run --isolated --extra dev --extra fsdp pytest \
  tests/backends/skyrl_train/utils/test_replay_utils.py -q \
  -p no:cacheprovider --ignore=tests/backends/skyrl_train/gpu
# 13 passed
```

Replay tests cover packed multi-row alignment, PP-local layer selection before
layout, compact `uint8`/`int16`/`int32` sources, required `int32` replay data,
padding-mask alignment, and sequence-parallel mask layout. `py_compile`, Ruff
(0.11.9), and Black (24.10.0) all pass on the changed files.


skyrl/backends/skyrl_train/utils/replay_utils.py
```diff
@@ -168,6 +162,27 @@ def _get_current_pp_stage_layer_range(model_config) -> tuple[int, int]:
     return offset, num_layers
 
 
+def _get_local_router_layer_indices(model_config, global_num_layers: int, instances: list) -> list[int]:
+    local_layer_offset, local_num_layers = _get_current_pp_stage_layer_range(model_config)
+    if local_num_layers == len(instances):
+        layer_indices = list(range(local_layer_offset, local_layer_offset + local_num_layers))
+    else:
```
erictang000, line 169:
```suggestion
    else:
        # Dense-layer mismatch: map each MoE router to its global layer index.
        # Prefer the patched layer_number; fall back to offset-based mapping
        # (assumes dense layers precede MoE layers).
```

erictang000 (approved):

PR #1918, opened 2026-07-17
[fix] Make FSDP HF export atomic on local paths

dinhxuanvu:
## Summary

`FSDPStrategy.save_hf_model` writes the HF export in place: `save_pretrained` emits `config.json` before the safetensors shards. A reader that treats `config.json` as the "export complete" signal (e.g. an async eval watcher polling the export dir) can therefore pick up a half-written checkpoint, and a crash mid-write leaves a partial export with a valid-looking `config.json` behind.

This stages the rank-0 export in a sibling temp dir on the same filesystem and atomically swaps it into place, so the final path only ever contains a complete export.

## Scope

- Local paths only get the stage-and-swap; cloud paths keep using `io.local_work_dir` (already stage-then-upload).
- The multi-writer megatron `distributed_save` path is intentionally untouched — all ranks write into one shared dir there, so a per-call atomic swap would be incorrect.

## Test plan

- [ ] FSDP SFT/RL run exports to a local `export_path`; confirm `global_step_*/policy` contains complete weights + `config.json` and no leftover `.tmp-hf-export-*` dirs.
- [ ] Resume/overwrite: re-export the same step and confirm the prior export is replaced cleanly.
- [ ] Cloud (`s3://`) export still works unchanged.

### DIFF
--- a/skyrl/backends/skyrl_train/distributed/fsdp_strategy.py
+++ b/skyrl/backends/skyrl_train/distributed/fsdp_strategy.py
@@ -3,7 +3,10 @@ import gc
 import json
 import os
 import random
+import shutil
+import tempfile
 from collections import defaultdict
+from contextlib import contextmanager
 from typing import List, Optional, Union
 
 import numpy as np
@@ -572,7 +575,7 @@ class FSDPStrategy(DistributedStrategy):
 
         # Step 4: Save on rank 0 only
         if self.is_rank_0():
-            with io.local_work_dir(output_dir) as work_dir:
+            with self._atomic_local_export_dir(output_dir) as work_dir:
                 # Save the model in HuggingFace format using safetensors
                 model_to_save.save_pretrained(work_dir, state_dict=output_state_dict, safe_serialization=True, **kwargs)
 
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
@SumanthRH PTAL

SumanthRH (approved):

SumanthRH:
> A reader that treats `config.json` as the "export complete" signal (e.g.
an async eval watcher polling the export dir) can therefore pick up a
half-written checkpoint, and a crash mid-write leaves a partial export
with a valid-looking `config.json` behind.

This is related more to the watcher setup than how checkpoints get saved. I wouldn't term this a fix.

One thing that is missing currently is a consistent way for marking the finish of a export - and this PR seems like a reasonable direction. We save the config json file at the end in the Megatron backend case, so we can make FSDP somewhat similar.

skyrl/backends/skyrl_train/distributed/fsdp_strategy.py
```diff
@@ -597,3 +600,38 @@ def save_hf_model(self, model: Union[HFModelWrapper, nn.Module], output_dir: str
...
+
+    @contextmanager
+    def _atomic_local_export_dir(self, output_dir: str):
+        """Rank-0 HF export dir that appears at ``output_dir`` only once complete.
+
+        HF ``save_pretrained`` writes ``config.json`` before the safetensors
+        shards, so an in-place write lets a reader that polls for ``config.json``
+        (e.g. an eval watcher) pick up a half-written export. For local paths,
+        stage in a sibling temp dir and atomically swap it in. Cloud paths already
+        stage-then-upload via ``io.local_work_dir``.
```
SumanthRH:
```suggestion
         Atomic write to the local export path provides a consistent way to signal export
         finish. Ex: An asynchronous watcher can watch for ``config.json`` file to populate
         in the export directory. 
```

PR #1919, opened 2026-07-17
[feat] Adding rollout router replay (R3) on tinker path

kailash109:
## Motivation 

Stabilizing MoE RL training logprob mismatch according to [R3 paper](https://arxiv.org/abs/2510.11370), adding same functionality that exists in SkyRL base training code to Tinker path. Addresses Tinker Engine issues in https://github.com/NovaSky-AI/SkyRL/issues/1846

## Changes 

The actual changes for passing router expert indices through to trainer are very small
- skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py: pass routed_experts through in sample() (natively returned by vllm). 
- skyrl/tinker/types.py / api.py / engine.py: routing_matrix on GeneratedSequence and LossFnInputs, dtype/shape on TensorData, threading through prepare_model_pass_batch.
- skyrl/backends/skyrl_train_backend.py: add rollout_expert_indices in _to_training_batch and add routing_matrix return to aggregate_sample_results
- skyrl/train/dataset/preprocess.py: extract padding/downcast logic into a shared pad_rollout_expert_indices helper
extract the padding/downcast logic into a shared pad_rollout_expert_indices helper (classic path behavior unchanged).

**Extras**: 
examples/tinker/moe/: Added Modal launcher (modal_moe_tinker.py) for Tinker server and GSM8K GRPO client (moe_smoke_client.py) with --enable-router-replay that runs Qwen3-30b-a3b. Had to make one change to tinker SDK to reattach routing_matrix in the JSON deserializer (default tinker implementation will just drop unknown response fields) 

One design consideration is whether the client should be responsible for receiving the router expert indices and pass it back in the forward_backward call, especially for longer sampling lengths where the indices tensor may grow in size. I decided to go this way for simplicity and because it seems to be what [Fireworks](https://fireworksai-docs.mintlify.app/fine-tuning/finetuning-intro#moe-models-and-routing-replay) does in their tinker service. 

## Validation

On the GSM8K GRPO run without router replay: 
image

With router replay: 

image

I haven't observed logprob diff collapse in these runs, but the logprob difference for without R3 seems to be > 2x than with it enabled. Will do more runs to see if i can observe a collapse situation

### DIFF
--- a/examples/tinker/moe/modal_moe_tinker.py
+++ b/examples/tinker/moe/modal_moe_tinker.py
@@ -1,0 +1,385 @@
+"""Serve an MoE model behind SkyRL's Tinker API server on Modal, and submit client jobs to it.
+
+
+demo -- one gpu container runs tinker server + gsm8k client 
+
+serve -- long lived gpu container running tinker server, exposed through modal tunnel 
+
+client -- cpu only container that runs the client against server's modal url 
+
+engine-test -- isolation test: boot ONE plain vLLM server (no Tinker, no megatron, empty GPUs) and time it to first /health 200. Distinguishes 'vLLM+model is slow/stuck' from 'the colocated Tinker path breaks engine boot'
+
+Qwen3-30B-a3B-base recipe: TP2/EP8/ETP1 for training, 2 vllm engines x TP4 for inference
+
+Usage (from the SkyRL repo root):
+    # long-lived server; prints the tunnel URL, stays up until you stop it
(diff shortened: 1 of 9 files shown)


casper-hansen:
I was working a similar feature, and I am not sure I like that `routing_matrix` is returned through the client and has to be passed through. In [my design](https://github.com/casper-hansen/SkyRL/tree/cursor/tinker-rollout-routing-replay), I cache and roll it and automatically apply it server-side. I observe similar logprob diff like you do, it reduces the diff by about 2x.

PR #1920, opened 2026-07-17
[tinker] Fix per-tenant blocking on forward/forward_backward calls

kailash109:
## Overview 

`forward_backward` / `forward` futures are currently completed only after entire cross-model batch finishes, so each client blocks on all other tenant's sub-batches even though backend executes per-model sub-batches serially anyways. 

Add per_model mode to `process_batch_requests` (enabled for forward and forward_backward), valid requests grouped by model_id and each model's futures are completed as soon as its own sub-batch returns, meaning clients unblock after their own batch finishes and can pipeline next request while other tenants' sub-batches are still running. 

GPU execution order is unchanged, and _sleep_inference_engines() calls across per-model sub-batches are no-ops (vllm early return if already sleeping). 

**Validation**: Tinker cpu tests pass


avigyabb:
Hey Kailash, thanks so much for putting this up! The speedup looks good on the SkyRL-Train path as we execute per-model sub-batches serially! However, with the JAX backend I believe we use a fused forward/backward that runs sequences from different tenants at once, which causes slowdown if we split the batch by model. Could we gate the per_model flag on this instead of unconditionally setting it in `process_pending_requests()`? Thanks!

avigyabb:
> Hey Kailash, thanks so much for putting this up! The speedup looks good on the SkyRL-Train path as we execute per-model sub-batches serially! However, with the JAX backend I believe we use a fused forward/backward that runs sequences from different tenants at once, which causes slowdown if we split the batch by model. Could we gate the per_model flag on this instead of unconditionally setting it in `process_pending_requests()`? Thanks!

can be ignored for now - will be removing batching logic across models for forward/backward in the jax path

erictang000 (approved):
lgtm thanks!

erictang000:
@kailash109 can you just resolve the test merge conflict then this should be good to merge

kailash109:
resolved!

pcmoritz:
@kailash109 Thanks a lot for fixing this, this is great!

PR #1923, opened 2026-07-17
[logging] Publish the training loop phase as a Prometheus gauge

eicherseiji:
## What does this PR do?

Publishes which macro-phase the fully async training loop is in as a Prometheus gauge. Today, joining "what was the loop doing" to "how busy were the GPUs" means hand-correlating the experiment tracker against Prometheus by wall-clock.

New metric:

- `skyrl_training_phase{phase=...}`: 1.0 for the active phase, 0.0 for the rest. Phases match the loop's `Timer()` keys (`wait_for_generation_buffer`, `convert_to_training_input`, `run_training`, `sync_weights`, `eval`, `save_checkpoints`) plus the default `generating`, so the wandb `timing/*` keys and the Prometheus label share one vocabulary.

Ray exports it to the same Prometheus that scrapes node GPU metrics, so per-phase GPU utilization is a single-store query:

    avg(ray_node_gpus_utilization) and on() (ray_skyrl_training_phase{phase="eval"} == 1)

Prometheus also survives a cluster restart, when the tracker may be unreachable.

## How

New module `skyrl/train/utils/phase_metrics.py` with `TrainingPhaseGauge`, following the one-module-per-concern layout of `ray_gpu_monitor.py`. The trainer constructs it in `__init__` and sets the phase at each existing `Timer()` boundary. Each transition writes only the two changed series; construction seeds all series so PromQL selectors never hit a missing one. Best-effort: it no-ops when Ray metrics are unavailable and warns on unknown phase names, so it never breaks training or tests.

## Tests

- `tests/train/utils/test_phase_metrics.py` (new): exactly one phase active, context-manager restore, unknown-phase rejection, no-op when Ray is unavailable. 5 passed.
- ruff and black clean on changed files. End to end needs a GPU cluster.


skyrl/train/utils/phase_metrics.py
```diff
@@ -0,0 +1,80 @@
...
+        try:
+            for phase in PHASES:
+                self._gauge.set(1.0 if phase == active else 0.0, tags={"phase": phase})
+        except Exception:
+            # Observability must never break training.
+            pass
+
+    def set_phase(self, name: str) -> None:
+        self._current = name
+        self._emit(name)
```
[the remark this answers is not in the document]
eicherseiji: The gauge is deliberately non-raising: observability should never break training, and a raise inside `with phase(...)` at a checkpoint boundary would kill a multi-day run over a label typo. Unknown names log a warning and keep the current phase consistent, and since all call sites are static literals, a typo surfaces in the first step of any run.

skyrl/train/utils/phase_metrics.py
```diff
@@ -0,0 +1,73 @@
...
+
+            self._gauge = Gauge(
+                "skyrl_training_phase",
+                description="1.0 for the active training-loop macro-phase, 0.0 otherwise.",
+                tag_keys=("phase",),
+            )
+            # Seed every series once so PromQL selectors never hit a missing series.
+            for phase in PHASES:
+                self._gauge.set(1.0 if phase == self._current else 0.0, tags={"phase": phase})
+        except Exception as e:
```
pcmoritz: Are there any practical settings where this would happen? Otherwise it might be cleaner to just remove the `try ... except`

skyrl/train/utils/phase_metrics.py
```diff
@@ -0,0 +1,73 @@
...
+
+from contextlib import contextmanager
+
+from loguru import logger
+
+# Macro-phases of one async training step. Names match the paired Timer keys so the wandb timing/*
+# keys and the Prometheus phase label share one vocabulary. "generating" is the default between
+# blocks, when the trainer is not blocking and generation proceeds in the background.
+PHASES = (
+    "generating",
```
pcmoritz: There is currently no equivalent to `generating` in the wandb `timing/*` metrics, right? Is there any way to align the metrics better?

pcmoritz (approved):
Two more suggestions:
- Put skyrl/train/utils/phase_metrics.py in skyrl/train/utils/metrics.py so the other metrics can also be put into this file going forward
- You could unify the `Timer` and `phase` context manager into a combined util context manager that internally calls both of them

skyrl/train/utils/phase_metrics.py
eicherseiji: Fixed

skyrl/train/utils/phase_metrics.py
eicherseiji: Ah true; fixed. Dropped `generating`. In fully async RL it doesn't really align well.

eicherseiji (commented):
Thanks @pcmoritz! Responded to comments

pcmoritz:
@eicherseiji There are currently some CI failures, can you fix before we merge the PR?

pcmoritz:
Looks like the remaining lint errors are unrelated to this PR, I'll merge it to unblock the next one

pcmoritz:
(I fixed the linting in https://github.com/NovaSky-AI/SkyRL/pull/1944)

PR #1924, opened 2026-07-17
[logging] Publish rollout buffer depth and group keep-rate metrics

eicherseiji:
## What does this PR do?

Logs the rollout buffer state to Prometheus. Generation and training run concurrently: generators fill a buffer and the trainer pulls `mini_batch_size` batches per step. The buffer depth tells which side is the bottleneck. Near zero means the trainer is starving for rollouts; near the cap means generation is paused at the staleness limit. Same idle-GPU symptom, opposite cause.

New metrics:

- `skyrl_gen_buffer_qsize` (Prometheus, per step): buffer depth at step start. The W&B counterpart is #1930's `async/gen_buffer_qsize_at_wait_start`.
- `skyrl_gen_buffer_maxsize`, `skyrl_mini_batch_size` (Prometheus, once at startup): run constants for dashboard panels, e.g. a fill ratio or capacity line.
- `async/keep_rate` (W&B) and `skyrl_gen_group_keep_rate` (Prometheus): `mini_batch_size / (mini_batch_size + dropped)` under `sample_full_batch`, the zero-variance drop rate for the last mini-batch, so a deep buffer of mostly droppable groups is distinguishable from a usable one.

## How

`ScalarGauges` (in `skyrl/train/utils/metrics.py`, alongside the phase gauge from #1923) lazily creates one Prometheus gauge per name at the first `set()`. The trainer publishes step-start values directly because the tracker flushes at step end, after training and weight sync, when the buffer has already moved on.

## Tests

- `tests/train/utils/test_metrics.py` (extended): gauge created once per name with its first description, values coerced to float.
- ruff and black clean on changed files. End to end needs a GPU cluster.


skyrl/train/utils/scalar_gauges.py
```diff
@@ -0,0 +1,29 @@
...
+            gauge = self._gauges.get(name)
+            if gauge is None:
+                from ray.util.metrics import Gauge
+
+                gauge = Gauge(name, description=description or name)
+                self._gauges[name] = gauge
+            gauge.set(float(value))
+        except Exception as e:
+            logger.warning(f"ScalarGauges disabled ({e}); scalar training metrics will not be published.")
+            self._enabled = False
```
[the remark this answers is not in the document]
eicherseiji: All current call sites are numeric by construction (`qsize()`/`.maxsize` are ints, `mini_batch_size` is an int from config, `keep_rate` is a float of ints), so `float()` cannot raise here. If a future consumer feeds arbitrary metric dicts, for example a tracker-to-Prometheus dual-write, non-numeric filtering belongs at that layer rather than as a speculative guard in this helper.

skyrl/train/fully_async_trainer.py
```diff
@@ -503,6 +509,17 @@ async def _watch_generators_done(tasks=generator_tasks, event=all_generators_don
                 trained_steps_this_epoch = self.async_train_dataloader.num_trained() // self.mini_batch_size
                 for _step_idx in range(self.global_step, (1 + epoch) * self.num_steps_per_epoch + 1):
                     with Timer("step", self.all_timings):
+                        qsize = generation_output_group_buffer.qsize()
+                        maxsize = generation_output_group_buffer.maxsize
+                        self.all_metrics["async/gen_buffer_qsize"] = qsize
+                        self.all_metrics["async/gen_buffer_maxsize"] = maxsize
```
SumanthRH: Oops I didn't see this PR and I already made #1930 . Can you rebase ?

SumanthRH: Also, `maxsize` should not be logged because it is a fixed constant

eicherseiji: The constant is for building dashboard panels

eicherseiji: I'll move it so it's logged only once

eicherseiji: Done: merged main and dropped this PR's W&B keys. Buffer depth in W&B is #1930's `async/gen_buffer_qsize_at_wait_start`; this PR now adds only the Prometheus side (`skyrl_gen_buffer_qsize` per step, plus the `skyrl_gen_buffer_maxsize`/`skyrl_mini_batch_size` constants published once at startup for dashboard panels, from the same attribute that sizes the queue).

skyrl/train/utils/scalar_gauges.py
```diff
@@ -0,0 +1,29 @@
...
+            return
+        try:
+            gauge = self._gauges.get(name)
+            if gauge is None:
+                from ray.util.metrics import Gauge
+
+                gauge = Gauge(name, description=description or name)
+                self._gauges[name] = gauge
+            gauge.set(float(value))
+        except Exception as e:
```
pcmoritz: Would it make sense to remove this `try ... except`?

eicherseiji: Removed, thanks!

pcmoritz:
Let's merge https://github.com/NovaSky-AI/SkyRL/pull/1923 and then also put this into `metrics.py`?

pcmoritz (approved):

PR #1389, opened 2026-03-25
[tinker] Support PPO loss with Tinker and add critic model in SkyRLTrainBackend

tamoghnokandar:
Fixes the first issue of #1380.

## Summary                                                                                                            
                                                                                                                        
  - Add critic model support to the Tinker backend, enabling actor-critic PPO training through the Tinker API           
  - Extend `LossFnInputs` with optional `values` and `returns` fields, and add `ppo_critic` loss type                   
  - Refactor `SkyrlTrainBackend` from single-model to multi-model registry (`model_id → role`), supporting both policy  
  and critic actor groups
  - Critic models are created via `create_model` with `model_role="critic"`, sharing the policy's base model with       
  independent training                                                                                                  
  - Add `register_actor_group()` and `set_algorithm_config()` to `WorkerDispatch` for dynamic critic registration
                                                                                                                        
  ## Test plan 

- [x] Unit tests for new `LossFnInputs` fields (`values`, `returns`) and `Datum.to_types()` conversion                
- [x] Unit tests for `prepare_model_pass_batch` with `ppo_critic` loss type
- [x] Unit tests verifying JAX backend rejects critic role and `ppo_critic` loss                                      
- [x] Verify existing policy-only workflows remain unaffected (values/returns default to empty)         
<!-- devin-review-badge-begin -->

---

<a href="https://app.devin.ai/review/novasky-ai/skyrl/pull/1389" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    <img src="https://static.devin.ai/assets/gh-open-in-devin-review-light.svg?v=1" alt="Open with Devin">
  </picture>
</a>
<!-- devin-review-badge-end -->

PR #1932, opened 2026-07-21
[tx] add missing model_role parameter to RayJaxBackend

andrewsykim:
https://github.com/NovaSky-AI/SkyRL/pull/1389 added model_role parameter to create_model but RayJaxBackend implementation was not updated


pcmoritz (approved):
Thanks for fixing this!

PR #1939, opened 2026-07-23
Keep prewarmed inference alive after model unload

j316chuck:
## Summary
- Start base inference during Tinker service startup when `--prewarm-inference` is enabled.
- On temporary model unload, stop only that model's training workers while retaining the service-owned vLLM/router runtime.
- Preserve existing full teardown behavior for services that were not prewarmed.

## Testing
```bash
python -m py_compile skyrl/backends/skyrl_train_backend.py skyrl/backends/skyrl_train/workers/worker.py skyrl/backends/skyrl_train/workers/worker_dispatch.py
git diff --check HEAD^ HEAD
```
Both passed. The focused pytest requires the SkyRL CUDA environment, which is not installed in this worktree.

### DIFF
--- a/skyrl/backends/skyrl_train/workers/worker.py
+++ b/skyrl/backends/skyrl_train/workers/worker.py
@@ -706,6 +706,13 @@ class PPORayActorGroup:
             raise RuntimeError("Cannot determine data-parallel size before actor group initialization.")
         return self._last_dp_size
 
+    def shutdown(self) -> None:
+        """Terminate all workers in this training actor group."""
+        for actor in self._actor_handlers:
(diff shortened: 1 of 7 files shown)


avigyabb:
Hey Chuck, thanks for putting this up! I was wondering: 

In the colocated case, prewarm_inference() in the TinkerEngine init creates a placement group and puts the vLLM engines on it. Then, in create_model() we create another placement group and use it to schedule the policy workers (which would not colocate it). I think in small enough clusters, this should hang and on a bigger cluster it would take up twice the needed amount of GPUs. Any chance I could be missing something here? Thanks!

PR #1945, opened 2026-07-25
[tinker] Return prompt and top-k prompt logprobs from sample requests

pcmoritz:
`sample` requests carrying `prompt_logprobs` / `topk_prompt_logprobs` came back with both fields unset, which the Tinker SDK surfaces as "Tinker response did not include prompt_logprobs".

Three gaps, all in the glue between the API layer and the backends:

- `api.py` accepted `topk_prompt_logprobs` and dropped it when building `types.SampleInput`, so it never reached a backend. A positive top-k now also implies `prompt_logprobs` (both come off the same prompt forward pass).
- The SkyRL-Train backend never put `include_prompt_logprobs` / `topk_prompt_logprobs` in the sample body, so vLLM was never asked for them, and then read the result as a list-of-lists when `RemoteInferenceClient.sample` returns flat per-prompt-token lists -- position 0 (always None) would have been returned as the whole field. Only the first sample of each request asks for prompt logprobs now, since all `num_samples` samples share one prompt.
- `types.SampleOutput.prompt_logprobs` was typed `list[float]`, which rejects the None at position 0.

The two forwarding clients in `extra/` also gain prompt-logprob support via vLLM's `/v1/completions`, and the vLLM -> Tinker conversion is now shared (`convert_vllm_prompt_logprobs`) so the three paths cannot drift.

The JAX backend returns an explicit error for `topk_prompt_logprobs`: tx's generator only produces the prompt tokens' own logprobs, so there is no per-position distribution to build top-k from.


bvolpato (approved):
LGTM! Nice end-to-end handling across API, SkyRL-Train, and both forwarding clients. Nonblocking follow-up: reject unsupported JAX `topk_prompt_logprobs` before generation, so known-failing requests avoid unnecessary work. Not blocking this PR.

erictang000 (approved):

PR #1955, opened 2026-07-28
[feat][tinker] Honor trainer.algorithm.loss_reduction on the Tinker-serving path

casper-hansen:
The Tinker API contract is that per-token losses are summed and clients send pre-normalized advantages: SkyRL-Train's loss_reduction handling lives in its native RL loop (trainer.py), which the Tinker-serving path bypasses. As a result, setting trainer.algorithm.loss_reduction via --backend-config had no effect for Tinker clients.

- Apply the configured reduction to client advantages when building the training batch in SkyRLTrainBackend._to_training_batch, reusing apply_loss_reduction_to_advantages_minibatch (the same helper SkyRL's example Tinker client uses client-side).
- Add "token_sum" (identity) as an explicit reduction mode and make it the serving-path default, so existing behavior is preserved unless a reduction is explicitly configured.
- Reject prompt_mean on the serving path (prompt groupings are not visible to the trainer) and mirror the validate_cfg requirement that seq_mean_token_sum_norm sets trainer.algorithm.max_seq_len.

### DIFF
--- a/skyrl/backends/skyrl_train/utils/ppo_utils.py
+++ b/skyrl/backends/skyrl_train/utils/ppo_utils.py
@@ -1163,7 +1163,7 @@ def apply_loss_reduction_to_advantages_minibatch(
             For step-wise training, minibatch_size can be variable.
         loss_mask: Mask of shape (minibatch_size, seq_len) indicating valid loss tokens.
         loss_reduction: One of "token_mean", "token_mean_legacy", "sequence_mean",
-            "seq_mean_token_sum_norm", "prompt_mean".
+            "seq_mean_token_sum_norm", "prompt_mean", "token_sum".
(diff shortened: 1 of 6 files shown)


casper-hansen:
@erictang000 requesting review

erictang000:
hey casper, thanks for the pr, i'm wondering if there is a reason that it would be more difficult to do the advantage normalization in the client code rather than in the engine?

if it's roughly the same complexity wise it might be better to just strictly enforce `token_sum` on the tinker code path and fail loudly if any other loss reduction is configured just to keep parity with the official tinker implementation

casper-hansen:
@erictang000 that is true, the only reason I wanted this is because the implementation was already there ready to go.

PR #1927, opened 2026-07-18
[feat][SFT] Ingest pretokenized datasets

avigyabb:
## What

Adds a pretokenized data path to the SFT trainer: point `pretokenized_dataset_paths` at local store(s) of already-tokenized rows and train directly, skipping online tokenization (`tokenize_chat_example` / `tokenize_sft_example`) entirely. Useful when a data pipeline tokenizes offline (e.g. on a Spark/Ray data cluster).

> **Scope note (per review):** this PR supports **local paths only**. Cloud-path (S3/GCS) ingestion is split into the follow-up draft #1933, where it lands once for both pretokenized stores and text-format datasets.

## How it works

**New module `skyrl/train/dataset/pretokenized.py`** — `load_from_pretokenized(path, max_length)`, the pretokenized counterpart of `SFTTrainer._load_and_tokenize` (same `list[dict]` return shape, so everything downstream — collators, sequence packing, samplers, checkpoint/resume — works unchanged):

- **Formats** (auto-detected, single file or directory of shards): Parquet, JSON-lines, raw Arrow IPC, or a HF `Dataset.save_to_disk` directory.
- **Row schema**: unpadded `input_ids` plus a full-sequence 0/1 `loss_mask` (same length). `num_actions` is *inferred* from the first nonzero mask entry; window-form masks, `num_actions` columns, and HF-style `labels` are rejected with clear errors. `attention_mask` is optional and must be all-ones (padding stays collation-internal). The mask form covers instruction-following (1s on the response) and multi-turn conversational data (1s on every assistant turn).
- **VLM ingestion**: rows carrying `pixel_values` + `image_grid_thw` pass through to the collator's `TensorList` path. Over-length VLM rows are dropped with a warning rather than truncated (consistent with the online VLM path); mixed text+VLM stores handle the null image columns parquet materializes on text rows.
- **`max_length`** truncation mirrors the online path (prompt prefix kept, action window shrinks, empty-loss rows dropped).

**Config (`SFTConfig`)** — integrated with multi-dataset SFT (#1883):

- `pretokenized_dataset_paths: List[str]` — multiple stores are concatenated and mixed per `train_dataset_weights` via `DataMixingSampler`, exactly like `train_datasets`. Exclusive with `train_datasets`/`train_dataset_splits` (explicit error, no silent precedence): a pretokenized store is the output of a preprocessing job, so the user is assumed to have pre-split/filtered it — HF split syntax does not apply.
[...]

PR #1961, opened 2026-07-29
[feat][SFT] Memory-map pretokenized stores instead of materializing rows

avigyabb:
## What

`load_from_pretokenized` now returns a map-style, arrow-backed **`PretokenizedDataset`** instead of a fully materialized `list[dict]`. Rows live in memory-mapped arrow files on disk; accessing row *i* pages in only the ~4 KB pages that hold it, so resident memory is the OS page cache (reclaimable) rather than the dataset.

This removes the two scaling walls of the materialized path for large pretokenized stores (#1927):
- **RAM = O(dataset)**: every tokenized row held as Python objects for the whole run.
- **x (dataloader_num_workers + 1)**: spawn-based dataloader workers each receive a full pickled copy of the list. The mmap dataset pickles as a file reference — workers re-map the same files, sharing pages instead of duplicating them.

## How

- **Validation stays eager, becomes vectorized**: chunked arrow scans (bounded memory, 200k rows per chunk) over lengths / mask values / attention_mask / VLM pairs — same fail-fast-at-load behavior and same error messages as before, without materializing rows.
- **Row dropping** (empty loss window, over-length VLM) becomes a vectorized keep-mask + `dataset.select()`.
- **Normalization** (`num_actions` inference, truncation, all-ones attention_mask) runs lazily per accessed batch via a picklable transform, so dataloader workers do it off the training critical path.
- **`sequence_lengths` come from arrow offsets** — `_log_dataset_stats` no longer materializes the store. Multi-store concat is a `torch ConcatDataset` view.
- **Map-style contract is preserved**, so samplers (random / `DataMixingSampler` / custom), weighted mixing, and the `data.pt` `StatefulDataLoader` resume flow are all unchanged — mid-epoch resume verified bit-exact, including with spawn workers.

## Measured (100k rows x 256 tokens, real FSDP training on 1xL4, ~3.6 s steps)

| | materialized `list[dict]` | mmap (this PR) |
|---|---|---|
| Load time | 20.8 s | **0.8 s** |
| Resident memory | O(dataset), x(workers+1) | **O(validation chunk); page cache thereafter** |
| Data cost per train step | 2.5 ms (w0) | 11 ms (w0), **0.8 ms (w2)** |

Row content is parity-tested identical to the materialized path, and identical-seed training reproduces its loss curve.

## Tests

`tests/train/test_sft_pretokenized.py` — full suite passes; covers format detection, schema validation/row-dropping parity, VLM rows, collation, `StatefulDataLoader` mid-epoch resume, and spawn-worker pickling.

## Follow-ups (stacked on this PR)

This is Part 1 of a trillion-token SFT dataloading plan: Part 2 adds cloud row-group random access (`RowGroupPretokenizedDataset`, fetches only the row groups a batch touches — no download), Part 3 adds a storage-aligned shuffle sampler (1x read amplification) and a disk cache tier. Complements #1933 (whole-store download path).


skyrl/train/sft_trainer.py
SumanthRH: You should make the return types consistent for `load_dataset`

Let's make the return type to be a `SFTDataset` in both the pretokenized and the text dataset cases. It's a relatively simple lift -> Convert the return type of `_load_and_tokenize` to be a Dataset that stores the full tokenized sequence in memory. Use `ConcatDataset` in cases with multiple training datasets.

```python

class SFTDataset(torch.utils.data.Dataset, abc.ABC):
		"""Base Dataset abstraction for SFT training with SkyRL"""

       @property
       @abc.abstractmethod
       def sequence_lengths(self) -> List[int]:
           raise NotImplementedError

class TextDataset(SFTDataset):
    pass

class PretokenizedDataset(SFTDataset):
    pass

class ConcatSFTDataset(SFTDataset, torch.utils.data.ConcatDataset):

    @property
    def dataset_lengths(self) -> List[int]:
        return [len(dataset) for dataset in self.datasets]
    
    @property
    def sequence_lengths(self) -> List[int]:
        """Sequence lengths from all the samples in the dataset"""
        seq_lengths = []
        for ds in self.datasets:
            seq_lengths.extend(ds.sequence_lengths)
        return seq_lengths

class SFTTrainer:
    def load_dataset(self) -> SFTDataset:
        ...

    def build_train_sampler(self, dataset: SFTDataset):
        is_multi_dataset = isinstance(dataset, ConcatSFTDataset)
        ...
```


We will have `TextDataset` and `PretokenizedDataset` as subclasses of `SFTDataset`. `TextDataset` currently materializes the entire dataset in memory, but it doesn't have to be -> Can be addressed in a follow-up

Obviously, these APIs are subject to change, but we can do some minimal legwork to have better APIs.

skyrl/train/dataset/pretokenized.py
```diff
@@ -125,104 +138,217 @@ def _load_as_hf_dataset(local_path: str) -> Dataset:
...
+        return len(self._dataset)
+
+    @staticmethod
+    def _strip_none(row: dict) -> dict:
+        return {k: v for k, v in row.items() if v is not None}
+
+    def __getitem__(self, idx) -> dict:
+        return self._strip_none(self._dataset[int(idx)])
+
+    def __getitems__(self, indices: list) -> list[dict]:
```
SumanthRH, line 349: Given that NormalizedTransform doesn't vectorize the processing, how much better is this batched fetch as opposed to row-wise?

Does `ConcatDataset` automatically use batched fetch on each dataset? Or will it default to row-wise?

SumanthRH (changes requested):
Nice! Only major comment is on providing a better API for text / pretokenized datasets to simplify logic in the `SFTTrainer`

```python
class SFTDataset(torch.utils.data.Dataset, abc.ABC):
		"""Base Dataset abstraction for SFT training with SkyRL"""

       @property
       @abc.abstractmethod
       def sequence_lengths(self) -> List[int]:
           raise NotImplementedError

class TextDataset(SFTDataset):
    pass

class PretokenizedDataset(SFTDataset):
    pass

class ConcatSFTDataset(SFTDataset, torch.utils.data.ConcatDataset):

    @property
    def dataset_lengths(self) -> List[int]:
        return [len(dataset) for dataset in self.datasets]
    
    @property
    def sequence_lengths(self) -> List[int]:
        """Sequence lengths from all the samples in the dataset"""
        seq_lengths = []
        for ds in self.datasets:
            seq_lengths.extend(ds.sequence_lengths)
        return seq_lengths

class SFTTrainer:
    def load_dataset(self) -> SFTDataset:
        ...

    def build_train_sampler(self, dataset: SFTDataset):
        is_multi_dataset = isinstance(dataset, ConcatSFTDataset)
        ...
```

SumanthRH (approved):

SumanthRH:
@avigyabb can you fix the test failures?

https://github.com/NovaSky-AI/SkyRL/actions/runs/30581508311/job/91002847129?pr=1961



```bash
FAILED tests/train/test_async_batch_collation.py::test_async_collation_matches_serial_default_collator - AttributeError: 'tuple' object has no attribute 'sequence_lengths'
FAILED tests/train/test_async_batch_collation.py::test_async_collation_matches_serial_packed_collator - AttributeError: 'tuple' object has no attribute 'sequence_lengths'
FAILED tests/train/test_async_batch_collation.py::test_async_collation_matches_serial_with_partial_tail - AttributeError: 'tuple' object has no attribute 'sequence_lengths'
FAILED tests/train/test_async_batch_collation.py::test_checkpoint_state_excludes_collated_ahead_batch - AttributeError: 'tuple' object has no attribute 'sequence_lengths'
```

skyrl/train/dataset/pretokenized.py
avigyabb: You're right, for this PR (part 1) since the transform isn't vectorized the batched fetch is only 1.2x better than row-wise when the dataset is mmap'd. However, if the data is in S3, batched fetches issue much fewer S3 requests as opposed to one S3 request per row in the batch.

`ConcatDataset` does not automatically use batched fetch, since it didn't have the `__getitems__()` method. I added in my latest commit.

examples/train/sft/README.md
```diff
@@ -41,6 +41,13 @@ Each row must contain:
 
 Rows are normalized to the trainer's internal format; `max_length` truncation still applies (rows whose loss
 window is fully truncated are dropped, and over-length VLM rows are always dropped rather than truncated).
+
+Pretokenized stores are **memory-mapped, not loaded into RAM**: schema validation runs eagerly at load time
+(vectorized, so malformed stores still fail fast), while row normalization happens lazily at access time.
+Controller memory stays bounded by the OS page cache regardless of dataset size, and the store files must stay
+on disk for the duration of the run. Set `dataloader_num_workers>=2` so the per-batch normalization is
```
SumanthRH, line 48: Can you add this into the docs page for SFT?

https://github.com/NovaSky-AI/SkyRL/blob/main/docs/content/docs/sft/overview.mdx#pretokenized-datasets

Issue #1315, opened 2026-03-11
[fully async] Inform/warn/break about log_probs=old_log_probs


SamuelGabriel:
Hi,

Unlike most algorithm descriptions that I have seen in papers, you don't keep the old weights around in the async setup. This should be fine, if we use the right loss (e.g. GLM5's agentic loss), but with the default settings in SkyRL this leads to clipping being turned off in our GRPO implementation.

I have a parallel [PR](https://github.com/NovaSky-AI/SkyRL/pull/1314) and maybe it would make sense that you force people to use the loss from there, or some loss that uses rollout_logprobs, when they use fully async.

PR #1965, opened 2026-07-31
[CI] Rebaseline gsm8k fully-async E2E regression thresholds

SumanthRH:
## What does this PR do?

Rebaselines the regression thresholds for the nightly `gsm8k` fully-async E2E CI run. The
`loss/avg_final_rewards` gate has been failing on **every** nightly run since 30th June
(22/22 of the last 22 runs), and `eval/all/avg_score` has been failing intermittently
(3/22). Both thresholds were calibrated on 23rd Feb 2026 (#1199) and were never updated
across three intentional changes to the fully-async recipe.

## Why the metrics moved

Pulled every run in the `gsm8k_fully_async_ci` wandb project (348 runs, back to 23rd April)
and diffed the full summary metric set across the step changes. Four discrete events, each
landing on the first nightly after a specific merge:

| regime | `loss/avg_final_rewards` | `eval/all/avg_score` | `reward/avg_pass_at_5` | `vllm/prefix_cache_hit_rate` | `clip_ratio` | `grad_norm` |
|---|---|---|---|---|---|---|
| 23 Apr – 17 Jun (original baseline) | **0.315** | 0.588 | 0.775 | 0.152 | 0.0000 | 0.30 |
| 19 Jun – 29 Jun | 0.285 | 0.500 | 0.745 | **0.921** | 0.0000 | 0.28 |
| 30 Jun – 13 Jul | **0.238** | 0.410 | 0.675 | 0.921 | **0.0133** | **0.163** |
| 14 Jul – 19 Jul | 0.238 | 0.500 | 0.678 | **0.152** | 0.0130 | 0.164 |
| 20 Jul – today | 0.240 | 0.575 | 0.680 | 0.152 | 0.0130 | 0.163 |

1. **#1798 (18 Jun)** added `trainer.fully_async.clear_kv_cache_on_weight_sync=false` to the
   recipe, so rollouts began reusing KV computed under stale weights.
   `vllm/prefix_cache_hit_rate` 0.15 → 0.92 and `ttft` 1.28s → 0.14s.
2. **#1850 (30 Jun)** set `policy_loss_type="rollout_is"`. This is the dominant cause and it
   has never recovered. Before it, the default `regular` loss recomputes old logprobs, so at
   `update_epochs_per_batch=1` the IS ratio is identically 1 — `clip_ratio` was **exactly
   0.0000 on every run for two months**, i.e. there was no importance-sampling correction at
   all despite `max_staleness_steps=4`. #1850 fixed exactly that (issue #1315) and added an
   assertion forbidding the old behaviour. Correcting it halved `grad_norm` and lowered
   measured step-9 reward.
3. **#1836 (14 Jul)** added the per-`global_step` KV cache salt, reversing #1798's stale-KV
   reuse (`prefix_cache_hit_rate` back to 0.152). `eval` recovered 0.41 → 0.50.
4. **#1929 (20 Jul)** moved metric logging to after eval. Pre-fix, metrics were logged
   *before* eval ran, so the final-step eval was never logged at all and the summary held
   the step-8 eval; post-fix it holds step 9. `eval` 0.50 → 0.575. This is a logging
   discontinuity, not a quality change.

So `eval/all/avg_score` is back at its pre-June level (0.575 vs 0.588) — but only just above
the old 0.56 gate, which is why it now trips intermittently. `loss/avg_final_rewards` is
legitimately lower because the recipe now applies the off-policy correction it was always
supposed to.

## New thresholds

Same methodology as #1199 (5% allowance from min/max of the last 20 runs), applied to the
current post-#1929 regime:

| threshold | old | new | last-20 min / max |
|---|---|---|---|
| `EVAL_ACC_MIN_VALUE` | 0.56 | **0.50** | 0.5292 / 0.5959 |
| `TRAIN_ACC_MIN_VALUE` | 0.32 | **0.20** | 0.2141 / 0.2609 |
| `AVG_NUM_TOKENS_MAX_VALUE` | 283 | 283 (unchanged) | 256.4 / 268.6 |
| `LOGPROBS_DIFF_MAX_VALUE` | 0.040 | 0.040 (unchanged) | 0.0160 / 0.0180 |

The two passing thresholds are left alone — tightening a gate that never fired isn't part of
fixing the regression and would only add flake risk.

### DIFF
--- a/tests/train/gpu_e2e_test/gsm8k_fully_async.sh
+++ b/tests/train/gpu_e2e_test/gsm8k_fully_async.sh
@@ -7,9 +7,35 @@ set -euo pipefail
 RUN_NAME="run_$(date +%Y%m%d%H%M%S)_$$"
 
 SCRIPT_DIR=$(dirname $(realpath $0))
-# Thresholds: 5% allowance from min/max of last 10 CI runs as of 23rd Feb 2026
-EVAL_ACC_MIN_VALUE=0.56
-TRAIN_ACC_MIN_VALUE=0.32
+# Thresholds: 5% allowance from min/max of the last 5 CI runs as of 31st July 2026
+# (eval min 0.5292, train reward min 0.2141).
+#
+# Rebaselined from the original 23rd Feb 2026 values (eval 0.56, train 0.32). Three
+# intentional changes to the fully-async recipe moved these metrics between then and now,
+# and the gate was never updated -- `loss/avg_final_rewards` had been failing on every
+# nightly run since 30th June:
+#
+#   #1798 (18 Jun)  stopped clearing the KV cache on weight sync, so rollouts reused KV
+#                   computed under stale weights (vllm/prefix_cache_hit_rate 0.15 -> 0.92).
+#                   eval 0.59 -> 0.50, train reward 0.32 -> 0.29.
+#   #1850 (30 Jun)  switched the recipe to policy_loss_type=rollout_is. Before this the
+#                   default `regular` loss recomputed old logprobs, so the IS ratio was
+#                   identically 1 at update_epochs_per_batch=1 -- clip_ratio was exactly
+#                   0.0 and there was no importance-sampling correction at all despite
+#                   max_staleness_steps=4. Correcting it halved grad_norm (0.28 -> 0.16)
+#                   and is the dominant cause of the lower reward: train 0.29 -> 0.24.
+#   #1836 (14 Jul)  per-global_step KV cache salt, reversing #1798's stale-KV reuse
+#                   (prefix_cache_hit_rate back to 0.15). eval 0.41 -> 0.50.
+#   #1929 (20 Jul)  logged eval metrics after eval instead of before, so the summary now
+#                   holds the final-step eval rather than the step-8 one. eval 0.50 -> 0.58.
+#
+# eval/all/avg_score has recovered to its pre-June level, but it sits close enough to the
+# old 0.56 gate that 3 of the last 22 runs fell below it, so it is rebaselined too.
+# avg_num_tokens and the logprobs diff were both comfortably inside their gates throughout
+# and are left as-is -- as in Feb, they are deliberately loose guardrails rather than tight
+# regression gates (the last 5 runs sit at 261-268 tokens and 0.0167-0.0174).
+EVAL_ACC_MIN_VALUE=0.50
+TRAIN_ACC_MIN_VALUE=0.20
 AVG_NUM_TOKENS_MAX_VALUE=283
 LOGPROBS_DIFF_MAX_VALUE=0.040
 


SumanthRH:
### Replay results confirming the attribution

Replayed three variants as Anyscale jobs on `l40s_test` (`g6e.12xlarge`, 4x L40S — same GPU
count as `l4_ci`, so the recipe runs unchanged), logging to a separate
`gsm8k_fully_async_ci_triage` wandb project. All reached `global_step=9` like the nightly.

| variant | `loss/avg_final_rewards` | `eval/all/avg_score` | `pass@5` | `prefix_cache_hit` | `clip_ratio` | `grad_norm` |
|---|---|---|---|---|---|---|
| **`jun17good`** — `2b31a747` (17 Jun, last green) on its matching ray 2.51.1 image | **0.3211** | 0.6035 | 0.8047 | 0.1517 | **0.0000** | 0.3082 |
| **`regularloss`** — today's `main`, only `policy_loss_type=regular` (pre-#1850) | **0.2906** | 0.6588 | 0.7695 | 0.1517 | **0.0000** | 0.2120 |
| **`kvclear`** — today's `main` + `clear_kv_cache_on_weight_sync=true` | **0.2172** | 0.5906 | 0.6289 | 0.1515 | 0.0126 | 0.1540 |
| _nightly `main` for reference (last 5)_ | _0.2141 – 0.2484_ | _0.5292 – 0.5883_ | _0.64 – 0.73_ | _0.1517_ | _~0.013_ | _~0.16_ |

Three things this settles:

1. **The 17 Jun commit still reproduces the original baseline: 0.3211** vs 0.3242 measured in CI
   that day. So the drop is entirely attributable to code/config changes — no external drift
   (dataset, vLLM, image) and no L40S-vs-L4 artifact.
2. **`clip_ratio` is exactly `0.0000` for both `regular`-loss runs and nonzero only with
   `rollout_is`**, directly confirming that before #1850 the recipe applied no
   importance-sampling correction at all despite `max_staleness_steps=4`. `grad_norm` splits
   the same way (0.21–0.31 for `regular`, 0.15 for `rollout_is`).
3. **Clearing the KV cache no longer changes anything.** `kvclear` gives `prefix_cache_hit`
   0.1515 vs unmodified `main`'s 0.1517 and lands inside the nightly reward range, so #1836's
   per-`global_step` cache salt fully closed the stale-KV hole that #1798 opened. The flag is
   effectively inert now — which also makes this run a valid stand-in for an unmodified control.

Measured decomposition of the 0.32 → 0.24 drop:

| step | train reward |
|---|---|
| 17 Jun baseline | 0.3211 |
| today's `main` with the pre-#1850 loss | 0.2906 (**−0.031** from everything else since 17 Jun) |
| today's `main` as CI runs it (`rollout_is`) | ~0.24 (**−0.05**, i.e. the bulk, from #1850) |

So #1850 is the dominant cause and it is a deliberate correctness fix, which is what justifies
rebaselining rather than reverting. The residual −0.031 is *not* stale KV (point 3 above); the
most likely candidate is #1802's fully-async consumer refactor, which is worth a separate look
but doesn't change the threshold decision.

One follow-up still running: `rollout_is` with the ScaleRL/CISPO band (`eps_clip_low=1.0`,
`eps_clip_high=4.0`) instead of the DAPO-tight `(0.8, 1.2)` it uses today. If that recovers
reward it's a tuning lead for the recipe, tracked separately from this CI fix.

PR #1988, opened 2026-08-05
[deps] Upgrade vllm-router and fix 0.1.15 startup

bvolpato:
## Summary

Replace temporary `vllm-router==0.1.14.post1` custom wheels with official `vllm-router==0.1.15` across SkyRL training, FSDP, and Megatron extras.

Version 0.1.15 includes the custom wheel's load-aware routing fix, publishes x86_64 and aarch64 wheels, and removes the old `pd_disaggregation` `RouterArgs` field. SkyRL's fallback access to that removed field crashed regular, non-PD startup after the router became healthy.

## Changes

- Pin official `vllm-router==0.1.15` and refresh the lockfile.
- Use the 0.1.15 `vllm_pd_disaggregation` field directly.
- Register SkyRL's typed `/inference/v1/generate` endpoint through router configuration.
- Cover regular and prefill/decode startup arguments.

## Testing

- Router tests (`10 passed`)
- Direct construction of regular cache-aware and prefill/decode `RouterArgs`
- Direct import of official `vllm-router==0.1.15`
- `uv lock --check`
- Ruff, Black, Gitleaks, and `git diff --check`
- Lockfile contains official manylinux x86_64 and aarch64 wheels


SumanthRH:
@bvolpato the custom wheel we're using also has our custom load aware router policy `sticky_least_loaded`: https://github.com/vllm-project/router/compare/main...SumanthRH:router:sticky_least_loaded

So we can't simply upgrade to vllm-router 0.1.15 directly . 

Let me make an issue to track the upgrade though - we should move to a custom wheel on top of 0.1.15 for now.

SumanthRH:
Created an issue to track #2070

bvolpato:
Makes sense! Thanks @SumanthRH

PR #1991, opened 2026-08-05
fix(lora): backload optimizer before adapter swap

pcmoritz:
The adapter swap copy_()s the live param buffer, grad buffer, fp32 main params and Adam state to/from CPU slots, so all of them must be resident. Megatron's grad offload frees grad_data in place (storage().resize_(0)) rather than moving it, so swapping while the optimizer half is offloaded reads a dangling device pointer and fails with a bare "CUDA error: invalid argument".

- ensure_active_adapter() now backloads with need_optimizer=True itself instead of trusting the caller, since paths like forward and save_weights_for_sampler deliberately keep the optimizer off GPU.
- Move ensure_active_adapter() ahead of _prepare_for_weight_sync(), which is what pushes those buffers back off GPU.
- Add _check_resident() in AdapterStore so a freed storage raises an actionable error instead of the opaque CUDA failure.

### DIFF
--- a/skyrl/backends/skyrl_train/workers/worker_dispatch.py
+++ b/skyrl/backends/skyrl_train/workers/worker_dispatch.py
@@ -86,11 +86,20 @@ class WorkerDispatch:
         No-op when ``model_id is None`` (single-tenant / FFT path) or when
         the workers don't have an AdapterStore (non-LoRA strategies).
 
-        Must be called *after* ``_ensure_on_gpu(role, ...)`` so the model
-        and optimizer storages are live before we tensor.copy_() into them.
+        The swap tensor.copy_()s the live param buffer, grad buffer, fp32 main
+        params and Adam state to/from CPU slots, so all of them must be
+        resident first. Megatron's grad offload *frees* grad_data in place
+        (``storage().resize_(0)``) instead of moving it, so swapping while the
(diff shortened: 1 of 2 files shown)


skyrl/backends/skyrl_train/workers/worker_dispatch.py
```diff
@@ -86,11 +86,20 @@ def ensure_active_adapter(self, role: str, model_id: Optional[str]) -> None:
...
+        optimizer half is offloaded reads a dangling device pointer and dies
+        with "CUDA error: invalid argument". Hence we backload with
+        ``need_optimizer=True`` here rather than trusting the caller: paths
+        like ``forward`` and ``save_weights_for_sampler`` deliberately keep the
+        optimizer off GPU. For LoRA (the only case with adapters) the grad
+        buffer and optimizer state cover adapter params only, so this is cheap.
         """
         if model_id is None or role not in self._actor_groups:
             return
+        self._ensure_on_gpu(role, need_optimizer=True, need_model=True)
```
erictang000, line 102: if we have the call order

```
self.forward("ref", ...)
self.forward(model="policy",...,model_id="lora1")
```

won't we unnecessarily run the lora1 forward with the optimizer state on GPU now? can we avoid that in some way even if the grad buffer/optim state should technically be small for lora

erictang000:
closed by #2000

PR #1994, opened 2026-08-05
Retry transient inference proxy unavailability

kalectory:
- Retry a missing or stale vLLM proxy until the engine republishes its URL instead of failing the sample immediately.
- Keep genuine vLLM response errors non-retryable and cover both paths with focused tests.

## Testing

```
uv run --extra dev --extra tinker pytest tests/tinker/skyrl_train/test_inference_forwarding_retry.py
uv run --extra dev pre-commit run --files skyrl/tinker/extra/skyrl_train_inference_forwarding.py tests/tinker/skyrl_train/test_inference_forwarding_retry.py
```

Verified 3 retry tests pass and all configured hooks pass.

### DIFF
--- a/skyrl/tinker/extra/skyrl_train_inference_forwarding.py
+++ b/skyrl/tinker/extra/skyrl_train_inference_forwarding.py
@@ -96,8 +96,8 @@ class SkyRLTrainInferenceForwardingClient:
             await session.commit()
 
     async def _forward_with_retry(self, sample_req, model_id: str, *, base_model: str | None) -> types.SampleOutput:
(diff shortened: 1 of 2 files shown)


avigyabb:
Hey Neil, thanks so much for putting your PR up: https://github.com/NovaSky-AI/SkyRL/pull/1994/changes! If I understand it correctly (🙏) I think my last PR (keeping multi-tenant LoRA runtime warm on last unload: https://github.com/NovaSky-AI/SkyRL/pull/2001) might eliminate the need for proxy retries because the proxy should no longer change even when the number of active tenants hit zero. Let me know if you think differently!

Issue #1654, opened 2026-05-12
[tinker] SkyRLTrainBackend tinker multi-tenancy improvements


erictang000:
1. Improve `delete_model` semantics for multi-tenancy - in the case that we are doing multi-tenant lora training, we should avoid unnecessary ray.shutdown() calls even when the number of active tenants goes to zero. However we should preserve the ray.shutdown behavior for the full parameter finetuning case.
2. Ensure that the case of an incoming request hitting an old proxy url in #1638 is not possible. We have to have some way of upserting new proxy urls for the full parameter fine tuning case, but we should make sure that updating logic is clean and we can never hit an old proxy url.
3. For the multi-tenant case if we can avoid unnecessary shutdowns, then we should only every have one proxy url upsert.

per discussion cc: @pcmoritz

PR #2001, opened 2026-08-07
[tinker] Keep the multi-tenant LoRA runtime warm on last unload

avigyabb:
Implements items 1 and 3 of #1654:

1. delete_model semantics: with the new backend-config key keep_runtime_warm_on_last_unload (default false), unloading the last LoRA policy drops just its adapter slot instead of calling ray.shutdown(); the shared Ray runtime, training workers, inference engines, and base model stay up. The next compatible create_model registers a fresh adapter against the warm runtime (the multi-LoRA registration gate now keys on the runtime being alive rather than on a policy model being registered). Full-parameter fine-tuning keeps the teardown-on-unload behavior.

3. Single proxy-URL upsert: because the inference engines are never torn down in warm mode, the vLLM proxy URL is published to EngineStateDB exactly once for the server lifetime and never cleared or re-upserted, so forwarded sample requests cannot race a teardown/rebuild.

Also:
- delete_model now unregisters the deleted tenant's LoRA adapter from vLLM (tracked via save_sampler_checkpoint), so the engines stop serving deleted tenants and adapter capacity is not leaked. This applies to the existing >1-tenant delete path too.
- Fix an AdapterStore seeding bug exposed by delete-then-create: after deleting the *current* adapter, the live GPU state still mirrors the deleted tenant, but create() treated current_id=None as "live is pristine" and handed the stale weights to the new adapter. The store now tracks live staleness and seeds new slots from pristine, leaving the restore to the next swap_to.

---

> [!NOTE]
> **Medium Risk**
> Changes multi-tenant model lifecycle, GPU adapter state, and vLLM registration; correctness is critical but the PR adds targeted tests and keeps teardown paths for FSDP/FFT.
> 
> **Overview**
> Adds **`keep_runtime_warm_on_last_unload`** on the SkyRL-Train backend (default **true**). For Megatron multi-LoRA policies, unloading the last registered model can drop only that tenant’s adapter instead of calling **`ray.shutdown()`**, so Ray workers, inference engines, and the base model stay resident. **`create_model`** now treats an alive runtime (`_dispatch` set) as enough to register another adapter, which matters when no models are registered but the warm runtime remains. Setting the flag to **`false`** restores scale-to-zero teardown and is the way to change the pinned LoRA **(rank, alpha)** signature without restarting the service.
> 
> **`delete_model`** also unloads LoRA adapters from vLLM when they were registered via **`save_sampler_checkpoint`** (tracked in **`_inference_adapter_ids`**), including the existing multi-tenant delete path. FSDP, full-parameter fine-tuning, and non-LoRA roles still tear down on last unload.
> 
> **`AdapterStore`** tracks **`_live_stale`** after deleting the current adapter so a subsequent **`create`** seeds new slots from pristine instead of treating stale GPU weights as authoritative.
> 
> Multi-tenancy docs describe the flag and warm vs cold lifecycle. CPU unit tests cover warm unload, inference unload, and adapter-store seeding.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3bc0802df46e8585862807c3d6a65ef75bba2c17. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


skyrl/backends/skyrl_train_backend.py
```diff
@@ -49,10 +49,16 @@
 class SkyRLTrainBackendOverrides(BaseModel, extra="allow"):
     """Configuration overrides for the SkyRL-Train backend.
 
-    All keys are applied as overrides to the default SkyRL-Train config.
+    Declared fields configure the backend itself; all extra keys are applied
+    as overrides to the default SkyRL-Train config.
     """
 
-    pass
+    keep_runtime_warm_on_last_unload: bool = False
```
avigyabb: should we allow users to set this?

pcmoritz: Should this be `True` by default? That seems like the more reasonable default to me, curious what you think.

pcmoritz (approved):
Besides the question about the right default, this PR looks good to me!

Might be good to have @erictang000 have a quick look too, he knows this code path very well.

skyrl/backends/skyrl_train_backend.py
avigyabb: I guess I just wanted to keep parity with previous versions, but I agree, I think keeping it True by default makes more sense.

PR #2002, opened 2026-08-07
[fix][ci] Format Tinker engine test

bvolpato:
## Status

Superseded by #2014, which merged the same Black formatting fix into `main`.


bvolpato:
this is now blocking code-quality checks on #1998 and the newer drafts (#2005, #2006, #2007, and #2010). The patch is formatting-only, and its local pre-commit and focused test validation are green. Could we land it so those branches can rerun against formatted `main`?

bvolpato:
Superseded by #2014, which merged the same Black formatting fix into main.

PR #2003, opened 2026-08-07
Initialize base inference before training model creation

j316chuck:
- Initialize non-colocated base-model inference when the service starts, before any training model exists.
- Publish a structured readiness event after the inference proxy reaches the shared engine-state database.
- Reuse the warm inference client when the first training model is created.

## Testing

```
ruff check skyrl/backends/skyrl_train_backend.py skyrl/tinker/engine.py tests/tinker/skyrl_train/test_text_only_batch_no_inference.py
```

Focused backend tests: 10 passed; two unrelated JAX engine tests require the JAX extra.

### DIFF
--- a/skyrl/backends/skyrl_train_backend.py
+++ b/skyrl/backends/skyrl_train_backend.py
@@ -429,6 +429,18 @@ class SkyRLTrainBackend(AbstractBackend):
             self._render_server.shutdown()
             self._render_server = None
 
+    def initialize_base_inference(self) -> None:
+        """Start base-model inference before the first training model is created."""
+        if self._inference_engines_initialized:
+            return
+        self._cfg = _build_skyrl_train_config(self.base_model, self.config)
+        if not ray.is_initialized():
+            logger.info("Initializing Ray for base-model inference")
+            initialize_ray(self._cfg)
+        self._colocate_pg = self._create_colocate_pg() if self._cfg.trainer.placement.colocate_all else None
+        self._create_new_inference_client()
+        self._inference_engines_initialized = True
+
     def _lora_signature_from(self, lora_config: types.LoraConfig) -> tuple:
         # Tinker's public LoraConfig only exposes rank + alpha (plus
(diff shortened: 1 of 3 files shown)


skyrl/backends/skyrl_train_backend.py
```diff
@@ -429,6 +429,18 @@ def _ensure_inference_engines(self):
...
 
+    def initialize_base_inference(self) -> None:
+        """Start base-model inference before the first training model is created."""
+        if self._inference_engines_initialized:
+            return
+        self._cfg = _build_skyrl_train_config(self.base_model, self.config)
+        if not ray.is_initialized():
+            logger.info("Initializing Ray for base-model inference")
+            initialize_ray(self._cfg)
+        self._colocate_pg = self._create_colocate_pg() if self._cfg.trainer.placement.colocate_all else None
```
avigyabb, line 440: It seems like `initialize_base_inference()` is only ever called in non-colocated mode. What is the purpose of having this check here?

avigyabb:
Also thought this might be an issue, but haven’t tested e2e yet:

When we call `_build_skyrl_train_config()` inside `intitialize_base_inference()`, I think the `self._cfg` that is returned does not include the LoRA configs. Since `lora.rank` defaults to 0, when we `build_new_inference_client()` with this config, `_uses_lora_weight_sync()` returns False and it starts vLLM with `enable_lora=False`. When we run `create_model()` on the first request, the configs get updated with the lora configurations, but `_ensure_inference_engines()` doesn’t `_create_new_inference_client()`. Therefore, on the first weight sync / sample, vLLM would error.

I think this is actually fine for Megatron, since by default it uses `merge_lora=True` which merges the LoRA weights with the base model weights (works with vLLM `enable_lora=False`).

What do you think?

j316chuck:
closed as this logic is implemented in https://github.com/NovaSky-AI/SkyRL/pull/2003

PR #2021, opened 2026-08-12
[tinker] Support tinker SDK proto wire format for fwd_bwd and sample

avigyabb:
## What does this PR do?

Fixes the nightly tinker cookbook e2e failures that started 2026-08-12:

```
ValueError: Server returned a JSON payload for self.model_cls=<class 'tinker.SampleResponse'> ...,
which this SDK version only supports as proto.
```

The e2e scripts install the tinker SDK unpinned, and **tinker 0.25.0 (released 2026-08-08) made its proto wire paths mandatory**:

1. **Responses**: `retrieve_future` results for `sample` / `forward` / `forward_backward` must be proto-serialized when the client sends `Accept: application/x-protobuf`. Since 0.18.0 the SDK has sent `Accept: application/x-protobuf, application/json` and deserialized whichever format the server returned — JSON was a true client-side fallback, which is why our JSON-only server worked. 0.25.0 removed that fallback for these two response types. (SDKs <0.18.0 have no proto module at all and speak pure JSON.)
2. **Requests**: `forward_backward` bodies are submitted as proto (`Content-Type: application/x-protobuf`), and forward-only passes are routed to `/api/v1/forward_backward` via the proto `forward_only` flag instead of `/api/v1/forward`. Unlike the response path, this was never a fallback arrangement before 0.25.0: the proto write path (present since ~0.22) was gated by the server-supplied `client_config` flag `proto_write_fwdbwd` (default false), and our stub `client_config` never opted in, so JSON requests were simply the default. 0.25.0 removed the flag and made proto requests unconditional.

## Changes

- **`skyrl/tinker/proto_serialization.py`** (new): mirrors the SDK's `tinker/proto/{request,response}_conv.py`. Serializers for `SampleResponse` / `ForwardBackwardOutput` (NaN encodes undefined prompt logprobs, `token_id=0 / logprob=-99999.0` sentinel fill for undefined top-k entries, batched int64 byte offsets for per-datum tensors), plus a parser for proto `ForwardBackwardRequest` bodies. The proto schema ships inside the `tinker` package, which the `tinker` extra already depends on -- no new dependency.
- **`skyrl/tinker/api.py`**: `retrieve_future` content-negotiates on the `Accept` header (errors and all other result types stay JSON); `forward_backward` accepts both wire formats and maps `forward_only=true` to the `FORWARD` request type. `poll_futures`/`wait_for_future` now carry `request_type` alongside `(status, result_data)`.
- Old JSON paths (including `/api/v1/forward`) are unchanged, so SDKs <=0.24.x keep working.

## Testing

- New `tests/tinker/test_proto_serialization.py` round-trips the server encoding through the installed SDK's own proto codecs (12 tests), plus a proto-path test in `test_future_waiting.py`.
- Manually verified a full loop (`forward_backward` -> `forward` -> `optim_step` -> `save_weights_for_sampler` -> `sample`) against a local JAX-backend server with **12 SDK versions spanning the supported range** (`tinker>=0.3.0`) -- all pass with this change; before it, 0.25.0 failed exactly like the nightly:

  | SDK versions | Wire behavior | Result |
  |---|---|---|
  | 0.3.0, 0.8.1, 0.13.1, 0.16.1 | pure JSON (no proto module) | ✅ |
  | 0.18.2, 0.20.0, 0.21.0, 0.22.0, 0.22.4, 0.23.4, 0.24.1 | proto responses (JSON-fallback), JSON requests | ✅ |
  | 0.25.0 | proto responses + proto requests, no fallback | ✅ |
- `tests/tinker/` CPU suite passes (`skyrl_train/` GPU-cluster tests excluded -- they fail in my workspace with a pre-existing Ray version mismatch unrelated to this change).

Notes for reviewers:
- SDK >=0.18.0 sends `Accept: application/x-protobuf, application/json` on `retrieve_future` and understands proto responses, so the existing `test_api.py` integration tests (locked SDK 0.22.4) now exercise the proto response path end-to-end.
- The SDK's zstd request compression (`proto_compress_fwdbwd`) is negotiated via `client_config`, which our server does not advertise, so compressed bodies never arrive.
- I left the e2e scripts' `--with tinker` unpinned on purpose: catching SDK/server incompatibilities is exactly what those nightlies are for. Happy to pin to `tinker==0.25.0` instead if determinism is preferred.

---

> [!NOTE]
> **Medium Risk**
> Changes the Tinker API wire contract on hot training paths (forward/backward and sampling) with binary encoding; mistakes could break all 0.25.0 clients while JSON fallbacks must stay correct for older SDKs.
> 
> **Overview**
> Fixes nightly cookbook failures against **tinker SDK 0.25.0**, which requires protobuf for `sample` / `forward` / `forward_backward` results on `retrieve_future` and sends `forward_backward` bodies as protobuf (including forward-only via a `forward_only` flag).
> 
> A new **`proto_serialization`** module mirrors the SDK’s proto codecs: it parses proto `ForwardBackwardRequest` bodies and serializes stored JSON results into `SampleResponse` and `ForwardBackwardOutput` wire bytes (tokens, logprobs, top-k sentinels, batched tensors).
> 
> **`retrieve_future`** negotiates on `Accept: application/x-protobuf` for SAMPLE, EXTERNAL, FORWARD, and FORWARD_BACKWARD; errors and other futures stay JSON. **`forward_backward`** reads JSON or protobuf and maps `forward_only=true` to the FORWARD request type. Future polling now passes **`request_type`** through waiters. JSON **`/forward`** and legacy SDK paths are unchanged.
> 
> Round-trip tests use the installed tinker SDK deserializers; **`EXTERNAL`** sample futures (external inference) serialize like **`SAMPLE`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd8043b979ce787ba6decb404e9bec0b61500a6e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## GPU e2e validation (2026-08-17)

Both CI e2e tests were run as Anyscale jobs on this branch (4xL4, exact CI configs, cookbook client installing the **unpinned** tinker SDK -> 0.25.0, the strict proto-only client):

- `gsm8k_tinker` (colocated, FSDP backend): **SUCCEEDED** (55m50s)
- `gsm8k_tinker_fully_async` (non-colocated): **SUCCEEDED** (56m28s)

The first attempt surfaced a real gap the local JAX-backend testing missed: `/asample` stores futures as `request_type=EXTERNAL` (not `SAMPLE`) whenever an external inference client is configured -- which is how the fsdp/megatron backends run sampling -- and those results bypassed the proto gate, reproducing the original failure. Fixed in `7c72fff0` (EXTERNAL results serialize as `SampleResponse`; failures still return JSON errors), with a regression unit test. Both jobs passed on the fixed commit, including the wandb reward assertions.

## Performance

Microbenchmark of the response paths at CI-realistic payload shapes (single CPU core, round-tripped through the SDK's own deserializers; script: `serialize_result` vs the raw-JSON passthrough):

| Payload | Wire size (JSON -> proto) | Server cost per result | Client parse (JSON -> proto) |
|---|---|---|---|
| sample, 4 seqs x 512 tokens (CI shape) | 51.5 KiB -> 16 KiB (3.2x) | +0.78 ms | 0.57 ms -> 0.02 ms |
| sample + top-20 prompt logprobs, 2048-token prompt | 1.17 MiB -> 344 KiB (3.5x) | +34 ms | 21.9 ms -> 0.07 ms |
| fwd_bwd, 512 datums x 600 tokens x 2 float fields | 11.4 MiB -> 2.4 MiB (4.8x) | +170 ms | 144.9 ms -> 2.8 ms |

- **Client + wire: strictly faster.** Proto payloads are 3-5x smaller and client-side deserialization is `np.frombuffer` over binary instead of JSON-parsing decimal floats (~50x faster on large payloads; the JSON column above is `json.loads` only, a lower bound -- the SDK adds model construction on top).
- **Server: a bounded regression vs the current JSON path, which #2017 made a zero-copy passthrough.** The proto path pays `json.loads(stored text) + validate + proto encode` per result: sub-millisecond for typical samples, ~170 ms worst-case for a full 512-datum fwd_bwd result. ~85% of that is the `json.loads` of the stored JSON text (a storage-format artifact, not a proto cost); the pre-#2017 baseline (`jsonable_encoder` + `json.dumps`) was ~300 ms on the same payload class, so this is still cheaper than the JSON path of two weeks ago. The SDK also chunks fwd_bwd requests (~5 MB estimated per chunk), so single results near the worst case are uncommon.
- **End-to-end: indistinguishable.** The GPU e2e runs are GPU-bound; proto-client (0.25.0) and JSON-client (0.24.x) runs both complete in 55-56 min.

Possible follow-ups (not in this PR): offload `json.loads + serialize` to a thread in `retrieve_future` so a worst-case result doesn't stall the event loop; longer term, store proto bytes in `FutureDB.result_data` for the four proto-serializable request types, which removes the parse entirely and makes every SDK >= 0.18.0 (all of which prefer proto via the Accept header) a zero-copy read.

### DIFF
--- a/.claude/docs/tinker.md
+++ b/.claude/docs/tinker.md
@@ -5,6 +5,7 @@ SkyRL's implementation of the [Tinker API](https://tinker-docs.thinkingmachines.
 ## Code Layout
 
 - **`skyrl/tinker/api.py`** -- FastAPI HTTP server. Receives Tinker SDK requests, writes them to SQLite/Postgres, returns future IDs.
+- **`skyrl/tinker/proto_serialization.py`** -- Proto wire-format conversion. SDK >= 0.25.0 sends `forward_backward` request bodies as protobuf (with forward-only passes routed to `/forward_backward` via a `forward_only` flag instead of `/forward`) and requires proto-serialized `SampleResponse`/`ForwardBackwardOutput` results from `retrieve_future`. Mirrors the SDK's `tinker/proto/{request,response}_conv.py`; round-trip tested in `tests/tinker/test_proto_serialization.py`.
 - **`skyrl/tinker/engine.py`** -- Background subprocess (`TinkerEngine`). Polls DB, batches compatible requests, dispatches to backend.
 - **`skyrl/tinker/types.py`** -- Internal Pydantic models (distinct from API request/response models in `api.py`). `LOSS_TYPES` dict defines valid loss functions.
 - **`skyrl/tinker/config.py`** -- `EngineConfig` Pydantic model. `add_model()` auto-generates argparse flags from Pydantic fields.
--- a/skyrl/tinker/api.py
+++ b/skyrl/tinker/api.py
@@ -1083,15 +1083,15 @@ async def get_training_run(model_id: str, session: AsyncSession = Depends(get_se
     )
 
 
-async def _read_forward_backward_request(req: Request) -> tuple[ForwardBackwardRequest, bool]:
+async def _read_forward_backward_request(request: Request) -> tuple[ForwardBackwardRequest, bool]:
     """Read a forward_backward body in either wire format.
 
     tinker SDK >= 0.25.0 submits the body as protobuf and routes forward-only
     passes here via the proto's ``forward_only`` flag instead of calling
     ``/api/v1/forward``; older SDKs keep sending JSON with forward_only False.
     """
-    body = await req.body()
-    if PROTO_CONTENT_TYPE in req.headers.get("content-type", ""):
+    body = await request.body()
+    if PROTO_CONTENT_TYPE in request.headers.get("content-type", ""):
         try:
             request_dict, forward_only = parse_forward_backward_request(body)
         except (DecodeError, ValueError) as e:
@@ -1108,17 +1108,17 @@ async def _read_forward_backward_request(req: Request) -> tuple[ForwardBackwardR
 
 
 @app.post("/api/v1/forward_backward", response_model=FutureResponse)
-async def forward_backward(req: Request, session: AsyncSession = Depends(get_session)):
+async def forward_backward(request: Request, session: AsyncSession = Depends(get_session)):
     """Compute and accumulate gradients (or run forward-only when the proto body asks for it)."""
-    request, forward_only = await _read_forward_backward_request(req)
-    await get_model(session, request.model_id)
+    req, forward_only = await _read_forward_backward_request(request)
+    await get_model(session, req.model_id)
 
     request_id = await create_future(
         session=session,
         request_type=types.RequestType.FORWARD if forward_only else types.RequestType.FORWARD_BACKWARD,
-        model_id=request.model_id,
-        request_data=request.forward_backward_input.to_types(),
-        seq_id=request.seq_id,
+        model_id=req.model_id,
+        request_data=req.forward_backward_input.to_types(),
+        seq_id=req.seq_id,
     )
 
     await session.commit()
--- a/skyrl/tinker/api.py
+++ b/skyrl/tinker/api.py
@@ -1,4 +1,5 @@
 import asyncio
+import json
 import os
 import random
 import re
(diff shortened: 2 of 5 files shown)


pcmoritz (approved):
This looks great, thanks a lot for implementing it. We can merge it once the merge conflicts are fixed and the tests pass.

PR #2023, opened 2026-08-13
[deps] Source causal-conv1d and mamba-ssm from Astral's GPU wheel index

erictang000:
Replaces the personally-built fork wheels for `causal-conv1d` and `mamba-ssm` with the official prebuilt wheels from Astral's GPU index (https://wheels.astral.sh), pinned to the cu128 / torch-2.11 builds.

Besides removing the dependency on a personal GitHub account, this widens platform coverage: the fork wheels were py3.12 / x86_64 only, while Astral publishes cp310-cp314 for both x86_64 and aarch64, so the `[tool.uv.sources]` entries no longer need a python/arch marker.

`flash-attn` and `transformer-engine-torch` stay on the fork wheels. Astral publishes torch-2.11 builds of both, but neither works with the transformer-engine 2.11.0 we pin — both blockers are documented inline and both clear once the TE stack moves to 2.16.

mamba-ssm 2.3.2.post1 pins tilelang==0.1.8 against vLLM 0.23.0's ==0.1.9, so tilelang is added to override-dependencies. 0.1.9 is what the lock already resolved to, so no other package's version changes.

Verified on H100: causal_conv1d_fn and Mamba2 forward/backward produce finite values, and TE still accepts flash-attn (is_installed=True).


pyproject.toml
```diff
@@ -255,6 +255,11 @@ override-dependencies = [
     # cp312 linux x86_64 wheel (only cp310/cp311 + aarch64/macOS), so resolving
     # to them breaks install on x86_64 py3.12. 0.2.3 has the wheel.
     "xgrammar==0.2.3",
+    # mamba-ssm 2.3.2.post1 pins tilelang==0.1.8 while vLLM pins ==0.1.9.
+    # 0.1.9 is what the environment already resolved to before mamba-ssm moved to
+    # the Astral wheel, so this only relaxes mamba-ssm's pin — no other package's
+    # resolved version changes.
```
erictang000:
```suggestion
```

PR #1854, opened 2026-07-01
[chore] bump vllm to 0.26.0

erictang000:
^

---

> [!NOTE]
> **High Risk**
> Touches critical inference startup and weight-sync paths (NCCL + delta), plus the core vLLM/flashinfer dependency stack used by all GPU training.
> 
> **Overview**
> Bumps **vLLM 0.23 → 0.26.0** (cu129 wheel) and applies the compatibility fixes needed to keep install, server startup, and weight sync working.
> 
> **Dependency / install fixes:** pins `flashinfer*` to 0.6.13 (MNNVL `layout_code` support; cubin has no 0.6.14), overrides `xgrammar==0.2.3` for a cp312 linux wheel, and poisons `flash_attn.cute` at import time so megatron-core's FA4 probe no longer crashes on cutlass-dsl 4.6.
> 
> **Server startup:** passes the new required `reuse_port=False` to `create_server_socket`, and seeds `VLLM_DP_MASTER_PORT` with a free ephemeral port so RayExecutorV2 no longer tries privileged port 100 when DP is off.
> 
> **Weight sync API drift:** NCCL sync now retargets the engine via `set_weight_update_target` + a `_LoadWeightsProxy` (0.26 dropped the `load_weights` callback). `DeltaWeightTransferEngine` adopts the 4-arg constructor and the full `update_weights` / start / finish lifecycle.
> 
> **Tests / misc:** multimodal `mm_serde` import falls back across the moved module path; GLM MLA coverage swaps the unservable tiny fixture for real `zai-org/GLM-4.7-Flash` in the h100 job. Adds `VLLM_0.26_UPGRADE_NOTES.md` documenting each required fix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8d61888b40a6a8e1e28e7e786cb1b0a2435bf588. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #2027, opened 2026-08-13
[megatron] 5/n towards Kimi K2.6: Kimi K2.5-family (KimiK25ForConditionalGeneration) support with INT4 QAT + LoRA

casper-hansen:
## What

- `KimiK25TextBridge`: DeepSeek-V3 bridge adapted to the unified VL checkpoint (`text_config` + `language_model.` weight prefix); text-only training with the vision tower left frozen in the inference engine (`language_model_only`), with a clear error for full-VLM requests.
- vLLM: declare `KimiK25ForConditionalGeneration` LoRA-capable (patch applied in the SkyRL worker extension import; the DeepSeek decoder already supports LoRA). The worker-side LoRA gate is an instance-level runtime Protocol check that requires every `SupportsLoRA` member (`is_3d_moe_weight`, `is_non_gated_moe`, `lora_skip_prefixes` too); we mirror the DeepSeek decoder's declarations and pin it with a protocol-membership test.
- Forward HF cache/auth env vars (`HF_HOME` & co) to Ray workers so model paths resolve identically on worker actors.
- GPU CI on published 2-layer slices of the real checkpoint: a `test_megatron_models` row running the production INT4-served / BF16-fake-quant path end to end, plus the adapter-export test.

## Why

Kimi K2.5-family checkpoints (K2.6, K2.7-Code) are unified VL checkpoints whose language model is DeepSeek-V3 architecture. The bridge megatron-bridge ships for the arch name builds a vision tower this backend cannot train, so training these checkpoints text-only needs its own bridge.

## Dispatch vs. the upstream VL bridge

megatron-bridge registers `KimiK25ForConditionalGeneration` for its own `KimiK25VLBridge` -> `KimiK25VLModel` (`megatron/bridge/models/kimi_vl/`). `model_bridges` registers `KimiK25TextBridge` under the same name; the dispatch registry is last-write-wins and this module is imported after `megatron.bridge.models`, so the text bridge overrides it.

Because that is import order rather than a guarantee, `init_configs` now asserts the resolved bridge is `KimiK25TextBridge` and raises otherwise, instead of logging that it happened. Verified both ways: the normal path resolves to the text bridge, and re-registering the upstream bridge last flips the dispatch and trips the assertion.

## Testing

Two 2-layer slices of `moonshotai/Kimi-K2.5` — layer 0 dense + layer 1 with all 384 routed experts (`first_k_dense_replace=1`), plus embedding/lm_head and the vision tower vLLM always builds:

- [`eatang/Kimi-K2.5-2layer`](https://huggingface.co/eatang/Kimi-K2.5-2layer) — INT4, compressed-tensors weights byte-identical to the release (16 GB)
- [`eatang/Kimi-K2.5-2layer-BF16`](https://huggingface.co/eatang/Kimi-K2.5-2layer-BF16) — routed experts dequantized with `dequantize_compressed_tensors_int4.py`, every tensor verified as a fake-quant fixed point (41 GB)

**`test_megatron_models.py::test_logprobs_matching_roundtrip[kimi-k2.5-2layer-int4-qat_h100_tp4_ep4]`** (4xH100, TP4/EP4/ETP1, vLLM TP4 colocated) runs the production recipe: vLLM serves the INT4 slice, Megatron loads BF16 masters through `fake_int4_qat.bf16_base_path` and fake-quantizes its grouped expert GEMMs to the same grid, and LoRA (`merge_lora=false`) syncs back as a PEFT adapter — INT4 base weights cannot take a bf16 broadcast. The Megatron-vs-vLLM bound is the signal: a forward on BF16 masters with fake-INT4 experts has to reproduce what the INT4 engine serves.

Measured: **0.029** mean |dlogprob| Megatron vs vLLM, **0.008** on the post-sync common-prefix check (bounds 1e-1 / 1e-1).

Two changes to the shared test body, both no-ops for existing rows: the pre-generation sync (previously keyed on `fp8_mode`) now also covers adapter-sync rows, since the policy is served under the adapter name and no adapter exists until the first sync; and the generator is given `policy_model_name` only when adapter sync is on. The `qwen3-moe` row was re-run unchanged to confirm.

**`test_kimi_k25_bridge.py`** previously required `moonshotai/Kimi-K2.7-Code` in the local HF cache (595 GB) plus a 2.1 TB dequantized master directory, so its checkpoint-gated tests never actually ran. They now default to the slices:

- `test_kimi_bridge_dispatch_and_provider` deleted — the parity row covers dispatch, the nested `text_config`, the rope/MoE provider fields and the `language_model.` prefix invariant harder, by requiring the resulting forward to match what vLLM serves.
- the adapter-export test stays: the parity row's LoRA is zero-initialized, so parity is blind to the export layout (per-expert keys, flat 2D tensors, an adapter pointing at the INT4 release). It loads ~41 GB of masters on one GPU, so it carries `pytest.mark.h100` and joins the H100 CI run; the L4 `-m megatron` job keeps the checkpoint-free protocol test and auto-skips the heavy one.

## Merge

`main` merged in, which moves this branch to megatron-bridge `8e5f0e1` and vLLM 0.28.0. The vLLM LoRA patch is still required there: `KimiK25ForConditionalGeneration` declares no `SupportsLoRA`, and all seven protocol members are missing before the patch is applied.

**Stacked on 4/n (#2026)** -- the GPU test exercises `normalize_moe_lora`; the diff includes 4/n until it merges. Part of the Kimi K2.x series (follow-up to #1862).

Made with [Cursor](https://cursor.com); the merge and the GPU CI entries above were made with [Claude Code](https://claude.com/claude-code).


casper-hansen:
Updated for the vLLM 0.26 bump (#1854, now on `main`) and merged `main` in (via the 4/n base branch):

- vLLM 0.26 added an annotation-only `lora_manager` member to the `SupportsLoRA` protocol. Native models are nominal subclasses and skip the structural check, but the patched `KimiK25ForConditionalGeneration` is checked structurally via a runtime-checkable Protocol `isinstance()`, which requires every member to *exist* on the class. The patch now sets `cls.lora_manager = None` (the documented "not yet set" state); the worker's LoRA mixin assigns the real manager at runtime.
- `KimiK25TextBridge.build_conversion_tasks` now forwards `*args, **kwargs` so it stays compatible with megatron-bridge 0.7.0 (#2042), which grows the base signature with a `weight_dtype` keyword. If #2042 lands first, its module-level None-task filter makes our per-bridge filter redundant but harmless; if this lands first, the filter is still needed on the current bridge pin. Heads-up that #2042 and this PR both touch the same import block in `model_bridges.py`, so whichever merges second will need a trivial conflict resolution.

The rest of the K2.5 bridge (mapping re-rooting via `get_common_mapping_list`, `_TextConfigShim`, worker-wrap patch application) is unchanged: the 0.7.0 `get_common_mapping_list` keeps its signature, and `hf_param` is still only read at resolve/export time, so the post-construction prefixing stays valid.

PR #2033, opened 2026-08-13
feat: support Nemotron 3.5 Lightning training

j316chuck:
## Fixes

- Pin vLLM, Megatron Bridge, and Megatron Core revisions that load Nemotron 3.5 Lightning.
- Start non-colocated inference before trainer GPU allocation, then reconnect weight sync after policy creation.
- Pass vLLM socket reuse, preserve explicit eager mode, and fall back from fused LM-head projection for HybridModel.
- Use the current Megatron checkpoint queue API so saves finalize.
- Remove the five-minute Tinker forwarding read timeout while retaining the ten-second connect timeout.

## Verification

- 6 focused SkyRL tests pass.
- Baked image: us-central1-docker.pkg.dev/useful-memory-477923-v7/skyrl/megatron@sha256:064cb5cef0b0be163d4f03af577002367c6515a3bd68ca72d125f7ff6b4d5e3c


erictang000:
hey @j316chuck, bumping megatron deps right now, will try to make sure the nemotron lightning is working when they're done

PR #2035, opened 2026-08-13
[fix][tinker] Make retried training requests idempotent

erictang000:
## Problem

The Tinker SDK wraps every training HTTP send in `execute_with_retries`, which retries for up to 5 minutes on any retryable exception. So when a `forward_backward` / `forward` / `optim_step` reaches the server but its response is lost — connection reset, 5xx, timeout — the SDK resends the identical request.

The server creates a fresh future for every POST, so that resend is executed a second time:

- a duplicated `forward_backward` adds its gradients to the pending accumulation twice
- a duplicated `optim_step` takes two optimizer steps

Both corrupt training silently. Nothing surfaces an error, and the client can't tell the difference — it just sees a successful response to its retry.

## Fix

The SDK already stamps every one of these requests with a per-request `seq_id` (`seq_id=request_id + 1`, drawn from one counter shared across request types on the training client). This PR threads that value through to `create_future` and uses it as an idempotency key:

- New `training_requests` table maps `(model_id, seq_id) -> request_id`, keyed on the pair, with `request_id` unique.
- A repeat POST carrying a `(model_id, seq_id)` already on file returns the **original** `request_id` instead of enqueueing new work. The client polls the original future and receives the original result, so a retry is observationally identical to the first attempt succeeding.
- The same `(model_id, seq_id)` arriving with a different request type or a different payload is a **409** rather than a silent aliasing of unrelated work.
- The mapping insert is guarded by the table's primary key. If a concurrent retry inserts first, the `IntegrityError` is caught, the session rolls back, and the winner's `request_id` is returned — so two racing retries still produce exactly one future.

`request_type` is deliberately *not* part of the key: the SDK's counter is shared across request types for a given training client, so `seq_id` alone is already unique per model. Keying on the pair and *verifying* the type gives a 409 on a genuine reuse instead of quietly partitioning it into a separate namespace.

## Backwards compatibility

`seq_id` is optional throughout. Requests that arrive without one — older clients, hand-built requests, endpoints not yet wired up — take the original path and get a fresh future every time. No mapping row is written, so nothing is deduplicated that wasn't before. Covered by `test_training_requests_without_sequence_numbers_remain_distinct`.

## Tests

`tests/tinker/test_training_request_idempotency.py`, 4 new tests against a real SQLite-backed session:

| Test | Asserts |
|---|---|
| `test_training_request_retry_returns_original_future` | A replayed request returns the first `request_id`; exactly one `FutureDB` row exists |
| `test_training_request_sequence_reuse_with_new_payload_fails` | Same `seq_id` + changed payload → `HTTPException` 409 |
| `test_concurrent_training_request_types_cannot_reuse_sequence` | Two concurrent requests racing on one `seq_id` (WAL enabled) → exactly one success, one 409, one future, one mapping |
| `test_training_requests_without_sequence_numbers_remain_distinct` | `seq_id=None` still yields distinct futures |

```bash
uv run --isolated --extra tinker --extra jax --extra dev pytest --forked -s \
  tests/tinker/test_training_request_idempotency.py
```

4 passed.

## Known limitation

The uniqueness of `(model_id, seq_id)` rests on one model_id being driven by a single SDK `TrainingClient` instance. `TrainingClient._request_id_counter` is per-instance and starts at 0, so two clients addressing the same model_id would both replay `seq_id` 1, 2, 3… — an identical payload would be treated as a retry and its work dropped, and a differing one would 409. Note that the SDK's *sampling* client guards against exactly this by seeding its counter with a random offset, while the training client does not.

This is a pre-existing property of the SDK's request numbering rather than something this PR introduces, and it is not reachable through the normal `create_lora_training_client` flow (each call gets a distinct model_id). Worth revisiting if we ever support attaching a second client to a live model.

## Note

Split out of #2008, which now carries only the Megatron cross-request gradient accumulation change. The two are independent and can merge in either order.

---

> [!NOTE]
> **Medium Risk**
> Changes how training futures are created and can affect gradient accumulation and optimizer steps on retries; behavior is well-tested but touches core training request handling.
> 
> **Overview**
> **Makes retried training requests idempotent** so SDK `execute_with_retries` cannot silently corrupt training when a request reaches the server but the response is lost.
> 
> Adds a `training_requests` table mapping `(model_id, seq_id)` to a single `request_id`. `create_future` accepts optional `seq_id`; a replay with the same model, sequence, request type, and payload returns the **original** `request_id` instead of a new future. The same `seq_id` with a different payload or type yields **409**. Concurrent retries race on the primary key; the loser rolls back and returns the winner’s id.
> 
> Optional `seq_id` is threaded through `forward_backward`, `forward`, and `optim_step` request models and handlers. Requests without `seq_id` keep the old behavior (distinct futures every time).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2e98fcb503bc9b014c91ba59844c81fbea074816. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/skyrl/tinker/db_models.py
+++ b/skyrl/tinker/db_models.py
@@ -100,6 +100,17 @@ class FutureDB(SQLModel, table=True):
     completed_at: datetime | None = Field(default=None, sa_type=DateTime(timezone=True))
 
 
+class TrainingRequestDB(SQLModel, table=True):
+    """Maps an SDK training sequence number to its single server-side future."""
+
+    __tablename__ = "training_requests"
+
+    model_id: str = Field(primary_key=True)
+    request_type: types.RequestType = Field(primary_key=True)
+    seq_id: int = Field(primary_key=True)
+    request_id: int = Field(foreign_key="futures.request_id", unique=True)
+
+
 class CheckpointDB(SQLModel, table=True):
     __tablename__ = "checkpoints"
 
--- a/skyrl/tinker/db_models.py
+++ b/skyrl/tinker/db_models.py
@@ -101,12 +101,11 @@ class FutureDB(SQLModel, table=True):
 
 
 class TrainingRequestDB(SQLModel, table=True):
-    """Maps an SDK training sequence number to its single server-side future."""
+    """Maps a model-global SDK training sequence number to one future."""
 
     __tablename__ = "training_requests"
 
     model_id: str = Field(primary_key=True)
-    request_type: types.RequestType = Field(primary_key=True)
     seq_id: int = Field(primary_key=True)
     request_id: int = Field(foreign_key="futures.request_id", unique=True)
 
--- a/skyrl/tinker/db_models.py
+++ b/skyrl/tinker/db_models.py
@@ -3,7 +3,7 @@
 from datetime import datetime, timezone
 from enum import Enum
 
-from sqlalchemy import DateTime, event
+from sqlalchemy import DateTime, UniqueConstraint, event
 from sqlalchemy.engine import url as sqlalchemy_url
 from sqlmodel import JSON, Field, SQLModel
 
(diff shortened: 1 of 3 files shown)


erictang000:
@cursor review

skyrl/tinker/db_models.py
```diff
@@ -100,6 +100,16 @@ class FutureDB(SQLModel, table=True):
     completed_at: datetime | None = Field(default=None, sa_type=DateTime(timezone=True))
 
 
+class TrainingRequestDB(SQLModel, table=True):
```
pcmoritz: Instead of introducing this new table, we might want to introduce an optional seq id in `FutureDB` (and we could have a uniqueness constraint on `(model_id, seq_id)`)

erictang000: done!

pcmoritz (approved):

PR #2042, opened 2026-08-14
[deps][megatron] bump megatron-bridge to 0.7.0 and megatron-core to 0.20.0

erictang000:
| Package | From | To |
| --- | --- | --- |
| `megatron-bridge` | `91a15142` (0.6.0, 2026-06-05) | `2b8cb21d` (0.7.0, 2026-08-13) |
| `megatron-core` | `71e418ea` (0.19.0, 2026-06-08) | `14346b65` (0.20.0, 2026-08-12) |

710 megatron-bridge commits and 439 megatron-core commits. `2b8cb21d` is the tip
of megatron-bridge `main`.

## The megatron-core rev is not a free choice

megatron-bridge vendors Megatron-LM as a git submodule at `3rdparty/Megatron-LM`
and resolves `megatron-core` through a path source pointing at it. **That
submodule pointer is the only megatron-core commit Bridge's CI ever builds
against**, and nothing in either package's published metadata expresses the
constraint — Bridge declares `megatron-core[dev,mlm]` unpinned, and uv ignores a
dependency's own `[tool.uv.sources]`. So the rev to use is:

```
$ git ls-tree 2b8cb21d90548a5796c748f1986c77fc1cfe2417 3rdparty/Megatron-LM
160000 commit 14346b65a2d0790e451919858f7771078105c5f0	3rdparty/Megatron-LM
```

`14346b65` is on Megatron-LM `main` (not `dev`), one day behind the `main` tip —
so "most recent" and "what Bridge validates" coincide here. The derivation is now
a comment above the pin so the next bump doesn't have to rediscover it.

For the record, the previous pin did not follow this rule: bridge `91a15142`'s
submodule was `64c3fb86`, but we pinned `71e418ea` (the Megatron-LM main tip at
the time).

No TE change was needed — megatron-core 0.20 asks for
`transformer-engine[pytorch,core_cu13]` unpinned and the 2.16 stack from #2039
satisfies it.

## Code changes required

Five upstream changes needed fixes. Two are ordinary API breaks; three are
behavioral, and **the last produced no error at all — just wrong numbers**.

### `dist_checkpointing.strategies.base` was deleted

megatron-core `a2496aa17` removed the module. We imported it, but not for
anything in it — it was being used as a namespace to stash our own global:

```python
ckpt_base.async_calls = AsyncCallsQueue(persistent=True)
```

`async_calls` was never a megatron-core attribute. The global now lives in
`megatron_strategy.py`. It stays module-level because `_finalize_async_calls` is
a `@staticmethod` invoked from a callback with no strategy handle.

### `build_conversion_tasks` gained `weight_dtype`

Passed by keyword from `load_weights_hf_to_megatron`, and overrides must forward
it. `GLM47FlashBridge` now does, matching upstream's own `kimi_k3_bridge` /
`kimi_k25_vl_bridge` overrides. Would have been a `TypeError` at GLM-4.7-Flash
weight-load time.

### `linear_attention_freq` became a per-layer list

`full_attention_interval_from_hf` changed its return type from `int` to
`int | list[int]`. It used to return an *interval* (e.g. `4`) that megatron-core
expanded against `num_layers`; it now returns an explicit pattern sized for the
**full** model. megatron-core asserts `len(pattern) == num_layers` (that assert
is unchanged — the bridge-side return type moved), so any `num_layers` override
on a hybrid model stopped building:

```
AssertionError: Invalid length of linear_attention_pattern: 40, expected 2
```

`init_configs` now truncates the pattern. Truncation is faithful to the old
behavior: for Qwen3.5-35B-A3B the old interval of 4 expanded to `[1, 1]` at
`num_layers=2`, exactly the first two entries of the new 40-entry pattern. Shrink
only — a pattern shorter than `num_layers` is a real misconfiguration and still
trips the upstream assert.

### Conversion tasks that no PP rank owns

0.7.0 registers **both** MoE expert layouts — grouped-GEMM
(`mlp.experts.linear_fc1`) and SequentialMLP
(`mlp.experts.local_experts.*.linear_fc1`); 0.6.0 registered grouped only. A
model uses one layout, so `get_conversion_tasks` now returns tasks for the other,
with `param_weight` `None` on *every* PP rank. `_init_param_buckets` broadcasts
each task's size across PP ranks to size weight-sync buckets, and hit
`ValueError: Object must exist on at least one PP rank`.

`broadcast_object_across_pp_ranks` gained an opt-in `allow_missing` so a caller
enumerating tasks can tell "nobody owns this" from an error; those tasks are
dropped when bucketing. They carry no weights, so this restores the 0.6.0 task
set exactly.

### Unmapped (None) conversion tasks

`build_conversion_tasks` is typed `List[None | WeightConversionTask]` and leaves
a None slot for any Megatron parameter its registry has no entry for. Upstream's
consumers guard `task.megatron_module is None` but not `task is None`:

```
AttributeError: 'NoneType' object has no attribute 'megatron_module'
```

The None-fill loop is byte-identical in 0.6.0 — what changed is *which*
parameters end up unmapped. This broke the router-replay tests
(Moonlight-16B-A3B, tp2/pp2/ep2): 27 of 1989 tasks unmapped, one at the same
per-layer offset in every layer.

We have carried this exact workaround in `GLM47FlashBridge` since megatron-bridge
0.3.1; it now applies to every bridge and logs the dropped parameter names once
per process. The router-replay tests are themselves megatron-vs-vLLM logprob
equivalence tests and pass with the drops applied, which is good evidence the
dropped entries are Megatron-internal state rather than weights.

### Packed vs per-expert MoE storage — silent numerical corruption

**This is the one to review carefully.**

Qwen3.5 routed experts are stored either fused (`experts.gate_up_proj`, one
stacked tensor per projection) or per-expert (`experts.<i>.gate_proj.weight`),
depending on the transformers version that wrote the checkpoint. Bridge 0.6.0
hardcoded the fused layout. 0.7.0 turned it into an argument:

```python
def _get_moe_lm_mappings(hf_prefix="model.", megatron_prefix="", experts_packed=False)
```

...defaulting to `False`, and upstream's own `Qwen35MoEBridge.mapping_registry`
now detects the layout with `moe_experts_stored_packed(...)`. Our
`Qwen35MoELMBridge` override did not, so it silently took the per-expert default.

Qwen3.5-35B-A3B stores its LM experts fused — 80
`model.language_model.layers.*.mlp.experts.gate_up_proj` keys in the safetensors
index. The generated mappings matched nothing, **nothing raised**, and every
routed-expert weight kept its initialized value:

```
test_logprobs_matching_roundtrip[qwen3.5-35b-a3b_h100_tp4_ep4]
AssertionError: Logprob diff should be less than 0.05, but is 2.838971
```

Fixed by detecting the layout the way upstream does, with the VLM key prefix.

**Lesson worth generalizing:** a *default argument value* changed the meaning of
an existing call. No signature check, import check, or type checker catches that
— only a numerical test does. Other overrides calling into upstream mapping
helpers deserve the same scrutiny. (`_get_dense_lm_mappings` was checked: its
parameters became keyword-only, which we already satisfy, and the dense path has
no packed/unpacked switch.)

## Dependency change

One override:

```toml
"fast-hadamard-transform; sys_platform == 'never'",
```

megatron-bridge 0.7.0 added it as a hard dependency and it breaks `uv sync`
outright (`ModuleNotFoundError: No module named 'torch'` at build time). It
publishes **no wheels** — sdist only — so every install would compile CUDA from
source, including every Ray runtime-env sync on every actor, against the
direction of #2039.

Dropping it preserves pre-bump behavior exactly: the only consumer is
megatron-core's experimental DSA (DeepSeek Sparse Attention) path, behind a
`try: from fast_hadamard_transform import ... except ImportError` guard, and
megatron-core 0.19 didn't depend on it either. The override comment records what
to do if DSA is ever wanted (git source + build config — upstream documents the
PyPI sdist as incomplete).

## Resolution delta

Four packages move; nothing else in the lock changes:

```
megatron-bridge   0.6.0+91a15142  -> 0.7.0+2b8cb21d
megatron-core     0.19.0+71e418ea -> 0.20.0+14346b65
imageio           2.37.4          -> (dropped)
imageio-ffmpeg    0.6.0           -> (dropped)
```

The two `imageio` packages are CVE-flagged codecs megatron-bridge deliberately
excludes from its own builds; they are simply no longer reachable from our graph.

## API audit

Beyond the suites, every megatron symbol we import (39 modules) was diffed
between the old and new revs — module presence, top-level function signatures,
and the method signatures of every class we subclass or call. Everything else
that changed is additive (new optional kwargs) and source-compatible:

- `AutoBridge.export_hf_weights` / `save_hf_weights` / `save_hf_pretrained` — `weight_dtype=None`
- `AutoBridge.export_adapter_weights` / `save_hf_adapter` / `export_adapter_ckpt` — `exclude_adapter_base_prefixes=None`
- `AutoBridge.import_ckpt` — `low_memory_save=False`
- `FullyParallelSaveStrategyWrapper` — `validate_access_integrity=True`
- `FullyParallelLoadStrategyWrapper` — `per_rank_object_load=False`
- `moe_utils.permute` / `unpermute` / `topk_routing_with_score_function` — batch-invariance kwargs

Three genuinely breaking megatron-core changes exist that we don't touch:
`get_batch_on_this_tp_rank`'s second positional renamed `is_sft` ->
`has_cu_seqlens`; `get_pretrain_batch_on_this_cp_rank` /
`get_sft_batch_on_this_cp_rank` removed (renamed by `8fa183119`); and
`log_on_each_pipeline_stage` grew required `level` / `msg` parameters. One
compatible relaxation we do touch: `get_default_save_sharded_strategy` lost its
`version` parameter, and we call it as
`get_default_save_sharded_strategy("torch_dist")`.

`GPTModel` is now deprecated in favor of `HybridModel` (megatron-core
`10cd30c0b`). It still exists and works; worth a follow-up, out of scope here.

## Testing

All suites run locally on 4×H100, compared against a `main` (`c53c17e86`)
baseline built in a separate worktree with its own environment.

| Suite | main | this PR |
| --- | --- | --- |
| H100 CI (megatron half) | 5 passed (29:49) | 5 passed (29:40) |
| `-m megatron_models` | 3 passed, 4 skipped | 3 passed, 4 skipped |
| H100 CI (fsdp half) | 2 passed | 2 passed |
| `-m megatron` | 130 passed, 2 skipped |  |

The H100 megatron half is the suite that exercised every regression above — both
router-replay crashes and the Qwen3.5 logprob corruption — and it now matches the
baseline exactly.

### Two flakes seen during this work, baselined as unrelated

- `ValueError: Unrecognized model in eatang/qwen3-moe-tiny-random`. The cached
  `config.json` is valid (`model_type: qwen3_moe`) and `AutoConfig.from_pretrained`
  resolves it fine standalone; a *different* parametrization failed on each run.
  Hub/cache concurrency, not a regression.
- `test_megatron_forward_qat_on_off` hitting the 30s placement-group timeout as
  the first test of a run, while Ray is still setting up per-actor runtime envs.
  Passes on `main` and in isolation.

---

> [!NOTE]
> **High Risk**
> Touches Megatron weight loading, checkpoint async I/O, and MoE export mappings—silent numerical corruption was possible before the Qwen3.5 experts_packed fix; dependency pins affect all Linux Megatron training installs.
> 
> **Overview**
> Bumps **megatron-bridge** to `2b8cb21d` (0.7.0) and **megatron-core** to `14346b65` (0.20.0), pinned to the Bridge submodule commit Bridge CI validates against. Adds a `fast-hadamard-transform` override so installs and Ray runtime-env syncs do not compile CUDA from source (DSA-only optional dep).
> 
> **Checkpointing:** megatron-core removed `dist_checkpointing.strategies.base`; async checkpoint queue now lives in module-level `_async_calls` in `megatron_strategy.py` instead of `ckpt_base.async_calls`.
> 
> **Weight conversion / bridges:** Global filter on `MegatronModelBridge.build_conversion_tasks` drops `None` unmapped tasks (fixes `AttributeError` on load). `broadcast_object_across_pp_ranks(..., allow_missing=True)` and skipping tasks with no parameter on any PP rank fix bucket sizing when 0.7.0 registers both MoE expert layouts. **Qwen35MoELMBridge** detects fused vs per-expert checkpoint layout via `moe_experts_stored_packed` (wrong default silently left expert weights uninitialized). **GLM47FlashBridge** forwards new `weight_dtype` kwarg.
> 
> **Model build:** Truncates `linear_attention_freq` when longer than overridden `num_layers` (hybrid Qwen3.5 + small `num_layers`).
> 
> Lockfile drops transitive `imageio` / `imageio-ffmpeg` from the Megatron graph.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d68a14038ecf8477f7148fde991c9430f006288d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


skyrl/backends/skyrl_train/distributed/megatron/megatron_strategy.py
```diff
@@ -46,6 +45,14 @@
 # Seed offset per pipeline-parallel rank, matching Megatron's standard practice.
 _PP_SEED_OFFSET = 100
 
+# Process-wide queue of in-flight async checkpoint writes. Megatron-core has no
+# global for this; SkyRL used to stash it on
+# `megatron.core.dist_checkpointing.strategies.base`, a long-deprecated module
+# that megatron-core 0.20 deleted, so it lives here now. It has to be
```
erictang000:
```suggestion
# global for this; It has to be
```

skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
```diff
@@ -172,6 +174,16 @@ def calculate_size_in_bytes(param, tp_size, ep_size):
...
 
         for idx, task in enumerate(weight_conversion_tasks):
+            # Skip tasks that own no parameter on any PP rank. megatron-bridge 0.7.0
+            # registers mappings for BOTH MoE expert layouts -- grouped-GEMM
+            # (`mlp.experts.linear_fc1`) and SequentialMLP
+            # (`mlp.experts.local_experts.*.linear_fc1`) -- so a model built with one
+            # layout still gets conversion tasks for the other. Those have no weights
+            # to export, and including them would break bucket-size accounting.
+            # Bridge 0.6.0 registered the grouped layout only, so this restores the
+            # previous task set exactly.
```
erictang000:
```suggestion
```

skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
```diff
@@ -490,6 +502,23 @@ def init_configs(
         for k, v in transformer_config_kwargs.items():
             setattr(provider, k, v)
 
+        # Keep a hybrid (linear-attention) layer pattern consistent with num_layers.
+        # megatron-bridge < 0.7.0 set `linear_attention_freq` to an *interval* (an int),
+        # which megatron-core expanded against whatever `num_layers` ended up being, so
+        # shrinking `num_layers` above just produced a shorter pattern. Since 0.7.0 the
+        # bridge resolves the HF config's `layer_types` into an explicit per-layer list
```
erictang000:
```suggestion
        # megatron bridge resolves the HF config's `layer_types` into an explicit per-layer list
```

pyproject.toml
```diff
@@ -233,6 +233,14 @@ no-build-isolation-package = [
 # unpinned dependencies.
 override-dependencies = [
     "nvidia-resiliency-ext; sys_platform == 'never'",
+    # New hard dependency of megatron-bridge 0.7.0. It publishes no wheels, so it
+    # would compile CUDA on every install and on every Ray runtime-env sync.
+    # megatron-core only uses it for the experimental DSA (DeepSeek Sparse
+    # Attention) variant, behind a `try: from fast_hadamard_transform import ...
+    # except ImportError` guard, so dropping it keeps the pre-bump behavior
+    # exactly. Remove this override (and add a git source + build config -- the
+    # PyPI sdist is documented upstream as incomplete) if DSA is ever needed.
```
erictang000:
```suggestion
    # megatron-bridge hard dependency that publishes no wheels, so it
    # would compile CUDA on every install and on every Ray runtime-env sync.
    # megatron-core only uses it for the experimental DSA (DeepSeek Sparse
    # Attention) variant, behind a `try: from fast_hadamard_transform import ...
    # except ImportError` guard, so dropping it keeps the pre-bump behavior
    # exactly. Remove this override (and add a git source + build config -- the
    # PyPI sdist is documented upstream as incomplete) if DSA is ever needed.
```

erictang000:
@cursor review

erictang000:
@cursor review

skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
```diff
@@ -176,6 +178,14 @@ def calculate_size_in_bytes(param, tp_size, ep_size):
         regular_task_indices: list[int] = []
 
         for idx, task in enumerate(weight_conversion_tasks):
+            # Skip tasks that own no parameter on any PP rank. megatron-bridge 0.7.0
+            # registers mappings for BOTH MoE expert layouts -- grouped-GEMM
+            # (`mlp.experts.linear_fc1`) and SequentialMLP
+            # (`mlp.experts.local_experts.*.linear_fc1`) -- so a model built with one
+            # layout still gets conversion tasks for the other. Those have no weights
+            # to export, and including them would break bucket-size accounting.
```
erictang000:
```suggestion
            # Skip tasks that own no parameter on any PP rank. megatron-bridge can
            # register mappings for BOTH MoE expert layouts -- grouped-GEMM
            # (`mlp.experts.linear_fc1`) and SequentialMLP
            # (`mlp.experts.local_experts.*.linear_fc1`) -- so a model built with one
            # layout still gets conversion tasks for the other. Those have no weights
            # to export, and including them would break bucket-size accounting.
```

PR #2043, opened 2026-08-14
[Tinker API] Restore variable-length result order

yapdianang:
## Summary

- Return each variable-length training result in the same order as its input example.
- Remove results produced only for padding.

## Before

Token-based batching returned results in packed microbatch order. A client could pair one example's mask with another example's log probabilities and fail before the optimizer step.

## After

SkyRL maps each result back to the original example index before it returns `loss_fn_outputs`. The change applies to Trajectory custom GSPO and any other client that reads per-example results.

## Testing

```bash
uv run --isolated --extra dev --extra tinker --extra skyrl-train \
  pytest tests/backends/skyrl_train/test_token_based_batching_utils.py -q
```

18 passed.

### DIFF
--- a/skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
+++ b/skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
@@ -1254,11 +1254,13 @@ class MegatronPolicyWorkerBase(MegatronWorker, PolicyWorkerBase):
             torch.cuda.empty_cache()
 
         # Aggregate metrics across micro-batches
-        all_loss_fn_outputs = []  # Handle separately from scalar metrics
+        loss_fn_output_batches = []
         for m_batch, metrics in zip(micro_buffer, metrics_list):
             # Extract loss_fn_outputs before reduce_metrics (it's not a scalar metric)
-            if "loss_fn_outputs" in metrics:
-                all_loss_fn_outputs.extend(metrics.pop("loss_fn_outputs"))
+            if metrics is None:
+                loss_fn_output_batches.append([])
+                continue
+            loss_fn_output_batches.append(metrics.pop("loss_fn_outputs", []))
             # Skip fully-padding microbatches: their metrics (clip_ratio=0, policy_entropy=0,
             # ...) are meaningless and would drag down the mean-reduced metrics. Summed
             # metrics (e.g. policy_loss) are unaffected since padding contributes 0, but
@@ -1302,6 +1304,13 @@ class MegatronPolicyWorkerBase(MegatronWorker, PolicyWorkerBase):
             if moe_metrics:
                 for k, v in moe_metrics.items():
                     status[k] = v
+
+        if not any(loss_fn_output_batches):
+            all_loss_fn_outputs = []
+        elif isinstance(microbatch_iterator, TokenBasedBatchIterator):
+            all_loss_fn_outputs = microbatch_iterator.reorder_and_combine_items(loss_fn_output_batches)
+        else:
+            all_loss_fn_outputs = [item for batch in loss_fn_output_batches for item in batch]
 
         return WorkerOutput(loss_fn_outputs=all_loss_fn_outputs, metrics=status)
 
--- a/skyrl/backends/skyrl_train/workers/worker_utils.py
+++ b/skyrl/backends/skyrl_train/workers/worker_utils.py
@@ -424,6 +424,18 @@ class TokenBasedBatchIterator(BaseBatchIterator):
         reordered_batch.metadata = ref_microbatch.metadata
         return reordered_batch
 
+    def reorder_and_combine_items(self, batches: List[List[dict]]) -> List[dict]:
+        """Restore per-sample microbatch outputs to input order."""
+        ordered = [None] * self.data.batch_size
+        for original_indices, items in zip(self._microbatches, batches):
+            if len(items) < len(original_indices):
+                raise ValueError("Microbatch output has fewer items than input samples")
+            for original_idx, item in zip(original_indices, items):
+                ordered[original_idx] = item
+        if any(item is None for item in ordered):
+            raise ValueError("Microbatch outputs do not cover every input sample")
+        return ordered
+
 
 def get_microbatch_iterator(
     data: TrainingInputBatch, micro_batch_size: int, max_tokens_per_microbatch: int
--- a/tests/backends/skyrl_train/test_token_based_batching_utils.py
+++ b/tests/backends/skyrl_train/test_token_based_batching_utils.py
@@ -160,6 +160,19 @@ class TestTokenBasedBatchIterator:
         for i in range(batch.batch_size):
             assert torch.equal(reordered["sequences"][i], batch["sequences"][i])
 
+    def test_reorder_and_combine_items_drops_padding(self):
+        batch = self._make_batch([10, 3, 8, 5])
+        iterator = TokenBasedBatchIterator(batch, max_tokens_per_microbatch=12)
+        output_batches = [
+            [{"sample": index} for index in indices] + [{"sample": "padding"}] for indices in iterator._microbatches
+        ]
+        iterator._num_padding_microbatches = 1
+        output_batches.append([{"sample": "padding-microbatch"}])
+
+        reordered = iterator.reorder_and_combine_items(output_batches)
+
+        assert [output["sample"] for output in reordered] == list(range(batch.batch_size))
+
     def test_get_microbatch_iterator_factory(self):
         batch = self._make_batch([10, 10, 5, 5])
 


skyrl/backends/skyrl_train_backend.py
```diff
@@ -829,6 +844,47 @@ def _normalize_policy_loss_request(
             normalized_config["eps_clip_high"] = clip_high_threshold - 1.0
         return ("regular" if loss_fn == "ppo" else "gspo"), normalized_config or None
 
+    @staticmethod
+    def _normalize_gspo_batch(batch: TrainingInputBatch) -> None:
```
j316chuck: should this code live here?

j316chuck: any chance we can reuse the loss functions written in the codebase to ensure bitwise parity?

erictang000: hmmm I think any advantage scaling should happen in the client code and not within SkyRL backend/server code - you should be able to implement this logic on the advantages before sending it to SkyRL's forward_backward.

skyrl/backends/skyrl_train_backend.py
```diff
@@ -857,6 +913,12 @@ def _forward_backward_single_model_batch(
         ):
             raise ValueError("Critic forward_backward requires values and returns for every response token")
         batch = self._to_training_batch(prepared_batch, role)
+        original_batch_size = batch.batch_size
+        original_token_counts = [int(count) for count in batch["attention_mask"].sum(dim=-1).tolist()]
+        submitted_indices = list(range(original_batch_size))
+        if loss_fn == "gspo":
+            self._normalize_gspo_batch(batch)
+            batch, submitted_indices = self._select_trainable_gspo_rows(batch)
```
erictang000: i think `_select_trainable_gspo_rows` is not necessarily GSPO specific right? looks like it's just gating on whether loss mask is 0 for rows? Ideally i'd like to avoid any loss specific logic here in this `_forward_backward_single_model_batch` function. If this can be made more generic or removed that would be good

skyrl/backends/skyrl_train_backend.py
```diff
@@ -914,10 +983,15 @@ def _forward_backward_single_model_batch(
                     loss_fn_outputs.append(formatted_output)
             else:
                 loss_fn_outputs = [{} for _ in range(end_idx - start_idx)]
+            request_metrics = dict(metrics)
+            if loss_fn == "gspo":
+                request_metrics.update(
+                    self._get_gspo_pruning_metrics(original_token_counts, submitted_index_set, start_idx, end_idx)
```
erictang000: this also seems not gspo specific, should make generic or remove

erictang000 (commented):
some initial comments, need to look closer at the seq_id/engine logic changes as well!

skyrl/backends/skyrl_train_backend.py
```diff
@@ -894,7 +956,14 @@ def _forward_backward_single_model_batch(
         if pad_size > 0 and per_sample_outputs:
             per_sample_outputs = per_sample_outputs[:-pad_size]
 
+        if len(submitted_indices) != original_batch_size:
+            restored_outputs = [{"logprobs": list(logprobs)} for logprobs in prepared_batch.all_sampling_logprobs]
+            for index, output in zip(submitted_indices, per_sample_outputs, strict=True):
+                restored_outputs[index] = output
+            per_sample_outputs = restored_outputs
```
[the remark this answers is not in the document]
yapdianang: The row-pruning and output-reconstruction branch is removed. Megatron now restores token-batched outputs directly to input order and handles an empty output list without this strict zip.

skyrl/tinker/engine.py
```diff
@@ -377,6 +378,56 @@ def _find_destructive_barriers(self, session: Session) -> dict[str, int]:
...
+
+        batchable = []
+        for model_id, requests in groupby(pending, key=lambda row: row.model_id):
+            requests = list(requests)
+            expected_seq_id = next_sequence_ids.get(model_id)
+            if expected_seq_id is None:
+                initial_seq_id = requests[0].seq_id
+                if initial_seq_id not in {0, 1}:
+                    continue
+                expected_seq_id = initial_seq_id
```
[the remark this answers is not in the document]
yapdianang: The official SDK starts every new TrainingClient at seq_id 1. A checkpoint resume creates a new model and client, then submits the load as seq_id 1. The scheduler now waits for 1 instead of bootstrapping from a later pending chunk; the regression test delays 1 while 2 and 3 are pending.

skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
```diff
@@ -1254,11 +1254,10 @@ def forward_backward(
             torch.cuda.empty_cache()
 
         # Aggregate metrics across micro-batches
-        all_loss_fn_outputs = []  # Handle separately from scalar metrics
+        loss_fn_output_batches = []
         for m_batch, metrics in zip(micro_buffer, metrics_list):
             # Extract loss_fn_outputs before reduce_metrics (it's not a scalar metric)
-            if "loss_fn_outputs" in metrics:
-                all_loss_fn_outputs.extend(metrics.pop("loss_fn_outputs"))
+            loss_fn_output_batches.append(metrics.pop("loss_fn_outputs", []))
```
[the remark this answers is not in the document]
yapdianang: Fixed. A pipeline stage that returns None now contributes an empty output batch and skips scalar metric processing.

skyrl/backends/skyrl_train_backend.py
yapdianang: Fixed. The server-side GSPO normalizer is removed. SkyRL preserves the submitted advantages, and a forward/backward boundary test verifies that client-scaled values reach dispatch unchanged.

skyrl/backends/skyrl_train_backend.py
yapdianang: Removed. SkyRL no longer performs GSPO-specific row selection in the backend.

skyrl/backends/skyrl_train_backend.py
yapdianang: Removed together with server-side row pruning. The backend no longer emits GSPO-specific submitted or skipped row metrics.

skyrl/tinker/engine.py
```diff
@@ -377,6 +378,53 @@ def _find_destructive_barriers(self, session: Session) -> dict[str, int]:
...
+            .where(FutureDB.status == RequestStatus.PENDING)
+            .order_by(FutureDB.model_id, FutureDB.seq_id)
+        ).all()
+
+        batchable = []
+        for model_id, requests in groupby(pending, key=lambda row: row.model_id):
+            requests = list(requests)
+            expected_seq_id = next_sequence_ids.get(model_id)
+            if expected_seq_id is None:
+                expected_seq_id = requests[0].seq_id
```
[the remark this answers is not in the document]
yapdianang: Fixed. A new model now expects seq_id 1, which matches the SDK. Pending IDs 2 and 3 remain blocked until the deliberately delayed ID 1 arrives; the new regression test covers this exact ordering.

hershg:
Downstream B300 live validation on this exact head (`ee28351f20f3a57ee2b0fb7a311b772b401138fa`):

- standalone SkyRL GLM 5.2: 16/16 variable-length rollouts, forward/backward, optimizer, and durable checkpoint;
- SkyRL–TCLI: XID 1048304 passed 16/16, XID 1048311 passed 64/64, and XID 1048316 passed 512/512 with all requested futures terminal;
- XID 1048334 completed three 64-rollout steps plus three 10-item eval rounds with zero stale or missing outcomes, three optimizer updates, and a durable step-3 checkpoint;
- focused local TCLI/GLM/config suite: 150 passed; the PR’s `skyrl_train_tests` and main test suites are green. The reported GPU check failure is Anyscale credential setup, not a test failure.

The GLM runtime also carries independent scale-aware FP8 import and DSA-sharing fixes, so this is downstream integration evidence rather than an isolated attribution claim. Across the corrected runs, sampler/trainer ratios remained near zero and no variable-length output/accounting regression appeared.

hershg:
GLM 5.2 B300 qualification is carrying this exact commit (`ee28351f20f3a57ee2b0fb7a311b772b401138fa`) because variable-length packed-microbatch ordering is a concrete learning-quality risk.

End-to-end evidence with this fix included:
- XID `1048304`: 16/16 terminal rollouts, mean log-ratio `-0.0094`, KL `0.0069`, ESS `0.99995`, grad norm `0.4286`, optimizer and checkpoint passed.
- XID `1048311`: 64/64 with healthy agreement/numerics.
- XID `1048316`: 512/512 with no missing future and healthy numerics.
- XID `1048334`: three 64-rollout steps plus three 10-item eval rounds, all terminal, three optimizer updates and durable checkpoint.

Those runs also include separate FP8-import and DSA fixes, so they are end-to-end integration evidence rather than an isolated attribution to this PR. The focused ordering tests here remain the direct regression. SkyRL #2099 includes this commit unchanged pending merge/rebase.

erictang000 (approved):

PR #2052, opened 2026-08-18
Enable GLM-5.2 FP8 LoRA on the current Megatron runtime

j316chuck:
- Add GLM-5.2 FP8 bridge/dequantization support and safe adapter export for LoRA training.
- Align packed FP8 token counts after TP sharding so Transformer Engine receives valid per-rank shapes.
- Pin the verified GLM dependency matrix: Transformers 5.8.1 patch, Megatron Bridge GLM support, and Megatron Core DSA/IndexShare support.

## Testing

```
uv run --isolated --extra tinker --extra dev --extra skyrl-train pytest -q tests/tinker/test_future_waiting.py tests/backends/skyrl_train/workers/megatron/test_adapter_store.py tests/backends/skyrl_train/workers/megatron/test_model_bridges.py tests/backends/skyrl_train/distributed/test_packing_utils.py tests/backends/skyrl_train/distributed/test_preprocess_packed_seqs_multiseq.py
```

32 passed, 1 skipped; validates current-main Tinker result handling plus the GLM adapter, bridge, and FP8 packing paths.


SumanthRH:
Closing after offline discussion: Fixes target CUDA 12, while SkyRL is now on CUDA 13 and these changes are no longer needed.

PR #2058, opened 2026-08-19
[Tinker API] Cache validated sampling targets

yapdianang:
## Summary

- Cache the validated external sampling target for each sampling session.
- Serialize the first lookup so a concurrent request burst performs one validation.
- Keep internal sampling and requests without a sampling session unchanged.

Depends on #2043.

## Before / After

| Case | Database validation reads |
| --- | ---: |
| Before: 512 concurrent requests for one sampling session | 512 |
| After: 512 concurrent requests for one sampling session | 1 |

The cache stores only a target that passed sampling-session, model, and checkpoint validation. A failed validation is not cached.

## Testing

```bash
uv run --extra dev --extra jax --extra tinker pytest \
  tests/tinker/test_sampling_target_cache.py \
  tests/tinker/test_api.py
uv run --extra dev pre-commit run --all-files
```

The API tests pass: 20 passed. All pre-commit hooks pass.

### DIFF
--- a/skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
+++ b/skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
@@ -1254,11 +1254,10 @@ class MegatronPolicyWorkerBase(MegatronWorker, PolicyWorkerBase):
             torch.cuda.empty_cache()
 
         # Aggregate metrics across micro-batches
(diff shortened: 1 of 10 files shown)


yapdianang:
Closing because #2061 includes this sampling-target validation change together with the training-request storage fix.

PR #2032, opened 2026-08-13
[docs] 10/n towards Kimi K2.6: Kimi K2.7-Code INT4 QAT + LoRA end-to-end example

casper-hansen:
## What

2-node (2x8 B300) DAPO example training Kimi-K2.7-Code at 262144-token context with INT4 fake-QAT + capacity-normalized expert LoRA and merge_lora=false adapter hot-loading, plus the docs page covering the full setup: dequantizing the INT4 release into BF16 masters (#2024) and the training/serving configuration (#2025, #2026, #2027) on the CUDA-13-native root lock (#2040; the separate deploy stack from #2028 is no longer needed).

Notable pins the example carries: `mla_prefill_backend=FLASHINFER` (FA4 CuTe-DSL prefill is incompatible with the locked nvidia-cutlass-dsl), `trainer.flash_attn=false` (MLA + sample packing needs TE cuDNN fused attention), `UV_PROJECT_ENVIRONMENT` + `--no-sync` for Ray's uv hook (workers execute from Ray's copied working dir where no .venv exists, and concurrent implicit syncs race), and a raised `SKYRL_WAIT_UNTIL_INFERENCE_SERVER_HEALTHY_TIMEOUT_S` (the 595 GB INT4 load + FULL_DECODE_ONLY capture exceeds the 600 s default on a cold page cache).

The docs page also spells out the multi-node Ray bring-up (which variables must be in the raylet environment on every node vs. what SkyRL re-exports through the runtime env), the file-backed frozen-master offload knob (`SKYRL_FROZEN_OFFLOAD_DIR`, #2062), and two host-level failure modes hit while verifying: a system cuDNN 9 of another 9.x than the pip one breaking TE fused attention with `CUDNN_STATUS_SUBLIBRARY_VERSION_MISMATCH`, and an NVSwitch fabric-state desync making NCCL's NVLS multicast bind fatal (`NCCL_NVLS_ENABLE=0` until the GPUs are reset).

## Why

Single runnable reference for the whole Kimi K2.x INT4+LoRA flow the series enables. The diff is self-contained (example script + docs only). Follow-up to #1862.

## Verification

Run end-to-end on 2 x 8 B300 against this branch (main @ 53b11553 merged in; torch 2.13 cu130 / vLLM 0.28 / TE 2.16 / megatron-bridge 8e5f0e1):

[...]

PR #2062, opened 2026-08-19
[megatron] 11/n towards Kimi K2.6: file-backed offload of frozen masters so Tinker API LoRA training fits host RAM

casper-hansen:
Part of the Kimi K2.6/K2.7 series (previous: #2024, #2025, #2026, #2027, #2029, #2031, #2032). Standalone; no dependency on the other open PRs, but #2026 (merge_lora=false adapter sync) and the upcoming adapter-only weight sync build on it.

## What

Two changes to the Megatron CPU-offload layer in `megatron_utils.py`, for LoRA training through the Tinker API on models whose frozen base does not fit host RAM twice:

- Frozen (`requires_grad=False`, non-adapter) params offload to **file-backed mmap tensors** instead of pinned/anonymous RAM. Written once per rank on first offload under `SKYRL_FROZEN_OFFLOAD_DIR` (default `/data/skyrl/frozen-offload`; set to `0` to disable and restore the previous pinned-RAM behavior). The masters are immutable for the whole run, so after the one-time write their pages are clean, evictable page cache instead of unreclaimable RAM; re-offloads are free and backloads read from hot cache.
- LoRA models **skip offloading the fused DDP buffers entirely**: Megatron's param/grad buffers only hold grad-requiring params, which for LoRA is the adapters (a few GB). They stay GPU-resident across offload/backload cycles, which is also what lets an adapter-only weight sync export the adapter without ever backloading the masters (follow-up PR).

## Why

Measured on our 2x8xB300 (sm103) cluster training Kimi K2.7-Code (INT4 QAT + LoRA, colocated vLLM engines) through the Tinker API: the trainer's CPU offload held ~1.34TB/node of pinned/anonymous RAM (~1.19TB BF16 masters + optimizer/misc) next to the vLLM engines' own ~1.3TB/node level-1 sleep backup — two full CPU copies of the model on a 2.9TB node, leaving <100GB of margin. Bring-up was a coin flip that ended in NUMA-local OOM kills (`CONSTRAINT_MEMORY_POLICY`: one socket full while the other had TBs free), or with swap enabled, compress/disk thrash until Ray's keepalive declared the workers dead.

With this change, peak-phase `MemAvailable` on the same nodes went from 80–380GB (repeated OOM kills across 8 consecutive bring-up attempts) to ~1.5TB, and bring-up passes reliably.

### DIFF
--- a/skyrl/backends/skyrl_train/distributed/megatron/megatron_utils.py
+++ b/skyrl/backends/skyrl_train/distributed/megatron/megatron_utils.py
@@ -230,7 +230,9 @@ def load_megatron_grads_to_gpu(models):
 # and re-read freely, instead of ~1.3TB/node of anonymous/pinned memory that
 # competes with the vLLM engines for physical RAM (the source of repeated
 # NUMA OOM kills and compress-swap stalls on TB-scale colocated models).
-# Files are written once per rank on first offload and reused afterwards.
+# Each param's file is written on first offload, mapped, and immediately
+# unlinked; the mapping keeps the inode alive until the process exits, and
+# sleep/wake cycles reuse the live mapping via ``param._offload_cpu_data``.
 # Set SKYRL_FROZEN_OFFLOAD_DIR=0 (or empty) to restore pinned-RAM offload.
 _FROZEN_OFFLOAD_DIR = os.environ.get("SKYRL_FROZEN_OFFLOAD_DIR", "/data/skyrl/frozen-offload")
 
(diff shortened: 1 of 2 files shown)


skyrl/backends/skyrl_train/distributed/megatron/megatron_utils.py
```diff
@@ -222,6 +224,70 @@ def load_megatron_grads_to_gpu(models):
...
+def _offload_frozen_param_to_file(name: str, param) -> bool:
+    """Move a frozen param's data to a file-backed mmap CPU tensor.
+
+    Returns True on success; False to let the caller fall back to pinned RAM.
+    """
+    try:
+        data = param.data.detach()
+        nbytes = data.numel() * data.element_size()
+        path = _frozen_offload_file(name, data)
+        if not (os.path.exists(path) and os.path.getsize(path) == nbytes):
```
avigyabb: If my understanding is correct, I believe there might be a bug in how we do file naming here. If we launch two jobs using models of the same architecture (lets say a fresh model and the SFT’d version of it), there is a chance that the second job may use the weights of the first job’s model since we skip writing to the file if it already exists.

avigyabb (changes requested):
Hey Casper, thanks again for all your contributions! Left a comment regarding some of the naming logic. If you're strapped on time, just lmk and I can take this over the line if you'd like!

avigyabb:
Also, if my understanding is correct, when we first `prepare_for_weight_sync` we call `_ensure_on_gpu` which calls `backload_to_gpu` which loads the entire model from disk to GPU and then when we weight sync, we discard the GPU copy of the frozen weights. For the adapter only sync path, we never needed to bring these onto the GPU to begin with. Maybe fixing this logic can be a follow up PR from our end?

https://github.com/NovaSky-AI/SkyRL/blob/b0f74aca4c3cdf9cd38fc70ec0798f9b3038c2e1/skyrl/backends/skyrl_train/workers/worker_dispatch.py#L169

avigyabb (approved):

SumanthRH:
Overall the PR is a good addition for memory constrained colocated training case. I think there are some good follow-ups here:
1. We should document the new way of offload/backload with LoRA + colocated training 
    a.  LoRA weights stay in GPU all the time, even after weight sync is done and we're generating with the vLLM engines. This is a bit misleading to users and we could have OOMs with high vLLM GPU util ratios
2. We should document env vars like `SKYRL_FROZEN_OFFLOAD_DIR` better, preferably in a recipe doc for large model training
3. Revisit offload/backload semantics for LoRA overall: We should book-keep trainable vs non-trainable weights differently and only backload trainable weights for weight sync. After weight sync, all weights should be offloaded. This will help avoid the issue in 1. as well.

SumanthRH (approved):
Let's address the follow-ups in separate PRs

SumanthRH:
Non-Megatron GPU CI tests passing: https://github.com/NovaSky-AI/SkyRL/actions/runs/34811362082
Megatron GPU CI tests passing: https://github.com/NovaSky-AI/SkyRL/actions/runs/34811366712
Megatron models GPU CI tests passing: https://github.com/NovaSky-AI/SkyRL/actions/runs/34811383590/job/103873258562?pr=2062

PR #2063, opened 2026-08-19
[megatron] 12/n towards Kimi K2.6: harden AdapterStore swaps for Tinker API multi-LoRA session churn

casper-hansen:
Part of the Kimi K2.6/K2.7 series (previous: #2062). Standalone: fixes apply to the existing multi-LoRA `AdapterStore` on main. Pairs naturally with #2062 (offload) and #2026 (merge_lora=false sync), which make the fixed code paths much hotter.

## What

Three fixes to `AdapterStore`, plus a CPU-side unit-test suite (`test_adapter_store_swap.py`) that fakes the DDP buffers via a monkeypatched `_iter_buffers` and replays the exact production ordering (create new model -> expiry-delete of the current one -> swap with grads offloaded):

- `_snapshot`/`_restore` skip `grad_data` copies while the DDP grad buffers are offloaded: `storage().resize_(0)` leaves a stale view behind, and reading/writing it is UB. Megatron zero-fills grad buffers on reload anyway, and grads are only ever offloaded post-step, so recording zeros is faithful.
- `create()` no longer adopts the live GPU state when `current_id` was cleared by `delete()`: a new `_live_dirty` flag distinguishes "live is pristine" (true first create) from "live still mirrors a deleted tenant" (must seed from the pristine snapshot; `swap_to` restores it).
- `register_pristine()` synchronizes after its `non_blocking` D2H snapshot so `create()`'s CPU-side `_copy_slot` cannot race the in-flight DMA.

## Why

Long-running Tinker API service with multiple LoRA sessions churning (clients create models, train, hit session expiry, reconnect) on our Kimi K2.7 B300 deployment. On 2026-08-18 we hit the failure chain this PR fixes: a `swap_to_adapter` crashed with CUDA `invalid argument` at the grad `copy_` (grad buffers had been offloaded post-step, leaving resized-to-zero storages), the run restarted, and the retry then *silently inherited the previous tenant's live adapter state* because `create()` treated the leftover live weights as pristine after the expiry `delete()` had cleared `current_id`. The second bug is the nastier one — no crash, just a new session fine-tuning on top of a dead tenant's weights.

### DIFF
--- a/skyrl/backends/skyrl_train/workers/megatron/adapter_store.py
+++ b/skyrl/backends/skyrl_train/workers/megatron/adapter_store.py
@@ -41,6 +41,22 @@ def _new_pinned_like(t: torch.Tensor) -> torch.Tensor:
     return torch.empty_like(t, device="cpu").pin_memory()
 
 
+def _grad_data_live(buf) -> bool:
+    """False while the buffer's grad storage is offloaded (freed).
+
+    DDP.offload_grad_buffers() frees grad_data via storage().resize_(0) while
+    keeping the tensor view intact; touching the view in that state is
+    undefined behavior (cudaMemcpyAsync on the stale pointer fails with
+    `invalid argument`). restore_grad_buffers()/zero_grad_buffer() reallocate
+    it zero-filled, i.e. Megatron already treats offloaded grads as discarded
+    — so snapshot/restore skip the grad copy instead of crashing. Grads are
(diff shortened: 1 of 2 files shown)


erictang000:
hey @casper-hansen i think this should be solved by #2000, closing this for now, feel free to re-open if #2000 looks insufficient

PR #2066, opened 2026-08-19
[tinker] 15/n towards Kimi K2.6: continuous sampling for the Tinker API engine under colocate_all

casper-hansen:
Part of the Kimi K2.6/K2.7 series. **Stacked on #2031 (9/n)** — its commits appear in this diff; review the last commit ("Continuous sampling: per-request future completion") for this PR's own changes.

## What

Replaces the Tinker API engine's batch-completion sampling with per-request future completion under `colocate_all`, keeping the engine-lifecycle safety property via admission control instead of serialization:

- `_ContinuousSampler` (engine): a daemon thread running an asyncio loop. Each admitted sample request executes as its own coroutine (`backend.sample_batch_async` -> per-request vLLM HTTP; vLLM continuous-batches internally), and its future is written the moment its own generation completes.
- The engine main loop admits pending samples whenever no blocking work exists, waking/bringing up engines on the main thread first (`backend.prepare_for_sampling`, so the sampler coroutines never touch engine lifecycle). Blocking work (forward/forward_backward, single requests, session cleanup that actually unloads models) drains in-flight samples first, then runs with the engines quiesced exactly as before. Admission filters stale requests first (serial-path parity), and no-op cleanup sweeps don't drain.
- Backend: `prepare_for_sampling` + `sample_batch_async` split out of the serial `sample()` path (shared validation; the async core reuses the loop-aware HTTP client with a persistent session on the sampler loop). `RemoteInferenceClient._get_session` closes the stale loop's connector synchronously on loop handoff.
- `SKYRL_TINKER_CONTINUOUS_SAMPLING=0` restores the serial loop; backends without the async surface (jax) keep it automatically via `hasattr` feature detection.

CPU unit tests in `tests/tinker/test_engine.py` cover admission/drain ordering, stale-request filtering, and the env-var opt-out.

## Why

Measured on our Kimi K2.7-Code agentic RL runs through the Tinker API (2x8xB300, colocated vLLM, ~64 lockstep rollouts): with batch-completion semantics each wave's turn latency was `max(generation in wave)` — a single 200s generation stalled 63 finished agents for its full tail, and GPU utilization between waves collapsed. The batching existed only to serialize sampling against ops that sleep the colocated engines; admission control preserves that property while letting every finished agent proceed immediately. The stale-request admission filter fixes a real bring-up failure we hit after a server restart: leftover sample requests from the previous server drove `prepare_for_sampling` into an engine build with no model registered, instead of failing fast with "model not loaded".


SumanthRH:
@greptileai review

SumanthRH (changes requested):
Nice! One issue with the current approach is that we've changed the ordering for new sample requests that come in before a barrier like optim_step / save_weights_for_sampler. 

The new code does the following:
```
process_pending_requests_once()
  -> sample_requests = find_batchable_sample()
  -> other_requests = find_single_requests()
  -> blocking_work = bool(forward_backward_requests or forward_requests or other_requests)

  if blocking_work:
    -> _ContinuousSampler.drain(...)
    -> sample_requests = {}

  -> process_batch_requests(forward_backward_requests, ...)
  -> process_batch_requests(forward_requests, ...)
  -> process_single_requests(other_requests)
```

If you have two requests [sample, save_weights_for_sampler], then in the above case, you can have the `save_weights_for_sampler` request run before we run `sample` which would be incorrect. 

This test fails right now:

```python
def test_continuous_sample_before_optim_step_runs_before_optim(continuous_engine):
    """A sample already queued before an optim_step must observe pre-optim
    weights. Continuous admission should run and drain that sample before the
    later mutating request."""
    engine = continuous_engine
    adam = types.AdamParams(learning_rate=1e-4, beta1=0.9, beta2=0.95, eps=1e-8, weight_decay=0.0)
    rid_sample, rid_optim = add_futures(
        engine,
        [
            (types.RequestType.SAMPLE, "model_a", sample_payload("")),
            (types.RequestType.OPTIM_STEP, "model_a", types.OptimStepInput(adam_params=adam).model_dump(mode="json")),
        ],
    )

    engine.process_pending_requests_once()
    engine.process_pending_requests_once()

    assert _wait_for(lambda: _future_status(engine, rid_sample) == RequestStatus.COMPLETED, timeout=3.0)
    assert _future_status(engine, rid_optim) == RequestStatus.COMPLETED
    assert engine.backend.order == [f"sample:{rid_sample}", "optim"]
```


@avigyabb  and I will try to make a quick fix

SumanthRH:
@greptileai review

SumanthRH:
@greptileai review

SumanthRH (approved):

PR #2067, opened 2026-08-19
[tinker] 16/n towards Kimi K2.6: accept zstd-compressed Tinker API forward_backward bodies (SDK >= 0.25)

casper-hansen:
Part of the Kimi K2.6/K2.7 series. Standalone, builds directly on the proto wire format from #2021. This is one of the two pieces salvaged from #2030 (closed as superseded by #2021), as flagged there.

## What

`_read_forward_backward_request` now handles `Content-Encoding: zstd`: the body is decompressed before wire-format dispatch (it applies to both proto and JSON bodies, since `Content-Encoding` is independent of `Content-Type`). Decompression is bounded at 1 GiB — far above any legitimate payload, but a small crafted body can no longer balloon into an arbitrarily large allocation. A malformed stream returns a clean 422.

Unit tests cover the SDK-encoded proto body compressed with zstd, a compressed JSON body, and the malformed-stream 422.

## Why

The tinker SDK >= 0.25 can submit the proto `forward_backward` body zstd-compressed. ASGI servers pass request bodies through undecoded, so on our Kimi K2.7 training service those requests failed proto parsing with a 422 until the server learned to decompress. The compression is worth supporting rather than disabling client-side: our RL forward_backward batches (long agentic transcripts) shrink several-fold on the wire, which matters when the training client is remote from the cluster.

### DIFF
--- a/skyrl/tinker/api.py
+++ b/skyrl/tinker/api.py
@@ -13,6 +13,7 @@ from uuid import uuid4
 
 import fastapi
 import psutil
+import zstandard
 from fastapi import Depends, FastAPI, HTTPException, Request, Response
 from fastapi.exceptions import RequestValidationError as FastAPIRequestValidationError
 from fastapi.responses import RedirectResponse, StreamingResponse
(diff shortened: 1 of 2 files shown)


avigyabb (approved):
LGTM! Thanks Casper!

PR #2071, opened 2026-08-19
[tinker] Restrict load_weights to a model's first request, matching Tinker

avigyabb:
**Stacked on #1996 — merge that first; this diff will then shrink to one commit.**

The real Tinker service only permits `LoadWeights` (`load_state` and `load_state_with_optimizer`) as a training client's **first** request. Any later attempt fails synchronously with `400: LoadWeights is not permitted with seq_id N` — verified empirically against the production service, including after a lone `forward_backward` (no optim step), a lone `save_state`, and a prior successful load. The sanctioned pattern is loading into a freshly created client (`create_training_client_from_state[_with_optimizer]`), which is why a weights-only load always lands on a zeroed optimizer there.

This PR enforces the same rule in `api.py`: the `load_weights` endpoint rejects the request if the model has any prior non-`create_model` future, echoing the SDK's `seq_id` in the error. `test_training_workflow` now asserts the rejection for both load flavors and restores through fresh clients — which also adds e2e coverage for the `optimizer=True` restore path from #1996.

**Note for reviewers:** this removes SkyRL's mid-run load support, which #1996's discussion leaned toward keeping as a deliberate extension. Happy to convert the restriction to an opt-in config flag (default permissive) if the team prefers the superset behavior.

## Testing

- `uv run --isolated --extra dev --extra jax --extra tinker pytest tests/tinker/` — 37 unit tests + full `test_api.py` e2e suite (17 tests) pass, including the rewritten `test_training_workflow` against a real server + tinker SDK client.

---

> [!NOTE]
> **Medium Risk**
> Changes checkpoint restore API semantics and training state (optimizer) on load; mid-run load is removed, which can break callers that relied on SkyRL’s previous behavior.
> 
> **Overview**
> Aligns **`load_weights`** with the real Tinker service: it is only allowed as a model’s **first** request (after `create_model`). If the model already has any other future (`forward_backward`, `save_weights`, etc.), the API returns **400** with `LoadWeights is not permitted with seq_id N`. Restores must use a **new** training client via `create_training_client_from_state` / `create_training_client_from_state_with_optimizer`.
> 
> The **`LoadWeightsRequest`** adds **`optimizer`** (default `true`) and optional **`seq_id`** for the error message. That choice flows through **`LoadWeightsInput.load_optimizer`** into **`AbstractBackend.load_checkpoint(..., load_optimizer)`** on JAX, Ray-JAX, and SkyRL-Train so weights-only loads skip optimizer/scheduler restore. Docs note the first-request-only rule.
> 
> Tests cover optimizer forwarding, SkyRL-Train load flags, and the e2e workflow (reject mid-run load; restore on fresh clients).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b0c5bdfbf789dc4cd013c2f76e6b0de1e32957f0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.claude/docs/tinker.md
+++ b/.claude/docs/tinker.md
@@ -36,7 +36,7 @@ All endpoints are under `/api/v1/`. Requests are async -- submit via POST, get a
 | `/asample` | POST | Generate samples from current or base model |
 | `/save_weights` | POST | Save full training checkpoint (weights + optimizer) |
 | `/save_weights_for_sampler` | POST | Sync weights to inference engines |
-| `/load_weights` | POST | Load a previously saved checkpoint |
+| `/load_weights` | POST | Load a previously saved checkpoint. Only permitted as a model's first request (matching the Tinker service); later attempts get a 400 -- load into a freshly created model instead |
 | `/training_runs/{unique_id}/checkpoints/weights/{checkpoint_id}` | DELETE | Delete a saved training checkpoint archive from disk |
 | `/training_runs/{unique_id}/checkpoints/sampler_weights/{checkpoint_id}` | DELETE | Delete a saved sampler checkpoint archive from disk |
 | `/retrieve_future` | POST | Long-poll for async result (300s timeout) |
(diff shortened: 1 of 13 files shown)


erictang000 (approved):
lgtm

PR #2090, opened 2026-08-22
Sharded RDT weight sync 

hao-aaron:
# Sharded-RDT (NIXL pull) weight-sync backend for SkyRL

## Context

Adds `generator.inference_engine.weight_sync_backend="sharded_rdt"` (alias `"rdt"`):
training→inference weight sync where the **vLLM inference workers pull** each slice
they consume from the trainer over NIXL/RDMA, instead of the trainer **pushing**
weights to every worker (the existing `nccl`, `cuda_ipc` and `delta` backends).

RDT is the pull-based counterpart to vLLM's trainer-send model (`WeightSource` +
`TrainerWeightTransferEngine` + `VLLMWeightSyncClient`). The trainer gathers weights
group-by-group and shares each group into a per-rank producer sidecar over CUDA IPC;
each inference worker pulls only the byte ranges its baked plan needs. For large
models this replaces an O(model) broadcast to every replica with a sharded,
pipelined, credit-limited pull.

Measured on Qwen3-235B-A22B (LoRA r128 merged, so every sync moves the full ~470 GB),
4 nodes — 2 trainer (Megatron tp4/pp2/ep8/etp1) + 2 inference (one vLLM engine,
dp16/tp1/ep16):

| backend | `timing/sync_weights` | vs NCCL |
|---|---|---|
| NCCL broadcast | 64.72 s | 1.00x |
| **sharded RDT** | **3.54 s** | **18.3x** |

Supports both **FSDP** and **Megatron** trainers, each via its own `WeightSource`.

### Base and vendoring

The RDT transport lives in the `vllm-rdt-weight-sync` fork. The pinned
`vllm==0.26.0` wheel *does* ship trainer-side ABCs at
`vllm/distributed/weight_transfer/base.py`, but a **narrower** version: no
`layerwise_groups`, and its `WeightSource` declares only `metadata()` / `__iter__`
without the ownership and group channels (`held_names` / `groups` / `iter_groups`)
this engine is built on. So the fork's versions are still vendored here.

This is an intermediate state. When SkyRL's pinned vLLM carries the fork's versions,
the vendored files are deleted and imports repointed (see Follow-ups).

### `sharded_rdt` as a first-class `WeightTransferStrategy` (`weight_sync/sharded_rdt_strategy.py`, new)

RDT now goes through `get_transfer_strategy_cls` **like every other backend**. This
replaces an earlier design in which RDT bypassed the strategy layer entirely and the
workers carried a backend conditional; that bypass is gone.

- **`ShardedRdtTransferStrategy`** — factory: `create_init_info` / `create_sender` /
  `get_vllm_transfer_engine`. The last is more than a reference mapping here: the
  consumers really do construct `ShardedRDTWeightTransferEngine` via vLLM's factory.
- **`ShardedRdtInitInfo`** — the config-derived args (`model_dtype`,
  `inference_world_size`, `trainer_actor_namespace`). Deliberately does **not**
  implement `for_servers` / `to_api_payload`: those exist for the driver-side
  `init_weight_update_communicator` fan-out, which this backend does not use.
- **`ShardedRdtWeightTransferSender`** — a thin wrapper over `RdtWeightSyncSender`.
  The transport is untouched; this only presents it through the shared interface.

Two places where the pull model genuinely does not fit the push backends' shape,
declared as **capabilities** so the workers hold no backend conditional:

| flag | value | why |
|---|---|---|
| `sender_initializes_receivers` | True | `trainer_init` opens the inference side itself (the bake needs the source metadata and must run under `set_current_vllm_config`), so worker rank 0 must **not** also call `init_weight_update_communicator` |
| `sender_needs_weight_extractor` | True | the rendezvous is eager, inside `create_sender`; deferring it to the first send deadlocks |
| `force_disable_expandable_segments` | True | the sidecar shares gathered tensors over CUDA IPC on **every** run, not only under colocation; VMM-backed memory makes that export/rebuild 5-10x slower per storage |
| `empty_cache_after_send` | False | publish buffers are reused by the next training step (measured 0.25-0.53 s/rank at 235B to scrub them for nothing); the worker still empties under `colocate_all` |

### Shared-ABC changes (`weight_sync/transfer_strategy.py`, `cuda_ipc_strategy.py`)

- **`WeightTransferSender.send(weight_extractor, dtype, **kwargs)`** — a new concrete
  *template method*. The default does exactly what the workers used to do inline:
  `extract_weights` + `get_weight_metadata` → `send_chunks`. RDT overrides `send()`
  and never materializes either argument.

  This is load-bearing, not tidiness. `get_weight_metadata` on the Megatron
  extractor is a whole-model `export_hf_weights` pass — the exact gather the RDT
  weight source exists to avoid (~20 s at 235B, and model-sized). If the shared call
  site computed it, RDT would pay it on every run. `send_chunks` raises on the RDT
  sender, with a message pointing at `send`.
- The four sender/strategy capability ClassVars above, plus
  `groups_chunks_by_module` on the strategy so the FSDP extractor reads the flag
  instead of re-deriving the backend from a string.
- **Cost to the other backends: 3 lines** — one ClassVar on
  `CudaIpcTransferStrategy`. Broadcast and delta are untouched; gating the
  extractor kwarg on a flag means their `create_sender` signatures never change.

### Trainer-side driver (`weight_sync/rdt_send.py`)

- **`RdtWeightSyncSender`** — built once per worker inside `create_sender`, which
  also performs the **eager rendezvous** (spawn each rank's producer sidecar, then
  the sender rank's bake). Eager is required: deferring to the first `send()`
  deadlocks — rank 0 blocks in the inference-side init RPC while the other ranks
  spin in gather collectives, and the sidecars (which share those ranks' GPUs) can
  then never finish NIXL agent creation, because libfabric's CUDA probe blocks
  behind the spinning kernels.
- **`_FsdpWeightSource`** — `WeightSource` over `FSDPWeightExtractor`. Reuses the
  extractor (not raw `named_parameters()`) so names, the FSDP2 all-gather and the
  dtype cast match what the consumer baked its plan over, and reorders into
  group-major order.
- **`MegatronStackedWeightSource`** — the production Megatron source. Serves
  **PP-local + EP-local**: each pipeline stage declares and serves only its own
  layers, each EP rank only its own experts, via `held_names()`; LoRA is merged
  locally. Replaces ~37k per-tensor gathers with stack-granularity ones.
- **`MegatronWeightSource`** — the plain fallback (whole-model bridge export).
  `make_weight_source()` selects between them and logs the **chosen class**, so an
  ablation cannot silently run the wrong source.

### Trainer-side control plane (`weight_sync/rdt_control_plane.py`, `inference_servers/rdt_control_protocol.py`)

The trainer engine drives the inference side through a **synchronous** blocking-HTTP
client (`SyncRdtControlPlaneClient`) rather than bridging onto SkyRL's async
`RemoteInferenceClient`. The engine then runs sync-to-sync in its worker thread with
zero event-loop involvement, which removes a deadlock class structurally rather than
by convention. `rdt_control_protocol` holds the method names and the per-server
`replica_rank` fan-out so nothing can drift.

Fan-out concurrency is required for **correctness**, not speed: consumers pull in
lockstep and a producer only frees a served group once every bound consumer has
pulled, so a serial `update_weights` would stall the gather loop and deadlock.

### Vendored transport (`weight_sync/sharded_rdt_{base,common,engine,fake,trainer}.py`)

Copied from the fork and kept in sync with it. Intended differences only:

- **import paths**;
- the **LIBFABRIC shim** (`rdt_libfabric_shim.py`, SkyRL-only) — without EFA ≥ 1.47
  NIXL silently falls off LIBFABRIC and every timing number is meaningless;
- a no-op **`trainer_send_weights`** on the consumer engine, which the pinned
  wheel's ABC still declares abstract (the fork's ABC dropped it);
- **no engine fault tolerance** — see below;
- black at 120 cols instead of the fork's 88.

`sharded_rdt_trainer.py` also carries the per-rank `_RDTProducerServer` Ray actor
(gather cache, serve rings, the per-group free barrier, the packed serve) and
cross-deployment **serve-slot sharing** (`_share_width` / `_sharers`): consumers that
are the same worker index of different inference deployments pull byte-identical
chunks, so they are served from one registered slot instead of one ring each. On the
235B 4-node shape that holds producer serve memory flat at ~5.0 GiB/GPU regardless of
replica count, instead of 14/28/54 GiB at 2/4/8 deployments — see
`.claude/docs/weight_sync.md` for the full derivation and the release-edge argument.

The vendored files are **not** excluded from `black`/`ruff` — they are formatted to
SkyRL's house style. (An earlier revision of this PR excluded them; that exclusion
is gone.)

### Deliberately NOT included: engine fault tolerance

The fork's trainer takes a live consumer set in `send_weights` and exposes
`get_worker_init_payload` so a restarted engine can rejoin. **Both are omitted
here.** The consumer set is the provisioned fleet, fixed at `trainer_init`. Do not
"restore" them when syncing with the fork.

The producer-side **stall watchdog IS kept** (`stall_timeout_s`, default 300 s): the
slot-sharing rendezvous relies on it as its failure mode, and without it a consumer
that stops pulling wedges every trainer rank in NCCL with no exception anywhere.

### Worker wiring (`workers/worker.py`, `fsdp/fsdp_worker.py`, `megatron/megatron_worker.py`, `train/trainer.py`)

- `Worker.init_weight_sync_state` has **no RDT branch at all** — it selects a
  strategy, passes the extractor when the strategy asks for it, and skips the
  driver-side receiver init when the sender owns the handshake.
- Both workers' three-way dispatch collapses to two branches (LoRA, else); the else
  branch is `sender.send(extractor, dtype, **kwargs)` for every backend. This also
  fixes a pre-existing FSDP/Megatron drift, where only Megatron skipped the
  post-send `empty_cache()`, by construction.
- The workers contain **zero** backend-string comparisons.
- `train/trainer.py` reports `timing/sync_weights` (the transfer alone) alongside
  `timing/sync_weights_total` (which adds vLLM's DP pause/resume bracket, ~1.4 s
  measured at 235B — not RDT's cost).

**Bug fixed in passing:** upstream reads
`self._weight_transfer_sender.handles_prefix_cache_reset` unconditionally at the top
of `broadcast_to_inference_engines`, but the old RDT path returned early from
`init_weight_sync_state` without ever assigning that attribute. RDT's first sync
would have raised `AttributeError` on this base. Routing RDT through the strategy
means the sender always exists.

### Inference-side (`inference_servers/new_inference_worker_wrap.py`, `layerwise_reload.py`, `utils.py`)

- `init_weight_transfer_engine_rdt` — runs the engine's one-time bake under
  `set_current_vllm_config` + `torch.device(self.device)`, rather than vLLM's native
  endpoint (the bake drives `model.load_weights` against `FakeRDTTensor`
  placeholders, and MoE `process_weights_after_loading` reads
  `get_current_vllm_config()`).
- `update_weights_rdt` — pulls this worker's slices. The engine defers GPU
  post-processing to background threads, so unlike ipc/nccl it does **not**
  synchronize here; `skyrl_finish_weight_update` drains via a `getattr`-probed
  `drain_pending` (the pinned wheel's worker ABC declares neither
  `defers_processing` nor `drain_pending`).
- Engine provisioning forces `distributed_executor_backend="ray"` and rejects
  `colocate_all` for this backend.

### Runtime registration (`weight_sync/rdt_vllm_register.py`)

Idempotently registers `ShardedRDTWeightTransferEngine` in vLLM's
`WeightTransferEngineFactory` under `"sharded_rdt"`. `vllm==0.26.0` types
`WeightTransferConfig.backend` as `Literal["nccl","ipc","sparse_nccl"] | str`, so no
Literal relaxation is needed — only the factory registration. No-op if vLLM is not
importable (non-Linux / CPU-only test envs).

## Config

- `generator.inference_engine.weight_sync_backend = "sharded_rdt"` (or `"rdt"`).
- Requires **non-colocated** placement (`trainer.placement.colocate_all=false`),
  `trainer.strategy` in {`fsdp`, `megatron`}, `backend=vllm`,
  `distributed_executor_backend=ray` (forced), Ray ≥ 2.56, and **EFA ≥ 1.47** on
  every GPU node for cross-node NIXL.
- `expert_tensor_parallel_size=1` for the optimized Megatron source: at ETP>1 no
  rank holds a whole expert, and it falls back to the plain source.
- Tuning knobs are `SKYRL_RDT_*` env vars, forwarded to every Ray worker by
  `prepare_runtime_environment`: `LOOKAHEAD` (gather credit depth, default 1),
  `NUM_BUFFERS` (ring depth, default 2), `STACKED_EXPERTS`, `EXPORT_RING`,
  `GC_FREEZE`, `SHARE_SLOTS`, `STALL_TIMEOUT_S`, `BUFFER_PRESIZE_GB`,
  `VERIFY_STACKED`. They are not yet `InferenceEngineConfig` fields (see
  Follow-ups). The sidecar does **not** inherit them — it is a Ray actor that
  inherits the raylet's environment — which is why the ring / lookahead / timeout
  knobs ride the init info instead.
- RDT receive buffers live **outside** `gpu_memory_utilization`: K x the largest
  chunk, per rank.

## Breaking changes / compatibility

- `nccl` / `cuda_ipc` / `delta` are behaviour-preserving. The only shared-code
  changes are additive: a concrete `send()` on the sender ABC whose default
  reproduces the previous inline sequence exactly, capability ClassVars with
  push-compatible defaults, and one ClassVar on `CudaIpcTransferStrategy`.
- New backend is opt-in; default remains `nccl`.
- `get_transfer_strategy_cls` no longer raises for `sharded_rdt` — it returns the
  strategy.

### CPU (244 RDT tests, 0 failures)

```bash
uv run --extra dev --extra fsdp pytest tests/backends/skyrl_train/weight_sync/
```

`test_sharded_rdt_plan.py` (the bake/plan/pack layout), `test_sharded_rdt_producer.py`
(the publish → serve → free lifecycle, slot sharing, the stall watchdog),
`test_sharded_rdt_source.py`, `test_rdt_control_plane.py`, and
`test_sharded_rdt_strategy.py` (the adapter: capability flags, eager rendezvous
ordering, and an explicit assertion that `send()` **never** touches
`get_weight_metadata` or `extract_weights`).

Whole-directory run: **446 passed, 6 skipped, 1 failed**, the one failure being a
pre-existing upstream delta-backend test (`test_delta_checkpoint.py`, concurrent Ray
actors) that fails identically on unmodified `origin/main`.

### GPU

```bash
# FSDP + RDT e2e (1 rank, and 2 ranks so full_tensor() is a real FSDP2 all-gather)
uv run --isolated --extra dev --extra fsdp pytest \
  tests/backends/skyrl_train/gpu/gpu_ci/test_policy_local_engines_e2e.py -k sharded_rdt

# Megatron + RDT, non-colocated, PP=2
uv run --isolated --extra dev --extra megatron pytest \
  tests/backends/skyrl_train/gpu/gpu_ci/megatron/test_megatron_worker.py -k sharded_rdt
```

Both **pass**. Note these tests read gsm8k from `$HOME/data/gsm8k`; CI generates it
as its first step (`uv run examples/train/gsm8k/gsm8k_dataset.py --output_dir
$HOME/data/gsm8k`).

### At scale — Qwen3-235B-A22B, 4 nodes

`bash rdt-multinode/runs/run_235b_4node_ablation_noft.sh v3 4`
(wandb `skyrl-rdt-multinode` / `noft_4node_235b_v3`)

| step | `timing/sync_weights` | `sync_weights_total` | logprob gap | avg_raw_reward | grad_norm |
|---|---|---|---|---|---|
| 1 | 3.578 s | 4.963 s | 0.0483 | 0.234 | 0.150 |
| 2 | 3.484 s | 4.857 s | 0.0546 | 0.391 | 0.176 |
| 3 | 3.562 s | 4.925 s | 0.0556 | 0.422 | 0.171 |
| **warm mean** | **3.541 s** | 4.915 s | 0.0528 | rising | non-zero |

Config verified live via the tripwire:
`source=MegatronStackedWeightSource lookahead_env=1 stacked_experts_env=1`,
`num_consumers=16 num_replicas=1`, 16 producer sidecars all on **LIBFABRIC**.

**Correctness.** `policy/minibatch_rollout_logprobs_abs_diff_mean` at 0.048-0.056
against a healthy ~0.05 (NCCL's independent reference on this shape is 0.0497). This
is the check a bad sync cannot fake: >1 means the weights in vLLM are not the
weights in the trainer. Two corroborating signals — reward rising 0.23 → 0.42 and
non-zero `grad_norm` — rule out the silent-corruption mode where reward looks normal
at step 0 and is then exactly 0 forever.

## Follow-ups

- **Un-vendor on a vLLM bump.** Once the pinned vLLM carries the fork's trainer-side
  ABCs *including* the group/ownership channels (`layerwise_groups`, `held_names`,
  `groups`, `iter_groups`), delete `sharded_rdt_{base,common,engine,fake,trainer}.py`
  + `rdt_vllm_register.py` and repoint imports. Every removal is flagged in-file.
- **Adapt NCCL/IPC to the trainer-engine shape**, then delete the
  `WeightTransferStrategy` layer and this adapter with it. The ABCs are already
  in-tree, so this is gated on nothing external;
  `BroadcastWeightTransferSender._send_chunks_vllm_native` already *is* the engine
  round trip.
- **Surface the `SKYRL_RDT_*` knobs as `InferenceEngineConfig` fields.**
- **FSDP + LoRA + sharded_rdt is unreachable**: the LoRA branch precedes the sender
  path in `fsdp_worker.py`, so sync silently takes the disk-adapter route. Either
  wire it or reject the combination in `validate_cfg`. (Megatron is safe in practice
  because `merge_lora=True` is the default.)
- **`teardown()` has no caller** on any backend — pre-existing, but now uniform
  enough that one call site would cover all four.
- **`sharded_rdt_fake.py` is behind the fork** by two changes not carried here: the
  fork removed the vestigial `_Scatter.nbytes` field and added a broadcasting-`copy_`
  numel guard at bake time. The guard prevents a silent producer/consumer pack-layout
  mismatch and is worth picking up, but it is a behaviour change wanting its own GPU
  validation.
- **Inference-side pipeline parallelism is untested.** The rank arithmetic is
  PP-aware and routing is per-name, but the EP validation asserts `dp * tp == ep` and
  no test exercises inference pp>1. Treat as unsupported.
- **Not validated:** the 6 nccl params of `test_policy_local_engines_e2e.py` as a
  push-backend GPU regression check (the CPU suite covers the ABC changes).
- Add `VLLM_CACHE_ROOT=/mnt/cluster_storage/vllm_cache` to the 235B recipes so both
  inference nodes share one torch.compile cache.

---

> [!NOTE]
> **High Risk**
> New RDMA/NIXL weight-transfer path with vendored vLLM engines, monkeypatches (LIBFABRIC, Megatron export), and trainer/inference handshake changes. Incorrect sync would silently desync policy weights; existing NCCL/IPC/delta paths stay default but share the sender ABC.
> 
> **Overview**
> Adds opt-in `weight_sync_backend=sharded_rdt` (alias `rdt`): inference workers **pull** only the slices they need from trainer ranks over NIXL/RDMA, instead of a full-model NCCL/IPC push. Non-colocated FSDP/Megatron + vLLM only; forces the Ray executor.
> 
> The new `ShardedRdtTransferStrategy` is a thin adapter so workers stay backend-agnostic. Real work is `RdtWeightSyncSender`: eager sidecar spawn + bake (deferred init deadlocks), stacked Megatron/FSDP `WeightSource`s, and a blocking HTTP control plane so the trainer engine never touches the async event loop. Consumers bake a static pull plan; GPU post-processing is drained in `skyrl_finish_weight_update`.
> 
> Vendored vLLM RDT engines plus a LIBFABRIC shim for EFA. Slot sharing keeps producer ring memory flat across replicas. Megatron examples (30B–235B, including DP+EP) and docs/tests land with the backend. Default remains NCCL; shared sender `send()` is additive so push backends keep the old extract-then-chunk path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 57299e823b204e77c440ae9eb0975aff4bee8dbd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


SumanthRH:
Taking a look now. Meanwhile, can you fix CI errors? 

You need to use the `vllm` marker for CPU tests that require vllm: 

```bash

==================================== ERRORS ====================================
_ ERROR collecting tests/backends/skyrl_train/weight_sync/test_sharded_rdt_plan.py _
ImportError while importing test module '/home/runner/work/SkyRL/SkyRL/tests/backends/skyrl_train/weight_sync/test_sharded_rdt_plan.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/hostedtoolcache/Python/3.12.14/x64/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
tests/backends/skyrl_train/weight_sync/test_sharded_rdt_plan.py:27: in <module>
    from skyrl.backends.skyrl_train.weight_sync.sharded_rdt_engine import (
skyrl/backends/skyrl_train/weight_sync/sharded_rdt_engine.py:69: in <module>
    from vllm.config.weight_transfer import WeightTransferConfig
E   ModuleNotFoundError: No module named 'vllm'
_ ERROR collecting tests/backends/skyrl_train/weight_sync/test_sharded_rdt_producer.py _
ImportError while importing test module '/home/runner/work/SkyRL/SkyRL/tests/backends/skyrl_train/weight_sync/test_sharded_rdt_producer.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/hostedtoolcache/Python/3.12.14/x64/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
tests/backends/skyrl_train/weight_sync/test_sharded_rdt_producer.py:30: in <module>
    import skyrl.backends.skyrl_train.weight_sync.sharded_rdt_trainer as trainer_mod
skyrl/backends/skyrl_train/weight_sync/sharded_rdt_trainer.py:51: in <module>
    from vllm.logger import init_logger
E   ModuleNotFoundError: No module named 'vllm'
```

tests/backends/skyrl_train/gpu/gpu_ci/conftest.py
```diff
@@ -58,8 +58,19 @@ def _ray_init(extra_env_vars: dict[str, str] | None = None):
...
+    # driver's, the cluster's workers won't have the venv's packages (vllm,
+    # skyrl, nixl). Setting SKYRL_RAY_PY_EXECUTABLE makes Ray launch this job's
+    # workers with that interpreter so actors import the same env as the driver.
+    # Opt-in only — default behavior (CI, fresh local cluster) is unchanged.
+    py_executable = os.environ.get("SKYRL_RAY_PY_EXECUTABLE")
+    if py_executable:
+        runtime_env["py_executable"] = py_executable
+
+    logger.info(f"Initializing Ray with runtime_env: {runtime_env}")
+    ray.init(runtime_env=runtime_env)
```
SumanthRH: This is not needed if you launch with `uv` as intended?

tests/backends/skyrl_train/gpu/gpu_ci/megatron/test_megatron_worker.py
```diff
@@ -116,28 +121,81 @@ def get_test_training_batch(batch_size=4) -> TrainingInputBatch:
...
+    """
+    gpu_nodes = [n for n in ray.nodes() if n.get("Alive") and n["Resources"].get("GPU", 0) >= 1]
+    reservations = []
+    for node in gpu_nodes[1:]:
+        gpus = int(node["Resources"]["GPU"])
+        pg = placement_group([{"GPU": 1}] * gpus, strategy="STRICT_PACK")
+        get_ray_pg_ready_with_timeout(pg, timeout=60)
+        reservations.append(pg)
+    print(f"[rdt-intranode] reserved {len(reservations)} node(s); 1 free GPU node for vLLM+policy")
+    return reservations
```
SumanthRH: This is again not needed to be added into CI? CI will run the tests on a single machine

tests/backends/skyrl_train/gpu/gpu_ci/megatron/test_megatron_worker.py
```diff
@@ -116,28 +121,81 @@ def get_test_training_batch(batch_size=4) -> TrainingInputBatch:
...
+        pytest.param("nccl", True, 4, 2, 2, 1, None, True, id="nccl_colocate_all_lora"),
+        # sharded_rdt (NIXL pull): non-colocated only. PP=2 exercises the
+        # stacked source's single mode: each stage declares and serves only its
+        # own layers, and on MoE models each rank serves only its own EP
+        # coordinate's experts. The naive whole-model extraction is the fallback
+        # selected by SKYRL_RDT_STACKED_EXPERTS=0.
+        # The harness sets policy world = num_gpus_per_node = inference_tp, so
+        # inference_tp must equal megatron_tp*megatron_pp (=4): 4 dedicated policy
+        # GPUs + 4 inference GPUs = 8, genuinely non-colocated.
+        pytest.param("sharded_rdt", False, 4, 2, 2, 1, None, False, id="sharded_rdt_non_colocated_pp2"),
```
SumanthRH: This will run on our 4xL4 CI machine - can you change the test for 4xL4s? 

Let's have these two tests:

1. (Inference TP2,  Trainer TP2) 
2. (Inference TP1, Trainer PP2). 

We'll test both cases of num_engines > 1 and trainer PP > 1 in the second test (can be separate but doesn't have to be)

tests/backends/skyrl_train/gpu/gpu_ci/conftest.py
SumanthRH: This is not needed if you launch with `uv` as intended?

skyrl/backends/skyrl_train/weight_sync/rdt_send.py
```diff
@@ -0,0 +1,1465 @@
...
+                if not isinstance(fc1, dict) or "gate" not in fc1 or "up" not in fc1 or not isinstance(fc2, str):
+                    return None
+                names.append(fc1["gate"])
+                names.append(fc1["up"])
+                names.append(fc2)
+            return names
+        except Exception:  # noqa: BLE001 - any registry shape we cannot read -> fall back
+            logger.warning("[rdt] expert-name lookup via the bridge failed; using the legacy names", exc_info=True)
+            return None
+
```
SumanthRH: When is this codepath hit? 

Can we just error out instead of using the legacy hardcoded path names and having any silent errors?

.claude/docs/weight_sync.md
```diff
@@ -60,6 +76,137 @@ re-read even for an empty delta.
...
+`SKYRL_RDT_STALL_TIMEOUT_S`). Without it, a consumer that stops pulling mid-sync never
+sends its `free_group` signals, the three waits block forever, and every trainer rank
+wedges in NCCL with no exception anywhere. On no publish/serve/free progress for the
+timeout, the producer fires the `set_gather_error` channel and the run fails with a real
+error. It does not recover; it makes the failure diagnosable. The slot-sharing rendezvous
+waits on the same channel, so a sharer that never arrives fails the same way.
+
+Vendored files (`sharded_rdt_*.py`) mirror
+`~/default/vllm-rdt-weight-sync`; keep both in sync. Entry-point doc:
+`~/default/RDT_WEIGHT_SYNC.md`.
```
SumanthRH: Remove references to local files

skyrl/backends/skyrl_train/workers/worker.py
```diff
@@ -334,16 +334,34 @@ def _set_expandable_segments(self, enabled: bool) -> None:
             logger.warning(f"Failed to set {setting!r}: {e}")
 
     @contextmanager
-    def _expandable_segments_disabled_for_sync(self):
+    def _expandable_segments_disabled_for_sync(self, force: bool = False):
```
SumanthRH, line 337: You should also update tests that depend on the specific signature of `_expandable_segments_disabled_for_sync`, for example:

https://github.com/NovaSky-AI/SkyRL/blob/a9564572eb8e91076be43c1bd5e4615e3de17265/tests/backends/skyrl_train/gpu/gpu_ci/test_prefix_cache_reset.py#L42-L46

skyrl/backends/skyrl_train/workers/fsdp/fsdp_worker.py
```diff
@@ -302,22 +315,24 @@ async def broadcast_to_inference_engines(
...
+            # cudaIpcGetMemHandle, which is incompatible with the VMM addresses
+            # expandable segments uses, and some senders (sharded_rdt) share GPU
+            # memory on every run and ask for the toggle unconditionally.
+            with self._expandable_segments_disabled_for_sync(
+                force=self._weight_transfer_sender.force_disable_expandable_segments
+            ):
+                await self._weight_transfer_sender.send(
+                    self.weight_extractor,
+                    generator_dtype,
                     **send_chunks_kwargs,
```
SumanthRH, line 341: Let's also update relevant tests here. 

e.g: 

https://github.com/NovaSky-AI/SkyRL/blob/a9564572eb8e91076be43c1bd5e4615e3de17265/tests/backends/skyrl_train/gpu/gpu_ci/test_prefix_cache_reset.py#L89-L91

https://github.com/NovaSky-AI/SkyRL/blob/a9564572eb8e91076be43c1bd5e4615e3de17265/tests/backends/skyrl_train/gpu/gpu_ci/test_prefix_cache_reset.py#L32-L37

We use a mocked sender and the test asserts that `send_chunks` was called, but we now call `send`

skyrl/train/trainer.py
```diff
@@ -490,8 +490,17 @@ async def train(self):
...
                             await self.dispatch.save_weights_for_sampler()
+                        # `sync_weights` is the transfer alone. The enclosing call
+                        # also pauses and resumes generation, which under vLLM DP
+                        # costs seconds of coordinator quiesce that is not weight
+                        # -sync work; `sync_weights_total` keeps the full bracket.
+                        engine_seconds = getattr(self.dispatch, "last_weight_sync_seconds", None)
+                        if engine_seconds is not None:
+                            self.all_timings["sync_weights"] = (
+                                self.all_timings.get("sync_weights", 0.0) + engine_seconds
+                            )
```
SumanthRH: Can we do this:

1. Leave the current `sync_weights` timing as is.
2. Have an additional timer `sync_weights_only_transfer`: Just the weight transfer time without pause/resume

SumanthRH: Let's add this in worker_dispatch?

```python
class WorkerDispatch:
     def get_timing_metrics(self):
            if self.last_weight_sync_seconds is None:
                return {}
            return {"sync_weights_only_transfer": self.last_weight_sync_seconds}
```

And then we can add it to the other trainers as well later if needed

SumanthRH (changes requested):
Great! Lots of changes, leaving comments as I make my way through the PR.

skyrl/backends/skyrl_train/workers/worker.py
```diff
@@ -491,18 +515,27 @@ async def init_weight_sync_state(
 
         # Create sender on all ranks
         # Strategy implementations may have different logic for different ranks
+        sender_kwargs = {}
+        if self._transfer_strategy_cls.sender_needs_weight_extractor:
+            # sharded_rdt rendezvouses inside create_sender (eagerly, to avoid a
+            # first-send deadlock), which needs the model. Both workers build the
+            # extractor before calling super(), so it is available here.
+            sender_kwargs["weight_extractor"] = getattr(self, "weight_extractor", None)
```
SumanthRH: Probably just modify the base class create_sender API to always take in the `weight_extractor` argument? 

I've done the same with `inference_engine_client` (not all engines need it).

examples/train/megatron/sharded_rdt/run_megatron_rdt_dpep_qwen3_235b_a22b_lora_4node.sh
SumanthRH, line 1: Nice! This is one of our primary example scripts. Can we add a README for this folder with instructions to run the examples? Specifically this example would be good - it is what we use in the blog.

skyrl/backends/skyrl_train/workers/worker_dispatch.py
```diff
@@ -600,11 +605,15 @@ def _broadcast_to_inference_engines(self, inference_engine_client, model_id: Opt
...
+        # Pass None for the client: workers cached it at init_weight_sync_state.
+        # It carries the HF tokenizer (~10MB — 0.13s pickle driver-side, 0.34s
+        # unpickle on EVERY worker), so shipping it per sync costs ~0.5s of the
+        # sync wall even via ray.put (deref still deserializes per worker).
         ray.get(
             self._actor_groups["policy"].async_run_ray_method(
                 "pass_through",
                 "broadcast_to_inference_engines",
-                inference_engine_client,
+                None,
```
SumanthRH, line 616: Curious if you used this change for FT as well? 

Because if a replica goes down and the inference cilent state is updated, then we need to send the new client again.

skyrl/backends/skyrl_train/inference_servers/new_inference_worker_wrap.py
```diff
@@ -48,6 +45,12 @@
 except ModuleNotFoundError:
     pass
 
+# Registering the sharded_rdt engine into vLLM's WeightTransferEngineFactory must
+# happen inside every worker process (GPUWorker.load_model builds the engine via
+# the factory). Importing here — the worker-extension module vLLM loads before
+# model init — guarantees it runs on each worker. No-op if vLLM is absent.
+from skyrl.backends.skyrl_train.weight_sync import rdt_vllm_register  # noqa: F401,E402
+
```
SumanthRH, line 62: should be in a try except incase registration fails (ex: on a trainer process)? Similar to the register_delta_weight_transfer_engine logic above.

skyrl/backends/skyrl_train/weight_sync/sharded_rdt/rdt_control_plane.py
SumanthRH, line 1: Can all the RDT files also be in a separate `weight_sync/sharded_rdt` folder given the amount of code for this engine?

tests/backends/skyrl_train/gpu/gpu_ci/megatron/test_megatron_worker.py
```diff
@@ -117,17 +120,49 @@ def get_test_training_batch(batch_size=4) -> TrainingInputBatch:
...
+        #   TP2/TP2:  2 policy + 1 engine x TP2  = 4. Trainer TP gather feeding an
+        #             inference-TP-sharded consumer plan. The model is dense and
+        #             pp == 1, so this covers the plain whole-model weight source.
+        #   TP1/PP2:  2 policy + 2 engines x TP1 = 4. Covers trainer PP>1, so the
+        #             stacked source's PP-local serving engages (each stage declares
+        #             only its own layers via held_names), AND num_engines>1, so the
+        #             consumer ids are offset per deployment by replica_rank -- the
+        #             M:N path that deadlocked before that fix.
+        pytest.param("sharded_rdt", False, 2, 2, 1, 1, None, False, 1, id="sharded_rdt_infer_tp2_trainer_tp2"),
+        pytest.param("sharded_rdt", False, 1, 1, 2, 1, None, False, 2, id="sharded_rdt_infer_tp1_trainer_pp2_2engines"),
```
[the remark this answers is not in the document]
SumanthRH: This is fine for CI - we do single node only.

SumanthRH (approved):

PR #2096, opened 2026-08-27
[fix][megatron] Restore token-batched output order

hershg:
- Restores per-sample forward/backward outputs to input order after token-balanced microbatch packing, matching the order expected by Tinker and the forward-only path.
- Drops dummy padded rows and padding-microbatch outputs before returning learner log-probabilities.

## Testing

```bash
uv run --isolated --extra dev --extra ray --with jaxtyping --with loguru pytest tests/backends/skyrl_train/test_token_based_batching_utils.py
```

All 18 token-batching tests pass; the new case covers uneven microbatches, padded rows, a padding microbatch, and original-order restoration.

---

> [!NOTE]
> **Medium Risk**
> Fixes incorrect per-sample learner outputs under token batching, which could silently corrupt on-policy training; scope is limited to Megatron token-batched forward_backward aggregation.
> 
> **Overview**
> When **token-based microbatching** is enabled, Megatron pads microbatches to a uniform size and may append **padding microbatches** for DP alignment. Per-sample **`loss_fn_outputs`** were accumulated in padded microbatch order, so downstream callers (e.g. Tinker) could see **wrong sample-to-logprob mapping** and extra dummy entries.
> 
> This PR adds **`restore_padded_microbatch_order`** on **`TokenBasedBatchIterator`**, which strips per-microbatch pad slots and ignores padding-microbatch outputs, then rebuilds the list in **original input sample order**. **`MegatronPolicyWorkerBase.forward_backward`** invokes it on aggregated outputs before returning **`WorkerOutput`**, aligning the training backward path with the existing forward-only reorder logic.
> 
> A unit test covers uneven microbatches, intra-microbatch padding, a padding microbatch, and order restoration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a1fa133ae2c3e3ebb40185009de54df50a6e67b9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
+++ b/skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
@@ -1268,6 +1268,9 @@ class MegatronPolicyWorkerBase(MegatronWorker, PolicyWorkerBase):
             for k, v in metrics.items():
                 all_metrics[k].append(v)
 
+        if use_token_batching and microbatch_iterator is not None and all_loss_fn_outputs:
+            all_loss_fn_outputs = microbatch_iterator.restore_padded_microbatch_order(all_loss_fn_outputs, micro_bsz)
+
         # Reduce across microbatches and all-reduce metrics across DP ranks
         # (metrics should be identical within DP groups, i.e., across TP/PP/SP ranks)
         # NOTE: Sum loss metrics because scaling is already applied before the worker reduction.
--- a/skyrl/backends/skyrl_train/workers/worker_utils.py
+++ b/skyrl/backends/skyrl_train/workers/worker_utils.py
@@ -1,5 +1,5 @@
 import math
-from typing import Dict, Iterator, List, Optional
+from typing import Dict, Iterator, List, Optional, TypeVar
 
 import torch
 import torch.distributed as dist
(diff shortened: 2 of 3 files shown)


hershg:
Thanks — I traced the PP path before changing this. forward_backward_mini_batch replaces non-last-stage results with a list of None values and then broadcasts the last stage complete metrics_list to every PP rank before returning. Therefore, by the time forward_backward aggregates outputs, every PP rank has the same full per-sample loss_fn_outputs, and restore_padded_microbatch_order has the expected length on each rank. Gating restoration to only the last stage would make the returned WorkerOutput ordering differ across PP ranks and would undo that broadcast invariant. The existing all_loss_fn_outputs guard also skips restoration when callers request no per-token outputs. I am keeping the current behavior for that reason.

hershg:
Superseded by #2043, whose exact head (`ee28351f`) is the implementation used by the qualified GLM CUDA 13 stack. Downstream variable-length live runs 1048304, 1048311, 1048316, and 1048334 completed with corrected sampler/trainer numerics, including 512/512 terminal rollouts. Closing this duplicate so the ordering fix has one review and merge path.

PR #2097, opened 2026-08-27
fix(tinker): avoid SQLite pool exhaustion under rollout load

hershg:
## Summary

- Keep externally forwarded sample futures and their terminal results in process for their bounded lifecycle; transient sampling never enters SQLite.
- Serialize the remaining high-concurrency SQLite writes, cache immutable sampler validation, and drain forwarding work during shutdown.
- Preserve Tinker 0.25 protobuf responses while moving CPU-heavy result serialization off the API event loop.

Stock SkyRL exhausts its 5+10 SQLite pool when Tau creates 512 samples alongside training futures and heartbeats. This PR removes sample creation/completion from that database hot path while preserving retry-safe terminal retrieval for the lifetime of the API process.

```text
sample request → in-memory future → vLLM
                                  ↓
client retrieve ← protobuf result ← terminal result
                 (bounded retry TTL; no SQLite write)
```

## Before / after

Each synthetic arm ran four waves of 512 model-path samples, 512 forward/backward futures, and 32 heartbeats.

| Arm | Result |
| --- | --- |
| Stock `27ae52bf` | Pool exhaustion in 4.01s |
| #2059 `92c3628d` | Fails wave 1; validation, training, and heartbeats still contend on SQLite |
| #2061 `2760b587` | Samples 2,048/2,048; training 511/2,048; heartbeats 7/128 |
| This PR through `20db38cf` | Samples 2,048/2,048; training 2,048/2,048; heartbeats 128/128; retry-safe retrieval passes |

[The comparison chart, exact commands, decision report, and raw CSV](https://www.notion.so/3c965e48eeef8192912fd489fec15716) preserve the full ablation.

For 512 simultaneous 8,192-token results, bounded off-event-loop protobuf serialization reduced the maximum event-loop gap from 1.284s to 0.007s (total batch time 1.352s versus 1.284s synchronously).

## Live integration

- [1048276](https://xmanager.trajectory.ai/experiments/1048276) and [1048319](https://xmanager.trajectory.ai/experiments/1048319): independent cold B300 runs; 10/10 steps, checkpoints 0/5/10, eval 500/500.
[...]

PR #2100, opened 2026-08-28
[Tinker] Harden forwarded rollouts under SQLite contention

j316chuck:
## Summary

- Keep active forwarded sample futures out of SQLite's coordination hot path, allocate durable negative request IDs, batch terminal persistence, and serialize SQLite mutations before connection checkout.
- Bound non-colocated inference with one absolute deadline and pre-dispatch-only retries; persist typed/redacted failures, advertise zstd trainer uploads, and map Qwen3.6 Megatron LoRA keys correctly for vLLM.
- Fence admission per model during unload, cancel and await that model's forwarding tasks, flush their terminal rows, and only then create the unload future.

This is the SkyRL half of [Trajectory #4257](https://github.com/Trajectorylabs/trajectory/pull/4257). Source under review is `d9d6ba84f1da61828a89479042e5e5f1826198f3`, based on `59d4daedee24c7b1a79d857bbf322a6d195c3792` and Hersh's hot-path work in #2097.

## Assumptions

- PostgreSQL and colocated/JAX paths retain their existing persistence and stale-session behavior.
- A forwarded completion is non-idempotent after dispatch. Only connect failures proven pre-dispatch may retry; read/write/protocol failures are terminal.
- Managed non-colocated Megatron/FSDP adapters are reclaimed by explicit unload or service shutdown. Periodic engine-side stale cleanup is disabled because that process cannot observe the API process's forwarding registry.
- On unload failure or drain timeout, safety wins: the adapter stays loaded and the model stays fenced.

## Before / after

| Problem | Before | Production change | Evidence |
| --- | --- | --- | --- |
| SQLite fan-out | Stock `59d4daed` failed wave 1 with 1,049 pool errors, 512 retrieval timeouts, 0/2,048 terminal samples, 15/2,048 training futures, and 0/128 heartbeats. | In-process active EXTERNAL futures, pre-checkout write serialization, batched terminal durability, cached immutable validation. | `d9d6ba84`: 2,048/2,048 samples and 2,048/2,048 training futures, 16 optimizer + 16 checkpoint operations, 128/128 heartbeats, zero errors in 7.16s. |
| Ambiguous futures | Negative request IDs and terminal results depended on a shared SQL hot path; failures often surfaced as generic retrieve 400s. | Crash-safe durable ID blocks, exact sequence idempotency, typed/redacted terminal errors, bounded persistence and shutdown. | Unit tests cover rollover/restart, duplicate sequence reuse, persistence failure, eviction, and exact terminal accounting. |
| Large forwarding traffic | Trainer uploads were uncompressed; one transport retry could run two 1,800s reads and hide duplicate work. | Advertise protobuf zstd, use one absolute inference deadline, retry only pre-dispatch connection failures, and keep router retries disabled for non-idempotent completions. | Exact-image Qwen run accepted real 512-trajectory compressed trainer waves; recovered pre-body disconnects created no duplicate or missing durable sequence. |
| Model unload | A completed job could remove a LoRA while old forwarded work still ran, producing post-unload `model not found`. | Per-model admission fence, cancel/await forwarding, flush terminal persistence, then enqueue unload; another model remains independent. | Deployment XID 1048603 drained with `forwarding_tasks=0`; create, sync, sample, real Megatron F/B, and unload were all durable and ordered, with 51/51 HTTP 200s. |

## Control flow

```text
before: request -> INSERT shared future -> forward -> UPDATE result
        unload -------------------------------------> delete adapter

after:  reserve durable ID -> register in memory -> forward under one deadline
        terminal -> batched durable flush
        unload -> fence model -> cancel/await -> flush model -> delete adapter
```

## Validation

- 134 passed, 23 skipped: broad Tinker suite.
- 56 focused forwarding/store/engine tests passed; causal unload suite 4/4 passed.
- Deterministic hot path: 2,048 sample + 2,048 training futures, optimizer/checkpoint/heartbeat load, zero errors.
- Synthetic full-payload gate: 262,112 prompt tokens plus 8,192-token training/result payloads, all durable.
- Ruff, Black, secret scan, and diff checks passed.

Immutable runtime:

```text
image sha256:04d2f97a0579bcfb48c1b734e83755066eaf7d3fc8dfa66c76c38e85e20ad78c
  -> clean embedded SkyRL d9d6ba84f1da61828a89479042e5e5f1826198f3
  -> no source overlay, startup patch, mutable mount, or runtime install
```

Live status: XID 1048603 passed the full deployment ladder. XID 1048617 is the ongoing 10-step, 512-concurrency Qwen3.6-27B stress run; optimizer steps 0-5 and checkpoint 5 are complete, and eval improved from 0.41 mean / 26% exact at step 0 to 0.77 / 72% at step 5 with 50/50 accounted in both rounds.

Known residual, not hidden: four `/forward_backward` uploads disconnected while Starlette was still reading the body. SDK retry recovered every logical shard with contiguous durable sequence IDs and no duplicates. A separate ~9-minute cluster-wide WireGuard outage also self-recovered. Neither was a SQLite failure, but the upload disconnect remains P2 transport/observability follow-up work.


hershg:
Decomposition update: the SQLite/external-future hot path is now isolated in #2097, and the typed forwarding deadline plus pre-dispatch-only retry policy is isolated in #2118. Please treat those portions of this umbrella as superseded. The remaining source-owned work here still needs to be split by concern (model unload fencing, Qwen LoRA key mapping, compression/diagnostics) before this PR is ready for maintainer review; we should not merge the umbrella as-is.

PR #2101, opened 2026-08-28
[Tinker] Keep active sampler state in process

j316chuck:
## Summary

- keep non-colocated forwarded sample futures in the API process instead of the shared SQL future table
- cache ready sampling-session routes in a bounded in-process LRU so hot `/asample` requests avoid sampling-session, model, and checkpoint database reads
- retain sequence idempotency, bounded retention/eviction, protobuf content negotiation, and database fallback for evicted routes
- leave database-backed behavior unchanged for other request and backend paths

## Why

Four concurrent LoRA trainers first saturated the SQLite connection pool through forwarded sampling futures. Moving those futures in process removed that path, but live validation showed `/asample` still exhausted the same pool while resolving a sampling session through three database reads. The SDK cannot sample until the random session ID has been returned by `save_weights_for_sampler`, so caching that route after the save commit is safe and removes the remaining hot-path database contention.

## Validation

- 98 passed, 23 skipped: full `tests/tinker` suite
- 7 passed: focused external future-store, ready-route LRU, and cached `/asample` tests
- cached-route test proves `/asample` does not query or commit through the database session
- Ruff and Black checks passed

## Deployment status

Draft pending an immutable image built from commit `9fc5579e9` and live multi-LoRA GPU validation. The paired Trajectory PR will pin the resulting digest and cross-link this PR.

### DIFF
--- a/skyrl/tinker/api.py
+++ b/skyrl/tinker/api.py
@@ -6,7 +6,9 @@ import re
 import shutil
 import signal
 import threading
+from collections import OrderedDict
 from contextlib import asynccontextmanager, suppress
+from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
 from typing import Annotated, Any, AsyncGenerator, ClassVar, Literal
 from uuid import uuid4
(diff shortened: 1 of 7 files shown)


j316chuck:
Superseded by #2100 so this work ships as one SkyRL PR. The paired scheduler and runtime integration is Trajectory #4257.

PR #2103, opened 2026-08-30
Raise a clear error when a batch is smaller than the DP shard stride

### LINKED ISSUE #1609: `RayPPOTrainer._remove_tail_data` silently truncates a training batch to empty and throws `ValueError`
With SkyRL v0.2.0, if [`RayPPOTrainer._remove_tail_data`][trainer-tail] computes a per-batch DP-shard `stride` larger than `train_batch_size`, it returns an empty list, the trainer calls the generator with zero prompts, and the run dies in [`get_rollout_metrics`](https://github.com/NovaSky-AI/SkyRL/blob/skyrl-v0.2.0/skyrl/train/generators/utils.py#L314) on `np.min` of an empty array.

The request here is to add an up-front validation and/or warning to prevent this.

## Repro

```none
trainer.placement.policy_num_gpus_per_node=6   # → policy DP world = 6
trainer.placement.ref_num_gpus_per_node=6      # colocated with policy
generator.inference_engine.tensor_parallel_size=2
generator.inference_engine.num_engines=1       # → vLLM DP = 1
generator.n_samples_per_prompt=4
trainer.train_batch_size=2
```

The ensuing trainer logs:

```none
Generating Trajectories: 0it [00:00, ?it/s]
... Finished: 'generate', time cost: 0.00s
ValueError: zero-size array to reduction operation minimum which has no identity
```

Why does this happen? From [`trainer.py:381-387`][trainer-tail]:

```python
stride = lcm_dp_size // math.gcd(lcm_dp_size, n_samples_per_prompt)
if stride <= 1:
    return entries
kept_prompts = (len(entries) // stride) * stride   # ← can return 0
```

For our config: `lcm_dp_size=6`, `gcd(6, 4)=2`, `stride=3`, `kept_prompts = (2 // 3) * 3 = 0`. Two prompts in, zero out.

## Workaround

The workaround is to bump `train_batch_size` to a multiple of `stride` (3, for the config above). Alternatively, pick a placement that yields `stride <= 1` (e.g. the 4+4 split with TP=4 → `gcd(4, 4)=4`, `stride=1`), or pick an `n_samples_per_prompt` that divides `lcm_dp_size`.

[trainer-tail]: https://github.com/NovaSky-AI/SkyRL/blob/skyrl-v0.2.0/skyrl/train/trainer.py#L365-L387

yurekami:
## Summary

- Fixes #1609: when `train_batch_size` is smaller than the DP shard stride (`lcm_dp_size / gcd(lcm_dp_size, n_samples_per_prompt)`), `RayPPOTrainer._remove_tail_data` silently truncated the batch to zero prompts, and the run died much later in `get_rollout_metrics` with an opaque `ValueError: zero-size array to reduction operation minimum which has no identity`.
- Raise a clear `ValueError` at the truncation point instead, naming the stride, the quantities it derives from, and the config knobs that fix it (`trainer.train_batch_size`, placement, `generator.n_samples_per_prompt`). This fires on the first batch of step 1, before any generation.
- `_remove_tail_data` is the single choke point: the sync trainer, the async example trainer, and the skyrl-agent integration all route through it. The train dataloader uses `drop_last=True` and the fully-async trainer (batch size 1) does not call this method, so a batch smaller than the stride is always a config error — there is no legitimate partial-tail case that could hit the new raise.

## Testing

New unit test `tests/train/test_remove_tail_data.py` covering the truncation behavior, the `stride <= 1` passthrough, and the issue's exact repro config (`lcm_dp_size=6`, `n_samples_per_prompt=4`, batch of 2 → stride 3):

```bash
uv run --extra dev --extra skyrl-train pytest tests/train/test_remove_tail_data.py -v
```

- Verified red/green: on `main` the new raise test fails with `DID NOT RAISE`; with this change all 3 pass.
- `uv run --extra dev --extra skyrl-train pytest tests/train/` on this branch: 743 passed, 1 skipped, and one pre-existing `ImportError` in `test_sft_tokenization.py::test_chat_vlm` that reproduces identically on `main` (4298730) in my environment (WSL Ubuntu, CPU-only).
- `pre-commit` (ruff, black, secrets) passes on the changed files.

Prepared with Claude Code (agent-assisted); all verification above was run locally.

---

> [!NOTE]
> **Low Risk**
> Validation-only change at an existing batch-truncation choke point; no change to successful sharding paths beyond clearer failure for invalid configs.
> 
> **Overview**
> Fixes **#1609** by changing `RayPPOTrainer._remove_tail_data` so a misconfigured batch no longer becomes an empty prompt list and fails later in rollout metrics.
> 
> When the prompt count is **non-zero** but below the DP shard stride (`lcm_dp_size / gcd(lcm_dp_size, n_samples_per_prompt)`), the method now raises a **`ValueError`** on step 1 (before generation) that names the stride, `lcm_dp_size`, `n_samples_per_prompt`, and suggests fixing **`trainer.train_batch_size`**, placement, or **`n_samples_per_prompt`**. **Empty** batches are still returned unchanged.
> 
> Adds **`tests/train/test_remove_tail_data.py`** for normal truncation, `stride <= 1` passthrough, the issue repro (batch of 2 with stride 3), and empty-batch behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3582a1f88172947ffe1ba510bfed90d38d4b61c9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/skyrl/train/trainer.py
+++ b/skyrl/train/trainer.py
@@ -630,6 +630,11 @@ class RayPPOTrainer:
         Each prompt produces `n_samples_per_prompt` samples. For data-parallel
         training we care that the total number of samples is nicely splittable
         across the (combined) data-parallel size of all enabled models.
+
+        Raises:
+            ValueError: If the batch has fewer prompts than the shard stride, so
+                truncation would leave zero prompts and the training step would
+                fail much later in rollout-metrics aggregation.
         """
         lcm_dp_size = self.dispatch.get_lcm_dp_size()
 
@@ -646,6 +651,15 @@ class RayPPOTrainer:
             return entries
 
         kept_prompts = (len(entries) // stride) * stride
+        if kept_prompts == 0:
+            raise ValueError(
+                f"Batch of {len(entries)} prompts cannot be sharded evenly: with the combined "
+                f"data-parallel size of the enabled models (lcm_dp_size={lcm_dp_size}) and "
+                f"generator.n_samples_per_prompt={n_samples_per_prompt}, the number of prompts "
+                f"per batch must be a multiple of {stride}. Increase trainer.train_batch_size "
+                f"to at least {stride}, or adjust the placement or n_samples_per_prompt so that "
+                f"fewer prompts are required per batch."
+            )
         return entries[:kept_prompts]
 
     def build_models(self, PolicyWorker, CriticWorker, RefWorker):
--- a/skyrl/train/trainer.py
+++ b/skyrl/train/trainer.py
@@ -651,7 +651,9 @@ class RayPPOTrainer:
             return entries
 
         kept_prompts = (len(entries) // stride) * stride
-        if kept_prompts == 0:
+        # An already-empty batch is passed through unchanged: zero prompts shard
+        # evenly, and it is not the truncation failure this error is about.
+        if entries and kept_prompts == 0:
             raise ValueError(
                 f"Batch of {len(entries)} prompts cannot be sharded evenly: with the combined "
                 f"data-parallel size of the enabled models (lcm_dp_size={lcm_dp_size}) and "
--- a/tests/train/test_remove_tail_data.py
+++ b/tests/train/test_remove_tail_data.py
@@ -49,3 +49,10 @@ def test_raises_when_batch_is_smaller_than_stride():
     stub = _stub_trainer(lcm_dp_size=6, n_samples_per_prompt=4)
     with pytest.raises(ValueError, match="train_batch_size"):
         RayPPOTrainer._remove_tail_data(stub, list(range(2)))
+
+
+def test_empty_batch_passes_through():
+    # An empty batch shards evenly and is not the truncation failure above,
+    # so it keeps its previous behaviour of being returned unchanged.
+    stub = _stub_trainer(lcm_dp_size=6, n_samples_per_prompt=4)
+    assert RayPPOTrainer._remove_tail_data(stub, []) == []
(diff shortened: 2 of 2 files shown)


yurekami:
Good catch — that was a real regression in the patch, and it is fixed in 3582a1f.

Before this PR an empty `entries` list fell through to `entries[:0]` and was returned unchanged. My guard changed that: `kept_prompts` is 0 for an empty batch too, so it raised and told the user to increase `trainer.train_batch_size`, which is both wrong (zero prompts shard evenly under any stride) and unrelated to the failure the error is meant to describe.

The guard is now `if entries and kept_prompts == 0:`, so the error is limited to the case it is about — a non-empty batch truncated to zero.

Added `test_empty_batch_passes_through` to pin the behaviour. Verified red/green in WSL:

```
# with the fix
tests/train/test_remove_tail_data.py::test_truncates_to_multiple_of_stride PASSED
tests/train/test_remove_tail_data.py::test_keeps_all_entries_when_stride_is_one PASSED
tests/train/test_remove_tail_data.py::test_raises_when_batch_is_smaller_than_stride PASSED
tests/train/test_remove_tail_data.py::test_empty_batch_passes_through PASSED
4 passed in 77.89s

# with the guard reverted to `if kept_prompts == 0:`
E  ValueError: Batch of 0 prompts cannot be sharded evenly: ... Increase
   trainer.train_batch_size to at least 3, ...
FAILED tests/train/test_remove_tail_data.py::test_empty_batch_passes_through
1 failed, 3 passed
```

Disclosure: prepared with the help of an AI code assistant; the test output above is from real runs.

PR #2120, opened 2026-09-01
[fix][lora] Remove unloaded adapters from vLLM engine

hershg:
- vLLM's native runtime unload removes only the serving registry entry, so warm reuse can retain engine-side LoRA state that points at an adapter directory SkyRL has deleted.
- Remove the engine entry before the registry entry, and fail model deletion closed when inference cleanup fails.
- Reproduced on GLM 5.3 B300 XID 1049461. This is a compatibility fix until vLLM PR #42634 is released.

## Testing

```
bash format.sh
```

```
pytest tests/backends/skyrl_train/inference_servers/test_vllm_server_actor.py
pytest tests/backends/skyrl_train/inference_servers/test_remote_inference_client.py::TestLoRAControlPlane
pytest tests/tinker/skyrl_train/test_warm_runtime_lifecycle.py
```

The focused suites passed 3, 5, and 10 tests respectively.

### DIFF
--- a/skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
+++ b/skyrl/backends/skyrl_train/inference_servers/remote_inference_client.py
@@ -1319,11 +1319,10 @@ class RemoteInferenceClient(InferenceEngineInterface):
 
     async def unload_lora_adapter(self, lora_name: str) -> Dict[str, Any]:
         """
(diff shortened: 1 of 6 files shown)


SumanthRH:
Closing after offline discussion: This is not strictly needed for enabling GLM 5.3 training

PR #2121, opened 2026-09-01
[fix][inference] Cancel abandoned generation

bvolpato:
## Summary

Cancel work started by `/skyrl/v1/generate` when its HTTP caller disconnects. Today a replay or evaluation timeout can close the connection while `engine.generate()` keeps running and consuming scheduler/KV-cache capacity.

## Changes

- Race generation against the request disconnect signal after consuming the body.
- Cancel and drain only that request's generation task. This lets vLLM's existing generator cancellation path abort its internal request ID.
- Shield cleanup from repeated handler cancellation.
- Return status 499 when SkyRL observes a disconnected caller. Connected callers and normal generation remain unchanged.

## Testing

- `39 passed`: disconnect cleanup, repeated handler cancellation, normal completion, endpoint integration, and existing generate-wire coverage.
- Full pre-commit suite passed: Ruff, Black, and Gitleaks.
- `uv lock --check` passed.

## Scope

This covers the non-streaming SkyRL replay endpoint. It does not add request timeouts or change streaming endpoints. Local tests prove cancellation reaches the engine generator; TP/GPU abort latency remains for hosted validation.

## Did this cause any problems?

Revert this commit to restore the previous endpoint behavior.

### DIFF
--- a/skyrl/backends/skyrl_train/inference_servers/request_cancellation.py
+++ b/skyrl/backends/skyrl_train/inference_servers/request_cancellation.py
@@ -1,0 +1,45 @@
+"""Tie non-streaming inference work to the lifetime of its HTTP caller."""
+
+import asyncio
+from collections.abc import Awaitable, Callable
+from typing import TypeVar
+
+from fastapi import Request
+
+T = TypeVar("T")
+
+
+class RequestDisconnectedError(Exception):
+    """The HTTP caller disconnected before inference completed."""
+
+
+async def _wait_for_disconnect(request: Request) -> None:
+    while True:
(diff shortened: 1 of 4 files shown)


bvolpato:
Validation update for `605909c`: CPU, training, gym, and code-quality checks passed. The [GPU workflow](https://github.com/NovaSky-AI/SkyRL/actions/runs/33595073057/job/100136763129) failed while submitting the Anyscale job with `Your user credentials are invalid`, before GPU tests ran; Vercel also requires authorization.

The current tests establish cancellation reaches the engine generator, not TP/GPU scheduler drain or abort latency. Those remain unverified until the hosted test can run.

PR #2114, opened 2026-09-01
vllm bump 0.28

hao-aaron:
# Upgrade vLLM 0.26.0 → 0.28.0

Minimal bump to keep everything working. Migrating onto vLLM's new trainer-send
abstraction is deliberately deferred to a follow-up PR.

## Concerns

**`deep_gemm` is unavailable.** vLLM 0.28 pins `torch==2.13.0`, but our CUDA extension
wheels top out earlier (flash-attn/causal-conv1d/TE at 2.12, mamba-ssm at 2.11), so torch
is overridden to 2.11. vLLM's main extensions are stable-libtorch-ABI and load fine, but
`vllm/third_party/deep_gemm/_C` is a version-specific build linked against torch 2.13's
`c10` and fails to import. vLLM falls back, so nothing breaks — but the DeepGEMM fused-MoE
and sparse-attn-indexer paths are gone. No test catches this (Qwen2.5-0.5B never reaches
those kernels). **Reviewers: flag if FP8 MoE matters here.**

**`packed` is now an init-time wire param.** 0.28 moved it from the per-round update info
to `NCCLWeightTransferInitInfo`, where it defaults to `False`. Getting this half-right does
not raise: the trainer packs, the worker unpacks one-by-one, and the transfer hangs inside
NCCL. `BroadcastInitInfo.to_api_payload()` now sends it. Only the GPU weight-sync test
covers this.

**flashinfer bumped, `flashinfer-cubin` dropped.** 0.28 imports
`flashinfer.autotuner.set_autotune_process_group`, absent in 0.6.13 — every engine failed
to start. Now `0.6.16.post3`. `flashinfer-cubin` was never published past 0.6.13 and
flashinfer hard-errors on a version mismatch, so it is dropped rather than pinned; it is
optional and falls back to the cubin cache. `torchvision` pinned to 0.26.0 (not stable-ABI,
hard-pins torch).

**`nccl_trainer_send.py` is vendored and temporary.** 0.28 deleted
`NCCLWeightTransferEngine.trainer_send_weights` and its `trainer_init` staticmethod. This
re-wraps `packed_nccl_broadcast_producer` and `nccl_common.trainer_init`, both unchanged in
0.28. Deleted by the follow-up PR. `sharded_rdt` needed no changes — it already vendors its
own trainer ABCs.

Also removes `patch_hybrid_fp8_kv_wake.py`, now redundant (vllm#41602 is upstream in 0.28).

## Testing

[...]

PR #2142, opened 2026-09-03
vLLM Stateful Weight Sync Sender

hao-aaron:
# Weight sync: migrate onto vLLM 0.28's trainer-send abstraction

Deletes SkyRL's `WeightTransferStrategy` / `WeightTransferSender` / `WeightExtractor`
layers and drives all four weight-sync backends through vLLM's own
`TrainerWeightTransferEngine`, over vLLM's native RLHF routes.
Follow-up to #2114.

---

## Why

vLLM 0.26 shipped only the *receive* half of the weight-transfer abstraction, so SkyRL
owned the send half: a strategy/sender hierarchy, its own extractor + chunk types, and a
`/collective_rpc` worker extension that wrapped every receive in
`set_current_vllm_config` because vLLM did not.

0.28 ships the send half (`TrainerWeightTransferEngine`, `WeightSource`,
`VLLMWeightSyncClient`, `WeightTransferTrainerFactory`) and wraps `set_current_vllm_config`
itself in `GPUWorker.start_weight_update` / `update_weights` / `finish_weight_update`. All
of SkyRL's parallel machinery becomes redundant.

## Shape of the change

**Before** — worker → SkyRL strategy → SkyRL sender → `/collective_rpc` → SkyRL worker
extension → vLLM engine.

**After** — worker builds a `WeightSource` → vLLM trainer engine → native RLHF routes →
vLLM engine (subclassed only where SkyRL needs a hook).

`init_weight_sync_state` is now four statements; `broadcast_to_inference_engines` is one
`engine.send_weights()` plus a memory bracket.

```python
self._weight_sync_engine = await self._weight_sync_thread(
    build_trainer_engine,
    ie_cfg=inference_engine_cfg,
    colocate_all=self.cfg.placement.colocate_all,
    rank=torch.distributed.get_rank(),
    inference_world_size=inference_world_size,
    source_factory=self._build_weight_source,
    server_urls=..., data_parallel_size=..., base_model_path=...,
)
```

## New modules (`weight_sync/`)

| file | lines | what |
|---|---|---|
| `sources.py` | 120 | `FsdpWeightSource` / `MegatronWeightSource` — vLLM's `metadata()` + `__iter__`, nothing else |
| `trainer_engines.py` | 310 | `build_trainer_engine`: resolve backend → build source → `trainer_init` |
| `control_plane.py` | 275 | `SkyrlWeightSyncClient` (blocking HTTP) + the per-server init rewrites |
| `skyrl_engines.py` | 195 | receive side: `skyrl_nccl` / `skyrl_ipc` + the drafter-reload proxy |
| `delta_trainer.py` | 199 | `DeltaTrainerWeightTransferEngine` |
| `patches/vllm/patch_model_runner_registry.py` | 59 | records `GPUModelRunner` so an engine can reach the drafter |

## Deleted

`transfer_strategy.py`, `broadcast_strategy.py`, `cuda_ipc_strategy.py`,
`delta_strategy.py`, `sharded_rdt_strategy.py`, `nccl_trainer_send.py`,
`weight_extractor.py`, `weight_extractor_utils.py`, `layerwise_reload.py`,
`rdt_control_plane.py`, `rdt_control_protocol.py`.

Also gone: `WeightChunk`, `WeightUpdateRequest`, `FSDPWeightExtractor`,
`MegatronWeightExtractor`, `get_transfer_strategy_cls`, `for_servers` /
`to_api_payload` / `get_vllm_transfer_engine`, `sender_initializes_receivers`,
`RemoteInferenceClient.init_weight_update_communicator` / `update_weights_nccl` /
`update_weights_ipc` / `start_weight_update` / `finish_weight_update`.

`sharded_rdt_base.py` shrank 410 → 207: the vendored ABCs now come from the wheel; what
remains is the two channels a *pull* backend needs and vLLM has no concept of
(`held_names()` ownership, `groups()` / `iter_groups()`).

## Backends

| logical | trainer engine | receive engine |
|---|---|---|
| `nccl` | vLLM's `NCCLTrainerWeightTransferEngine` | `skyrl_nccl` |
| `ipc` | vLLM's `IPCTrainerWeightTransferEngine` | `skyrl_ipc` |
| `delta` | `DeltaTrainerWeightTransferEngine` | `delta` |
| `sharded_rdt` | `ShardedRDTTrainerWeightTransferEngine` | `sharded_rdt` |

The receive side takes new names for NCCL/IPC because SkyRL subclasses vLLM's engines and
`register_engine` refuses an already-registered name. `get_transfer_strategy` stays and is
called from **both** sides — the driver (to configure the servers) and
`build_trainer_engine` (to pick the trainer engine) — so the two cannot disagree.

## The worker extension

`NewInferenceWorkerWrap` goes from 8 weight-sync methods to 2. Both remaining are limits of
*dispatch*, not of the engine abstraction:

- `fetch_weights` — `/collective_rpc` reaches worker methods by name only, and no native
  route can invoke an engine method.
- `skyrl_sleep_for_weight_sync` / `skyrl_wake_for_weight_sync` — `EngineCore.sleep`
  hardcodes `clear_prefix_cache = level >= 1`, and `CuMemBackend.suspend` cannot express
  "discard weights, offload kv_cache".

Gone with the rest: `skyrl_start/finish_weight_update`, `update_weights_nccl`,
`update_weights_ipc`, `init_weight_transfer_engine_rdt`, `update_weights_rdt`, and
`LayerwiseReloadWorkerMixin` (0.28's `get_numel_loaded` is character-identical to SkyRL's
patched copy, so the `CopyCounter` patch is no longer needed).

## Behavior changes worth reviewing

1. **No bucketing.** The Megatron export is one whole-model `export_hf_weights` call.
   Bucketing never bounded memory (it *accumulates* a bucket before handing it on); its
   real purposes — IPC handle count and fused-loader grouping — are subsumed by vLLM's
   `packed_ipc_producer`. One call also satisfies `_accumulate_grouped_export`'s
   "every task of a `group_key` in one call" by construction.

2. **`packed=True` forced on IPC**, overriding vLLM's default. The unpacked path holds a
   contiguous copy of *every* parameter until past `finish_weight_update` — the whole model
   resident on the trainer.

3. **Packed buffer sized from the source.** vLLM's 1 GiB default is smaller than a
   large-vocab embedding (Qwen3-235B: 1.24 GiB), which *raises* on the IPC path. Pulls the
   Megatron `metadata()` dry export forward to init.

4. **Delta gained a drafter reload it never had.** The old
   `DeltaWeightTransferEngine.receive_weights` was a bare `self.model.load_weights(...)`, so
   MTP + delta silently drafted with stale weights. Now re-streams `iter_tensors` a second
   time. *This is the one place the PR is not behavior-preserving — happy to split it out.*

5. **Delta owns its layerwise lifecycle.** Previously no-ops, with the `/collective_rpc`
   wrap doing it; now `start/finish_weight_update` run `initialize/finalize_layerwise_reload`.

6. **`_weight_sync_thread`** selects this rank's CUDA device inside the worker thread.
   `send_weights` runs off the event loop, the current device is thread-local and defaults
   to 0, and under `RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES` this rank's device is not
   0 — so the NCCL communicator, IPC handles and gathers would land on the wrong GPU.

7. **Capability probes replace declared flags.** `getattr(engine, "skyrl_…", default)`,
   because two of the four trainer engines are vLLM's own classes and cannot carry SkyRL
   attributes. `sender_initializes_receivers` is gone — vacuous now that every engine
   rendezvouses in its own `trainer_init`.

## Spec-decode drafter

Unchanged in effect, moved in mechanism. `spec_decode_utils.py` is untouched.

vLLM's native `/start_draft_weight_update` is a **separate session** with its own data —
right for EAGLE (separate checkpoint), wrong for MTP, where the head's weights are already
in the policy's stream and `share_embeddings` unconditionally aliases the target's
`embed_tokens`/`lm_head` into the draft (so a draft session would re-send the single
largest tensor). Instead the drafter is reloaded from the same stream in the same session,
via a `_LoadWeightsProxy` installed only when this process actually has a drafter.

Per backend: `nccl` proxy → proxy (same); `ipc` direct call → proxy (same effect);
`delta` none → second pass (**new**); `sharded_rdt` none → none (**still a gap**).

**Known limitation, pre-existing:** none of these run `initialize_layerwise_reload` on the
*draft* model, so `process_weights_after_loading` never runs there. Fine for bf16 MTP
heads; wrong for a quantized drafter.

## Tests

Deleted with their subjects: `test_transfer_strategies.py` (969 lines),
`test_sharded_rdt_strategy.py`, `test_weight_chunk.py`, `test_weight_extractor_utils.py`,
`test_rdt_control_plane.py`.

Added: `test_control_plane.py` (the init rewrites — a bad NCCL `rank_offset` hangs in the
rendezvous rather than erroring), `test_sources.py` (channel agreement + laziness),
`test_trainer_engines.py` (init-info choices, backend resolution, capability probes).

Rewritten: `test_prefix_cache_reset.py` (probes, with a fake engine rather than a Mock so
the *defaults* are actually exercised), GPU `test_weight_sync.py` (all four backends now go
through `build_trainer_engine`, i.e. the production path),
`test_megatron_extractor_consistency.py` → `test_megatron_weight_source.py`.

| suite | result |
|---|---|
| CPU `-m "not vllm"` | 1,484 passed |
| CPU `-m "vllm"` | 257 passed |
| GPU `test_weight_sync.py` | **4 passed** — NCCL TP=2, NCCL 1P1D PD, colocated IPC, sharded RDT |
| GPU `test_delta_weight_sync_e2e.py[fsdp]` | 1 passed |
| GPU `test_offload_kv_weight_sync.py` | passed |
| ruff + black | clean |

Not run — needs a real Megatron model: GPU `megatron/test_megatron_weight_source.py` and
the megatron half of the delta e2e. Apply `run_train_gpu_ci` and
`run_train_megatron_gpu_ci`.

The delta e2e needs `~/data/gsm8k/validation.parquet`
(`python examples/train/gsm8k/gsm8k_dataset.py --output_dir ~/data/gsm8k`). Without it the
test fails in `get_test_prompts` *after* the sync has already succeeded, which reads like a
weight-sync failure and is not one.

## Known gaps

- **No GPU run has had a speculative-decoding drafter**, so the proxy branch has never
  executed. A `speculative_config` parametrization on the delta or offload e2e would close it.
- **No Megatron source on GPU** — `MegatronWeightSource` is unit-tested against a fake
  bridge only. `NCCLTrainerWeightTransferEngine._checked_iter` now validates
  `metadata()` against iteration at runtime, so expect it to surface latent ordering bugs
  on the first real Megatron MoE run; that is the feature working.
- **Single trainer rank in every GPU test** — `is_sender=False` paths (IPC handle
  all-gather, delta's result gather) are unexercised.
- **PP-collective cache at `pp > 1`** — megatron-bridge's `broadcast_from_pp_rank` takes a
  `cache_key`, and the old extractor rebuilt tasks each sync "so mapping objects start with
  clean PP-collective caches". A single long whole-model export may accumulate per-mapping
  cached tensors. Needs a peak-memory measurement.

## Docs

`.claude/docs/weight_sync.md` rewritten; `architecture.md`, `backends/fsdp.md` and
`docs/content/docs/getting-started/inference_architecture.mdx` updated.

---

> [!NOTE]
> **High Risk**
> Touches every training-to-inference weight path (NCCL, IPC, delta, sharded RDT), NCCL rank-offset arithmetic, and distributed worker threading/device context—misconfiguration tends to hang or silently desync rollouts rather than fail fast.
> 
> **Overview**
> **Replaces SkyRL’s custom weight-sync stack** (`WeightTransferStrategy` / senders / `WeightChunk` / extractors and `/collective_rpc` NCCL/IPC/RDT paths) with **vLLM 0.28’s trainer-send model**: each training rank builds a `WeightSource` (`FsdpWeightSource` / `MegatronWeightSource`) and a `TrainerWeightTransferEngine` whose `send_weights()` drives the native RLHF lifecycle (`/init_weight_transfer_engine` → start → update → finish).
> 
> **New trainer-side plumbing** includes `build_trainer_engine` (`weight_senders.py`), blocking HTTP `SkyrlWeightSyncClient` with NCCL/RDT per-server init rewrites (`control_plane.py`), engine registration (`register.py`), and receive subclasses `skyrl_nccl` / `skyrl_ipc` with drafter reload (`weight_receivers.py` + `patch_model_runner_registry.py`). Delta moves under `weight_sync/delta/` with `DeltaTrainerWeightTransferEngine` owning publish, pre-pause `fetch_weights`, pause/resume, and layerwise reload on the receive engine.
> 
> **`NewInferenceWorkerWrap` is narrowed** to dispatch-only hooks: checkpoint-delta `fetch_weights` and direct `CuMemAllocator` sleep/wake for KV offload; layerwise reload and packed NCCL/IPC receive logic are removed. **`RemoteInferenceClient`** no longer fans out weight-sync init/chunk RPCs—the driver/async client keeps pause, `fetch_weights`, LoRA, etc., while trainers use the sync control plane.
> 
> **Docs and ops**: large rewrite of `.claude/docs/weight_sync.md` and public inference architecture; FP8 Hopper example switches rollout to **4× TP4** (blockwise FP8 shard constraints), adds env/W&B knobs and Megatron CPU optimizer offload defaults.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2481fe657552c7b4f99fe0ddac0f2359d1947b1f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


skyrl/backends/skyrl_train/inference_servers/new_inference_worker_wrap.py
```diff
@@ -284,71 +150,28 @@ def skyrl_wake_for_weight_sync(self, tags: list) -> None:
...
-
-        from vllm.config import set_current_vllm_config
-
-        typed_update_info = self.weight_transfer_engine.parse_update_info(update_info)
-        with set_current_vllm_config(self.vllm_config), torch.device(self.device):
-            self.weight_transfer_engine.receive_weights(typed_update_info)
+    def _skyrl_draft_model(self):
+        """The spec-decode drafter module, or None when there is no drafter."""
+        get_draft_model = getattr(self.model_runner, "get_draft_model", None)
+        return get_draft_model() if callable(get_draft_model) else None
```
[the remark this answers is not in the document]
SumanthRH: This is incorrect: https://github.com/vllm-project/vllm/blob/2cf0a6915ce544dc493a0990f2ea38d81601128a/vllm/v1/worker/gpu/model_runner.py#L456-L460

tests/backends/skyrl_train/weight_sync/test_transfer_strategies.py
SumanthRH, line 1: I had an agent review the deletions in this file. It looks like you've deleted a bunch of tests for code that is still present. Can we revert that? We should always keep tests for code that lives in skyrl even if there's something similar on the vLLM side for RDT. 

| Deleted class | Still-live code it covered |
|---|---|
| `TestPpLocalOwnership` (~120 lines) | `rdt_send.py:961` `held_names`, `_pp_local`, `_demoted`, walk reorder |
| `TestExpertOwnership` (~75 lines) | EP stamping, "expert count not divisible by ep_size raises" (`rdt_send.py:1053+`) |
| `TestHeldNamesComposition` | ownership composition across pp × ep |
| `TestStampedYieldValidation` | held-name/tensor agreement checks |
| `TestQkvIndexDeviceCtx` | `rdt_send._qkv_index_device_ctx` (`rdt_send.py:53`) |
| `TestShardedRdtVllmRegistration` | `rdt_vllm_register.ensure_registered()` |
| `TestLoraLoadRequest` | `weight_sync/base.py` `LoraLoadRequest` |

SumanthRH: Can you simple move these RDT specific tests to a separate file and not delete them

skyrl/backends/skyrl_train/weight_sync/delta_trainer.py
```diff
@@ -0,0 +1,198 @@
...
+    """Register ``delta`` in ``WeightTransferTrainerFactory`` (idempotent)."""
+    from vllm.distributed.weight_transfer.factory import WeightTransferTrainerFactory
+
+    if DELTA_BACKEND in WeightTransferTrainerFactory._registry:
+        return
+    WeightTransferTrainerFactory.register_engine(
+        DELTA_BACKEND,
+        "skyrl.backends.skyrl_train.weight_sync.delta_trainer",
+        "DeltaTrainerWeightTransferEngine",
+    )
```
SumanthRH, line 198: This is not called anywhere?

skyrl/backends/skyrl_train/workers/fsdp/fsdp_worker.py
```diff
@@ -205,42 +102,27 @@ def init_model(self, model_path, num_training_steps: int = None):
...
+        are the names vLLM expects; the outer wrapper's are prefixed differently.
+        ``weight_prefix`` covers the one case where they still differ: syncing a
+        CausalLM backbone into a vLLM multimodal namespace.
+        """
         weight_prefix = "language_model." if self._is_multimodal_lm_only else ""
-        self.weight_extractor = FSDPWeightExtractor(
-            self.model.model,
-            enable_bucketing=enable_bucketing,
-            batch_size_threshold_gb=(
-                inference_engine_cfg.weight_transfer_threshold_cuda_ipc_GB if enable_bucketing else 0.0
```
SumanthRH, line 235: Deleting this means that SkyRL doesn't set the bucket size for packed transfer anymore. `weight_transfer_threshold_cuda_ipc_GB` becomes unused. 

Can we pass this to vLLM in init info? 

https://github.com/vllm-project/vllm/blob/8a728663c1c3eeace834a95f5654fa653cc1998c/vllm/distributed/weight_transfer/ipc_engine.py#L60

SumanthRH: Also, the current logic has a common weight extractor for all weight sync backends. Which means that `weight_transfer_threshold_cuda_ipc_GB` also sets the bucket size for NCCL. 

So we should use the same field `weight_transfer_threshold_cuda_ipc_GB` to set the packed transfer size even in the case of NCCL: https://github.com/vllm-project/vllm/blob/8a728663c1c3eeace834a95f5654fa653cc1998c/vllm/distributed/weight_transfer/nccl_engine.py#L80

I'll work on renaming the config in SkyRL meanwhile to avoid confusion

skyrl/backends/skyrl_train/inference_servers/base.py
```diff
@@ -125,19 +122,16 @@ async def sleep(self, *args: Any, **kwargs: Any):
...
+        """Load weights the engine can reach itself, rather than transferring them.
 
-        Args:
-            init_info: WeightSyncInitInfo from the sender containing all info needed
-                to create the appropriate receiver.
+        The only caller is the LoRA path, which passes a
+        :class:`LoraLoadRequest` naming an adapter directory on disk. Tensor
+        transfer goes through the trainer-side engines (see
+        ``weight_sync/trainer_engines.py``).
         """
```
SumanthRH, line 132: Remove these docstring edits for `update_named_weights`? 

They are wrong

skyrl/backends/skyrl_train/weight_sync/sharded_rdt/rdt_send.py
SumanthRH, line 1: One pattern I see is some weird docstring comments from the agent that references now deleted code. 

I have added many guidelines in the `.claude/` folder in SkyRL to specifically avoid such docstrings. 

Can you point your agent to the guidelines in `.claude/docs` and specifically the guide here: 
asking it to make edits?

https://github.com/NovaSky-AI/SkyRL/blob/c516f3a5634701f2d157753cb47bc3f8271b0f11/.claude/docs/contributing.md?plain=1#L51-L157

SumanthRH (changes requested):
Did a first pass! Take a look!

SumanthRH:
Overall given the large set of changes, we should also run some e2e tests. It would be good to: 

1. Run the Qwen 3 235B weight sync benchmark with RDT + Megatron + vLLM, repeating the same settings in the blog: https://vllm.ai/blog/2026-08-22-rdt-weight-transfer
2. Do a before and after comparison for Qwen 3 235B weight sync for the NCCL broadcast backend with like Megatron TP 8 EP 8 and vLLM TP 4. Ideally this is a real E2E training config with GSM8K for like 2 steps. This is to see if any of the weight export changes on the trainer broke something.

SumanthRH:
Ran both the E2E benchmarks:

1. Qwen 3 235B weight sync benchmark with RDT + Megatron + vLLM over 4 H100 nodes, following[ this script ](https://github.com/NovaSky-AI/SkyRL/blob/52063b6567a350e02d326699cd7d0f380bbad466/examples/train/megatron/sharded_rdt/run_megatron_rdt_dpep_qwen3_235b_a22b_lora_4node.sh#L10): Weight sync time averaged around 3.856 s across ten updates, compared to 3.49s from the benchmark
2. Qwen 3 235B weight sync benchmark over NCCL with Megatron TP8 PP2 EP8 and vLLM TP4 PP2: Post-refactor weight sync is around 23.80 s vs 70.24 s on `main` across two real GSM8K updates each: 66.1% faster. I believe this is largely due to the change in weight sync iterator logic in this PR.

SumanthRH:
The primary outstanding TODO in this branch is to merge from `main` and resolve conflicts after FP8 RL changes in https://github.com/novasky-ai/skyrl/pull/1898

SumanthRH:
ran e2e FP8 training on Blackwell, reward and logprob diff are in expected range compared to main. timing is in fact after the weight source related changes

Screenshot 2026-09-18 at 2 36 18 PM

Screenshot 2026-09-18 at 2 36 46 PM

Screenshot 2026-09-18 at 2 37 23 PM

Screenshot 2026-09-18 at 2 37 34 PM

SumanthRH:
Ran E2E FP8 Training on Hopper, in line with results on main:


Screenshot 2026-09-18 at 2 39 19 PM

Screenshot 2026-09-18 at 2 39 36 PM

Screenshot 2026-09-18 at 2 39 54 PM

SumanthRH (approved):

PR #2155, opened 2026-09-04
[deps] bump transformers to 5.16.1

erictang000:
## What

Bumps `transformers` from 5.8.0 to 5.16.1.

- `pyproject.toml`: `transformers>=5.6.1,<=5.8.0` -> `>=5.6.1,<=5.16.1`, in both the
  project dependencies and the linux `override-dependencies` entry.
- `tokenizers` floor moves `0.21.2` -> `0.23.1`. transformers 5.16.1 requires
  `tokenizers>=0.23.1`; without raising the floor the resolver holds transformers at
  5.15.1 (which caps `tokenizers<=0.23.0`) rather than forking tokenizers across
  platforms.
- `docker/pyproject.amd.toml` mirrors both pins.
- `uv.lock` relocked: the only resolved versions that change are
  `transformers 5.8.0 -> 5.16.1` and `tokenizers 0.22.2 -> 0.23.2`. No git dependency
  moved — Megatron-Bridge, Megatron-LM and the rest are on the same commits as `main`.

No `skyrl/` or `tests/` file is touched: `git diff main..HEAD -- skyrl/ tests/` is empty.

### CPU — green

| Suite | Result |
| --- | --- |
| `cpu_skyrl_train`, non-vllm | 1568 passed, 13 skipped |
| `cpu_skyrl_train`, vllm | 193 passed, 1 skipped |
| `cpu_skyrl` (tx / tinker / utils) | 228 passed, 5 skipped |

The `cpu_skyrl` run additionally reported 6 failures that are artifacts of running a CPU
workflow on a GPU box, not of the bump — 4 re-ran clean with `CUDA_VISIBLE_DEVICES=""`,
and `test_llama3[1]`/`[2]` fail identically on `main` and are `pytest.skip`-ed on CI.
Details in the comments.

### GPU — 11 of 12 green

Run as Anyscale jobs off this commit, using each workflow's own config and entrypoint.

- [x] `gpu_skyrl` (tests/tx/gpu)
- [x] `gpu_skyrl_train`
- [x] `gpu_skyrl_train_megatron` — fails the same 4 tests on `main`; no regression
- [x] `gpu_skyrl_train_megatron_models`
- [x] `gpu_ci_h100`
- [x] `tinker_skyrl_train_backend_gpu`
- [x] `gpu_e2e_ci`
- [x] `gpu_e2e_ci_fully_async`
- [x] `gpu_e2e_ci_megatron`
- [x] `gpu_e2e_ci_sft`
[...]

PR #2093, opened 2026-08-26
Target GatedDeltaNet projections with Megatron LoRA

kalectory:
## Summary

- Include GatedDeltaNet `in_proj` and `out_proj` modules in Megatron’s existing `all-linear` LoRA target expansion.
- This enables ordinary LoRA adapters on hybrid Qwen3.5/3.6 GDN layers without adding an algorithm-specific training path.

## Testing

```bash
uvx --from ruff==0.11.9 ruff check skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py skyrl/train/config/config.py
```

Ruff passes. End-to-end Qwen3.6-27B B300 runs validated TP4 training, GDN adapter attachment, 2xTP2 vLLM loading and live adapter sync for both standard LoRA and PiSSA; the matched 100-step GPQA runs completed successfully.

---

> [!NOTE]
> **Medium Risk**
> Changes which modules receive LoRA and adapter export/sync for GDN layers; wrong classification could break training or silently omit adapter weights on Qwen3-Next-style models.
> 
> **Overview**
> Megatron **`all-linear`** LoRA now also targets GatedDeltaNet **`in_proj`** and **`out_proj`** (standard and canonical LoRA lists), so hybrid models like Qwen3.5 can train GDN adapters through the existing Megatron path.
> 
> **`in_proj`** is removed from that list when **`gdn_in_proj_lora_is_safe`** reports the bridge uses the fused Qwen3-Next-style **`in_proj_qkvz`/`in_proj_ba`** mapping—PEFT export cannot split or merge those adapters correctly. Separate **qkv/z/b/a** layouts and models without GDN mappings still get **`in_proj`**.
> 
> Config docs for **`target_modules`** describe this behavior. New Megatron GPU CI tests pin bridge classification and document that fused-layout **`in_proj`** adapters still fail or drop weights on export until Megatron-Bridge gains support.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8d7fa6cdda26ecaee223369332463e9631a4c5ff. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #2159, opened 2026-09-04
[fix][megatron] Fix the 4 failing tests in the Megatron GPU CI suite

erictang000:
## What

Fixes the 4 tests failing in `gpu_skyrl_train_megatron`. That workflow has been red on
every push to `main` since at least Sept 1; both failures predate the transformers 5.16.1
bump (#2155) and reproduce on `main` before it.

### 1. `test_lora_gdn_targets.py` — 3 assertions that never ran

```
AttributeError: 'types.SimpleNamespace' object has no attribute 'hf_pretrained'
skyrl/backends/skyrl_train/distributed/megatron/megatron_utils.py:216
```

`gdn_in_proj_lora_is_safe` installs the bridge's `hf_pretrained` onto the model bridge
before the mapping lookup, for the bridges whose `mapping_registry` probes the
checkpoint. The fake bridge these tests build carries only `_model_bridge`, so all three
cases raised before reaching their assertion — since #2093, where the read and the
fixture landed together.

Fixed by giving the fake the attribute. These registries are constructed directly and
never read it, so `None` is the honest value.

### 2. `test_vlm_sft_hf_parity` — OOM on CI's 22GB L4

```
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 7.93 GiB.
GPU 0 has a total capacity of 22.03 GiB of which 5.62 GiB is free.
  ... in _ParamAndGradBuffer.__init__ -> self.grad_data = torch.zeros(...)
```

The test runs fp32 (deliberately — it needs numerical comparability with the HF forward)
on a single GPU with tp=pp=1, so nothing shards. It still built the DDP param/grad
buffers and the optimizer's fp32 master + AdamW state, which add ~15GB on top of the 8GB
model. It only ever forwards, so it now takes `inference_only_init`, matching what the
other forward-only Megatron tests do (`test_megatron_models.py`, `test_mtp_*.py`).

Measured on one GPU:

| | peak GPU memory |
| --- | --- |
| before | 31,100 MiB (30.4 GiB) — over the L4's 22.03 GiB |
| after | 15,250 MiB (14.9 GiB) |

The parity numbers are identical either way — `max_abs_diff=0.009659`,
`mean_abs_diff=0.002209` — so this drops only training state, nothing the test measures.

### 3. Docstring

`inference_only_init`'s docstring described only the optimizer half of what the Megatron
path skips. It also skips the DDP wrap, which is where most of the memory goes.

## Testing

```
tests/backends/skyrl_train/gpu/gpu_ci/megatron/test_megatron_vlm_init.py
tests/backends/skyrl_train/gpu/gpu_ci/megatron/test_lora_gdn_targets.py
15 passed in 1065.59s (0:17:45)
```

`test_lora_gdn_targets.py` alone goes 3 failed / 3 passed -> 6 passed.

Full `-m megatron` suite on `l4_ci`, the hardware that actually OOMs, same compute config
and entrypoint as the workflow:

```
== 163 passed, 4 skipped, 152 deselected, 263 warnings in 12628.18s (3:30:28) ==
```

vs the `main` baseline through the same harness — 159 passed / 4 failed, with the same 4
skipped and 152 deselected. The four failures converted to passes and nothing else moved.

### DIFF
--- a/skyrl/train/config/config.py
+++ b/skyrl/train/config/config.py
@@ -639,8 +639,9 @@ class PolicyConfig(BaseConfig):
     Backend-specific behavior:
     - FSDP: initialize weights in bf16 instead of fp32 (skipping the fp32 master weights
       that mixed-precision training requires) and skip optimizer/LR-scheduler construction.
-    - Megatron: skip optimizer/LR-scheduler construction (DistributedOptimizer eagerly
-      materializes fp32 master + AdamW state on GPU)."""
+    - Megatron: skip the DDP wrap, whose ``_ParamAndGradBuffer`` holds a param and a
+      grad copy of the model, and skip optimizer/LR-scheduler construction
+      (DistributedOptimizer eagerly materializes fp32 master + AdamW state on GPU)."""
 
 
 @dataclass
--- a/tests/backends/skyrl_train/gpu/gpu_ci/megatron/test_lora_gdn_targets.py
+++ b/tests/backends/skyrl_train/gpu/gpu_ci/megatron/test_lora_gdn_targets.py
@@ -57,7 +57,13 @@ QWEN35_NAMES = [
 
 
 def _bridge_with_registry(registry: MegatronMappingRegistry):
-    return SimpleNamespace(_model_bridge=SimpleNamespace(mapping_registry=lambda: registry))
+    # `gdn_in_proj_lora_is_safe` installs the bridge's `hf_pretrained` onto the
+    # model bridge before the lookup, for the bridges whose `mapping_registry`
+    # probes the checkpoint. These registries are built directly and never read it.
+    return SimpleNamespace(
+        hf_pretrained=None,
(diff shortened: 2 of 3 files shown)


erictang000:
### Verified on `l4_ci` — the workflow is green

Ran the full `-m megatron` suite on the hardware that actually OOMs, same compute config
and entrypoint as the workflow:

```
== 163 passed, 4 skipped, 152 deselected, 263 warnings in 12628.18s (3:30:28) ==
```

Against the `main` baseline from the same harness:

| | passed | failed |
| --- | --- | --- |
| `main` | 159 | 4 |
| this branch | 163 | 0 |

Same 4 skipped and 152 deselected on both sides, so the four failures converted to
passes with nothing else moving.

That closes the loop on the part my local box could not prove: the 30.4GB -> 14.9GB peak
was measured on an H100, so the claim that it now fits a 22GB L4 was an inference until
this run. It fits.

PR #2167, opened 2026-09-05
[chore] Re-enable vllm compile cache

### LINKED ISSUE #2137: Re-enable torch compile cache with vLLM
# Summary

SkyRL historically disabled saving torch compile cache with vLLM due to race conditions with multiple vllm engines in older vllm versions (0.9.2 at the time): 

https://github.com/NovaSky-AI/SkyRL/blob/25afc279fe24b5d05e4c4679113a6c697aa656f9/skyrl/train/utils/utils.py#L752-L757

We should check if such race conditions are still an issue and enable saving compile cache if possible

SumanthRH:
Closes #2137.

## What

Stops force-setting `VLLM_DISABLE_COMPILE_CACHE=1` for every vLLM run. An explicit user setting is now forwarded to the Ray workers instead.

## Why the workaround existed

Added in #95 (vLLM 0.9.0, July 2025) after multi-node runs failed at compile time with `RuntimeError: vLLM failed to compile the model ... corrupted compilation artifact`.

## Why it's safe now

`torch >= 2.10` writes inductor artifacts atomically ([pytorch#162432](https://github.com/pytorch/pytorch/pull/162432)) which seems to be the failure mode that produced corrupt artifacts. We pin torch 2.11. Also it looks like vLLM ships the same fix as a backport for older torch.

## Measurements: Standalone tests with concurrent engines 

Hardware: 8xB200, `vllm 0.28.0` / `torch 2.11.0+cu130`, engines sharing one `VLLM_CACHE_ROOT`.

**Startup — 4 concurrent engines:**

| | eng 0 | 1 | 2 | 3 |
|---|---|---|---|---|
| cold | 63.90s | 63.89s | 63.94s | 63.82s |
| warm | 30.46s | 30.48s | 30.46s | 30.57s |

**~52% faster engine startup**, ~33s saved per engine (Qwen2.5-0.5B; should grow with model size).

Repeated the test 5 times × 8 concurrent engines, cache wiped each round so all 8 compile and write simultaneously. No failures were observed

## Measurements: Tests using SkyRL's serve entrypoint

Hardware: 8xH200 node, `vllm 0.28.0` / `torch 2.11.0+cu130`, engines sharing one `VLLM_CACHE_ROOT`.

Configurations under test

| id | model | engines | TP | DP | GPUs |
|---|---|---|---|---|---|
| exp1 | Qwen/Qwen2.5-1.5B-Instruct | 8 | 1 | 1 | 8 |
| exp2 | Qwen/Qwen2.5-1.5B-Instruct | 4 | 2 | 1 | 8 |
| exp3 | Qwen/Qwen3-30B-A3B | 4 | 1 | 2 | 8 |

### Results

**exp1 — Qwen2.5-1.5B, TP=1, 8 engines**

| arm | init engine | compilation | 
|---|---|---|
| cold | 27.86 s | 12.70 s |
| warm | **18.38 s** | **3.47 s** |

**exp2 — Qwen2.5-1.5B, TP=2, 4 engines**

| arm | init engine | compilation |
|---|---|---|
| cold | 30.31 s | 13.12 s |
| warm | **17.89 s** | **3.84 s** |

**exp3 — Qwen3-30B-A3B, DP=2, 4 engines**

| arm | init engine | compilation |
|---|---|---|
| cold | 57.42 s | 24.30 s |
| warm | **31.50 s** | **5.87 s** |

More importantly, no errors were observed over multiple re-runs, so we should be good. 

---

> [!NOTE]
> **Medium Risk**
> Changes default vLLM startup/compile behavior for all training runs without an explicit override; regressions could surface as compile failures or corrupt cache on older PyTorch stacks, though the project now pins torch 2.11 where atomic writes mitigate that.
> 
> **Overview**
> **Stops forcing `VLLM_DISABLE_COMPILE_CACHE=1` on every vLLM Ray runtime** when the inference backend is vLLM. vLLM’s default torch compile cache is left enabled unless the driver process already sets `VLLM_DISABLE_COMPILE_CACHE`, in which case that value is logged and forwarded to workers (same pattern as other optional vLLM env vars).
> 
> This removes a workaround that always disabled the compile cache after vLLM 0.9.0 compile failures, so warm starts can reuse cached inductor artifacts again. Users who still need the old behavior can export `VLLM_DISABLE_COMPILE_CACHE=1` before launch; several example scripts already do that.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit df97857a6a9a0757c5a1f5a45bfebe70b07327cf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


SumanthRH:
For future reference:

I also had Claude analyse the cache directories with vLLM - what gets saved and how the cache is keyed. Here's the writeup from Claude:



## vLLM torch.compile cache analysis

Everything lives under `VLLM_CACHE_ROOT` (default `~/.cache/vllm`), in
`torch_compile_cache/`. There are two sibling trees, written by different code paths, keyed
differently, and used at different moments.

#### Folder 1 — the AOT cache (`torch_aot_compile/`)

```
torch_compile_cache/torch_aot_compile/<64-hex hash>/
├── rank_0_0/model          the serialized ahead-of-time compiled forward
├── rank_1_0/model          one per TP/DP rank in the engine
└── inductor_cache/         TORCHINDUCTOR_CACHE_DIR is redirected here (decorators.py:558)
```

The whole compiled model is saved as one artifact and loaded back wholesale on the next start (`Directly load AOT compilation frompath …`). Note that Inductor's cache is nested *inside* it — so when this folder is bypassed,
you lose the Inductor cache along with it.

**Its name is `sha256` of three things** (`decorators.py:541-543`):

| component | what it covers | changes when… |
|---|---|---|
| `env_hash` | every vLLM env var not on the ignore list, `envs.compile_factors()` | you change any `VLLM_*` setting — **including `VLLM_RAY_BUNDLE_INDICES`**, which is why the cache fragments per engine |
| `config_hash` | `vllm_config.compute_hash()` — model, dtype, parallelism, compilation and kernel config | you change the model, TP/DP, `max_num_batched_tokens`, `enforce_eager`, … |
| `_model_hash_key` | `sha256(vllm.__version__ + forward.__qualname__ + forward.__code__.co_firstlineno)` | you upgrade vLLM, or edit the model file such that `forward`'s first line moves |

#### Folder 2 — the Dynamo/backend cache (`<hash>/`)

```
torch_compile_cache/<10-hex hash>/
└── rank_{tp_rank}_{dp_rank}/<prefix>/          e.g. rank_0_1/backbone/
    ├── vllm_compile_cache.py                   the index: {(range, subgraph, compiler): artifact path}
    ├── artifact_compile_range_1_8192_subgraph_0    one per piecewise subgraph
    ├── artifact_compile_range_1_8192_subgraph_1
    ├── artifact_compile_range_1_8192_subgraph_48
    ├── computation_graph.py                    the traced FX graph, for debugging
    └── cache_key_factors.json                  the key's raw inputs — very useful for diffing
```

This is the older, finer-grained path: each piecewise subgraph is cached separately and
looked up individually (`Directly load the compiled graph(s) … from the cache`).

**Its name is the first 10 hex chars of `sha256` of four things** (`backends.py:1063-1066`):

| component | what it covers | changes when… |
|---|---|---|
| `env_hash` | same as above | same as above |
| `config_hash` | same as above | same as above |
| `code_hash` | for every traced source file: **its absolute path**, then its contents, concatenated and hashed (`backends.py:1044-1055`) | you edit model code — **or the file simply moves**, which is exactly what `uv run --isolated` does on every process |
| `compiler_hash` | `get_inductor_factors()`: host/system summary, `torch_key()`, and portable dumps of the Inductor and functorch configs | you upgrade torch, change host/GPU, or change an Inductor setting |

Note that both keys contain `env_hash`, so both fragment per engine under
`VLLM_RAY_BUNDLE_INDICES`. Only this one additionally breaks when the interpreter path changes — and the interpreter path changes on every single process under `uv run --isolated`. i.e every re-run with `uv` leads to a different hash, and thus a cache miss. 

My comment: Dynamo backend cache will pretty much never get used in SkyRL because of this. Compilation takes like 1.65s for a 30B on a 8xH100 node, so it's not that bad right now, but it can matter at large scale.

PR #2180, opened 2026-09-08
[deps][megatron] bump megatron-bridge to 8e5f0e1 and megatron-core to b3393bbb

erictang000:
megatron-bridge 2b8cb21d (8/13) -> main HEAD 8e5f0e1 (9/8), and megatron-core 14346b65a -> b3393bbb -- the `3rdparty/Megatron-LM` submodule pin of that bridge rev, and the only core commit Bridge's CI validates against. b3393bbb is on Megatron-LM main; main is 3 commits ahead of it, and those 3 rework `experimental_attention_variant/dsa.py` (NVIDIA/Megatron-LM#6402), so this stays on the validated rev rather than main's tip.

Refreshes `dsa_index_share_recompute.patch` (the NVIDIA/Megatron-LM#6793 backport) against the new `megatron/core/recompute.py`. b3393bbb threads `cp_layout_state` and `packed_sequence_cp_metadata` through `checkpointed_forward`, adding a `cp_layout_state.finalize_layer()` call immediately above the `return custom_forward` that the patch's second hunk anchors on, so that hunk stopped applying ("1 out of 2 hunks FAILED"). #6793 is still not upstream -- b3393bbb's `recompute.py` has no index-share carrier -- so the backport is still required.

This failure mode is silent: `patch_dsa_index_share` logs at INFO and returns False, and `test_dsa_index_share_recompute.py` skips itself when the patch does not apply, so CI stays green while DSA models (GLM 5) under activation recompute on the non-packed path lose the fix and get silently wrong logits. With the refreshed patch the backport applies again and all 5 regression tests run and pass instead of skipping.

No other SkyRL code needed changing: `patch_mla_thd_v_pad` still applies, the GLM-4.7/Qwen3.5 bridge subclasses still register and dispatch, upstream's `build_conversion_tasks(..., weight_dtype=None)` signature and `_megatron_global_param_names_all_pp_ranks` are unchanged, and the megatron worker, strategy and model-wrapper modules all still import.


erictang000:
h100 ci: https://github.com/NovaSky-AI/SkyRL/actions/runs/34289181088  🟩

megatron models ci: https://github.com/NovaSky-AI/SkyRL/actions/runs/34289328040/job/102272100608?pr=2180 🟩

PR #2181, opened 2026-09-08
[GLM tests] Runtime-controlled B300 profiling benchmark

hershg:
- Keeps merged #2190 as the sole trainer-profiler lifecycle, with acknowledged start/status/stop, owner-only optimizer-step advancement, overlap rejection, error-preserving cleanup, and selected-rank Megatron timers that cannot enter L1 barriers skipped by other ranks.
- Adds transport-neutral monotonic client receipts and two honest server aggregates: `inference_engine_initialization_aggregate` and `sampler_weight_sync`. Missing vLLM substages and standalone activation stay absent; distributed attribution uses rank maximum, never rank sum.
- Exposes `skyrl.tinker.qualification.run_lora_qualification` for the shared #2182 evidence sequence, and updates the two-trainer-node 256K recipe to TP4/CP2/PP2.

## Testing

```text
uv run pytest examples/tinker/glm53/tests/test_profiles.py tests/backends/skyrl_train/utils/test_phase_timing.py -q  # 24 passed
uv run pytest examples/tinker/glm53/tests tests/backends/skyrl_train/inference_servers/test_build_vllm_cli_args.py tests/backends/skyrl_train/utils/test_phase_timing.py tests/backends/skyrl_train/utils/test_profiler.py tests/tinker/skyrl_train/test_profiling.py tests/tinker/skyrl_train/test_publication_timing.py tests/tinker/test_engine.py tests/tinker/test_model_info_tokenizer.py tests/tinker/test_profiling_receipts.py tests/tinker/test_qualification.py tests/tinker/test_server_timing.py -m 'not vllm' -q  # 250 passed, 4 deselected
```

Changed-file pre-commit and compileall passed. The local full Tinker collection reached 222 passed/22 skipped; its three setup errors were the CPU render fixture failing because optional local `vllm` was absent. Exact-head [SkyRL-CPU](https://github.com/NovaSky-AI/SkyRL/actions/runs/35021851516) and [SkyRL-Train-CPU](https://github.com/NovaSky-AI/SkyRL/actions/runs/35021851380) are green, including full pytest, engine benchmarks, code quality, gym, without-vLLM, with-vLLM, and Tinker jobs.

## Qualification status

Head: `80a3d95d34519bb2bb4ea7210138f6820d19423f`  
Base: `a7637de1fd71e1696f08cae2b9ffcacb3297b9a0`

Profiling contract v10 is published and validator-clean: `ready.json` SHA-256 `54a932acd3e4b4f38591247e5984b17881b976eb31fe704a442ec454cf086350`; `releases/v10/SHA256SUMS` SHA-256 `7e4fe907399eeb917e23784afdb2d30e80fed4019b45b0be38ae422dd35df60c`. It preserves accepted v9 contract 1.3.0, both schema versions 1.1.0, #2190 dependency metadata, public endpoint/helper identities, and unchanged qualification source bytes while repinning current main/head and camera-readiness evidence. The timer and 256K topology review threads are resolved; no review threads remain open.

No image build or GPU/live campaign was run for this rebased head, by instruction. The automatic [Anyscale GPU workflow](https://github.com/NovaSky-AI/SkyRL/actions/runs/35021851330) failed on an empty token/invalid credentials before job submission or any product test; Vercel is an irrelevant deployment-authorization check for this Python-only PR.

Historical roughly 600-second warm-publication measurements remain old-stack context, not a current denominator. The final performance comparison is matched profiler-disabled baseline Arm A versus the selected independently numerics-qualified native candidate; profiler-enabled runs are attribution-only. Retired Arms B/D and #2195 are excluded.


examples/tinker/glm53/common.json
```diff
@@ -0,0 +1,90 @@
...
+  "generator.inference_engine.data_parallel_size": 1,
+  "generator.inference_engine.distributed_executor_backend": "mp",
+  "generator.inference_engine.weight_sync_backend": "nccl",
+  "generator.inference_engine.enforce_eager": false,
+  "generator.inference_engine.max_num_seqs": 8,
+  "generator.inference_engine.max_num_batched_tokens": 32768,
+  "generator.inference_engine.enable_prefix_caching": false,
+  "generator.inference_engine.enable_chunked_prefill": true,
+  "generator.inference_engine.engine_init_kwargs": {
+    "kv_cache_dtype": "fp8",
```
hershg: use bf16 and not override to fp8

examples/tinker/glm53/common.json
```diff
@@ -0,0 +1,90 @@
...
+    "disable_custom_all_reduce": true,
+    "linear_backend": "triton",
+    "moe_backend": "triton",
+    "lora_target_modules": [
```
hershg: drop these lora_target_modules; vllm will infer from disk the appropriate modules it needs trainer.policy.model.lora.target_modules

examples/tinker/glm53/common.json
```diff
@@ -0,0 +1,90 @@
...
+  "generator.inference_engine.run_engines_locally": true,
+  "generator.inference_engine.language_model_only": true,
+  "generator.inference_engine.num_engines": 1,
+  "generator.inference_engine.tensor_parallel_size": 8,
+  "generator.inference_engine.pipeline_parallel_size": 1,
+  "generator.inference_engine.data_parallel_size": 1,
+  "generator.inference_engine.distributed_executor_backend": "mp",
+  "generator.inference_engine.weight_sync_backend": "nccl",
+  "generator.inference_engine.enforce_eager": false,
+  "generator.inference_engine.max_num_seqs": 8,
```
hershg: drop this

examples/tinker/glm53/configs/common.json
```diff
@@ -0,0 +1,90 @@
+{
```
hershg, line 1: need a client-side tinker script w/ profiler enabled that does the full_context training (https://github.com/NovaSky-AI/SkyRL/tree/main/examples/train_scripts/full_context) -style process for fast iteration

examples/tinker/glm53/run_client.py
```diff
@@ -0,0 +1,174 @@
...
+import httpx
+import tinker
+from tinker import types
+
+
+def build_datum(tokens: list[int], context: int) -> types.Datum:
+    """Repeat a token fixture into exactly `context` input/target positions."""
+    if not tokens or context < 2:
+        raise ValueError("a nonempty token fixture and context >= 2 are required")
+    sequence = (tokens * ((context + 1 + len(tokens) - 1) // len(tokens)))[: context + 1]
```
hershg: magic formula very hard to understand/reason about

examples/tinker/glm53/256k-3n.json
```diff
@@ -0,0 +1,13 @@
+{
+  "trainer.placement.policy_num_nodes": 2,
+  "trainer.policy.megatron_config.pipeline_model_parallel_size": 2,
+  "trainer.policy.megatron_config.transformer_config_kwargs": {
+    "num_layers_in_first_pipeline_stage": 38
+  },
+  "trainer.remove_microbatch_padding": true,
+  "trainer.max_tokens_per_microbatch": 262144,
+  "generator.inference_engine.gpu_memory_utilization": 0.95,
```
hershg: 0.95 is too high right? both here and in 2node

examples/tinker/glm53/glm53_b300.md
```diff
@@ -0,0 +1,136 @@
+# GLM 5.3 on B300: native SkyRL examples
```
hershg: should probably drop / clean this up / streamline to readme/report

examples/tinker/glm53/run_client.py
```diff
@@ -0,0 +1,212 @@
...
+                },
+            }
+            for datum in data
+        ],
+        allow_nan=False,
+    )
+
+
+@contextmanager
+def measure(report, name: str):
```
hershg: 'name' is bad param name

examples/tinker/glm53/run_client.py
```diff
@@ -0,0 +1,212 @@
...
+                continue
+            response.raise_for_status()
+            result = response.json()
+            if result["type"] != "unload_model" or result["model_id"] != model_id:
+                raise RuntimeError(f"unexpected unload result: {result}")
+            return
+    raise TimeoutError(f"unload did not finish for {model_id}; inspect the server before reusing it")
+
+
+def run(args) -> None:
```
hershg, line 14: design should look more opinionated right? / more generic framework for the various phases (create, unload, fwd/bkwds, ...) and the kinds of tests we're testing on (if there's multiple tests/runs/... that we want to run)

current run() is an unreadible huge blob

examples/tinker/glm53/run_client.py
```diff
@@ -0,0 +1,212 @@
...
+        with measure(report, "create_model") as record:
+            trainer = service.create_lora_training_client(base_model=args.model_path, rank=32, seed=0)
+            record["model_id"] = trainer.model_id
+        try:
+            info = trainer.get_info()
+            if not info.is_lora or info.lora_rank != 32:
+                raise ValueError("expected a rank-32 LoRA training client")
+            tokenizer = trainer.get_tokenizer()
+            datums = [
+                build_datum(tokenizer.encode(text, add_special_tokens=False), args.context) for text in FIXTURE_TEXTS
```
hershg: i thought we're testing at full size context right - these fixtures aren't that.

examples/tinker/glm53/run_server.py
```diff
@@ -0,0 +1,90 @@
...
+            "skip_first": 0,
+            "wait": 0,
+            "warmup": 1,
+            "active": 1,
+            "repeat": 1,
+            "record_shapes": True,
+            "profile_memory": True,
+            "with_stack": False,
+        }
+    return config
```
hershg, line 47: deployment-side configs should go here, whereas job-side configs should go on the client, right?

examples/tinker/glm53/run_client.py
```diff
@@ -0,0 +1,212 @@
...
+                        raise ValueError("sampling returned no sequence")
+                    record["output_tokens"] = result.sequences[0].tokens
+            with measure(report, "checkpoint") as record:
+                record["path"] = trainer.save_state("full-context-final").result().path
+        finally:
+            with measure(report, "unload"):
+                unload_model(args.base_url, trainer.model_id)
+
+
+def main() -> None:
```
hershg, line 64: how is warmth being handled in the profiling; is it client or server side too?

examples/tinker/glm53/common.json
hershg: Handled in 49fc5bf: inference model_dtype is bfloat16 and kv_cache_dtype is auto, so KV follows BF16. The FP8 override is removed; this is config cleanup, not a live capacity qualification.

examples/tinker/glm53/common.json
hershg: Handled in 49fc5bf: removed only inference-side lora_target_modules. The trainer's attention/MLP target list is preserved; vLLM reads the exported adapter metadata.

examples/tinker/glm53/common.json
hershg: Handled in 49fc5bf: removed max_num_seqs overrides from common and the 32K profile. The README distinguishes the upstream 1024 ceiling from demonstrated concurrent capacity.

examples/tinker/glm53/run_client.py
hershg: Changed in 1fa19d8. build_full_context_datum now uses cycle/islice to take exactly context_length + 1 tokens, then names input_tokens and target_tokens explicitly. No integer repeat-count formula. CPU tests cover 32,768 and 262,144 scored positions.

examples/tinker/glm53/256k-3n.json
hershg: Reduced both 256K profiles from 0.95 to 0.80 in 1fa19d8, matching the 32K starting point. 0.95 left little headroom; 0.80 is a conservative candidate, not a proof that BF16 KV at 256K fits. The README keeps that qualification explicitly open.

examples/tinker/glm53/glm53_b300.md
hershg: Replaced glm53_b300.md with a shorter README in 1fa19d8: phase flow, setup/run commands, config ownership, artifact meanings and evidence limits. Internal research bookkeeping stays outside the example.

examples/tinker/glm53/run_client.py
hershg: Renamed the helper to measure_phase and its argument to phase_name in 1fa19d8.

examples/tinker/glm53/run_client.py
hershg: Refactored in 1fa19d8. run() is now the short top-level phase sequence, first in the file. prepare_full_context_inputs, run_gspo_step and publish_and_sample own the substantial operations; measure_phase is the single timing/failure-recording primitive. Kept one explicit benchmark protocol rather than adding a configurable test registry. Tests cover order and cleanup failures at publication, sample, reference, backward, optimizer and checkpoint.

examples/tinker/glm53/run_client.py
hershg: The short strings are seeds, not the training tensors. Made that explicit in 1fa19d8: SEED_TEXTS -> build_full_context_datum -> exactly C input tokens, C shifted targets and C unit mask entries. prepare_inputs logs both lengths, datums.json saves the expanded tensors, and CPU tests assert all positions at C=32,768 and 262,144. Only the post-publication sample is intentionally short; it does not qualify full-context inference.

examples/tinker/glm53/run_server.py
hershg: Yes: context/batch/update count/LR are client settings; topology, precision, memory/packing budgets and worker trace setup are server settings. There is a current API limitation: the existing profiler lifecycle is server-runtime-scoped, with no public Tinker per-job start/reset/configure API. The README and CLI help now state this explicitly. Leaving this thread open for agreement on whether this fresh-service profiling protocol is enough for the baseline or we should separately add a job-scoped profiler API; the client does not silently pretend to configure server workers.

examples/tinker/glm53/run_client.py
hershg: Clarified in 1fa19d8. The client now separates initial publication/sample, one real warmup optimizer update, then measured updates; --steps 3 means 1 warmup + 2 measured. The server profiler starts at runtime creation, warms up through optimizer 0, and captures until optimizer 1 (including intervening publication and reference/backward passes). It does not reset for another client on a warm service. Use a fresh dedicated service for the profiled comparison, and unprofiled runs for speed claims. Tradeoff: initial publication adds one sync (4 total by default). Leaving open for review of this protocol; GPU trace collection is still unverified.

examples/tinker/glm53/run_client.py
```diff
@@ -0,0 +1,234 @@
+"""Profile dataset-independent, full-context GSPO through SkyRL's Tinker API."""
```
hershg: we can first test this on small model and check/verify any timing/performance mismatches, and only later then graduate to GLM 5.3
- can in parallel run on both tinker (for baseline) and our own nodes

enable torch profiler for backend config

verify not doing two backwards per step; also profiling/... artifacts should still be collected even in case of OOM during warmup etc...

hershg: Updated in 6c4b5a4, with the OOM guard separately reviewable in a0e28cb. run() now shows one batched reference forward, one batched GSPO forward/backward, one optimizer, and publication/sample per step; default batch size is two full-context sequences total, with one warmup plus two subsequent updates. No extra accumulation requests. Backend torch profiling is enabled on all trainer ranks with no recording warmup gap. Megatron forward/backward/optimizer catch PyTorch OOM locally, attempt active-trace export, then re-raise; a CPU test verifies a real trace before any optimizer step and preserves the original OOM. SIGKILL, failed export, cold model load, and vLLM remain outside this guarantee. Added independent qwen3-0.6b.json (one trainer GPU plus one inference GPU) using the same client; no new test framework. 23 client/config tests and 16 focused profiler tests pass; distributed/GPU tests remain unrun. Hosted Tinker parity is not implemented: this client uses SkyRL-specific GSPO/unload APIs, so matching supported loss semantics is required before a comparison. Leaving this thread open for pre-launch review and small-model GPU validation. No jobs launched.

examples/tinker/glm53/configs/glm53-256k-3n.json
```diff
@@ -0,0 +1,13 @@
+{
```
hershg, line 1: note - for 256k 2node, try PP2, and trade off TP for CP (TP 4 CP 2, instead of TP 8 CP 1)

(point agent at skyrl repo `.claude/skills/parallelism-strategies/SKILL.md` pp picker - could be useful, might be pre-training centric rn though)

examples/tinker/glm53/run_client.py
```diff
@@ -0,0 +1,236 @@
...
+import tinker
+from tinker import types
+
+SEED_TEXTS = (
+    "A river flows past a stone bridge. Trees grow along the bank and birds gather in the branches. ",
+    "Calculate the area of a rectangle: multiply its length by its width. Explain each arithmetic operation. ",
+)
+
+
+def run(args) -> None:
```
hershg, line 14: this got more verbose now and is really hard to read again. needs better design/clean up

examples/tinker/glm53/run_server.py
```diff
@@ -0,0 +1,99 @@
...
+import argparse
+import json
+import os
+from pathlib import Path
+import sys
+
+
+def build_config(profile: str, model_path: Path, state_dir: Path, profile_dir: Path) -> dict:
+    root = Path(__file__).parent / "configs"
+    config = {} if profile == "qwen3-0.6b" else json.loads((root / "common.json").read_text())
```
hershg: note, bit of hack used here

examples/tinker/glm53/run_server.py
```diff
@@ -0,0 +1,99 @@
...
+        "profile_memory": True,
+        "with_stack": False,
+    }
+    return config
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "profile", choices=["qwen3-0.6b", "glm52-32k-2n", "glm53-32k-2n", "glm53-256k-2n", "glm53-256k-3n"]
```
hershg: bit of hack / ideally should be discovered

hershg: Fixed in 98f24a575fbab018a7b9bea4809137255173408c: CLI profile choices are discovered from configs/*.json, excluding the shared common file. No hardcoded model list remains. New-profile discovery and existing profile rendering tests pass.

skyrl/backends/skyrl_train/utils/profiler.py
```diff
@@ -19,6 +20,23 @@ def build_profiler_from_policy_cfg(trainer_cfg):
...
+
+    @wraps(method)
+    def wrapped(self, *args, **kwargs):
+        try:
+            return method(self, *args, **kwargs)
+        except torch.OutOfMemoryError:
+            logger.exception(f"OOM in {method.__name__}; flushing the local worker's profiler")
+            if self.profiler is not None:
+                self.profiler.stop()
+                self.profiler.start()
```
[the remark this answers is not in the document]
hershg: Not applicable to the decorated call sites: Worker.__init__ initializes self.profiler=None (skyrl/backends/skyrl_train/workers/worker.py:296), and the decorator only wraps the Megatron forward/backward/optimizer methods, not construction. Keeping that explicit class contract rather than silently accepting an uninitialized worker. The disabled-profiler and real CUDA-OOM paths have coverage.

examples/tinker/glm53/run_server.py
hershg: Fixed in 98f24a575fbab018a7b9bea4809137255173408c: GLM profiles explicitly declare extends=common; standalone Qwen has no parent. Removed the profile-name special case. Tests cover a newly discovered model inheriting only when explicitly requested, and ensure extends never reaches the backend.

examples/tinker/glm53/run_client.py
```diff
@@ -0,0 +1,249 @@
...
+        while time.monotonic() < deadline:
+            response = client.post("api/v1/retrieve_future", json={"request_id": request_id})
+            if response.status_code == 408:
+                continue
+            response.raise_for_status()
+            result = response.json()
+            if result["type"] != "unload_model" or result["model_id"] != model_id:
+                raise RuntimeError(f"unexpected unload result: {result}")
+            return
+    raise TimeoutError(f"unload did not finish for {model_id}; inspect the server before reusing it")
```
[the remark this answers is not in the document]
hershg: Fixed in 98f24a575fbab018a7b9bea4809137255173408c: each retrieve poll uses the remaining cleanup budget instead of inheriting the 35s HTTP default. ReadTimeout becomes an explicit cleanup TimeoutError with its cause preserved; the unload mutation is never resubmitted. Tests cover 115s then 30s remaining budgets, terminal model identity, and timeout propagation. Example suite: 29 passed + 4 subtests.

examples/tinker/glm53/run_client.py
```diff
@@ -0,0 +1,249 @@
...
+    """Use SkyRL's HTTP unload endpoint; the public SDK has no unload method."""
+    deadline = time.monotonic() + 120
+    with httpx.Client(base_url=base_url.rstrip("/") + "/", timeout=35) as client:
+        response = client.post("api/v1/unload_model", json={"model_id": model_id})
+        response.raise_for_status()
+        request_id = response.json()["request_id"]
+        while time.monotonic() < deadline:
+            response = client.post("api/v1/retrieve_future", json={"request_id": request_id})
+            if response.status_code == 408:
+                continue
```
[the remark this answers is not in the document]
hershg: The canonical endpoint long-polls: retrieve_future waits via wait_for_future/ExternalFutureStore.wait with RETRIEVE_FUTURE_TIMEOUT_SECONDS=300 before returning 408. It does not immediately return not-ready. An additional sleep is therefore not needed for this path. The separate, valid client/server read-timeout mismatch is fixed in 98f24a57.

examples/tinker/glm53/configs/glm53-256k-3n.json
hershg: Reviewed the parallelism skill. The two-node disaggregated profile has 8 trainer GPUs and already uses TP4/CP2/PP1. TP4/CP2/PP2 would require 16 trainer GPUs (three total nodes including inference), so a two-node PP2 candidate must change another axis, e.g. TP2/CP2/PP2 or TP4/CP1/PP2, and recheck expert layout. Leaving this experimental suggestion open rather than silently changing the benchmark target; no new 256K qualification is claimed.

skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
```diff
@@ -1445,15 +1451,23 @@ def forward_backward(
...
-            temperature=self.cfg.algorithm.temperature,
-            loss_fn=loss_fn,
-            loss_fn_config=loss_fn_config,
-            return_per_token_outputs=return_per_token_outputs,
-        )
+        phase_context = nullcontext(None)
+        if self.profiler is not None and self.profiler.check():
+            model_config = get_model_config(self.actor_module[0])
+            assert model_config is get_mcore_model_config(self.model.actor_module[0])
+            phase_context = measure_megatron_schedule(model_config, Timers(log_level=2, log_option="all"))
```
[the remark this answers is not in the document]
hershg: Fixed in c661283f. The phase timers remain rank-local, but measure_megatron_schedule now temporarily sets model_config.barrier_with_L1_time=False while they are installed, so a selected profiling rank never enters Megatron L1 timer barriers that unselected ranks skip. The original value is restored in finally on both success and failure. The regression covers both initial True and False values; the full focused #2181 suite passes (201 tests), with Ruff, Black, diff checks, and commit hooks green. A new immutable image and distributed validation will use this head.

examples/tinker/glm53/configs/glm53-256k-3n.json
hershg: Resolved on the rebased head in 80a3d95d. The `glm53-256k-3n` profile has two 8-GPU trainer nodes (the third node is inference), so TP4/CP2/PP2 uses all 16 trainer ranks, keeps TP within one NVLink domain, and adds context sharding for 256K. `test_profiles_preserve_nested_common_settings` now pins `(nodes=2, TP=4, CP=2, PP=2, context=262144)`; the profile/timer focused suite passed 24 tests and the broader affected suite passed 250 tests (4 deselected). No GPU qualification is claimed.

skyrl/backends/skyrl_train/workers/megatron/megatron_worker.py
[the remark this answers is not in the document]
hershg: Revalidated after rebasing onto `a7637de1` at head 80a3d95d. `measure_megatron_schedule` still forces `barrier_with_L1_time=False` only while rank-local timers are installed and restores the original value in `finally`; the regression covers both initial True and False plus failure cleanup. The focused profile/timer suite passed 24 tests and the broader affected suite passed 250 tests (4 deselected). No selected-rank GPU run is claimed in this CPU-only camera-readiness update.

Issue #2116, opened 2026-09-01
[tinker] Retain only the K most recent checkpoints


ktanishqk:
## Summary

Add configurable retention for the K most recent Tinker checkpoints, as listed in the Q3 roadmap (#1846). This would build on the deletion support added in #1840.

### Proposed behavior:

- Retention is disabled by default

- After a checkpoint saves successfully, keep the newest K completed checkpoints for each model

- Failed or unfinished checkpoints do not count

- Never delete an older checkpoint until the new one has saved successfully

- Reuse the existing checkpoint deletion logic where possible

---

One question is how sampler checkpoints should work. Some are saved to disk, while others are only used temporarily to update the inference engine. SkyRL currently records both in the same way, so we need to ensure retention deletes only files that were actually saved.

Would it be better to support training checkpoints first and add saved sampler checkpoints in a follow-up stacked PR, or should both be handled together? I’m interested in implementing this once the preferred scope is confirmed.

PR #2186, opened 2026-09-09
fix(tinker): finalize failed saves for unloaded models

LOGO127:
## Summary

Move the unloaded-model checks into the existing checkpoint-status context for
training and sampler saves. The request's future and checkpoint then both reach
FAILED with the same error instead of leaving the checkpoint PENDING.

When a queued save is dispatched after its model is no longer loaded, the current
early ErrorResponse completes the future as FAILED but bypasses checkpoint
finalization. Deleting that checkpoint returns 425, and reusing its name returns
409. This change allows the failed checkpoint record to be deleted and retried.

The two-file patch retains successful-save behavior and ephemeral sampler
`persist=False` behavior. It does not repair historical orphan records, change
retention policy (#2116), or implement Megatron checkpoint publication (#2145).

## Validation

- Added 12 CPU regression/control cases: success, backend failure and unloaded
  model across training, persistent sampler and ephemeral sampler saves, plus
  actual ASGI/SQLite delete-and-retry paths. Only the backend is mocked in these
  committed tests.
- Same tests on baseline `0b286bac`: 6 failed / 6 passed. Failures are three
  PENDING-vs-FAILED statuses and three delete responses of 425 instead of 204.
- Signed candidate: all 12 pass. An earlier selected CPU suite including these
  tests passed 94 tests (not an additional 94).
- Supplementary, uncommitted real-JAX lifecycle probe: 6 pass, versus 3 failures
  and 3 passing controls on baseline. This deliberately stages unload before
  dispatch; it does not claim that ordinary FIFO requests reorder themselves.
- Supplementary real TCP smoke: the unmodified API launches its JAX engine child;
  Tinker SDK 0.24.0 creates a tiny-Qwen LoRA model and successfully saves a training
  checkpoint. One test passed in 44.90s. This is a successful-save control, not a
  cross-process reproduction of the unload condition.
- Repository Ruff, Black and secret-detection hooks passed on the reviewed patch;
  signed contents match that patch, and `git diff --check` passes.

Local validation used WSL Ubuntu, Python 3.12 and an independent locked CPU
environment (JAX 0.11.0, Flax 0.12.8, CPU Torch 2.11.0, Tinker 0.24.0), not the
full production dependency installation. Supplementary probes and environment
files are not included in this scoped patch. No GPU training, PostgreSQL,
SDK >=0.25 protobuf compatibility, model quality or performance is claimed.

To run the committed regression suite in a configured JAX development environment:

```bash
uv run --isolated --extra dev --extra jax pytest tests/tinker/test_checkpoint_status.py -q
```

## Attribution

AI-assisted investigation, implementation and test automation with Codex. The
contributor reviewed the patch and personally signed commit `8671853`.

---

> [!NOTE]
> **Medium Risk**
> Changes checkpoint save error handling and DB finalization for training/sampler paths; behavior for successful saves is intended to stay the same but failed-async edge cases are security-adjacent to data lifecycle.
> 
> **Overview**
> **Fixes checkpoint rows stuck in `PENDING` when a queued save runs after the model was unloaded.**
> 
> Training and sampler save handlers used to return an `ErrorResponse` before entering `_checkpoint_status_context`, so the request failed but the checkpoint row never moved to `FAILED`—blocking delete (425) and name reuse (409). The unloaded-model check now runs **inside** that context and raises `_ModelNotLoadedError`, so the same failure path updates the DB and the client still gets a failed result via `process_single_requests`.
> 
> Expected stale-request cases skip full `logger.exception` noise for `_ModelNotLoadedError`. Return types for the two save methods no longer include `ErrorResponse` because failures are exceptional. **Regression tests** cover training/sampler unloaded saves and backend `ValueError`, asserting `FAILED` status and successful checkpoint delete over ASGI/SQLite.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d022965b3a35707506447ed1a4475d6f4473fcbb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/skyrl/tinker/engine.py
+++ b/skyrl/tinker/engine.py
@@ -645,20 +645,17 @@ class TinkerEngine:
 
         return types.LoadWeightsOutput(type="load_weights")
 
-    def process_save_weights(
-        self, model_id: str, request_data: types.SaveWeightsInput
-    ) -> types.SaveWeightsOutput | types.ErrorResponse:
+    def process_save_weights(self, model_id: str, request_data: types.SaveWeightsInput) -> types.SaveWeightsOutput:
         """
         Saves a clean training checkpoint by converting the trimmed NNX graph
         to a pure dictionary before serialization, following official Flax docs.
         """
-        if not self.backend.has_model(model_id):
-            return _model_not_found_error(model_id)
-
         checkpoint_id = request_data.path
         output_path = self.config.checkpoints_base / model_id / f"{checkpoint_id}.tar.gz"
 
         with self._checkpoint_status_context(model_id, checkpoint_id, types.CheckpointType.TRAINING):
+            if not self.backend.has_model(model_id):
+                raise ValueError(_model_not_found_error(model_id).error)
             self.backend.save_checkpoint(output_path, model_id)
             logger.info(f"Saved trimmed training checkpoint for model {model_id} to {output_path}")
 
@@ -669,11 +666,8 @@ class TinkerEngine:
 
     def process_save_weights_for_sampler(
         self, model_id: str, request_data: types.SaveWeightsForSamplerInput
-    ) -> types.SaveWeightsForSamplerOutput | types.ErrorResponse:
+    ) -> types.SaveWeightsForSamplerOutput:
         """Process a save_weights_for_sampler request and save model weights."""
-        if not self.backend.has_model(model_id):
-            return _model_not_found_error(model_id)
-
         # Make sure the user cannot store checkpoints in places like ../../<important file>
         checkpoint_id = Path(request_data.path).name
         output_path = self.config.checkpoints_base / model_id / "sampler_weights" / f"{checkpoint_id}.tar.gz"
@@ -684,6 +678,8 @@ class TinkerEngine:
         persist = request_data.sampling_session_seq_id is None
 
         with self._checkpoint_status_context(model_id, checkpoint_id, types.CheckpointType.SAMPLER):
+            if not self.backend.has_model(model_id):
+                raise ValueError(_model_not_found_error(model_id).error)
             self.backend.save_sampler_checkpoint(output_path, model_id, persist=persist)
             logger.info(f"Saved sampler checkpoint for model {model_id} to {output_path}")
 
--- a/skyrl/tinker/engine.py
+++ b/skyrl/tinker/engine.py
@@ -30,6 +30,10 @@ from skyrl.utils.log import logger
 _MAX_IDS_PER_QUERY = 500
 
 
+class _ModelNotLoadedError(ValueError):
+    """A stale request targets a model that is no longer loaded."""
+
+
 def _model_not_found_error(model_id: str) -> types.ErrorResponse:
     """Log and return an ErrorResponse for a request targeting a model that isn't loaded."""
     logger.info(
@@ -321,7 +325,8 @@ class TinkerEngine:
             status = CheckpointStatus.COMPLETED
             error_message = None
         except Exception as e:
-            logger.exception(f"Error saving checkpoint for model {model_id}, checkpoint {checkpoint_id}: {e}")
+            if not isinstance(e, _ModelNotLoadedError):
+                logger.exception(f"Error saving checkpoint for model {model_id}, checkpoint {checkpoint_id}: {e}")
             error_message = str(e)
             raise
         finally:
@@ -655,7 +660,7 @@ class TinkerEngine:
 
         with self._checkpoint_status_context(model_id, checkpoint_id, types.CheckpointType.TRAINING):
             if not self.backend.has_model(model_id):
-                raise ValueError(_model_not_found_error(model_id).error)
+                raise _ModelNotLoadedError(_model_not_found_error(model_id).error)
             self.backend.save_checkpoint(output_path, model_id)
             logger.info(f"Saved trimmed training checkpoint for model {model_id} to {output_path}")
 
@@ -679,7 +684,7 @@ class TinkerEngine:
 
         with self._checkpoint_status_context(model_id, checkpoint_id, types.CheckpointType.SAMPLER):
             if not self.backend.has_model(model_id):
-                raise ValueError(_model_not_found_error(model_id).error)
+                raise _ModelNotLoadedError(_model_not_found_error(model_id).error)
             self.backend.save_sampler_checkpoint(output_path, model_id, persist=persist)
             logger.info(f"Saved sampler checkpoint for model {model_id} to {output_path}")
 
@@ -754,7 +759,8 @@ class TinkerEngine:
                 try:
                     result = self.process_single_request(request_type, model_id, request_data)
                 except Exception as e:
-                    logger.exception(f"Error processing request {request_id}: {e}")
+                    if not isinstance(e, _ModelNotLoadedError):
+                        logger.exception(f"Error processing request {request_id}: {e}")
                     result = types.ErrorResponse(error=str(e), status="failed")
             results[request_id] = result
         self._complete_futures(results)
--- a/tests/tinker/test_checkpoint_status.py
+++ b/tests/tinker/test_checkpoint_status.py
@@ -1,0 +1,158 @@
+import asyncio
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+import httpx
+import pytest
+from sqlalchemy.ext.asyncio import create_async_engine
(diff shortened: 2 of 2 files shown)


LOGO127:
Thanks for taking a look! A brief CI note: the [GPU Actions job](https://github.com/NovaSky-AI/SkyRL/actions/runs/34348582506/job/102455984091) stopped while submitting its Anyscale job, with `Your user credentials are invalid`, before the GPU tests ran. I am not counting this as GPU validation or attributing it to the patch. The CPU workflow is still running at this check.

Could a maintainer please advise the normal way to run this check for a fork PR when convenient? I have left the workflow and permissions unchanged; I am not asking for secrets to be shared or checks to be bypassed. Local CPU evidence and its limitations are in the PR description. Thank you! This CI investigation and follow-up were prepared with AI assistance.

skyrl/tinker/engine.py
```diff
@@ -645,20 +645,17 @@ def process_load_weights(
...
         """
-        if not self.backend.has_model(model_id):
-            return _model_not_found_error(model_id)
-
         checkpoint_id = request_data.path
         output_path = self.config.checkpoints_base / model_id / f"{checkpoint_id}.tar.gz"
 
         with self._checkpoint_status_context(model_id, checkpoint_id, types.CheckpointType.TRAINING):
+            if not self.backend.has_model(model_id):
+                raise ValueError(_model_not_found_error(model_id).error)
```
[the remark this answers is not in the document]
LOGO127: Thanks for catching the noisy stale-request logs. Addressed in `e0f30b5d8676e3a69e1321ac71cc3c2444478574` with a dedicated internal `_ModelNotLoadedError` in both save paths. The two exception handlers omit tracebacks only for that subtype; the existing `_model_not_found_error` helper still emits one info message, without an additional warning. Checkpoint/future failure finalization and error text are preserved.

The regression matrix covers training, sampler and ephemeral sampler saves: unloaded requests produce no traceback and one info message, while ordinary backend `ValueError` and `OSError` retain both existing exception logs. Successful saves remain covered. On the exact signed commit, the checkpoint suite plus `test_engine.py` reran with **33 passed in 28.61s** in the pinned local CPU environment. This does not claim GPU validation or resolve the previously reported upstream Anyscale credentials issue.

The contributor personally reviewed/signed the increment; implementation and verification used Codex assistance. The signed file blobs match the reviewed and previously hook-checked candidate; no history rewrite was performed.

skyrl/tinker/engine.py
```diff
@@ -684,6 +678,8 @@ def process_save_weights_for_sampler(
         persist = request_data.sampling_session_seq_id is None
 
         with self._checkpoint_status_context(model_id, checkpoint_id, types.CheckpointType.SAMPLER):
+            if not self.backend.has_model(model_id):
+                raise ValueError(_model_not_found_error(model_id).error)
```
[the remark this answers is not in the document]
LOGO127: Addressed in the same signed e0f30b5 increment as the training-save path. The sampler and persist=False ephemeral sampler cases are included in the logging/status regression matrix; full details and validation limits are in my reply on the first thread.

LOGO127:
Thanks for reviewing. I investigated the CPU failure at `e0f30b5`: the two failures are `test_qwen3_moe_layer_lora[1-1]` and `[2-1]`, in the merged-weight numerical comparison. All 33 checkpoint/engine tests passed in that run.

The MoE code/tests, dependency files, and CPU workflow are unchanged by this PR. I ran the three MoE LoRA configurations on both the pre-patch baseline `0b286bac` and PR head in the same pinned local CPU environment; both passed. This is not an exact reproduction of the GitHub runner, so I cannot yet establish the cause or rule out an interaction.

Could a maintainer please rerun the failed CPU job when convenient? My rerun attempt was denied for insufficient repository permissions. I have left the test tolerances and workflow unchanged. If it fails again, I can continue investigating with the runner environment and a reproducible input. Thank you!

This investigation and message were prepared with AI assistance.

tests/tinker/test_checkpoint_status.py
```diff
@@ -0,0 +1,171 @@
...
+            )
+        )
+        session.commit()
+    yield engine
+    engine.db_engine.dispose()
+
+
+@pytest.mark.parametrize("outcome", ["success", "backend_error", "backend_value_error", "unloaded"])
+@pytest.mark.parametrize("mode", ["training", "sampler", "ephemeral_sampler"])
+def test_checkpoint_and_future_reach_matching_terminal_status(checkpoint_engine, mode, outcome):
```
ktanishqk: the grid is easy to grow and hard to read. imo: unloaded × training/sampler + one `ValueError` (for the `_ModelNotLoadedError check`) is enough

tests/tinker/test_checkpoint_status.py
```diff
@@ -0,0 +1,171 @@
...
+
+    if outcome == "unloaded":
+        captured_log.exception.assert_not_called()
+        captured_log.info.assert_called_once()
+        assert "model not loaded" in captured_log.info.call_args.args[0]
+        captured_log.warning.assert_not_called()
+    elif outcome in ("backend_error", "backend_value_error"):
+        assert captured_log.exception.call_count == 2
+    else:
+        captured_log.exception.assert_not_called()
```
ktanishqk: seems brittle. i wouldn't test logs here.
if someone later adds a `logger.info` on this path, this test fails even though the checkpoint status is still right.
i’d drop the counts and just assert the DB rows.

tests/tinker/test_checkpoint_status.py
```diff
@@ -0,0 +1,171 @@
...
+@pytest.mark.asyncio
+@pytest.mark.parametrize("mode", ["training", "sampler", "ephemeral_sampler"])
+async def test_failed_save_can_be_deleted_and_retried(checkpoint_engine, monkeypatch, mode):
+    engine = checkpoint_engine
+    engine.backend.has_model.return_value = False
+    async_db = create_async_engine(get_async_database_url(str(engine.db_engine.url)))
+    monkeypatch.setattr(api.app.state, "db_engine", async_db, raising=False)
+    monkeypatch.setattr(api.app.state, "engine_config", engine.config, raising=False)
+    monkeypatch.setattr(api.app.state, "sampler_checkpoint_validation_lock", asyncio.Lock(), raising=False)
+    monkeypatch.setattr(api.app.state, "validated_sampler_checkpoints", set(), raising=False)
```
ktanishqk, line 65: can we not just check for checkpoint `PENDING` -> `FAILED` and then delete returns `204`? the rest aren't this bug and give the test extra ways to fail.

ktanishqk (changes requested):
fix is right, move it in the `with` statement. however, testing is a bit too extensive and adds failure points that might hurt more than help. going to ask for some consolidation and reduction there - check reviews!

ktanishqk:
hey @LOGO127 - i had started working on this issue after creating it because i was waiting on scope confirmation before opening a pr.
however, thanks for starting work on this. going to turn into a reviewer instead so we can coordinate and land the best version!

LOGO127:
Thanks @ktanishqk for coordinating and taking the reviewer role. The test-only follow-up has now been consolidated locally into three cases: unloaded training save, unloaded sampler save, and one ordinary backend ValueError. It checks the PENDING -> FAILED database transition and DELETE 204, without logger-count, duplicate/retry, or successful-save matrices.

The reduced tests still reproduce the original bug: unmodified pre-fix source gives 2 failed / 1 passed; the candidate plus test_engine.py gives 21 passed. Ruff, Black and secret-detection hooks pass. These are local CPU results in the previously documented independent pinned environment, not a resolution of the existing MoE/Anyscale CI failures.

The follow-up is local pending contributor review/sign-off; this PR's published head remains e0f30b5. Thanks for helping keep the regression focused.

LOGO127:
Addressed the review in e9940cf by reducing the checkpoint regression coverage to the three requested cases: unloaded training save, unloaded sampler save, and one ordinary backend ValueError. The test now only checks PENDING -> FAILED plus DELETE 204; the logger-count, retry/duplicate, ephemeral, and success matrices were removed.

Validation: old production code + the reduced tests gives 2 failed / 1 passed (the two unloaded cases stay PENDING); current PR head + the reduced tests and test_engine.py gives 21 passed. Ruff, Black, and hardcoded-secret hooks pass as well.

Thanks for the direction — the production fix is unchanged.

PR #2209, opened 2026-09-15
[fix] Restore input order for FSDP forward_backward loss_fn_outputs under token batching

avigyabb:
## Summary

With `max_tokens_per_microbatch > 0`, the base `PolicyWorkerBase.forward_backward` (inherited by the FSDP policy workers) extended per-sample `loss_fn_outputs` in packed microbatch order, so clients paired one example's mask with another example's logprobs / elementwise loss. Entries produced for padding microbatches also leaked into the results.

#2043 fixed this exact bug in the megatron worker's `forward_backward`, but the same un-reordered `extend` remained in the shared base worker path.

## Changes

- `skyrl/backends/skyrl_train/workers/worker.py`: collect `loss_fn_outputs` per microbatch and map them back to input positions with `TokenBasedBatchIterator.reorder_and_combine_items` (which also drops padding entries), mirroring the megatron worker's `forward_backward`. Sample-based batching keeps the plain flatten.
- `tests/backends/skyrl_train/test_token_based_batching_utils.py`: CPU regression tests that drive `PolicyWorkerBase.forward_backward` through real token-based packing with a forced padding microbatch, asserting per-sample outputs come back in input order with padding entries dropped; the sample-based and metrics-only branches are covered as well.

The critic worker's `forward_backward` returns no per-sample outputs, so it is unaffected.

## Testing

```
uv run --isolated --extra dev --extra skyrl-train pytest tests/backends/skyrl_train/test_token_based_batching_utils.py --noconftest
20 passed
```

The new ordering test fails on `main` without the fix and passes with it.

---

> [!NOTE]
> **High Risk**
> Incorrect per-sample logprobs/loss under token batching directly corrupts RL/SFT training signals; the change is localized to output reordering but affects a critical training path when token microbatching is enabled.
> 
> **Overview**
> Fixes **misaligned per-sample `loss_fn_outputs`** on the shared FSDP `PolicyWorkerBase.forward_backward` path when `max_tokens_per_microbatch > 0`. Token packing runs microbatches out of input order and can append padding microbatches; the old code concatenated outputs in iteration order, so callers could pair the wrong sample’s logprobs/elementwise loss with masks (Megatron was fixed in #2043).
> 
> **`forward_backward`** now accumulates `loss_fn_outputs` per microbatch, then restores **input order** and **drops padding** via `TokenBasedBatchIterator.reorder_and_combine_items`; sample-based batching still uses a simple flatten, and the metrics-only path returns an empty list when there are no per-token outputs.
> 
> CPU regression tests stub `PolicyWorkerBase.forward_backward` with real token packing (including a forced padding microbatch) to assert ordering and the non-token / no-outputs branches.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit da0729edbc7bba496bc97c9053693d48b8d2fdbd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


SumanthRH (approved):

SumanthRH:
@avigyabb can you resolve merge conflicts?

PR #2221, opened 2026-09-16
[weight_sync] In-memory LoRA adapter sync for Megatron (lora.sync_mode=memory)

avigyabb:
## Summary

Adapter-only LoRA weight sync (Megatron, `merge_lora=false`) can now ship the adapter through the existing weight-sync transport instead of writing PEFT files to `lora_sync_path` and having vLLM read them back. Opt in with `trainer.policy.model.lora.sync_mode=memory`; `disk` remains the default and is unchanged.

Weight sync gains a **target** axis next to the existing **transport** axis. `WeightExtractor.receive_target` (`None` = base model) travels once per sync to `skyrl_start_weight_update`, and the worker picks a sink: `model.load_weights` for the base model, or a staging dict for a LoRA adapter. No new transport is added.

vLLM 0.28 can only load an adapter from a directory, so `patches/vllm/patch_lora_in_memory.py` wraps `WorkerLoRAManager._load_adapter` at worker import and serves `skyrl-memory://<name>` from staged GPU tensors via `LoRAModel.from_lora_tensors`, keeping vLLM's own validation, EP expert filtering and name mapping. This runs against the stock 0.28 wheel — no fork, no pinned dependency changes. `patches/vllm/lora_in_memory_upstream.patch` is the same change shaped as an upstream diff.

## Measured results

GLM-4.7-Flash (`glm4_moe_lite`, 47 layers, 64 routed experts, top-4), LoRA rank 32, 2.0 GB / 18,040-tensor public adapter. One node, 4xH100: Megatron 2 GPUs EP2 -> vLLM 1 engine TP2, non-colocated, NCCL. Three optimizer-backed syncs per mode after the initial sync.

| Sync | disk | memory | disk logprob gap | memory logprob gap |
|---|---:|---:|---:|---:|
| 0 (pre-training) | 10.00 s | 0.51 s | - | - |
| 1 | 6.84 s | 0.50 s | 0.0372 | 0.0347 |
| 2 | 5.15 s | 0.47 s | 0.0314 | 0.0333 |
| 3 | 5.06 s | 0.48 s | 0.0318 | 0.0303 |
| **steady state** | **5.108 s** | **0.475 s** | 0.0316 | 0.0318 |

**10.7x faster in steady state**, with `rollout_train_logprobs_abs_diff_mean` indistinguishable between the two modes. The memory run wrote nothing: the adapter file from the preceding disk run was still untouched at the end of it.

Per-sync stage split on the memory path: `exported+sent 0.36-0.39 s, registered on vLLM 0.09-0.10 s`.

Shared expert adapters are deduped structurally while the bridge export streams (grouped per module / expert span / `lora_A|lora_B`, verified with `torch.equal`), so ~928 unique tensors go on the wire instead of 18,040. The adapter is built on the GPU, which keeps vLLM's per-key `.to()` a no-op so aliases keep sharing storage; building on CPU would pin a private copy per key and re-inflate the adapter to its public size.

On a dense model the same path gives 0.073 s vs 0.103 s (Qwen3-0.6B, rank 8, 9.7 MB adapter) — the speedup scales with adapter size and key count, so it is not a fixed multiplier.

### GLM-5.3, two B300 nodes

GLM-5.3-BF16 (`glm_moe_dsa`, 78 layers, 256 routed experts, top-8), LoRA rank 32, 30.77 GB / ~38K-tensor public adapter (0.62 GB unique). Two nodes, 8x B300 each: Megatron 8 GPUs TP8/EP8 on one node -> vLLM 1 engine TP8 on the other, non-colocated, NCCL (Libfabric, GDR). Three optimizer-backed syncs per mode after the initial sync; disk and memory were separate runs on the same cluster and config.

| Sync | disk | memory | disk logprob gap | memory logprob gap |
|---|---:|---:|---:|---:|
| 0 (pre-training) | 513.70 s | 17.45 s | - | - |
| 1 | 408.81 s | 15.95 s | 0.0234 | 0.0240 |
| 2 | 375.79 s | 12.34 s | 0.0208 | 0.0234 |
| 3 | 384.60 s | 11.48 s | 0.0201 | 0.0191 |
| **steady state** | **389.73 s** | **13.26 s** | 0.0214 | 0.0222 |

**29.4x faster in steady state**, with the logprob gap again indistinguishable between modes. The disk baseline reproduces #2196's GLM-5.3 baseline on the same hardware (387.40 s) within 1%.

Per-sync stage split on the memory path: `exported+sent 9.46-14.08 s, registered on vLLM 0.42-0.48 s`. The export dominates: the bridge streams the public adapter (every expert per layer, 30.77 GB per rank) and the extractor dedupes it to 0.62 GB while streaming, so sync cost scales with public key count rather than unique bytes -- #2196's rank-local export (~1.2 s) avoids that walk, which is the gap between the two memory paths. On the disk path the 28.7 GiB `adapter_model.safetensors` takes ~1m46s to write; the remaining ~5 min per sync is vLLM reading and registering it.

Each resident vLLM adapter slot is ~22.4 GiB per inference GPU on this model (see `.claude/docs/weight_sync.md`). To fit on B300 the run used `max_loras=1`, `gpu_memory_utilization=0.80`, `enforce_eager=true`, `max_model_len=4096`, `max_num_batched_tokens=1024`; none are on the sync path. Rollout: gsm8k, 512-token prompts / 512 generated.

## Testing

- CPU: `tests/backends/skyrl_train/weight_sync/`, `tests/backends/skyrl_train/inference_servers/`, `tests/train/test_config.py` — 718 passed, 5 skipped.
- GPU: `tests/backends/skyrl_train/gpu/gpu_ci/inference_servers/test_weight_sync.py` — 4 passed, covering base-model sync over NCCL broadcast, NCCL with prefill/decode disaggregation, colocated CUDA IPC, and sharded RDT. This is the regression surface for the `receive_target` plumbing, which touches the shared push senders.
- End-to-end disk-vs-memory comparison on GLM-4.7-Flash and on Qwen3-0.6B, three steps each per mode.
- `ruff==0.11.9` and `black==24.10.0` clean.

`tests/train/test_config.py::test_megatron_validation_rejects_mxfp8_with_fp8_param` fails on this branch and identically at the base commit on non-Blackwell hosts — its expected message depends on GPU architecture. Not related to this change.

## Scope and limitations

Validated on one MoE architecture and one topology. The following are rejected by `validate_inference_engine_cfg`: FSDP, `merge_lora=true`, and the `delta` / `sharded_rdt` backends (neither has a chunk stream to carry the adapter).

Not rejected but **untested**: PP>1, multi-node trainers, colocated runs (CUDA IPC — note `_prepare_for_weight_sync` backloads the full policy model under `colocate_all` even for adapter-only sync), fp8 rollout dtype (which turns the no-op cast into a real one and would materialize every alias), and `max_loras>1` multi-tenant churn.

Dedupe degrades safely: a model whose exported keys do not match the `.experts.<N>.` naming, or whose provider does not expose `num_moe_experts`, simply sends every key. Group membership is verified tensor-by-tensor, so a wrong grouping costs bandwidth, never correctness.

## Repro

```
bash tests/train/gpu_e2e_test/lora_sync_mode_compare.sh
```

Runs the same short GRPO job once per `sync_mode` and prints per-sync timings, the memory path's stage split, and the logprob gap for each. `ci/anyscale_gpu_lora_sync_compare_h100.yaml` is the Anyscale job spec.


erictang000:
can close in favor of #2250

PR #2250, opened 2026-09-21
[weight_sync] In-memory LoRA adapter sync for Megatron (lora.sync_mode=memory)

avigyabb:
Brings #2221's in-memory LoRA adapter sync onto the weight-sync sender refactor (#2142, now on main). #2142 deleted the `WeightExtractor` / `WeightTransferStrategy` layer that #2221 hooked into, so this is a re-expression on the new abstractions rather than a rebase of the original branch.

## What it does

Adapter-only sync (Megatron `merge_lora=false`) currently round-trips through the filesystem: the trainer writes `adapter_model.safetensors` to `lora_sync_path`, and every vLLM worker reads it back and materializes the adapter on pinned CPU memory. `trainer.policy.model.lora.sync_mode=memory` instead ships the adapter tensors down the transport the base model already uses (NCCL broadcast when non-colocated, CUDA IPC when colocated) and builds the `LoRAModel` from the received GPU tensors. Nothing is written. `sync_mode=disk` stays the default and is unchanged.

## How it maps onto the refactor

Weight sync gains a second axis: the transport says *how* tensors move, the **target** says what they are and how the receiver applies them.

| piece | where |
|---|---|
| `MegatronLoraAdapterSource` — a `WeightSource` whose stream is a PEFT adapter, deduped as exported | `weight_sync/sources.py` |
| the wire target (name, adapter config, alias map) + dedupe helpers | `weight_sync/lora_target.py` (torch-only; both halves read it) |
| `SkyrlLoraStagingMixin` — stages the round's tensors instead of loading them | `weight_sync/weight_receivers.py` |
| `WorkerLoRAManager._load_adapter` serving `skyrl-memory://<name>` from the stage | `patches/vllm/patch_lora_in_memory.py` |

Three things worth a reviewer's attention:

**The target rides `/collective_rpc`, not the transfer.** `send_weights()` is vLLM's and its per-round payload is a fixed dataclass of names, dtypes and shapes. Adding a channel would mean subclassing the update info of every transport, so the adapter's name/config/aliases are armed out of band through the worker extension — which exists for exactly this class of dispatch problem. Arming is consumed by `finish_weight_update`, so a missing or failed arm can never apply an adapter stream to the base model, or the reverse.

**The adapter source must be prepared before the send.** The dedupe decides which keys exist, so one export has to serve `metadata()` and iteration both (vLLM's `_checked_iter` requires they agree), and the alias map it produces has to reach the workers first. `prepare()` runs the export on every rank; iteration drops the result, and reading the source afterwards raises rather than re-exporting — an implicit re-export would be a collective run by one rank alone, which hangs the job for the full NCCL watchdog timeout instead of failing.

**Deduplication is what makes many-expert MoEs affordable.** With `share_expert_adapters` one adapter serves every expert an EP rank owns and the PEFT export replicates it per expert. Only unique tensors go on the wire; the receiver re-attaches the public names to the same storage. Membership is verified with `torch.equal`, so a wrong group size costs bandwidth, never correctness.

`finalize_adapter` applies the same two rewrites as `_save_lora_adapters_and_sync`, in the same order (rank-scale fold, then the fused-expert 3D→flat rewrite), so the two paths hand vLLM the same adapter.

I also folded the two receive engines' identical lifecycle wrapping into one `SkyrlReceiveLifecycleMixin` rather than duplicating the new LoRA branch into both.

## Validation

`sync_mode=memory` is rejected for anything but Megatron adapter-only sync over the `nccl` backend: `delta` publishes checkpoint diffs and `sharded_rdt` bakes a pull plan into model params, so neither has a stream to carry an adapter. FSDP is not implemented.

## Testing

**CPU (new):** `test_lora_target.py` (dedupe, aliases, wire target), `test_lora_receive.py` (the armed/unarmed branch and the staging clone, against a stand-in engine — so the receive side has CPU coverage it did not have before), `test_sources.py::TestLoraAdapterWeightSource` (channel agreement under real vLLM 0.28, plus the unprepared-read guards), and 7 config-validation cases.

Full CPU suite green on both CI jobs. Three failures are **pre-existing on main**, reproduced on a pristine worktree: `test_config.py::test_megatron_validation_rejects_mxfp8_with_fp8_param` and the two `test_lora_export_names.py::test_full_weight_sync_uses_inference_namespace` cases.

**GPU, on 4xH100-80G:** `tests/train/gpu_e2e_test/lora_sync_mode_compare.sh` with GLM-4.7-Flash (31B, 64 routed experts), Megatron EP2 → vLLM TP2, non-colocated, LoRA rank 32. Mean `timing/sync_weights` over three syncs, run twice (once before the rebase, once on exactly this branch):

| | disk | memory |
|---|---|---|
| pre-rebase | 6.05 / 7.11 / 7.07 → **6.74 s** | 0.487 / 0.481 / 0.491 → **0.486 s** |
| this branch | 6.43 / 5.15 / 5.92 → **5.83 s** | 0.491 / 0.486 / 0.493 → **0.490 s** |
| logprob gap (this branch) | 0.0340 | 0.0322 |

**12–14x, ~5.5 s saved per sync.** The disk path is what varies between runs (a file write, then a read per worker); the memory path is stable to a millisecond. The rollout-vs-trainer logprob gap is unchanged between modes, which is what says the adapter actually landed. Rank-0 stage split for memory: export 0.28–0.29 s, send 0.10–0.11 s, register 0.09–0.10 s. Dedupe on this model: 928 unique tensors against 17112 public (19.4x) — per-expert keys alias 32-to-1 under EP2, attention and shared-expert keys never alias.

Two caveats worth knowing:

- The disk baseline is the *favourable* case: trainer and engines shared a node, so `lora_sync_path` was local NVMe with a warm page cache, not the shared mount a real non-colocated run needs.
- This job runs `normalize_moe_lora=false`, so every module sits at `config_rank` and **the rank-scale fold is a no-op in both paths here** — it is covered by its own unit tests, not by this run.

Running it is also what found the two defects in the third commit: a worker-side log line that could never appear (only loggers under vLLM's own namespace are configured in a worker process), and a post-send `metadata()` that re-ran a collective on one rank and hung the job for 30 minutes.

## Known follow-up

Every rank retains the deduped adapter on GPU through the send, though only rank 0 transmits. main recently stopped the disk path from keeping adapter state on non-writer ranks because the CPU spike could OOM a node; the in-memory path has a milder version of the same shape (deduped, on GPU) and could drop it on non-senders.

---

> [!NOTE]
> **Medium Risk**
> Touches distributed weight sync, vLLM LoRA internals via monkeypatch, and strict ordering (arm → send → register); mis-arming is mitigated by one-round targets, but GPU memory and multi-tenant LoRA budgeting need careful ops tuning.
> 
> **Overview**
> Adds **`trainer.policy.model.lora.sync_mode=memory`** for Megatron adapter-only training (`merge_lora=false`): LoRA weights go over the existing **NCCL / CUDA IPC** weight-sync path instead of writing PEFT files to `lora_sync_path`. **`sync_mode=disk`** stays the default.
> 
> The change introduces a **LoRA receive target** (`lora_target.py`) with shared-expert **dedupe and alias expansion**, **`MegatronLoraAdapterSource`** (prepare-once export so `metadata()` and iteration agree), and receive-side **staging** (`SkyrlLoraStagingMixin`) that skips base-model reload when armed. Adapter name, config, and aliases are armed out of band via **`/collective_rpc`** (`skyrl_set_lora_receive_target`) before `send_weights()`, then registered with **`load_lora_adapter(..., in_memory=True)`** using a vLLM runtime patch (`patch_lora_in_memory.py`, `skyrl-memory://` paths).
> 
> Config validation restricts memory mode to **Megatron + nccl backend** (not delta/sharded_rdt/FSDP). Unload best-effort calls **`skyrl_discard_in_memory_lora`**. CPU tests cover dedupe/receive staging; a **4xH100** compare script and Anyscale CI job exercise disk vs memory end-to-end.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec9e3fde9da0fc65aaf5060868f94eb768ea9247. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


skyrl/train/config/config.py
```diff
@@ -106,7 +106,15 @@ class SkyRLLoraConfig(BaseConfig):
     """Dropout probability applied to LoRA layers, to help prevent overfitting."""
     lora_sync_path: str = "/tmp/skyrl_lora_sync"
     """Directory where LoRA adapter weights are saved and synchronized between the training and inference processes.
-    Must be accessible to all workers in distributed setups."""
+    Must be accessible to all workers in distributed setups. Unused when ``sync_mode="memory"``."""
+    sync_mode: str = "disk"
```
avigyabb, line 110: set default to memory?

skyrl/train/utils/utils.py
```diff
@@ -608,6 +608,26 @@ def validate_inference_engine_cfg(cfg: SkyRLTrainConfig):
...
 
+    lora_cfg = cfg.trainer.policy.model.lora
+    if lora_cfg.sync_mode not in {"disk", "memory"}:
+        raise ValueError(f"trainer.policy.model.lora.sync_mode must be 'disk' or 'memory', got {lora_cfg.sync_mode!r}")
+    if lora_cfg.sync_mode == "memory":
+        # The adapter rides the base-model transport (NCCL broadcast / CUDA IPC)
+        # into the receive engine, which stages it for vLLM's LoRA manager. The
+        # other backends have no such stream to carry it: delta publishes
+        # checkpoint diffs and sharded_rdt bakes a pull plan into model params.
+        if cfg.trainer.strategy != "megatron":
```
avigyabb, line 619: fall back to disk?

skyrl/backends/skyrl_train/patches/vllm/lora_in_memory_upstream.patch
```diff
@@ -0,0 +1,277 @@
+
```
avigyabb, line 1: remove?

skyrl/backends/skyrl_train/weight_sync/lora_target.py
```diff
@@ -0,0 +1,159 @@
+"""LoRA adapter as a weight-sync *target*.
```
avigyabb, line 1: design - should we move this inline into sources.py? should we make soruces a dir and put the different classes as files within?

ci/anyscale_gpu_lora_sync_compare_h100.yaml
SumanthRH, line 1: Nit: This can be removed

tests/train/gpu_e2e_test/lora_sync_mode_compare.sh
SumanthRH, line 1: can be removed

skyrl/backends/skyrl_train/patches/vllm/patch_lora_in_memory.py
```diff
@@ -0,0 +1,207 @@
...
+
+``add_lora`` still flows API server -> engine core -> every worker, so the
+adapter name in the OpenAI-compatible model registry stays correct and
+``load_inplace=True`` replaces the previous generation exactly as the
+directory path does. Staged tensors stay registered after the load: the
+``LoRAModel`` shares their storage (same device and dtype), so this costs
+nothing, and it lets vLLM rebuild the adapter after an LRU eviction the way it
+would re-read a directory. The next sync for the same name replaces them;
+``discard_in_memory_adapter`` frees them on unload.
+
```
SumanthRH, line 38: I guess this might work for us right now with `share_expert_adapters` but the general case with vLLM is going to be different: We could have `share_expert_adapters=False` and in this case the GPU memory usage would be pretty bad: temporary memory usage during weight sync would be > 30GB per GPU because we sync the full weights and this could easily lead to OOMs. 

I would prefer to figure out how to load the LoRA tensors chunk by chunk after receiving on GPU memory. Let's do this for the vLLM upstream PR

SumanthRH (changes requested):
Looking good for now!

I think this works as an initial unblock. I would want to redesign a bit for the general case, can be a follow-up. 


Let's also add a GPU test for this? We can just modify the current lora weight sync to add a path for in-memory: 

`tests/backends/skyrl_train/gpu/gpu_ci/test_lora.py::test_policy_local_engines_e2e`

Should add one for colocated and non-colocated for megatron.

skyrl/backends/skyrl_train/patches/vllm/patch_lora_in_memory.py
avigyabb: Hmm, yea makes sense. Sg!

skyrl/backends/skyrl_train/patches/vllm/patch_lora_in_memory.py
```diff
@@ -0,0 +1,207 @@
...
+    lora = manager._lora_model_cls.from_lora_tensors(
+        lora_model_id=lora_request.lora_int_id,
+        tensors=tensors,
+        peft_helper=peft_helper,
+        device=manager.device,
+        dtype=manager.lora_config.lora_dtype,
+        model_vocab_size=manager.vocab_size,
+        weights_mapper=weights_mapper,
+        skip_prefixes=skip_prefixes,
+    )
```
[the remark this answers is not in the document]
SumanthRH: @avigyabb I haven't had a chance to dig into the vLLM lora code yet, can you look into this?

SumanthRH:
@avigyabb one thing that would be useful is to show parity for `memory` and `disk` for LoRA training for a small model like Qwen 2.5 1.5B Instruct on GSM8K. We should expect parity on pretty much every metric.

SumanthRH (approved):

SumanthRH:
@avigyabb can you fix the failing CPU CI tests?

PR #2254, opened 2026-09-22
[ci] Rename JAX workflows to SkyRL-JAX-{CPU,GPU}, give SkyRL-{CPU,GPU} to SkyRL-Train, and scope JAX CI to the tx code path

erictang000:
## Summary

Two things:

**1. Swap the CPU/GPU check namespaces.** `SkyRL-CPU` / `SkyRL-GPU` were the JAX/tx suites, while the SkyRL-Train suites — the ones almost every PR actually cares about — were namespaced `SkyRL-Train-CPU` / `SkyRL-Train-GPU`. Now the default-looking check name belongs to the default code path:

| Before | After | File |
|---|---|---|
| `SkyRL-Train-CPU` | **`SkyRL-CPU`** | `cpu_skyrl_train.yaml` |
| `SkyRL-Train-GPU` | **`SkyRL-GPU`** | `gpu_skyrl_train.yaml` |
| `SkyRL-CPU` | **`SkyRL-JAX-CPU`** | `cpu_skyrl.yaml` → `cpu_jax.yaml` |
| `SkyRL-GPU` | **`SkyRL-JAX-GPU`** | `gpu_skyrl.yaml` → `gpu_jax.yaml` |

**2. Scope the JAX workflows to the tx code path** so they run as rarely as possible. They previously triggered on `skyrl/**` + `tests/**` + `ci/**` + `pyproject.toml` — i.e. on essentially every PR, including `SkyRL-JAX-GPU`, which submits a *paid Anyscale job* on each one. Both now trigger only on:

```
skyrl/tx/**
skyrl/backends/{backend,jax,ray_jax}.py
skyrl/utils/**            # tx imports skyrl.utils.{log,storage} directly
skyrl/tinker/types.py     # tx imports skyrl.tinker.types directly
tests/tx/**
tests/backends/test_jax_backend.py
<the workflow file itself>
```

plus `skyrl/benchmarks/**`, `pyproject.toml` and `uv.lock` on the **CPU** one only — a resolution change is the main way a JAX break lands without anyone touching tx, and that runner is free. The GPU one deliberately leaves those out (costs credits) and keeps `workflow_dispatch` for on-demand runs.

## Keeping coverage: the non-JAX suites move to `SkyRL-CPU`

`SkyRL-CPU`-as-was ran `tests/` minus a set of ignores, which is more than just tx. Two of those suites are not JAX-owned and would have silently stopped running under the narrowed filter, so they move to a new `skyrl_tinker_tests` job on `SkyRL-CPU`:

- `tests/tinker` (top level) — the Tinker API server tests. They default to the JAX backend, but the server itself is actively maintained, so they should keep running on every PR.
- `tests/utils`

It's a separate job rather than a step on `skyrl_train_tests` because `jax` and `fsdp` are conflicting backend extras and need separate envs, and because `tests/tinker/test_api.py` re-execs `uv run` inheriting the parent's flags.

**The split is exact** — verified by collection counts, not by eye:

| | tests collected |
|---|---|
| old `SkyRL-CPU` selection | 221 |
| new `SkyRL-JAX-CPU` selection | 104 |
| new `skyrl_tinker_tests` selection | 117 |

104 + 117 = 221, so nothing is dropped and nothing runs twice.

## Also

- Dropped the JAX workflow's standalone `uvx ruff@0.11.9 check` step. `check_code_quality` on `SkyRL-CPU` already runs the same pinned ruff over all files via pre-commit, and it runs on the PRs where the JAX workflow now doesn't.
- Updated `.claude/docs/ci.md` with a namespace table and the scoping gotcha.

## Notes for review

- **Branch protection**: if any of these are required status checks, the required-check list needs updating by hand — `SkyRL-Train-CPU` / `SkyRL-Train-GPU` no longer exist, and `SkyRL-CPU` / `SkyRL-GPU` now point at different workflows.
- I left `SkyRL-Train-GPU-Megatron` (`gpu_skyrl_train_megatron.yaml`) alone since you only named the two. Say the word if you want it to become `SkyRL-GPU-Megatron` for consistency.
- The `push:` and `pull_request:` path lists are duplicated in each JAX workflow because GitHub Actions doesn't support YAML anchors. Noted in a comment at the top of each file.
- Both JAX workflows should run on this PR (their own files changed), which is the validation for the new filters.


erictang000:
Reopening from a branch on the upstream repo -- GitHub Actions doesn't fire for fork PRs here.

PR #2261, opened 2026-09-22
[Examples] Harbor through skyrl-capture: durable token-in/token-out rollouts

kouroshHakha:
**Draft, for early feedback.** Sharing the shape for early feedback and planning --  not proposing this as mergeable.

## What this adds

Two things, which is why it is large:

1. **`skyrl-capture/`** — the capture package itself, ~65k lines including its
   own tests, viewer and docs, added as a sibling package alongside
   `skyrl-gym`. Three commits: a straight copy, then the wiring (license to
   Apache-2.0, a `harbor-capture` extra, a path source), then CI that fires
   only on `skyrl-capture/**`.
2. **`examples/train_integrations/harbor_capture/`** — the integration, ~2.5k
   lines including tests and the validation scripts.

A second Harbor integration, `examples/train_integrations/harbor_capture/`, sitting beside the existing `harbor/` one. Harbor runs **completely unmodified, in text space**. A per-trajectory HTTP route renders its prompt, calls `/skyrl/v1/generate` with token IDs, and keeps a message graph — so training gets the exact tokens the model was conditioned on, and Harbor never learns what a token is.

The guarantee is narrow and stated without reference to anything else:

> Every token the trainer trains on is a token the engine sampled from.

That is worth a proxy at all because the harness talks messages, so prompts must be re-derived each turn. If re-deriving perturbs anything the model already saw, the training sequence contains tokens that were never in any prompt the engine sampled from — and nothing downstream can detect it.

## How it differs from `harbor/` today

Today's integration sets `collect_rollout_details=True` and records the `prompt_token_ids` vLLM reports per call. Each row is faithful. What it cannot do is summarization:

```python
assert len(traj.rollout_details) == 1, "Expected exactly one rollout segment"
# TODO(Charlie): Support summarization.
```

Compaction breaks its token accounting, so it is banned outright. A proxy-side graph doesn't have the problem: a rewritten history stops matching at the last unchanged message and branches there, and each branch becomes its own training row.

## How it differs from #2143

#2143 does the same thing in-tree — a generator-owned in-process proxy, a `Trace` message graph, a composer. Same core insight, and the two agree on more than they disagree.

The difference is where the state lives. #2143 keeps the trace in memory in the generator. Here it's an **out-of-process service with durable per-trajectory persistence**: an append-only journal per in-flight trajectory, a compiled record per finished one, CRC-framed and crash-safe. That buys things an in-memory trace can't offer — a run survives the trainer restarting, rollouts are inspectable while they happen, records are exportable afterwards, and there's a viewer over them. It costs an extra process and a serialization boundary.

Neither is obviously right. #2143 is simpler and has fewer moving parts. This is better if you want rollouts to be durable artifacts you can go back to. Worth deciding deliberately rather than by whichever lands first.

**One thing #2143 has that this doesn't: end-to-end training runs with summarization on, compared by reward curve.** That's the more convincing evidence for the branching claim, and I don't have it yet.

## Validation

Everything below ran on the real path — SkyRL's own `vllm_server_actor` serving
`/skyrl/v1/generate`, capture in front of it, Harbor unmodified in Modal
sandboxes, `compose` producing the batch. Same rollout path as training, minus
the gradient step.

### Setup

```bash
# 1. the engine SkyRL itself runs
VLLM_BATCH_INVARIANT=1 python -m skyrl.backends.skyrl_train.inference_servers.vllm_server_actor \
    --model Qwen/Qwen3-4B-Instruct-2507 --tensor-parallel-size 1 \
    --host 127.0.0.1 --port 9500 --max-model-len 65536 \
    --gpu-memory-utilization 0.85 --enable-auto-tool-choice --tool-call-parser hermes

# 2. capture in front of it, on the skyrl upstream
PYTHONPATH=$PWD uv run --extra harbor-capture skyrl-capture serve \
    --host 0.0.0.0 --port 8080 --mode tokens \
    --upstream-type skyrl --upstream-url http://127.0.0.1:9500 \
    --model Qwen/Qwen3-4B-Instruct-2507 --tokenizer Qwen/Qwen3-4B-Instruct-2507 \
    --max-model-len 65536 --record-dir /tmp/icap-validation \
    --upstream-module examples.train_integrations.harbor_capture.upstream

# 3. a rollout phase: 16 prompts x 16 repetitions, code_contests via Terminus-2
python validation/rollout_phase.py --task-dir /path/to/harbor/tasks \
    --tasks 16 --group-size 16 --turns 8 --timeout 180 --concurrency 16
```

`VLLM_BATCH_INVARIANT=1` makes sampled tokens independent of how requests batch
together, so a difference between runs is a real difference and not a scheduler
artefact. It uses slower kernels, which is why no latency number is quoted here.

### What actually got captured

314 trajectories, 1,712 graph nodes, 856 exchanges. Turn counts per trajectory:

| turns | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 |
|---|---|---|---|---|---|---|---|---|
| trajectories | 26 | 210 | 17 | 15 | 7 | 12 | 8 | 17 |

Here is one 8-turn rollout in full — 16 nodes, an agent writing and running
Python in a sandbox:

```
turn   prompt   completion   reused_prefix   stop
   0     1291          508               0   stop
   1     2099          593            1799   stop
   2     2914          495            2692   stop
   3     3941          537            3409   stop
   4     4720          578            4478   stop
   5     5540          549            5298   stop
   6     6331          573            6089   stop
   7     7268          585            6904   stop
```

**`reused_prefix` is exactly the previous `prompt + completion`, every turn:**
1291+508 = 1799, 2099+593 = 2692, 2914+495 = 3409, and so on to 6331+573 =
6904. That column is the whole mechanism. The harness sent back text; capture
did not re-render it, it carried the model's own tokens forward and rendered
only what was new. A 7,268-token prompt at turn 7 contains 6,904 tokens that
are byte-for-byte the tokens the engine produced and consumed earlier.

That trajectory exports as one training row:

```
input_ids 7853   trainable 4418   rollout_logprobs 7853

node_spans [[0,1288] [1288,1799] [1799,2096] [2096,2692] [2692,2911] [2911,3409]
            [3409,3938] [3938,4478] [4478,4717] [4717,5298] [5298,5537] [5537,6089]
            [6089,6328] [6328,6904] [6904,7265] [7265,7853]]
```

16 spans tiling `[0, 7853]` with no gap and no overlap, alternating
client-authored and model-sampled. `trainable` is 4,418 — the exact sum of the
eight completions (508+593+495+537+578+549+573+585), so the loss mask selects
every sampled token and nothing else.

### Check 1 — token fidelity

```bash
python validation/token_fidelity.py /tmp/icap-validation
```

For each model-authored node it takes the span the export assigned it, and
compares against the `prompt_token_ids` capture sent to the engine and the
`completion_ids` the engine returned for that node. Matched on node id, not on
completion text — two turns of one rollout can sample identical text.

```
312 trajectories, 312 paths, 856 sampled turns
  trained tokens == tokens the engine sampled from : 856/856
  sampled turns with no matching exchange          : 0
  turns that differed                              : 0
```

A mismatch here is training on a prompt the model was never conditioned on.

### Check 2 — the batch is well formed

```bash
python validation/validate_compose.py /tmp/icap-validation <run-id>
```

Composes the record the way the generator does and checks the result rather
than trusting it: the eight arrays a trainer indexes in parallel agree row by
row; each rollout's rows are contiguous under one `TrajectoryID` and marked
`is_last_step` exactly once, so one advantage can reach all of them; every row
of a rollout carries the same reward; every row either trains on something or
is an explicit masked placeholder.

```
255 rollouts -> 255 rows
  rollouts represented        : 255/255
  rows with trainable tokens  : 227
  fully masked placeholders   : 28
  trainable tokens            : 411,306
  checks failed               : 0
```

The 28 placeholders are the 26 agent timeouts plus 2 errors — present as masked
rows rather than dropped, which is what keeps rewards and rollouts aligned.

### The 256-rollout run

```
256 rollouts -> 256 rows in 1842s
  rows with gradient : 226/256
  trainable tokens   : 410,664
  stop reasons       : complete 226, agent_timeout 26, error 4
  reward mean/max    : 0.207 / 1.000
  groups with spread : 11/16
```

`groups with spread: 11/16` is the number that says this task set would train:
a GRPO group whose 16 rewards are all equal gives no signal, and 11 of 16 have
spread.

A later 32-rollout run after the fixes below: 32/32 gradient-bearing, all
`complete`, and **32 engine sessions released for 32 rollouts** — the run before
the fix released 278 for 256.

## Why the numbers are worth trusting

Because running it at scale found real bugs, each of which is fixed and each of
which the checks above would not have caught on their own:

- **A finish error was masking the `TimeoutError` behind it.** `finish` ran in
  a `finally` and was allowed to raise, so its 503 replaced the timeout already
  in flight — and the retry policy matches on `TimeoutError` to mask a timeout
  *without* retrying, precisely because "retrying buys nothing and costs a
  sandbox". Every timeout that raced an in-flight turn bought that sandbox.
  Visible as engine sessions released: **278 for 256 rollouts** before, **32 for
  32** after.

- **26 trajectories were stranded mid-finish.** An agent that times out
  abandons a generation the engine is still running; `finish` correctly refuses
  to compile a record missing that turn and returns a retryable 503 — but the
  retry goes to a *fresh* trajectory, because the graph is append-only, so
  nobody is left who knows the old id. The listing surfaced **one** of them; the
  other 25 were journals on disk that nothing showed. Capture now sweeps them on
  an interval, and re-runs the recorded request rather than a new one (a finish
  is identified by a hash over its labels, annotations and command result, so
  asking with fresh arguments reads as a *contradicting* outcome and is
  rejected — which is how the first version of the sweep failed all 26).

- **Two prompt-rendering divergences** between capture's renderer and the
  model's own chat template: the OpenAI `{"type": "function", ...}` tool
  wrapper, and key order from capture's canonicalisation. Neither breaks
  fidelity — capture trains on exactly what it sent — but both put the model
  slightly off its post-training distribution. Currently covered by fixtures
  only; see the limits below.

## What this does not cover

- **No tool-schema traffic.** Terminus-2 drives the shell through its own text protocol, so `tools` is empty on every turn and both rendering divergences above are covered only by fixtures.
- **No branching run.** `--summarize` hasn't been run end to end, so the capability that most distinguishes this from `harbor/` has no live evidence here.
- **No performance numbers.** Correctness runs used `VLLM_BATCH_INVARIANT=1`, whose deterministic kernels make latency unquotable.
- **Branching is what most distinguishes this from `harbor/`, and it has no live evidence here.** Every one of the 314 captured trajectories exported exactly one path, because `--summarize` was never on. The graph code that makes branching work is exercised by unit tests and by the `harbor/`-banned case in principle, not by a run.

## Tests and CI

`tests/integrations/harbor_capture/` — 32 tests, on the CI path already
(`cpu_skyrl.yaml` collects `tests/` wholesale; I added
`examples/train_integrations/harbor_capture/**` to its `paths:` so a change to
the example triggers the tests that import it). They need neither capture nor a
cluster: the conftest stubs capture's protocol base class, since capture is an
optional dependency of one example.

`cpu_skyrl_capture.yaml` is new and fires **only** on `skyrl-capture/**` — a
change to the Megatron backend should not rebuild a container. Five jobs, of
which one is worth pointing at: **`base-install`** installs the package without
extras and asserts that importing the runtime pulls in no `transformers`,
`renderers` or `torch`. That is what lets capture sit beside a trainer pinning
its own transformers, and every other job runs `--all-extras` and would never
notice a renderer import escaping to module scope.


kouroshHakha:
# skyrl-capture v1 — design discussion (2026-09-22)

Goal: land a **minimal** `skyrl-capture` in a new repo (`NovaskyAI/skyrl-capture`),
in small PRs. This doc shows the end state first, then the open decisions,
then the build order. Each decision comes with a proposal. Reply inline with `→ ok`,
`→ no, because…`, or `→ cut`.

---

## 0. Scope at a glance

Status vs `inference-capture` today: **✅ exists** (port it) · **🟡 exists, changes** (port, then simplify) · **🆕 new** (write it).

### Keep (v1)

| Item | Why | Today |
| --- | --- | --- |
| Many concurrent trajectories per server | A training step runs hundreds of rollouts. One process per rollout would mean hundreds of tokenizers loaded in memory | ✅ One process holds every live trajectory in memory, keyed by the id in the route URL. TITO keeps each trajectory's exact token trace in memory next to it |
| Graph bookkeeping (forks, roots, subagents, summarization, thinking-strip) | This is the product. Without it, each rollout is just a flat list of calls, and multi-path samples can't be exported | ✅ Text: after replying, hash each message together with tools and model, walk the longest matching prefix, and append the rest. TITO: nodes come from the renderer's per-message spans, and the next turn extends the previous turn's exact tokens. Both produce the same one-node-per-message forest |
| Text **and** TITO upstreams (one per process) | TITO is for training (exact tokens). Text is for observing any OpenAI endpoint (eval, API models). One mode per process keeps each code path simple | ✅ The mode is fixed by the upstream config at startup. Text forwards bytes untouched and parses them afterwards. TITO renders, calls a token-in/token-out engine (vLLM wire, with a pluggable subclass for SkyRL's path), and rebuilds a chat response |
| [`renderers`](https://github.com/PrimeIntellect-ai/renderers) lib as the only tokenizer path | Its per-message token spans are what the TITO graph is built from. And model quirks (reasoning placement, tool-call framing) are handled in explicit per-model `renderers` code, not by patching Jinja chat templates | 🟡 TITO already renders through `renderers`, with a pool and bridge-extension. It also ships a built-in ChatML/byte renderer for tests and benchmarks. That second renderer goes |
| OpenAI SDK retry handling | Without it, every SDK retry adds a fork (despite having a `x-stainless-retry-count`). That's a false training sample, and SDKs retry by default | 🟡 Retries are inferred after the fact from the graph: if a repeat produces the same reply, it adds no nodes and is flagged as a abandoned. **New:** replaying the cached response for a retry, and making a duplicate wait for the call still in flight. Today a TITO retry samples again, gets a different reply, and forks |
| Disk record **for viewing**, written once on finish | Debugging rollouts needs the trajectories after the job ends. It doesn't need them mid-run or after a crash, so one write per trajectory is enough. Files are split by weight so a viewer never decompresses what it doesn't show: `{id}.json.zst` (meta, annotations, failures, nodes with messages, per-call info with sampling params, and array offsets), `{id}.tokens.zst` (token ids + logprobs), `{id}.experts.zst` (R3, MoE only), `{id}.sampling_mask.zst` (per-token sampling support, when top-p/top-k is on). Open trajectories are flushed with `status: open` on graceful shutdown | 🟡 Today a background coordinator appends each exchange to a CRC-framed per-trajectory journal, fsynced on every write, and `finish` compiles it into one zstd record atomically. Token arrays and routed experts sit inline in that record. v1 has no journal and splits the heavy arrays into sidecars |
| Round-robin pool of N servers | One process has one GIL, so rendering will cap out at scale. Round-robin gives scale-out without a router | 🆕 Today it's one pod. Scale-out is only described in the docs: replicas behind a load balancer that has to hash on the trajectory id, with nothing enforcing it in code |
| Routed experts (R3) + sampling mask per token | Without them the trainer's logprobs don't match what inference sampled from: MoE routing differs between kernels, and top-p/top-k truncates the distribution. See *Training-fidelity arrays* | 🟡 R3 is captured today (it must cover prompt + completion, or the turn is rejected) and stored inline. 🆕 Sampling masks aren't requested or stored today |
| Token-sample export per root→leaf path | This is what SkyRL consumes. It's the output that justifies building the rest | ✅ Walks every root→leaf path and concatenates node tokens into `input_ids` / `loss_mask` / logprobs. Each sampled node trains in exactly one row, and abandoned branches can be masked. Today `finish` already returns one trajectory's rows directly, and bulk exports run as jobs. v1 keeps only the per-trajectory path |
| Minimal control API + free `meta` dict | Only create and finish are needed. `finish` seals the trajectory and returns its samples and status in one call | 🟡 Today's control plane creates (with an idempotent client-chosen id), finishes, annotates and labels trajectories, and reads them back under a project/run/trajectory hierarchy. v1 keeps only create / finish / samples |
| Viewer (late milestone) | It's how people trust the graph. It reads only the finished files, so it can land last without blocking anything. Listings and the tree/timeline views read only `{id}.json.zst`. The token view loads `{id}.tokens.zst` on demand. Nothing in the viewer reads R3. Running trajectories are visible through the server's `GET /trajectories/{id}`, not on disk | 🟡 A web UI served by the capture process over a read API that merges live journals and finished records (timeline, tree, paths). It needs porting to finished-only files and lazy sidecar loads |
| Leaked-trajectory TTL | With writes only on finish, a trajectory nobody finishes is never written and holds memory until the process exits. An idle TTL (no request and no finish for e.g. 1 h) writes it as `status: abandoned` and drops it from memory | 🆕 Today there's deliberately no TTL: an unfinished journal stays on disk and in memory forever, and a background sweeper finishes the ones whose `finish` was refused mid-turn |

### Drop / defer

Everything here exists today, except where noted. Dropping it means not porting it.

| Item | Why | Revisit when | Today |
| --- | --- | --- | --- |
| WAL, CRC journal, recovery on restart | This is the biggest chunk of complexity. A failed rollout is cheap: SkyRL masks it and moves on | Server deaths show up in real runs | ✅ A request for a trajectory this process doesn't hold replays its journal from disk, after truncating any torn tail |
| `delivery_uncertain` / fail-closed close-on-durable | It only exists to make recovery exact. Without recovery there's nothing left for it to resolve | Along with the WAL | ✅ A TITO response doesn't close until its exchange is fsynced. After a crash, an exchange that's durable but might not have been delivered is flagged, excluded from training, and resolved only by a later request that contains it |
| Load-balancer hash routing, replica rebuild | The URL already names the server, so stickiness comes for free | Never, unless we need a single public endpoint | LB hashing exists only in the docs. ✅ Rebuild: a replica that lands a turn for an unfamiliar trajectory rebuilds its token trace from the graph |
| Builtin ChatML/byte renderer | It's a second renderer to keep consistent, and nobody trains on it. Tests can use a tiny HF tokenizer or a stub | — | ✅ An offline, deterministic ChatML renderer over a byte tokenizer, used by the test suite and the benchmarks |
| Anthropic + Responses dialects | Harbor and SkyRL speak OpenAI chat. Each extra dialect adds parsing, SSE handling and materialization rules | A harness that needs one (e.g. Claude Code, codex) | ✅ Pluggable text-protocol adapters. Non-message context (Anthropic `system`, Responses `instructions` and items) is materialized into nodes |
| Commit coordinator, backpressure/gap accounting | With best-effort writes there's no ordered durable queue to manage | Along with the WAL | ✅ A bounded queue, ordered per trajectory. When it's full the work is refused and recorded as a gap, which surfaces in health checks and exports |
| project/run/step as indexed columns | A free `meta` dict covers the grouping. SkyRL already knows its own step | The viewer needs to query across runs | ✅ First-class fields on every trajectory, indexed for listing and grouping |
| Export job runner + multiple formats | One format (token samples) and one consumer (SkyRL). The samples `finish` returns are enough | A second consumer, or offline dataset builds | ✅ File-backed async jobs producing four formats (`token-samples`, `text-samples`, `graph`, `replay`) |
| Custom HTTP/1.1 transport | Its per-request speedup is small next to engine latency, and it's a lot of code to own. A standard async HTTP stack is enough | A profile shows the transport matters | ✅ A hand-written HTTP/1.1 client with keep-alive, chunked bodies and TLS, built to keep client-library overhead off the request path |
| Control-plane idempotent create hashing | The server mints the id, so a lost create just leaks one empty trajectory | Creates get retried across a flaky network | ✅ The SDK picks the id up front, and the server persists a hash of the create body. The same id with the same body returns the same trajectory; with a different body it's a conflict |
| `finish` as a barrier + background finish sweeper | `finish` becomes a seal: it closes the trajectory right away and cancels in-flight handlers, so a late engine result is dropped rather than committed. The harness had already given up on that turn. That removes the `503`, the recorded finish request and the sweeper | — | ✅ `finish` refuses with a retryable `503` while a turn is in flight. A sweeper every 60 s finishes the ones whose caller never retried (26 of 256 in a real run) |
| Narrower per-trajectory lock (engine call outside the lock) | Needs a pending-turn object across the unlocked engine call, commit-time dedupe of nodes another turn may have created, and interleaving tests, all in the TITO path. The benefit is unproven | A profile shows parallel subagents in one trajectory queuing on the lock | Not built. Today TITO holds one lock per trajectory for the whole turn, engine call and response included, and v1 keeps that |

**Net:** the new code is the pool, the retry replay cache, the leaked-trajectory TTL and sampling-mask capture. Everything else in v1 is porting and cutting.

---

## 1. End state

### 1.1 Components

```mermaid
flowchart LR
  subgraph Driver["SkyRL driver (entrypoint)"]
    Gen["Generator<br/>(e.g. HarborCaptureGenerator)"]
    Pool["CapturePool<br/>round-robin over URLs"]
    Gen --> Pool
  end

  subgraph S1["capture server k (1 process, 1 upstream)"]
    App["HTTP app<br/>control + /t/{id}/v1/*"]
    Store["TrajectoryStore<br/>id → Trajectory"]
    Up["Upstream<br/>Text | Tito"]
    Rec["RecordWriter<br/>on finish / TTL / shutdown"]
    App --> Store
    App --> Up
    Store --> Rec
  end

  Harness["Harness / agent<br/>(unchanged OpenAI client)"]
  Eng["vLLM / SkyRL engines<br/>text: /v1/chat/completions<br/>tito: /generate"]
  Disk[("record_dir/<br/>{id}.json.zst<br/>{id}.tokens.zst<br/>{id}.experts.zst<br/>{id}.sampling_mask.zst")]
  Viewer["Viewer (reads disk)"]

  Pool -- "POST /trajectories" --> App
  Pool -- "base_url = serverK/t/{id}/v1" --> Gen
  Gen -. "base_url" .-> Harness
  Harness -- "chat.completions" --> App
  Up --> Eng
  Rec --> Disk --> Viewer
  Gen -- "finish → status + samples" --> App
```

**Why this is minimal:** the URL names the server, so stickiness comes for free.
There's no load balancer, no hashing, no shared state between servers, and no recovery.

### 1.2 Data model (the part to get right first)

```mermaid
classDiagram
  class Trajectory {
    id
    status: open|finished|failed
    graph: MessageGraph
    meta: dict
    annotations: dict
    failures: list~Failure~
    replay: RetryCache
  }
  class MessageGraph {
    nodes: dict~id,Node~
    children: dict~key,id~
    match(msgs) Match
    extend(leaf, msgs) list~Node~
    paths() list~Path~
  }
  class Node {
    id
    parent
    role
    author: client|model
    message
    delta_hash
    created_at
    tokens: NodeTokens
    calls: list~CallInfo~
  }
  class NodeTokens {
    token_ids
    sampled_start
    logprobs
    routed_experts
    sampling_mask
  }
  class CallInfo {
    t_start, t_end
    usage
    finish_reason
    model
    sampling
  }
  class Failure {
    t
    status
    error
    input_leaf
  }
  class Sample {
    trajectory_id
    node_path
    input_ids
    loss_mask
    logprobs
  }
  Trajectory --> MessageGraph
  Trajectory --> Failure
  MessageGraph --> Node
  Node --> NodeTokens
  Node --> CallInfo
  MessageGraph ..> Sample : paths()
```

- One graph per trajectory. Text and TITO use the same `Node`; only TITO sets `tokens`.
- `delta_hash` = hash(message, tools, model). In TITO it's hash(token delta).
- Subagents, summaries, thinking-strip, and resamples all just come out as forks or new roots. None of them needs its own code.
- **No exchange entity.** The graph is a set of token prefixes, so identical outputs under one parent are one node. A successful call appends a `CallInfo` to the model node it produced (usually one call per node). The call's input is that node's `parent`. A call that produced no node (engine error, context overflow, TITO attribution failure) goes in `Trajectory.failures`.
- Identical samples therefore give one training row. `len(calls)` is kept for anyone who wants to weight by it.
- The viewer timeline is the model nodes' calls sorted by `t_start`. The wait between calls comes from the parent chain, as today. SDK retries are replayed by the retry cache and never touch the graph, so they only show up as a metric.

### 1.3 One request

```mermaid
sequenceDiagram
  participant H as Harness
  participant S as Server
  participant T as Trajectory
  participant U as Upstream
  H->>S: POST /t/{id}/v1/chat/completions
  S->>T: retry? → replay cached response
  S->>T: lock: match(msgs) → leaf
  alt TITO
    S->>U: render/bridge from leaf → prompt_ids
    U-->>S: completion ids + logprobs
    S->>T: extend(graph) then respond
  else Text
    S->>U: forward bytes
    U-->>S: response
    S-->>H: respond first
    S->>T: extend(graph) after
  end
  Note over T: kept in memory, written to disk on finish
```

---

## 2. Decisions to make


| Question | Proposal | Alt |
| --- | --- | --- |
| Where does round-robin live? | `CapturePool` Python lib in the driver: `pool.create(meta) → Trajectory(id, base_url)` | A router process (adds a hop, and brings routing back) |
| Who launches the servers? | The repo ships `skyrl-capture serve` only. SkyRL wraps N of them in Ray actors on the SkyRL side. The repo stays Ray-free | The repo ships a Ray launcher |
| A server dies | Its trajectories fail loudly: the generator sees an error and masks the rollout. No recovery | Journal + rebuild (the WAL we're cutting) |
| One node type for both modes? | Yes. `tokens` is optional. Text mode builds by message-hash prefix match; TITO builds from renderer message spans | Two graph types |
| What does the per-trajectory lock cover? | **Decided:** keep today's whole-turn lock in TITO (engine call included). Text commits only under the lock. The narrower lock is deferred (see Drop / defer) | Narrow lock with engine call outside, deferred |
| TITO token mismatch / unattributable turn | Mark the trajectory `failed` and return the completion anyway. `finish` still returns the samples with `status: failed`, and the generator decides whether to train on them | Current `poisoned` + 410 on later turns |
| Retries | An `x-stainless-retry-count>0` or `Idempotency-Key` hit replays the cached response, and an in-flight duplicate waits on the original. Every other repeat is a real fork, and "abandoned" = a leaf nobody continued from | Byte-identity dedupe (can't tell a retry from a resample) |
| Durability | **Decided:** nothing is written until the trajectory ends (finish, TTL or graceful shutdown). Then the document and each heavy array go to their own zstd file, written to a temp name and renamed. A hard crash loses the in-flight trajectories | Snapshot on a debounce while open. Append per call |
| Client dialects in v1 | OpenAI chat completions only, with SSE synthesized for TITO and relayed for text. `/v1/models` passes through | + Anthropic / Responses |
| Renderer | The `renderers` library (what today's capture uses), plus a renderer pool. vLLM `/render` API later behind the same interface | vLLM `/v1/chat/completions/render` first (what `skyrl/backends/renderer.py` and William's PR use). Unverified whether it returns per-message token spans, which the TITO graph needs |
| HTTP stack | aiohttp server + aiohttp client with a pooled keepalive. Drop the custom transport | Starlette + custom client (today) |
| Control API | `POST /trajectories {meta}` → `{id, base_url}`. `POST /trajectories/{id}/finish {annotations}` (e.g. `{"reward": 1.0}`) seals the trajectory, releases its engine session, writes it to disk, and returns `{status, samples}`. `GET /trajectories/{id}` for live inspection. `meta` is a free dict, with no project/run/step schema | Keep the hierarchy as columns |
| Export | **Decided:** `finish` returns `{status, samples}`. That's one row per trainable root→leaf path (`input_ids`, `loss_mask`, `logprobs`, `routed_experts`, `sampling_mask`), with the train-once rule. There's no separate samples endpoint. Grouping/`is_last_step` stays in SkyRL | Separate `GET …/samples`, export formats + job runner |
| Viewer | Last milestone. It reads only `record_dir` (document always, tokens on demand, never R3), and the record format is frozen at M3 | Ship it alongside M2 |
| Trajectory id | Server mints it and the pool never retries a create. The generator puts the SkyRL `TrajectoryID` in `meta` | Client-supplied ids + idempotent create |
| Session affinity upstream | **Decided:** send `X-Session-ID = trajectory id` (TITO) so the prefix cache stays warm, and **capture releases the engine session on finish** (and on TTL), so generators don't have to | Generator releases it (today) |

### HarborGenerator on this surface (decided)

```python
async with self.pool.trajectory(meta) as t:          # round-robin create; seals with an error on exception
    config = _with_capture_route(template, base_url=t.base_url, cache_salt=cache_salt)
    results = await self._run_harbor(config, prompt)
    reward = float(results.get("reward", 0.0))
    result = await t.finish({"reward": reward, "stop_reason": results.get("stop_reason", "complete")})
rows = result.samples if result.status == "finished" else []   # the generator's call
```

- **Samples come from `finish()`.** It returns samples + status, and the generator decides what to train on.
- **Capture releases the engine session on finish.**
- **No `t.complete(messages)` sugar.** A caller without a harness points an ordinary OpenAI client at `t.base_url`.
- **What this removes from today's generator:** hand-built session ids, annotating before finish, swallowing a refused finish and relying on the sweep, `_trainable_rows` integrity checks, and `_release_session`. `generate()`, the retry loop and `compose()` are unchanged.

### Training-fidelity arrays (v1 requirement)

Both of these are **required in v1**. Without them the trainer's logprobs aren't computed over the same thing inference sampled from.

| | Routed experts (R3) | Sampling mask |
| --- | --- | --- |
| **Why** | MoE routing differs between the inference and training kernels. Replaying the rollout's expert choices in the trainer's forward pass removes that mismatch | With top-p / top-k / min-p, inference samples from a *truncated* distribution. The trainer computes a full-vocab softmax, so the importance ratio compares two different distributions and is biased. The fix is to set logits outside the rollout's support to −inf and renormalize, over the same support ("keep sampling mask", as in DeepSeek-V3.2 §3.3) |
| **What the engine returns** (vLLM token-in/token-out `generate`) | `routed_experts`: base64 `.npy`, `[num_tokens − 1, layers, k]` covering prompt **and** completion (the last token isn't forwarded yet). Needs `enable_return_routed_experts` | `sampling_mask`: `list[list[int]]`, one support set per generated token. Needs the engine flag `return_sampling_mask` |
| **Engine constraints** | — | Model Runner V2, `logprobs_mode=processed_logprobs`, no speculative decoding, no custom logits processors, and per request `temperature > 0` and `top_k > 0` (bounds the mask size) |
| **Per node** | The node's own token span of the routing array (prompt nodes too, since the trainer replays the whole sequence) | Only positions with `sampled_mask = 1`. Client and scaffold tokens don't need one (they're loss-masked); a trainer that wants dense rows uses a singleton `{token}` |
| **Storage** | `{id}.experts.zst`: uint8/uint16, flat, with offsets | `{id}.sampling_mask.zst`: CSR, i.e. flat int32 `ids` + per-token `offsets`. Up to `top_k` ids per sampled token (e.g. 50 → ~200 B/token) |
| **In the sample** | `routed_experts` aligned to `input_ids` | `sampling_mask` in CSR, aligned to the loss-masked positions |

Three rules this adds to the graph:

1. **The sampling params that shape the support are part of a model node's identity** (`temperature`, `top_p`, `top_k`, `min_p`). The same tokens sampled under a different `top_p` have a different support, so they're a different node. Otherwise the dedupe would pair one call's tokens with another call's mask.
2. **A node's routing comes from the call that created it.** A later call re-reports routing for its whole prompt. Where it overlaps existing nodes, that copy is dropped, not merged, so each path's R3 is exactly what was stored per node.
3. **Stored logprobs change meaning.** With `processed_logprobs` they're logprobs under the truncated, renormalized distribution. Record `logprobs_mode` on the trajectory so a trainer can't mix the two.

Out of capture's scope, but needed for training: SkyRL's trainer doesn't consume a sampling mask today (miles applies it with a `masked_fill(−inf)` before the log-softmax).

---

## 3. Build order (one PR per box)

```mermaid
flowchart TD
  M0["M0 skeleton<br/>pkg, CI, lint"] --> M1
  M1["M1 graph core (pure)<br/>Message hash, Node, MessageGraph,<br/>match/extend/paths, fixtures"] --> M2
  M1 --> M4
  M2["M2 text server<br/>aiohttp app, TrajectoryStore,<br/>create/finish, forward + graph"] --> M3
  M3["M3 record<br/>write on finish / TTL / shutdown,<br/>doc + sidecars, RecordReader"] --> M7
  M2 --> M4
  M4["M4 TITO<br/>RendererPool, TitoUpstream(vllm/skyrl),<br/>NodeTokens + R3 + sampling mask,<br/>bridge, samples, session release"] --> M5
  M2 --> M5
  M5["M5 retries<br/>retry-count / Idempotency-Key<br/>replay + coalesce"] --> M6
  M6["M6 CapturePool<br/>round-robin, N servers"] --> M8
  M7["M7 viewer<br/>port over record_dir"]
  M8["M8 SkyRL HarborCaptureGenerator<br/>(in SkyRL repo), parity run"]
```

| M | What lands | Done when |
| --- | --- | --- |
| M1 | `graph/` only, with no I/O | Fixture tests pass for next-turn, resample, compaction, subagent, tool change, and thinking-strip. The fixtures come from today's `examples/agents/` |
| M2 | Text mode end to end | A mock OpenAI upstream + 32 concurrent trajectories give a correct graph, and `finish` returns text paths with a status |
| M3 | Record on disk + TTL | `finish` writes the document and sidecars. Reading a listing opens only the documents. An idle trajectory is written as `abandoned` after the TTL. Graceful shutdown writes the open ones |
| M4 | TITO | Token-exact parity with today's `inference-capture` TITO on recorded Harbor turns. R3 and sampling masks round-trip from vLLM into samples, aligned to `input_ids` / sampled positions |
| M5 | Retries | An SDK retry storm gives zero extra nodes, and a resample still forks |
| M6 | Pool | Trajectories spread across 4 servers, and each `base_url` works on its own |
| M7 | Viewer | Today's viewer runs against the new files and never opens `experts` |
| M8 | SkyRL | Eval parity against the harness-side TITO collector |

